PR #3, opened 2022-08-10
Rename spec to match repository name

kontura:
This should help with: https://github.com/rpm-software-management/ci-dnf-stack/pull/1138


inknos:
Since we just discussed that the specfile should be named dnf5 and not libdnf this PR is perfect. Thanks

PR #5, opened 2022-08-10
Repoquery: show all packages that resolve selected advisories

kontura:
Instead of the exact versions that match advisory packages show all
exact matches plus higher versions (which also resolve the advisory).

dnf4 shows equal or if there is none then the first higher. Given that
this is not documented anywhere I think we could change it to show equal
plus all higher. Which seems to make more sense to me for repoquery.

For: https://github.com/rpm-software-management/ci-dnf-stack/pull/1101


j-mracek:
LGTM, Please feel free to merge it when all tests pass.

jrohel:
Seems OK.
I also re-ran all jobs and the CI tests passed.

PR #6, opened 2022-08-10
Fix building against sdbus-cpp-1.2

m-blaha:
sdbus-cpp-1.2 is landing in rawhide and one of changes was making used
sdbus::Error constructor explicit. Therefore not usable for copy-initialization.

### DIFF
--- a/dnf5daemon-server/threads_manager.hpp
+++ b/dnf5daemon-server/threads_manager.hpp
@@ -67,9 +67,9 @@ public:
                 } catch (const sdbus::Error & ex) {
                     reply = call.createErrorReply(ex);
                 } catch (const std::exception & ex) {
-                    reply = call.createErrorReply({dnfdaemon::ERROR, ex.what()});
+                    reply = call.createErrorReply(sdbus::Error(dnfdaemon::ERROR, ex.what()));
                 } catch (...) {
-                    reply = call.createErrorReply({dnfdaemon::ERROR, "Unknown exception caught"});
+                    reply = call.createErrorReply(sdbus::Error(dnfdaemon::ERROR, "Unknown exception caught"));
                 }
                 bool success = false;
                 std::string error_msg;


jrohel:
> one of changes was making used sdbus::Error constructor explicit.

one of the changes was the addition of a new explicit constructor `sdbus::Error(const std::string& name, const char* message = nullptr)`

dnf5daemon-server/threads_manager.hpp
```diff
@@ -67,9 +67,11 @@ class ThreadsManager {
                 } catch (const sdbus::Error & ex) {
                     reply = call.createErrorReply(ex);
                 } catch (const std::exception & ex) {
-                    reply = call.createErrorReply({dnfdaemon::ERROR, ex.what()});
+                    sdbus::Error err{dnfdaemon::ERROR, ex.what()};
+                    reply = call.createErrorReply(std::move(err));
```
jrohel, line 71: Since the new constructor is explicit, it must be called explicitly. No more is needed.
Replace the line `reply = call.createErrorReply({dnfdaemon::ERROR, ex.what()});` with `reply = call.createErrorReply(sdbus::Error(dnfdaemon::ERROR, ex.what()));`

It also applies to the next occurrence.

dnf5daemon-server/threads_manager.hpp
```diff
@@ -67,9 +67,11 @@ class ThreadsManager {
...
                 } catch (...) {
-                    reply = call.createErrorReply({dnfdaemon::ERROR, "Unknown exception caught"});
+                    sdbus::Error err{dnfdaemon::ERROR, "Unknown exception caught"};
+                    reply = call.createErrorReply(std::move(err));
```
jrohel, line 74: See above.

dnf5daemon-server/threads_manager.hpp
m-blaha: Thanks! Will change it.

jrohel:
Thanks for fixes.

PR #8, opened 2022-08-11
Move call to load libdnf plugins from application to base.setup()

jrohel:
We want the libdnf library to load the plugins automatically without the need for a call in the application.


j-mracek:
LGTM, but there are more problem with CI or clang test. Please can you double check that everything is ok with the test suite.

m-blaha:
I'm sure that the CI failure is unrelated to this PR.

PR #9, opened 2022-08-11
Fix formating of solv_userdata->checksum in logging message

jrohel:
The checksum is not a string, but a byte array.
Convert it to a hexadecimal string.


m-blaha (approved):
Thanks a lot for fixing this annoying issue!

PR #13, opened 2022-08-17
builddep: Warning when using macros with source rpms

jan-kolarik:
Follows up fix from dnf4 from https://github.com/rpm-software-management/dnf-plugins-core/pull/461.

Originally reported in https://bugzilla.redhat.com/show_bug.cgi?id=2077820.


m-blaha (approved):
Thanks, LGTM.

PR #16, opened 2022-08-18
test: Compare trimmed toml output

jan-kolarik:
Ignore stray newlines from the `toml` file as `toml::format` output has changed for `toml` library versions between fedora-35 and fedora-36.

### DIFF
--- a/test/libdnf/system/test_state.cpp
+++ b/test/libdnf/system/test_state.cpp
@@ -28,41 +28,40 @@ CPPUNIT_TEST_SUITE_REGISTRATION(StateTest);
 
 using namespace libdnf;
 
-static const std::string packages_contents{R"""(
-[packages]
+static const std::string packages_contents{R"""([packages]
 "pkg.x86_64" = {reason="User"}
 "unresolvable.noarch" = {reason="Dependency"}
 "pkg-libs.x86_64" = {reason="Dependency"}
-
 )"""};
 
-static const std::string nevras_contents{R"""(
-[nevras]
+static const std::string nevras_contents{R"""([nevras]
 "unresolvable-1.2-1.noarch" = {from_repo="repo2"}
 "pkg-libs-1.2-1.x86_64" = {from_repo=""}
 "pkg-1.2-1.x86_64" = {from_repo="repo1"}
-
 )"""};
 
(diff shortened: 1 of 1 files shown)


test/libdnf/system/test_state.cpp
```diff
@@ -63,6 +63,16 @@ enabled_stream = "stream-1"
 
 )"""};
 
+static inline void erase_newlines(std::string & str) {
+    str.erase(std::remove(str.begin(), str.end(), '\n'), str.cend());
```
lukash, line 67: Does this remove all newlines? You don't want to do that... The previous version was better, though I'd only trim the trailing newlines, as this is a workaround and ideally you don't want any stray newlines at either end, the extraneous newline should actually ideally be fixed in `toml11`... (I haven't reported a bug for it yet)

test/libdnf/system/test_state.cpp
```diff
@@ -63,6 +63,16 @@ enabled_stream = "stream-1"
...
+}
+
+static void assert_equal_without_newlines(std::string expected, std::string actual) {
+    erase_newlines(expected);
+    erase_newlines(actual);
```
lukash, line 72: You only really need to remove the trailing newline from the stored tomls, not from the static strings that we compare against (or store in the test files).

test/libdnf/system/test_state.cpp
```diff
@@ -63,6 +63,16 @@ enabled_stream = "stream-1"
...
+
+    CPPUNIT_ASSERT_EQUAL(expected, actual);
```
lukash, line 74: The issue with doing this is that whenever you get a failure, you'll get the line number of this `CPPUNIT_ASSERT_EQUAL()` call, and you'll be quite clueless which one is actually failing.

I'd return a trimmed copy from a `trimmed(...)` function and call that function on the file contents inside all the ASSERTs.

test/libdnf/system/test_state.cpp
jan-kolarik: OK I wanted to simplify things here. The problem is there is/is not a trailing newline between each section, so for the `modules_contents` case, the trailing remove is not enough.

PR #17, opened 2022-08-18
modules: dependencies and computing static contexts

pkratoch:


include/libdnf/module/module_dependencies.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+#include <vector>
+
+struct _ModulemdDependencies;
+
+
+namespace libdnf::module {
+
+
+class ModuleDependencies {
```
j-mracek, line 33: Do we really need that class? If yes do we really need to store inside `ModulemdDependencies * dependencies`? We can replace it by a `vector` with `map`

include/libdnf/module/module_item.hpp
```diff
@@ -138,13 +140,15 @@ class ModuleItem {
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getStaticContext()
     bool get_static_context() const;
 
-    /// @return The list of strings of modules requires (e.g. "nodejs", "nodejs:12", "nodejs:-11").
+    std::vector<ModuleDependencies> get_module_dependencies() const;
```
j-mracek, line 143: I was thinking whether is it correct or nor and I can confirm that it is according to libmodulemd specifications.
Dependencies contain `std::vector<std::map<string, vector<string>>>`.

I have let say less correct proposal. At the end we use all elements (name and stream) with the same impact and we do not care whether they were in the same map or a different one. Therefore we can simplify the structure to `std::vector<std::pair<std::string, std::vector<std::string>>>`. But `std::pair<std::string, std::vector<std::string>>` is not nice and not transparent therefore we can replace it by class Require or Dependency that will contain stream name (string) and streams (sorted vector with strings). I can imagine that it will have 3 methods - get_name, get_streams, and to_string. Because it will require a default public constructor for binding I recommend when someone will call get_name() and inside will be an empty string I suppose to raise a logic error or a libdnf assert.  As I said it is not a correct representation but may be it can simplify the usage of dependencies. What do you think?

pkratoch: Hi, thanks for your suggestion, I changed it to the vector of ModuleDependency objects.

include/libdnf/module/module_dependencies.hpp
pkratoch: Changed according to the suggestion below.

include/libdnf/module/module_dependency.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <string>
+#include <vector>
+
+namespace libdnf::module {
+
+
+class ModuleDependency {
+public:
+    ModuleDependency(std::string module_name, std::vector<std::string> streams)
```
j-mracek, line 31: Small issue
`ModuleDependency(std::string module_name, std::vector<std::string> streams)`
vs
ModuleDependency(std::string & module_name, std::vector<std::string> & streams)
`std::string module_name` pass a copy of `module_name` to constructor => redundant copy

include/libdnf/module/module_dependency.hpp
```diff
@@ -0,0 +1,53 @@
...
+        : module_name(module_name),
+          streams(streams) {}
+
+    /// @return Name of the required module.
+    /// @since 5.0
+    const std::string get_module_name() const { return module_name; };
```
j-mracek, line 37: Please return a reference - `const std::string & get_module_name() const { return module_name; };`

include/libdnf/module/module_dependency.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+    /// @return Vector of streams. Prefix '-' denotes conflicting stream, otherwise, the stream is one of required.
+    ///         If there are no other streams, any active stream of the module can satisfy the require.
+    /// @since 5.0
+    const std::vector<std::string> get_streams() const { return streams; };
```
j-mracek, line 42: Please reurn a reference `&`

include/libdnf/module/module_dependency.hpp
j-mracek: What about to also add a move constructor?
`ModuleDependency(std::string && module_name, std::vector<std::string> && streams)`

include/libdnf/module/module_dependency.hpp
```diff
@@ -0,0 +1,53 @@
...
+
```
j-mracek, line 43: What about to `add to_string` method? - then we do not need in ModuleItem get_dependencies that will return vector<string>.

include/libdnf/module/module_item.hpp
```diff
@@ -138,13 +140,19 @@ class ModuleItem {
...
 
-    /// @return The list of strings of modules requires (e.g. "nodejs", "nodejs:12", "nodejs:-11").
-    /// @param remove_platform When true, the method will not return requires with stream "platform"
-    //                         (default is false).
+    /// @return Module dependencies.
     /// @since 5.0
     //
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getRequires(bool removePlatform=false)
-    std::vector<std::string> get_requires(bool remove_platform = false);
+    std::vector<ModuleDependency> get_module_dependencies() const;
```
j-mracek, line 147: What about to add optional argument `bool remove_platform = false`?

include/libdnf/module/module_item.hpp
```diff
@@ -179,16 +191,34 @@ class ModuleItem {
     std::string get_name_stream() const;
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getNameStreamVersion()
     std::string get_name_stream_version() const;
+    std::string get_name_stream_context() const;
 
     std::vector<ModuleProfile> get_profiles_internal(const char * name) const;
-    static std::vector<std::string> get_requires(_ModulemdModuleStream * md_stream, bool remove_platform);
+
+    std::vector<ModuleDependency> get_module_dependencies(_ModulemdModuleStream * md_stream) const;
+    std::string get_module_dependencies_string(_ModulemdModuleStream * md_stream, bool remove_platform);
```
j-mracek, line 199: I don't know why you have those two methods?

j-mracek: May be the word `static` is missing in the definition.

include/libdnf/module/module_item.hpp
```diff
@@ -179,16 +191,34 @@ class ModuleItem {
...
+
     static std::string get_name_stream(_ModulemdModuleStream * md_stream);
 
+    // For compatibility with older modules that didn't have static contexts
+    std::string computed_static_context;
```
j-mracek, line 204: I have to check the usage to understand it.

include/libdnf/module/module_item_container.hpp
```diff
@@ -23,11 +23,17 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
 #include "libdnf/module/module_item.hpp"
+#include "libdnf/module/module_item_container_weak.hpp"
 
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
 
+struct s_Pool;
+typedef struct s_Pool Pool;
```
j-mracek, line 34: What about to create an impl?

libdnf/module/module_item.cpp
```diff
@@ -101,6 +110,16 @@ std::string ModuleItem::get_name_stream_version() const {
 }
 
 
+std::string ModuleItem::get_name_stream_context() const {
+    return libdnf::utils::sformat(
+        "{}:{}:{}",
```
j-mracek, line 115: I am not sure whether it is the fastest way how to merge multiple strings. Please can you add TODO to play here more with performance?

include/libdnf/module/module_dependency.hpp
j-mracek: It can be useful for outputs like `dnf module info` and it can be internally used by get_module_dependencies_string()

include/libdnf/module/module_item.hpp
```diff
@@ -138,13 +140,19 @@ class ModuleItem {
...
-    std::vector<std::string> get_requires(bool remove_platform = false);
+    std::vector<ModuleDependency> get_module_dependencies() const;
+
+    /// @return The string representing module dependencies (e.g. "ninja;platform:[f29]" or "nodejs:[-11]").
+    /// @param remove_platform When true, the method will not return dependencies with stream "platform"
+    ///                        (default is false).
+    /// @since 5.0
+    //
+    // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getRequires(bool removePlatform=false)
+    std::string get_module_dependencies_string(bool remove_platform = false);
```
j-mracek, line 155: What about to move it to private section? I guess it is too specific to keep it in public API.

include/libdnf/module/module_item_container.hpp
```diff
@@ -75,10 +79,16 @@ class ModuleItemContainer {
     std::vector<std::string> get_default_profiles(std::string module_name, std::string module_stream);
 
 private:
+    friend ModuleItem;
+
     BaseWeakPtr base;
+    Pool * pool;
+    std::map<std::string, int> repositories;
```
j-mracek, line 86: Please add a description what repositories contain and how it is used.

j-mracek: Or add TODO

include/libdnf/module/module_item_container.hpp
```diff
@@ -75,10 +79,16 @@ class ModuleItemContainer {
...
     WeakPtrGuard<ModuleItemContainer, false> data_guard;
 
     BaseWeakPtr get_base() const;
+
+    std::vector<std::unique_ptr<ModuleItem>> modules_without_static_context;
```
j-mracek, line 91: Please add comment why we need it.

include/libdnf/module/module_dependency.hpp
```diff
@@ -0,0 +1,61 @@
...
+    /// @return Vector of streams. Prefix '-' denotes conflicting stream, otherwise, the stream is one of required.
+    ///         If there are no other streams, any active stream of the module can satisfy the require.
+    /// @since 5.0
+    const std::vector<std::string> & get_streams() const { return streams; };
+
+private:
+    friend class ModuleItem;
+
+    std::string to_string();
```
j-mracek, line 51: Is there any reason to have it private?

include/libdnf/module/module_dependency.hpp
j-mracek: Nice

pkratoch: I am only not sure if it's used correctly here: https://github.com/rpm-software-management/dnf5/pull/17/files#diff-e1c5753bcfca8e49721f0b42e243d88368343645cf917ed472164466592b0740R214

include/libdnf/module/module_item.hpp
pkratoch: I don't know how to test it if I move it to private. But I can remove it from the test for now, until there is another way.

include/libdnf/module/module_item_container.hpp
pkratoch: I moved this to Impl and added a comment.

include/libdnf/module/module_dependency.hpp
pkratoch: For the same reason as `get_module_dependencies_string`? But it's possible you said it's also used elsewhere, but I can't remember where, sorry. If that's the case, I can make it public.

j-mracek: Outputs will require it and in general I think it will be useful to have it public

PR #18, opened 2022-08-19
system state: Add versioning, rpmdb cookie, make API private

lukash:


test/libdnf/private_accessor.hpp
```diff
@@ -0,0 +1,85 @@
+/*
+Copyright Contributors to the libdnf project.
```
lukash, line 2: I am actually not sure about this copyright notice, since the code is copy-pasted from someone else...

include/libdnf/base/base.hpp
```diff
@@ -95,19 +95,18 @@ class Base {
...
-    // TODO(lukash) if it stays on public API, use WeakPtr
-    libdnf::system::State & get_system_state() { return *system_state; }
-
     void add_plugin(plugin::IPlugin & iplugin_instance);
     void load_plugins();
     plugin::Plugins & get_plugins() { return plugins; }
 
     libdnf::BaseWeakPtr get_weak_ptr() { return BaseWeakPtr(this, &base_guard); }
 
+    class Impl;
```
j-mracek, line 104: It is not an issue, but the forward declaration in other header file is just above `ImplPtr<Impl> p_impl;`. may be it would be better to do it by the same approach. What do you think?

test/libdnf/private_accessor.hpp
j-mracek: Good point, because of movement of the project, the information is outdated. Anyway, for the PR I have no problem to keep it like i is.

include/libdnf/base/base.hpp
lukash: This is a limitation of the workaround I've used to access the private member in unit tests, the `Impl` must be declared public...

test/libdnf/private_accessor.hpp
lukash: Hmm, I'm not sure you understand, this is unrelated to moving the project, I'm talking about the code in this file, which is copy-pasted from https://gist.github.com/dabrahams/1528856

I don't think it'll cause issues, but I thought I'd mention it.

j-mracek: Thank you for additional information. Well in this case it can be a problem. Do you know the license under which the code was published?

lukash: The linked gist doesn't have a license notice, so no. It is itself based on someone else's code: http://bloglitb.blogspot.com/2010/07/access-to-private-members-thats-easy.html (so asking one author may not be enough...)

Inferring from the way it's published I'd say it's public domain. I just thought I'd mention it for completeness...

jrohel: I tried to come up with my own version. I thought of several solutions. They all use the same idea. The address of private members can be obtained during explicit instantiation of the class.

```
// A getter generating structure
template<typename AccessTag, auto value>
struct CreateGetter {
    friend constexpr auto get(AccessTag) { return value; }
};

// Helper macro
#define create_getter(AccessTag, class_member_ptr) \
struct AccessTag {}; \
constexpr auto get(AccessTag); \
template struct CreateGetter<AccessTag, class_member_ptr>;
```

Example of usage:
```
// Class with private member
class A {
private:
    int priv_value = 10;
};

// Just use the macro to create the AccessTag and the getter
create_getter(A_priv_value, &A::priv_value)

// And use getter
int main() {
    A a;
    std::cout << a.*get(A_priv_value()) << std::endl; // read value
    a.*get(A_priv_value()) = 5;  // set value
    std::cout << a.*get(A_priv_value()) << std::endl; // read value
    return 0;
}
```

j-mracek: Is it possible for tests import a private class from libdnf5/libdnf?

jrohel:
Thanks

PR #19, opened 2022-08-22
Support aliases of command line arguments from config file, remove hardcoded aliases

jrohel:
PR implements support for defining aliases of command line arguments using configuration files.
It removes most of the existing aliases from the source code and defines them in the configuration file.

The configuration files are in TOML format and are loaded from the following directories:
- "{INSTALL_PREFIX}/lib/dnf5/aliases.d" -> "/usr/lib/dnf5/aliases.d" - for distribution configs
- "{SYSCONFDIR}/dnf/dnf5-aliases.d" -> "/etc/dnf/dnf5-aliases.d" - for system wide user config
- "{XDG_USR_CONFIG_DIR}/dnf5/aliases.d" -> "{user_home_dir}/.config/dnf5/aliases.d" - for individual users

Inside the directories, the order of loading is determined by alphabetical order by file name.


dnf5/config/usr/lib/dnf5/aliases.d/compatibility.conf
```diff
@@ -0,0 +1,125 @@
+########################################
+# Global options compatibility aliases
+########################################
```
kontura, line 3: I can see there still are a couple of commands specific options compatibility aliases. Namely the advisory shared options: `--advisory`, `--bz` and `--cve`. 
These are per command aliases so `upgrade`, `repoquery` and `install` all have them, and `advisory` command also has two.

I think these would be possible to define with something like:
```
['repoquery.advisory']
type = 'cloned_named_arg'
long_name = 'advisory'
source = 'repoquery.advisories'
complete = false
```
but it would be copied for each command - 4 times. 
Do we plan to do this as well? Then the code base wouldn't have any hard-coded aliases.

dnf5/cmdline_aliases.cpp
```diff
@@ -0,0 +1,640 @@
...
+
+                        if (group) {
+                            group->register_argument(alias_arg);
+                        }
+                        element_parent_cmd->register_named_arg(alias_arg);
+                        break;
+                    }
+
+                    // Creates a new named argument
+                    case ElementType::NAMED_ARG: {
```
kontura, line 391: Maybe I have missed something but how is the `NAMED_ARG` used?
I don't see how it can connect to any functionality.

Is there a plan to have all commands defined entirely in the config so that users can use completely custom naming, descriptions etc..? and dnf5 will just hook into some expected ids?

kontura (commented):
I think we should also have some test. Maybe a few ci-dnf-stack tests could be created?

include/libdnf-cli/argument_parser.hpp
```diff
@@ -168,6 +168,12 @@ class ArgumentParser {
         /// Can contain this argument. Groups of conflicting argument can be used.
         void set_conflict_arguments(std::vector<Argument *> * args) noexcept { conflict_args = args; }
 
+        /// Adds a conflicting argument. Also adds reverse conflict.
+        void add_conflict_argument(Argument & conflict_arg);
+
+        /// Add conflict arguments from another argument. Also adds reverse conflicts.
+        void add_conflict_arguments_from_another(Argument & src_arg);
```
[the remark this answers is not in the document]
jrohel: I also originally wanted to name the method `copy_conflict_arguments_from_another`. But then I wondered if it would be confusing. If conflicting arguments already exist (eg added explicitly with `add_conflict_argument`) using `copy_conflict_arguments_from_another` will add more conflicts and leave the existing ones. That's why I called `add`, that it will not replace existing conflicts, but add new ones. But maybe I'm overthinking it.

dnf5/cmdline_aliases.cpp
jrohel: The `cloned_named_args` is a simplified version of the named argument alias. It basically copies the existing argument. The user defines which argument (source), where and new names (short, long). And the group and whether to offer bash completions.

The `named_arg` is more powerful. Defines a new named argument that can represent a combination of existing named arguments (defined using the `attached_named_args` table) and pass a different value to them.
So far the replaced aliases were simple and therefore it was enough to use 'cloned_named_args'.

dnf5/config/usr/lib/dnf5/aliases.d/compatibility.conf
```diff
@@ -0,0 +1,125 @@
...
+
+['options-compatibility-aliases']
+type = 'group'
+header = "Options Compatibility aliases:"
+
+['nobest']
+type = 'cloned_named_arg'
```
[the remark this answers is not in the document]
jrohel: Yes. Documentation is missing. But the release is fast approaching. And so I prefer to finish the API as best as possible and document it later than not to deliver the API. Documentation can be completed later. But redesigning an API after it's already released is more painful. Yes we can create the JIRA.

The good thing is that users can use the already existing definition of aliases from this PR as a template.

libdnf-cli/argument_parser.cpp
```diff
@@ -457,14 +448,20 @@ void ArgumentParser::Command::print_complete(
     }
 }
 
-void ArgumentParser::Command::parse(
-    const char * option, int argc, const char * const argv[], const std::vector<NamedArg *> * additional_named_args) {
+void ArgumentParser::Command::parse(const char * option, int argc, const char * const argv[]) {
```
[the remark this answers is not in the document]
jrohel: Ok. I enhanced the commit message.

dnf5/config/usr/lib/dnf5/aliases.d/compatibility.conf
jrohel: Yes. My idea is source code without hard-coded aliases. A few aliases left in the source code, this PR is already so big, but I'll look at it again.

dnf5/cmdline_aliases.cpp
kontura: I see, I confused this together with the COMMAND.
Thanks for explanation.

dnf5/config/usr/lib/dnf5/aliases.d/compatibility.conf
kontura: It is not a problem, it can be done in the next PR.

I will merge once the CI finishes.

PR #24, opened 2022-08-26
Improvements of documentation

j-mracek:


include/libdnf/base/transaction.hpp
```diff
@@ -33,17 +33,20 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
 
-
+/// Error related to processing transaction
 class TransactionError : public Error {
 public:
     using Error::Error;
+    /// @return Return a domain of the Error Class
     const char * get_domain_name() const noexcept override { return "libdnf::base"; }
+    /// @return Return a name of the Error
     const char * get_name() const noexcept override { return "TransactionError"; }
```
inknos, line 43: I would rephrase it as "Return Error Class's domain name" and "Return Error Class' name"

include/libdnf/base/goal_elements.hpp
```diff
@@ -95,29 +99,46 @@ enum class GoalAction {
...
 };
 
 struct GoalJobSettings : public ResolveSpecSettings {
 public:
+    /// Return used value for the strict
     GoalUsedSetting get_used_strict() const { return used_strict; };
+    /// Return used value for the best
     GoalUsedSetting get_used_best() const { return used_best; };
+    /// Return used value for the clean_requirements_on_remove
     GoalUsedSetting get_used_clean_requirements_on_remove() const { return used_clean_requirements_on_remove; };
```
inknos, line 129: I would rephrase as "Return used value for strict" or "Return used values for the strict option". Same for best and clean_requirements_on_remove

include/libdnf/base/goal_elements.hpp
```diff
@@ -95,29 +99,46 @@ enum class GoalAction {
...
 
+    /// Setting whether hints should be reported
```
inknos, line 131: Please, use "Set whether..."

include/libdnf/base/goal_elements.hpp
```diff
@@ -95,29 +99,46 @@ enum class GoalAction {
...
     bool report_hint{true};
+    /// Setting for strict, AUTO means that it is handled according the the default behavior
     GoalSetting strict{GoalSetting::AUTO};
+    /// Setting for best, AUTO means that it is handled according the the default behavior
     GoalSetting best{GoalSetting::AUTO};
+    /// Setting for clean_requirements_on_remove, AUTO means that it is handled according the the default behavior
     GoalSetting clean_requirements_on_remove{GoalSetting::AUTO};
```
inknos, line 138: Same here. "Set strict" or "Set strict options" sound better. Same for other options documented as "Setting"

j-mracek:
Thank you very much for your comments. I hope I fixed all of them.

include/libdnf/base/transaction.hpp
inknos: I just realized I corrected slightly wrong. It should be “class'” not “class's”. My mistake. Since all other corrections are good, I pushed this one character correction and I will merge it now.

PR #25, opened 2022-08-31
Fix issues find by running with sanitizers

kontura:
Even with these patches running with sanitizers still doesn't pass cleanly. However I believe the remaining reports are not problems with dnf5 or I wasn't able to figure it out at this time.

Couple of notes for future reference:

Note 1
---

Just about every run of dnf5 binary produces:
```
==16==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x7f2bebc2e68f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)
    #1 0x7f2be83a373c  (<unknown module>)
```
I am not sure to what does this refer to but it is somehow connected to dynamic loading of plugins (`dlopen()` - `dlclose()`) because when the `dlclose()` is commented out the reported leak disappears. I was able to find that running sanitizers with dynamic library loading is problematic but I didn't find any solutions.

Note 2
---

Similar report also happens in unittests first: (this error is correct and we do it on purpose)
```
/root/dbox/src/dnf5/test/libdnf/run_tests.cpp:70:53: runtime error: member access within address 0x608000001ea0 which does not point to an object of type 'HackTestCaller'
0x608000001ea0: note: object is of type 'CppUnit::TestCaller<AdvisoryAdvisoryTest>'
 00 00 00 00  e8 5c bd 00 00 00 00 00  78 5d bd 00 00 00 00 00  d0 20 00 00 40 60 00 00  23 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'CppUnit::TestCaller<AdvisoryAdvisoryTest>'
```
and later:
```
==86903==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 128 byte(s) in 2 object(s) allocated from:
    #0 0x7fdb31ded595 in __interceptor_realloc.part.0 (/lib64/libasan.so.8+0xb9595)
    #1 0x7fdb3157e90a in d_growable_string_callback_adapter (/lib64/libstdc++.so.6+0xb090a)
```
This seems to be due to the workaround in `LogCaptureListener` for accessing the `BaseTestCase` to get logs. Not using the [`LogCaptureListener`](https://github.com/rpm-software-management/dnf5/blob/main/test/libdnf/run_tests.cpp#L66-L89) fixes the errors. I think the second error might be in fact this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80578

Note 3
---

There are also some memory leaks in `zchunk`. Either librepo is using it wrong or there are problems in the `zchunk` library itself. I haven't looked into it yet. (Detected only on rawhide host - not in a f36 container)

Note 4
---

Finally to build with `tito` I had to remove `-Werror` (otherwise it failed - complaining some std library functions may be used uninitialized) and disable the `LogCaptureListener`.

### DIFF
--- a/include/libdnf-cli/output/environmentinfo.hpp
+++ b/include/libdnf-cli/output/environmentinfo.hpp
@@ -89,6 +89,7 @@ static struct libscols_table * create_environmentinfo_table(EnvironmentType & en
     scols_symbols_set_right(sy, "  ");
     scols_symbols_set_vertical(sy, "");
     scols_table_set_symbols(table, sy);
+    scols_unref_symbols(sy);
 
     add_line_into_environmentinfo_table(table, "Id", environment.get_environmentid().c_str(), "bold");
     add_line_into_environmentinfo_table(table, "Name", environment.get_name().c_str());
--- a/libdnf/repo/solv_repo.cpp
+++ b/libdnf/repo/solv_repo.cpp
@@ -84,7 +84,8 @@ bool SolvRepo::can_use_solvfile_cache(fs::File & solvfile_cache) {
     int dnf_solv_userdata_len_read;
 
     int ret_code = solv_read_userdata(solvfile_cache.get(), &dnf_solv_userdata_read, &dnf_solv_userdata_len_read);
-    std::unique_ptr<SolvUserdata> solv_userdata(reinterpret_cast<SolvUserdata *>(dnf_solv_userdata_read));
+    std::unique_ptr<SolvUserdata, decltype(&solv_free)> solv_userdata(
+        reinterpret_cast<SolvUserdata *>(dnf_solv_userdata_read), &solv_free);
     if (ret_code != 0) {
         auto & pool = get_pool(base);
         logger.warning(
--- a/include/libdnf-cli/output/groupinfo.hpp
+++ b/include/libdnf-cli/output/groupinfo.hpp
@@ -102,6 +102,7 @@ static struct libscols_table * create_groupinfo_table(GroupType & group) {
     scols_symbols_set_right(sy, "  ");
     scols_symbols_set_vertical(sy, "");
     scols_table_set_symbols(table, sy);
+    scols_unref_symbols(sy);
 
     add_line_into_groupinfo_table(table, "Id", group.get_groupid().c_str(), "bold");
     add_line_into_groupinfo_table(table, "Name", group.get_name().c_str());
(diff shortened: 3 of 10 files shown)


Conan-Kudo (approved):

libdnf/repo/solv_repo.cpp
```diff
@@ -84,7 +84,8 @@ bool SolvRepo::can_use_solvfile_cache(fs::File & solvfile_cache) {
     int dnf_solv_userdata_len_read;
 
     int ret_code = solv_read_userdata(solvfile_cache.get(), &dnf_solv_userdata_read, &dnf_solv_userdata_len_read);
-    std::unique_ptr<SolvUserdata> solv_userdata(reinterpret_cast<SolvUserdata *>(dnf_solv_userdata_read));
+    std::unique_ptr<SolvUserdata, decltype(&free)> solv_userdata(
+        reinterpret_cast<SolvUserdata *>(dnf_solv_userdata_read), &free);
```
jrohel, line 88: Libsolv gets memory by `solv_malloc`! Yes, `solv_malloc` uses `malloc` internally. However, this is an implementation detail of libsolv.
Memory allocated with `solv_malloc` should be freed with `solv_free`.

kontura: My mistake. I have updated it.
Thanks!

include/libdnf-cli/output/environmentinfo.hpp
```diff
@@ -109,6 +109,7 @@ template <typename EnvironmentType>
 void print_environmentinfo_table(EnvironmentType & environment) {
     struct libscols_table * table = create_environmentinfo_table(environment);
     scols_print_table(table);
+    scols_unref_symbols(scols_table_get_symbols(table));
```
jrohel, line 112: No. You are fixing the memory leak in the wrong place. Memory leak occurs earlier in the `create_environmentinfo_table` function.

Symbols use reference counting. The `create_environmentinfo_table` function does:
1. `auto sy = scols_new_symbols();` - allocates new symbols with reference counter set to 1
2. `scols_table_set_symbols(table, sy);` - add reference to `sy` from the table, symbols reference counter is incremented to 2
3. The variable `sy` is not longer used and there is a memory leak, please add the line `scols_unref_symbols(sy);` immediately after `scols_table_set_symbols(table, sy);` to decrement the reference counter back to 1.

And then `scols_unref_table(table);` will properly decrement the reference counter to 0 and memory will be freed.

Please fix it on other places as well.

kontura: I see, fixed.

kontura:
Rebased.

jrohel:
Thanks

PR #28, opened 2022-09-02
Add reports of skipped packages and uninstallable best candidates 

j-mracek:


j-mracek:
I still have to try broken-dependencies-report.feature.

libdnf/base/transaction.cpp
```diff
@@ -216,6 +216,16 @@ void Transaction::Impl::set_transaction(rpm::solv::GoalPrivate & solved_goal, Go
     auto solver_problems = process_solver_problems(base, solved_goal);
     if (!solver_problems.empty()) {
         add_resolve_log(GoalProblem::SOLVER_ERROR, solver_problems);
+    } else {
+        // TODO(jmracek) To improve performance add a test whether it make sence to resolve transaction in strict mode
+        // Test whether there were skipped jobs or used not the best candidates due to broken dependencies
+        rpm::solv::GoalPrivate solved_goal_copy(solved_goal);
+        solved_goal_copy.set_run_in_strict_mode(true);
+        solved_goal_copy.resolve();
```
jan-kolarik, line 224: I don't know the internals properly yet, but it seems to me like kind of a workaround solution, because we are resolving the goal twice. 

Could we implement classifying/grouping the solver problems to strict and non-strict ones and when processing the `process_solver_problems` result we would just have a check and based on that call `add_resolve_log` with `SOLVER_ERROR` or `SOLVER_PROBLEM_STRICT_RESOLVEMENT`?

j-mracek: It is correct, it is workaround. The problem with the solver is that it provide a solution or not for the particular setting. It means if the strict resolvement fails there will be no transaction at all. If we request relax setting, then solver will provide a solution but it can skip packages, job without a report that it happen. Solver translate requests into set of the rules and according to rules it provide a solution. If a problem appears it reports only broken rules. If we say that rule is optional, skipping the rule is allowed therefore no problem for solver to report. or may be I do not understand your suggestion.

jan-kolarik: Thanks for the explanation, I expected the reason for doing this is something like that, but anyway I wanted to ask :slightly_smiling_face:

jan-kolarik:
LGTM, I've added a above mentioned PR for the tests changes which are passing locally.

PR #29, opened 2022-09-06
Depend on toml11 via toml11-static

ppisar:
According to Fedora packaging guidelines
<https://docs.fedoraproject.org/en-US/packaging-guidelines/#_packaging_header_only_libraries>,
header-only libraries must be pulled in with a -static RPM dependency
symbol. A rationale is that this form of static linking must be
tracked for security purposes.

This patch fixes it.


jrohel:
Thank you

PR #33, opened 2022-09-08
System state: correctly calculate the GROUP reason

m-blaha:
Also adds a new APIs:
- get_package_groups(na)
- get_packages_by_reason(reason)

### DIFF
--- a/libdnf/system/state.cpp
+++ b/libdnf/system/state.cpp
@@ -238,6 +238,27 @@ void State::set_package_reason(const std::string & na, transaction::TransactionI
 }
 
 
+std::set<std::string> State::get_packages_by_reason(const std::set<transaction::TransactionItemReason> & reasons) {
+    std::set<std::string> packages;
+    if (reasons.contains(transaction::TransactionItemReason::GROUP)) {
+        auto & package_groups = get_package_groups_cache();
+        for (const auto & pkg : package_groups) {
+            packages.emplace(pkg.first);
+        }
+    }
+    std::set<std::string> reasons_str;
+    for (const auto & reason : reasons) {
+        reasons_str.emplace(transaction::transaction_item_reason_to_string(reason));
+    }
+    for (const auto & [na, pkg_state] : package_states) {
+        if (reasons_str.contains(pkg_state.reason)) {
+            packages.emplace(na);
+        }
+    }
+    return packages;
+}
+
+
 void State::remove_package_na_state(const std::string & na) {
     package_states.erase(na);
 }
--- a/libdnf/system/state.cpp
+++ b/libdnf/system/state.cpp
@@ -284,6 +284,16 @@ void State::remove_group_state(const std::string & id) {
     package_groups_cache.reset();
 }
 
+std::set<std::string> State::get_package_groups(const std::string & name) {
+    auto & package_groups = get_package_groups_cache();
+    auto it = package_groups.find(name);
+    if (it == package_groups.end()) {
+        return {};
+    } else {
+        return it->second;
+    }
+}
+
 
 std::string State::get_module_enabled_stream(const std::string & name) {
     auto it = module_states.find(name);
--- a/libdnf/system/state.cpp
+++ b/libdnf/system/state.cpp
@@ -200,13 +200,26 @@ State::State(const std::filesystem::path & path) : path(path) {
 
 
 transaction::TransactionItemReason State::get_package_reason(const std::string & na) {
+    transaction::TransactionItemReason packages_reason = transaction::TransactionItemReason::NONE;
     auto it = package_states.find(na);
-    if (it == package_states.end()) {
-        return transaction::TransactionItemReason::NONE;
+    if (it != package_states.end()) {
+        // TODO(lukash) this allows more reasons than valid here, assert?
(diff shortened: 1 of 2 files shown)


libdnf/system/state.cpp
```diff
@@ -222,8 +235,30 @@ void State::set_package_reason(const std::string & na, transaction::TransactionI
         reason_str);
 
     package_states[na].reason = reason_str;
+
+    // invalidate the package by reasons cache
```
j-mracek, line 239: I think the set operation invalidate the cache of package reasons. But the comment is somehow cryptic for me.

libdnf/system/state.hpp
```diff
@@ -112,6 +114,11 @@ class State {
     /// @since 5.0
     void set_package_reason(const std::string & na, transaction::TransactionItemReason reason);
 
+    // @return All NA of packages installed with specified reason.
+    // @param reason The reason to search for
+    /// @since 5.0
+    std::set<std::string> get_packages_by_reason(const transaction::TransactionItemReason & reason);
```
j-mracek, line 120: Because the PR does not contain the usage therefore I am fine with future change of the method.

libdnf/system/state.cpp
```diff
@@ -262,6 +275,8 @@ GroupState State::get_group_state(const std::string & id) {
 
 void State::set_group_state(const std::string & id, const GroupState & group_state) {
     group_states[id] = group_state;
+    // invalidate the package groups cache
+    package_groups_cache.reset();
```
lukash, line 301: You also need this in `remove_group_state()` right below. Maybe it should also be added to `State::load()` for completeness, though it may not ever get used that way...

libdnf/system/state.cpp
```diff
@@ -235,8 +235,30 @@ void State::set_package_reason(const std::string & na, transaction::TransactionI
         reason_str);
 
     package_states[na].reason = reason_str;
+
+    // invalidate the package by reasons cache
+    packages_by_reason_cache.reset();
```
lukash, line 240: Same here, needs to be added to `remove_package_na_state()` and maybe `load()`.

libdnf/system/state.hpp
```diff
@@ -223,6 +228,9 @@ class State {
     // map caches groups for packages. Package name is the key, set of group ids is the value
     const std::map<std::string, std::set<std::string>> & get_package_groups_cache();
     std::optional<std::map<std::string, std::set<std::string>>> package_groups_cache;
+    // map caches packages by the reason. The key is reason (as string), value set of NA
+    const std::map<std::string, std::set<std::string>> & get_packages_by_reason_cache();
+    std::optional<std::map<std::string, std::set<std::string>>> packages_by_reason_cache;
```
lukash, line 233: Is the cache really needed? Maybe it would be enough to just calculate it every time, assuming this will just be used for a potential command and not frequently in code in general. Though the cache probably doesn't hurt anything.

libdnf/system/state.cpp
m-blaha: oh, righ. good catch. I'll fix it.

libdnf/system/state.hpp
m-blaha: For `group remove` usecase I need to ask for every package installed with the group, so I consider the `package_groups_cache` useful. I'm not that convinced about the `package_by_reason_cache` though.

m-blaha: Do you prefer to remove the cache? Actually the `get_packages_by_reason()` function is called twice, but once with `GROUP` argument where this cache does not apply... Hm, maybe it should be called third time with `EXTERNAL_USER` reason. But the API can still be changed to accept a list of reasons...

lukash: I don't see the calls, are they in the other PR?

I guess you can keep it, apart from the small memory consumption it doesn't hurt anything, just adding some complexity which I'm not sure is warrantied.

m-blaha: I've added a commit with cache removal. I can squash them if you like it.
Yes, the calls are in other PR. Here is the draft - https://github.com/rpm-software-management/dnf5/pull/35. It needs to be adjusted to the latest changes.

lukash: Yeah, if you remove it, please squash, thanks!

m-blaha: Removed and squashed.

j-mracek:
LGTM

j-mracek (approved):

libdnf/system/state.hpp
```diff
@@ -210,6 +212,12 @@ class State {
     std::map<std::string, GroupState> group_states;
     std::map<std::string, ModuleState> module_states;
     SystemState system_state;
+
+    // Caches to speed-up searching the maps by values
+    //
+    // map caches groups for packages. Package name is the key, set of group ids is the value
+    const std::map<std::string, std::set<std::string>> & get_package_groups_cache();
```
lukash, line 229: Just a nitpick, I'd prefer the method to be in a block with the other private methods and the member variables in a separate block below them. Unless you prefer this way for a reason...

m-blaha: I've moved them. Originally there were two caches so I wanted to keep them together.

PR #34, opened 2022-09-09
HistoryDB: create pkg_names table and reference it in rpm and comps_group_package

jrohel:


jan-kolarik:
Reviewed and LGTM. Thanks!

PR #35, opened 2022-09-09
User-installed packages are not removed as unused dependencies

m-blaha:


libdnf/rpm/solv/goal_private.hpp
```diff
@@ -116,6 +118,9 @@ class GoalPrivate {
 
     libdnf::solv::Pool & get_pool() { return libdnf::get_pool(base); };
 
+    /// @return True if SOLVER_CLEANDEPS flag was set for any of jobs
+    bool get_clean_deps_present() { return clean_deps_present; }
```
j-mracek, line 122: [minor] I know it is difficult to name it and it is not wrong. What about to not use `get` word but something else. Like
`is_clean_deps_present`, `using_clean_deps`, `contain_clean_deps`, ... 
But as I said, it is not a problem.

m-blaha: I guess `is_clean_deps_present` will do. I'm grateful for every advice, because coming up with a name is the hardest part :)

j-mracek:
LGTM, but there is an error with integration tests. I will try to rerun all tests

m-blaha:
Hm, looks like some infra error:
```
Error: Error downloading packages:
2022-09-13T11:28:10.1803530Z   Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f36&arch=x86_64 (IP: 152.19.134.142)
```

Let's see the new run...

j-mracek:
All tests passed

PR #39, opened 2022-09-13
Add template command

inknos:


lukash:
I'm not sure what's the purpose of this tutorial, as adding a command straight to dnf5 seems more like a contribution than a usage (adding a dnf5 command from outside dnf5 would be as a plugin).

But right now the tutorials in `doc/tutorial` are for using the libdnf API, so some restructuring is probably needed to keep it clean and clear.

inknos:
Agree with you on restructuring. Maybe this could go to some specific section or subsection for contributors at the end of the documentation's page.

The idea of the template came up this morning while looking at the code with @j-mracek. We thought it might be useful to document a template command. Here's why.

- It is not clear why c++ pointers and smart pointers are used in the same scope. This clarifies that it is safe for an external contributor to use it in this context. I wouldn't make sense to comment in a single command or in all of them.
- It would give some sort of guideline on how to write dnf5 commands. This could eventually prevent writing bad code and consequently longer review times. I believe having a template would help us too to avoid such mistakes.

Following this example, plugins would be documented as well.

praiskup:
In Copr team we'd appreciate this kind of a simple "how-to write a new DNF plugin command".  The plan is to keep the support for `dnf copr` for the DNF4 => DNF5 transition.  CC @xsuchy

mcurlej:
@lukash i don't think that the tutorial section is strictly for usage of DNF5. Tutorials are tutorials. Just create a subsection for usage of DNF5 and a section for writing plugins. The more examples about everything we have the better. If in the future we decide to make a different structure of tutorials we can. The most important part is, the more documentation we have the better.

@inknos +1

doc/templates/command/template.cpp
```diff
@@ -0,0 +1,42 @@
+#include "template.hpp"
+
+#include <iostream>
+
+namespace dnf5 {
+
+using namespace libdnf::cli;
+
+void TemplateCommand::set_argument_parser() {
+    auto & ctx = get_context();
```
j-mracek, line 10: What about to add a comment what the context contains?

doc/templates/command/template.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+
+namespace dnf5 {
+
+// This would define the command template with two options:
+// --foo
+// --bar
+class TemplateCommand : public Command {
+public:
+    explicit TemplateCommand(Command & parent) : Command(parent, "template") {}
```
j-mracek, line 20: What about to mentioned that constructor sets the name of the command?

doc/templates/command/template.cpp
```diff
@@ -0,0 +1,42 @@
...
+    auto & parser = ctx.get_argument_parser();
+
+    auto & cmd = *get_argument_parser_command();
```
j-mracek, line 13: Please add a comment to each block of the code - like // set short description for the command. it will be showed in help or something else.

doc/templates/command/template.cpp
```diff
@@ -0,0 +1,42 @@
...
+    cmd.set_short_description("A command that prints its name and arguments' name");
+
+    foo_option = dynamic_cast<libdnf::OptionBool *>(
```
j-mracek, line 16: Please could you add a description what the block does?

doc/templates/command/template.hpp
```diff
@@ -0,0 +1,50 @@
...
+    libdnf::OptionBool * foo_option{nullptr};
+
+    // Second: STL Unique Pointer
+    // Create a unique_ptr<BarOption>
+    //
+    // This might be needed in case you need to do some evaluation of the options
+    // after the TemplateCommand creation before any funciton call like
+    // set_argument_parser.
+    // In order to use a unique_ptr, BarOption has to be defined (in this example
+    // the definition is in arguments.hpp)
```
j-mracek, line 42: I am somehow confused by the description.

doc/templates/command/template.cpp
```diff
@@ -0,0 +1,42 @@
...
+    bar_option = std::make_unique<BarOption>(*this);
+
+    cmd.register_named_arg(foo);
+}
+
+void TemplateCommand::run() {
+    // auto & ctx = get_context();
+    std::cout << "Template Command" << std::endl;
+
+    if (foo_option->get_value()) {
```
j-mracek, line 34: What about to add a comment - that will tell that the value of the options is accessible like this

doc/templates/command/template.hpp
```diff
@@ -0,0 +1,50 @@
...
+private:
+    // There are two ways to specify options for a command
+
+    // First: C++ Pointer
+    // Create a OptionBool pointer for the option you need.
+    //
+    // There is no problem using a pure C++ pointer since the OptionBool class
+    // will take care of moving the pointer ownership to the parser, which will
+    // be in charge of handling the memory deallocation.
+    libdnf::OptionBool * foo_option{nullptr};
```
j-mracek, line 33: The pointer is only required to keep access to the value after option parser process all options and arguments. May be it would be nice to mentioned when the parser does it.

j-mracek:
LGTM

PR #43, opened 2022-09-16
HistoryDB: Rename tables "pkg_names" and "archs" to singular form

jrohel:
To be consistent with other table names.


m-blaha (approved):
Thanks!

PR #44, opened 2022-09-16
modules: Load modules during repository loading

pkratoch:


include/libdnf/repo/repo.hpp
```diff
@@ -374,6 +374,7 @@ class Repo {
     void make_solv_repo();
 
     void load_available_repo(libdnf::repo::LoadFlags flags);
+    void load_repo_modules();
```
j-mracek, line 377: Do we want to split `load_available_repo` and `load_repo_modules`? Is there any advantage for that approach?

j-mracek:
It looks like that the build fail is not related to PR. (unused variable in progress bar but only in rawhide -> guessing a change in compilator)

include/libdnf/repo/repo.hpp
pkratoch: I changed it.

pkratoch:
There are some failing tests because of failure to load a repository. I think this might be because the `modules.yaml` is not loaded correctly, so I am setting "blocked" and I'll investigate.

pkratoch:
> There are some failing tests because of failure to load a repository. I think this might be because the `modules.yaml` is not loaded correctly, so I am setting "blocked" and I'll investigate.

The problem is with `modules.yaml.gz`. I made some changes, but it still doesn't work, even though I pass `use_solv_xfopen=true` when opening the `modules.yaml` and it should be able to handle `.gz`. I get `filesystem error: error retrieving file position: Bad file descriptor` when trying to read it and I don't see what's wrong.

j-mracek:
LGTM

j-mracek:
But clang has a problem - Could you modify the include order in `libdnf/repo/repo.cpp`

pkratoch:
> But clang has a problem - Could you modify the include order in `libdnf/repo/repo.cpp`

Sure, I forgot to check paths that don't have "module" in them. :)

PR #45, opened 2022-09-16
Update spec to release in Fedora

inknos:


dnf5.spec
```diff
@@ -6,14 +6,15 @@ Name:           dnf5
...
-License:        GPLv2+
+License:        GPL-2.0-or-later
 URL:            https://github.com/rpm-software-management/dnf5
 Source0:        %{url}/archive/%{version}/dnf5-%{version}.tar.gz
 
 Requires:       libdnf5%{?_isa} = %{version}-%{release}
 Requires:       dnf-data
 Recommends:     bash-completion
 BuildRequires:  bash-completion
+BuildRequires:  pkgconfig(smartcols)
```
j-mracek, line 17: Why did you move smart calls requires? Somehow I think that they are required for build of libdnf5 CLI part and also DNF5.

j-mracek: Please can you describe more information about changes in commit message? Mostly why changed spec is a better option.

inknos: > Please can you describe more information about changes in commit message? Mostly why changed spec is a better option. Like why LGPL-2.1-or-later is a better taq

I am adding more comments to the commit message.

> Why did you move smart calls requires? Somehow I think that they are required for build of libdnf5 CLI part and also DNF5.

I forgot to put the PR back to work in progress, it is not actually finished. Likewise, I actually missed a `%if %{with libdnf_cli}`. The reason is that `BuildRequires` “should” be in the main section. This is a preference of many fedora packagers as far as I understood, but I am considering if it would work for us, it's a bi change. I am actually thinking to move it back to the original place.

Actually, I would ask an opinion to the team in today's meeting.

inknos: As discussed, I grouped `BuildRequires` at the beginning of the specfile.

j-mracek:
LGTM

PR #48, opened 2022-09-19
Add setter for allow_erasing to Goal

j-mracek:
It is more transparent to have a setter then a bool argument for resolve. The setter does not accept the value because it prevents unwanted reset of already set value.

### DIFF
--- a/dnf5/commands/remove/remove.cpp
+++ b/dnf5/commands/remove/remove.cpp
@@ -64,6 +64,8 @@ void RemoveCommand::run() {
         auto option = dynamic_cast<libdnf::OptionString *>(pattern.get());
         goal->add_rpm_remove(option->get_value());
     }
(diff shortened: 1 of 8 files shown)


jrohel:
With this API change, in the future it will not be possible to run 2 resolvs in parallel (in threads) with different allowerasing settings over one Goal.
On the other hand, the question is whether such parallelism will ever be supported by libsolv.

j-mracek:
> With this API change, in the future it will not be possible to run 2 resolvs in parallel (in threads) with different allowerasing settings over one Goal. On the other hand, the question is whether such parallelism will ever be supported by libsolv.

Correct, but I am not sure whether it is a problem or not. I do not remember any plan related to such a functionality.

jrohel:
I'm thinking about the future.
But, I'll merge it.

PR #49, opened 2022-09-20
dnf5: Add option "--allowerasing"

jrohel:
The option was added to commands for which it makes sense.
This was a global option in dnf4, though it was ignored for many commands.


j-mracek:
LGTM, please feel free to merge it when all tests pass.

PR #51, opened 2022-09-21
dnf5: Create MemoryBufferLogger before Context 

jrohel:
We want to capture all messages from the context in the log.

### DIFF
--- a/libdnf/base/base.cpp
+++ b/libdnf/base/base.cpp
@@ -39,8 +39,9 @@ namespace libdnf {
 static std::atomic<Base *> locked_base{nullptr};
 static std::mutex locked_base_mutex;
 
-Base::Base()
-    : repo_sack(get_weak_ptr()),
+Base::Base(std::vector<std::unique_ptr<Logger>> && loggers)
+    : log_router(std::move(loggers)),
+      repo_sack(get_weak_ptr()),
       rpm_package_sack(get_weak_ptr()),
       transaction_history(get_weak_ptr()),
       vars(get_weak_ptr()),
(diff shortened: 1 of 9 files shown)


jrohel:
I pushed the commits again. The only change is that I added doc strings for the Base and Context constructors.

libdnf/base/base.cpp
```diff
@@ -48,6 +48,16 @@ Base::Base()
...
+Base::Base(std::vector<std::unique_ptr<Logger>> && loggers)
+    : log_router(std::move(loggers)),
+      repo_sack(get_weak_ptr()),
+      rpm_package_sack(get_weak_ptr()),
+      transaction_history(get_weak_ptr()),
+      vars(get_weak_ptr()),
+      p_impl(new Impl(get_weak_ptr())) {
+    load_defaults();
+}
+
```
jan-kolarik, line 60: I was thinking if using the constructors delegation wouldn't be better (also on other places in the code, not only this commit). Like having fully defined the constructor with largest number of parameters and other ones delegate to this one, f.e. in this case we could have just something like this:

`Base::Base()
    : Base(std::vector<std::unique_ptr<Logger>>()) {}`

So we avoid the redundancy in the code. What do you think about that?

jan-kolarik: Also we could use just a single constructor with a default value in this case:

`Base(std::vector<std::unique_ptr<Logger>> && loggers = {});`

jrohel: Removing the parameterless constructor will change the public ABI.  But we can still do it.  I will delete it.

jan-kolarik (commented):
LGTM, thanks.

PR #52, opened 2022-09-21
libdnf/utils/locker: Add support for read/write locking

lukash:
Just a small commit that adds read/write locking support to the `Locker` class I did in preparation for locking the system state.


Conan-Kudo (approved):

m-blaha (approved):
Thanks!

PR #53, opened 2022-09-21
Print reason in transaction info, use reason from system state for Replaced packages

lukash:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1161


j-mracek:
LGTM

PR #54, opened 2022-09-21
Base: Hide "pool" from public

jrohel:
The "get_pool(const libdnf::BaseWeakPtr & base)" function is only intended for internal use within the libdnf library. That's why the Base class didn't have a public "get_pool()" method. Which didn't solve anything. The "get_pool" is still on the public API, only it is not a method of the Base class, but a function in the "libdnf" namespace. And disallowing the export of this function for other languages in SWIG is just a crutch. It does not address the C++ API.

This PR moves "pool" to p_impl and removes the "get_pool" function from the public API.


j-mracek:
@jrohel May I ask you to rebase the code?

j-mracek:
LGTM - there was not related fail with tests therefore I rerun them.

PR #55, opened 2022-09-22
dnf5 specfile fedora review

inknos:


bindings/ruby/CMakeLists.txt
```diff
@@ -16,7 +16,9 @@ function(add_ruby_module LIBRARY_NAME MODULE_NAME)
     set_property(SOURCE ../../${LIBRARY_NAME}/${MODULE_NAME}.i PROPERTY SWIG_MODULE_NAME ${MODULE_NAME})
     set(CMAKE_SWIG_FLAGS -module ${MODULE_NAME} -ruby)
     swig_add_library(${TARGET_NAME} LANGUAGE ruby SOURCES ../../${LIBRARY_NAME}/${MODULE_NAME}.i)
+    set(RUBY_SO_VERSION 1)
```
j-mracek, line 19: I am not sure whether to define multiple variable for so name with value 1 is a good idea. In case that we will change so name it will require to modify multiple variables. What about to use one variable for version of all so names?

j-mracek: Or at least for all bindings to use one variable

inknos: I moved the so variable to the CMakeFile in the parent dir

j-mracek (approved):
LGTM

PR #58, opened 2022-09-23
Make sure to set installroot when Base is used in tests

lukash:
Otherwise, it works with host root files, which influence the tests.


m-blaha (approved):
Thanks for fixing this! Last week I also ran into this issue.

PR #63, opened 2022-09-27
Module state - read from /etc/dnf/modules.d

m-blaha:


include/libdnf/module/module_sack.hpp
```diff
@@ -67,6 +67,19 @@ class ModuleSack {
     /// @since 5.0
     std::vector<std::string> get_default_profiles(std::string module_name, std::string module_stream);
 
+    /// @return `true` the module stream is enabled on the system, `false` otherwise.
+    /// @param name Name of the module
+    /// @param stream Name of the module stream
+    /// @since 5.0
+    // @replaces libdnf:ModulePackageContainer.hpp:method:ModulePackageContainer.isEnabled()
+    bool is_enabled(const std::string & name, const std::string & stream) const;
```
j-mracek, line 75: @m-blaha @pkratoch Do you think that we need `in_enabled` and `is_disabled` on two placec - `ModuleItem` and `ModuleSack`? Or would be better to have more general method - get_state() - But one item can have multiple states at the same time - enabled, default, active, disabled ...

libdnf/system/state.hpp
```diff
@@ -190,6 +190,10 @@ class State {
     /// @since 5.0
     void save();
 
+    // TODO(mblaha) - needed to override modules state from dnf4 persistor
+    // remove the method from API once not needed
+    void reset_module_states(std::map<std::string, ModuleState> new_states) { module_states = new_states; }
```
j-mracek, line 195: Please could you add a comment what it does or when it is supposed to be used?

include/libdnf/module/module_sack.hpp
pkratoch: We agreed that the `ModuleState` enum (AVAILABLE, DEFAULT, ENABLED, DISABLED) consists of mutually exclusive states and "active" will be entirely different. So, I think it would be better to have a method `get_state()` and then separate method `is_active()`. But maybe it might still make sense to ask if a whole module is disabled and that's something that would have to be in `ModuleSack`.

libdnf/module/module_sack.cpp
```diff
@@ -326,6 +328,27 @@ void ModuleSack::Impl::make_provides_ready() {
 }
 
 
+bool ModuleSack::is_enabled(const std::string & name, const std::string & stream) const {
+    try {
+        auto & system_state = get_base()->p_impl->get_system_state();
+        return system_state.get_module_state(name) == module::ModuleState::ENABLED &&
```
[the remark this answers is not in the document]
m-blaha: Yeah, this is quite common behavior of system state - when you ask for the state of something not present, `StateNotFoundError` exception is raised.

include/libdnf/module/module_sack.hpp
m-blaha: My original intention was to have it only as a `ModuleItem` method. But then I realized then in dnf4 there is use case where only name and stream is known (https://github.com/rpm-software-management/libdnf/blob/891380467c2b9bcee0f52371ba78e4bfb1017b9e/libdnf/module/ModulePackageContainer.cpp#L1826) . I'm not sure whether this is also gonna be present in dnf5 implementation, though.

j-mracek: Marek, thank you for finding the example when search for name string is required. It means lets keep it in ModuleSack

j-mracek: From performance perspective - for module processing also get_state - make sense.

m-blaha: How about removing those two commits with `is_enabled`/`is_disabled` methods from PR?  They are just examples of modules state usage. The crucial part here is overriding system state from `/etc/dnf/modules.d` directory which I need for future work.

pkratoch: I see. And is there a reason to have `is_enabled()` if there is `get_state()`? Now that I think about it, it would make sense to me if there were just both `get_state(const std::string & name)` and `get_state(const std::string & name, const std::string & stream)`. I don't know where and how it's used, but maybe there could be just these `get_state()` methods until there is a reason for adding `is_enabled()`, `is_disabled()`?

pkratoch: Sorry, I didn't notice your comment until I sent mine. Sounds good to me to remove them until they are needed.

libdnf/system/state.hpp
m-blaha: OK, I've documented the method and moved it into private.

include/libdnf/module/module_sack.hpp
j-mracek: I also agree about removal.

m-blaha: it's gone already :)

libdnf/utils/dnf4convert/dnf4convert.cpp
```diff
@@ -0,0 +1,68 @@
...
+            libdnf::system::ModuleState state;
+            state.enabled_stream = module_config.stream.get_value();
+            state.installed_profiles = module_config.profiles.get_value();
+            state.state = libdnf::module::ModuleState::AVAILABLE;
+            auto status_value = module_config.state.get_value();
+            if (status_value == "enabled") {
+                state.state = libdnf::module::ModuleState::ENABLED;
+            } else if (status_value == "disabled") {
+                state.state = libdnf::module::ModuleState::DISABLED;
+            }
```
j-mracek, line 60: I guess that when value is something else then the code will ignore it silently. I remember than empty value can be there - valid value, but another string?

m-blaha: The `state` in in `ConfigModule` an OptionEnum with allowed values `"", "enabled", "disabled". So in case of invalid value, the empty string is returned by default (and a warning about not allowed value of state option is written. Such module is then considered `AVAILABLE`.

j-mracek: It is handled by config parser

PR #67, opened 2022-09-29
Add modular filtering

j-mracek:


libdnf/module/module_sack_impl.hpp
```diff
@@ -34,6 +34,10 @@ class ModuleSack::Impl {
     Impl(const BaseWeakPtr & base) : base(base), pool(pool_create()) {}
     ~Impl() { pool_free(pool); }
 
+    const std::vector<std::unique_ptr<ModuleItem>> & get_modules();
+
+    void module_filterring();
```
pkratoch, line 42: There should be only one "r" in the word "filtering".

j-mracek: Thank you

j-mracek: Fixed

libdnf/module/module_sack.cpp
```diff
@@ -135,7 +140,6 @@ void ModuleSack::create_module_solvables() {
     // dnf_sack_set_considered_to_update(moduleSack);
 }
 
-
```
pkratoch, line 138: Why is this removed?

libdnf/module/module_sack_impl.hpp
```diff
@@ -36,6 +37,25 @@ class ModuleSack::Impl {
 
     const std::vector<std::unique_ptr<ModuleItem>> & get_modules();
 
+    /// Apply modular filtering to package set. For proper functionality it requires loaded of all repositories and
+    /// resolve of active modules (modular solver)
```
pkratoch, line 41: For proper functionality it requires loaded of all repositories  and resolve of active modules (modular solver)
-> For proper functionality, all repositories must be loaded and active modules must be resolved (modular solver).

libdnf/module/module_sack_impl.hpp
```diff
@@ -36,6 +37,25 @@ class ModuleSack::Impl {
...
+    void module_filtering();
+
+    /// Supporting methods that iterates over all modules and creates filtering sets for modular filterring
```
pkratoch, line 44: Supporting methods -> Supporting method
filterring -> filtering

libdnf/module/module_sack.cpp
```diff
@@ -145,6 +149,150 @@ BaseWeakPtr ModuleSack::get_base() const {
...
 
+std::tuple<
+    std::vector<std::string>,
+    std::vector<std::string>,
+    std::vector<std::string>,
+    std::vector<std::string>,
+    rpm::ReldepList>
+ModuleSack::Impl::collect_data_for_modular_filtering() {
+    base->get_module_sack()->add_modules_without_static_context();
+    auto & allPackages = get_modules();
```
pkratoch, line 160: Since we renamed `ModulePackage` to `ModuleItem`, can we not call this `allPackages`, but rather e.g. `all_module_items`? (Also, use snake_case for variables.) The same for the `packageID` variable.

libdnf/module/module_sack.cpp
```diff
@@ -145,6 +149,150 @@ BaseWeakPtr ModuleSack::get_base() const {
...
+            //         bool found = false;
+            //         for ( auto & demodularized : it->second) {
+            //             if (nevra.get_name() == demodularized) {
+            //                 found = true;
+            //                 break;
+            //             }
+            //         }
+            //         if (found) {
+            //             continue;
+            //         }
```
pkratoch, line 211: Could it be simplified like this or would it be less effective?
```
//     if (std::find(it->second.begin(), it->second.end(), nevra.get_name()) != it->second.end()) {
//         continue;
//     }
```
Also, this whole "else" branch is almost the same as the "if" branch, but I suppose it would be slower if the check was inside, hence the redundancy, right?

libdnf/module/module_sack.cpp
```diff
@@ -145,6 +149,150 @@ BaseWeakPtr ModuleSack::get_base() const {
...
+                for (const auto & rpm : artifacts) {
+                    auto nevras = rpm::Nevra::parse(rpm, {rpm::Nevra::Form::NEVRA});
+                    if (nevras.empty()) {
+                        // TODO(jmracek) Unparsable NEVRA - What to do?
+                        continue;
+                    }
+                    auto & nevra = *nevras.begin();
+                    auto arch = nevra.get_arch();
+                    if (arch == "src" || arch == "nosrc") {
+                        src_names.push_back(nevras.begin()->get_name());
```
pkratoch, line 187: `nevras.begin()->get_name()` can be replaced with `nevra.get_name()`.

libdnf/module/module_sack.cpp
```diff
@@ -145,6 +149,150 @@ BaseWeakPtr ModuleSack::get_base() const {
...
+    exclude_src_names_query.filter_name(src_names);
+    exclude_src_names_query.filter_arch({"src", "nosrc"});
+
+    // Required to filtrate out source packages and packages with incompatible architectures.
+    exclude_names_query.filter_name(names);
+
+    // unifying with exclude_src_names_query will remove additional removal of included packages. Remove included
+    // modular packages to not exclude modular packages from active modules
+    exclude_names_query.update(exclude_src_names_query);
+    exclude_names_query.difference(include_query);
```
pkratoch, line 285: I don't understand these comments. I thought the `names` are excluded, so that we exclude all packages that have the same name as the modular packages from active modules (but different from the modular packages from active modules obviously).
And why do we distinguish between `names` and `src_names` when we then exclude them both? Is it used somewhere else? If so, it deserves to be commented.

libdnf/module/module_sack.cpp
j-mracek: No idea, may be it is a consequence of code modification that was removed

libdnf/module/module_sack.cpp
j-mracek: Thanks

libdnf/module/module_sack.cpp
j-mracek: Thank you for note. I would prefer to modify the code during implementation of demodularization. Is it OK for you?

pkratoch: Sure, I was just wondering about it as I was trying to understand what's going on there.

libdnf/module/module_sack.cpp
j-mracek: I provided additional text above. Hopefully it will be more explanatory.

libdnf/module/module_sack.cpp
j-mracek: fixed

PR #68, opened 2022-09-29
Fix rpmlint dbus-policy-allow-without-destination errors

m-blaha:
Let's use simple best-practice policy file from dbus documentation.


inknos (approved):

inknos:
LGTM

PR #70, opened 2022-09-29
remove Provides for perl

inknos:
Provides are auto-generated
https://docs.fedoraproject.org/en-US/packaging-guidelines/Perl/#_build_dependencies

Testing on copr...

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -265,7 +265,6 @@ Development files for libdnf5-cli.
 %package -n perl5-libdnf5
 Summary:        Perl 5 bindings for the libdnf library
 License:        LGPL-2.1-or-later
-Provides:       perl(libdnf) = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


dnf5.spec
```diff
@@ -265,7 +265,7 @@ Development files for libdnf5-cli.
 %package -n perl5-libdnf5
 Summary:        Perl 5 bindings for the libdnf library
 License:        LGPL-2.1-or-later
-Provides:       perl(libdnf) = %{version}-%{release}
+# Provides:       perl(libdnf) = %{version}-%{release}
```
j-mracek, line 268: Why did you only commented out the provide? May be it would be nice to know why you keep it if it is auto-generated.

inknos: My mistake, I was locally testing. now the review passed and this is RFR

j-mracek:
LGTM, please feel free to merge it

j-mracek (approved):

PR #74, opened 2022-10-03
Write output of RPM scriptlets to logger

jrohel:
The output of the RPM scriptlets was  written to the "scriptlet.out" file in the current working directory and was owerwritten every time dnf5 was run.
Now there is a thread that processes output from the RPM scriptlets and writes it to the logger.

Logging of RPM transaction callbacks was also added. It is possible to read from the log which scriptlet the given output is from.


jan-kolarik:
Gone through the code and tested locally, everything seems fine. Thanks!

PR #75, opened 2022-10-03
Revert so versioning for bindings and fix perl subpackage

inknos:
https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2120661


j-mracek:
LGTM

PR #76, opened 2022-10-03
[dnf5] Add version "1.0" to cfg file for aliases of command line arguments

jrohel:


dnf5/cmdline_aliases.cpp
```diff
@@ -99,8 +99,38 @@ void load_aliases_from_toml_file(Context & context, const fs::path & config_file
         const auto arg_parser_elements =
             toml::parse<::toml::discard_comments, libdnf::PreserveOrderMap, std::vector>(config_file_path);
 
+        try {
+            const auto & version = toml::find(arg_parser_elements, "version");
+            const std::string version_str = version.as_string();
+            if (version_str != "1.0") {
```
m-blaha, line 105: What do you think about defining a constant with supported aliases.toml version? I'd prefer it over `"1.0"` literal here.
Not imporant right now (but TODO is nice): once we extend the aliases.toml format and bump the version to 1.1, library should still be able to parse 1.0 toml files. In libdnf/system/state.cpp there is some code to parse versions which can be used also here. Simple string comparison doesn't allow such support (e.g. support any toml with major version == 1 and minor version <= 2).
But this is not priority now (until we bump the version).

m-blaha: Also this load_aliases_from_toml_file() function is quite long (ca 550 lines) and over-complicated so it should probably be refactored and split to several sub-functions. But again - not now, maybe just add TODO?

jrohel: @m-blaha 
- I defined the supported version as a constant.
- Once we extend the aliases.toml format and bump the version to 1.1, than we can add parsing. Currently only one version "1.0" is supported. As you wrote, parsing the version is not a priority now.

PR #77, opened 2022-10-04
Do not read "/etc/yum/vars", "/etc/yum/protected.d/", "/etc/yum/repos.d" directories

jrohel:


jan-kolarik:
Great, thanks for the fix.

PR #78, opened 2022-10-04
Add "get_user_state_dir" func,  move non-root user log file to "~/.local/state/" ($XDG_STATE_HOME) 

jrohel:
It is a question of where to save the logger files in XDG. Basically, there are 2 options - STATE category and DATA category.
Following this proposal https://wiki.debian.org/XDGBaseDirectorySpecification I moved the logger file to the STATE category.


jan-kolarik:
I also agree with the `$XDG_STATE_HOME` location, it seems to be recommended also from other sources (e.g. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

j-mracek:
LGTM

PR #81, opened 2022-10-06
goal: Reason change correctly returns resolve problems

m-blaha:
In case the spec fails to resolve to a package,
add_reason_change_to_goal() method returns problem. As a result also `dnf5 mark` command returns exit code 1 in case the specified package does not exist.


jan-kolarik:
LGTM, tested locally. Should've noticed that within my previous review :frowning_face:

PR #83, opened 2022-10-10
Remove set_reason() from transaction public API

jan-kolarik:
Method `set_reason()` was removed from the public API and also a visibility of related members was fixed to (hopefully) originally planned level. Due to these changes also some header includes reordering and other small refactoring was necessary.

### DIFF
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -288,12 +288,14 @@ void Transaction::Impl::set_transaction(rpm::solv::GoalPrivate & solved_goal, Go
 
     // Add groups to the transaction
     for (auto & [group, action, reason] : solved_goal.list_groups()) {
-        groups.emplace_back(group, action, reason);
+        TransactionGroup tsgrp(group, action, reason);
+        groups.emplace_back(std::move(tsgrp));
     }
 
     // Add reason change actions to the transaction
     for (auto & [pkg, reason, group_id] : solved_goal.list_reason_changes()) {
-        packages.emplace_back(pkg, TransactionPackage::Action::REASON_CHANGE, reason, group_id);
+        TransactionPackage tspkg(pkg, TransactionPackage::Action::REASON_CHANGE, reason, group_id);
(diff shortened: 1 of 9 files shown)


libdnf/base/transaction.cpp
```diff
@@ -288,12 +288,14 @@ void Transaction::Impl::set_transaction(rpm::solv::GoalPrivate & solved_goal, Go
 
     // Add groups to the transaction
     for (auto & [group, action, reason] : solved_goal.list_groups()) {
-        groups.emplace_back(group, action, reason);
+        TransactionGroup tsgrp(group, action, reason);
```
jan-kolarik, line 291: Because this ctor is not public anymore and it is a friend only in the `Transaction::Impl` scope.

libdnf/base/transaction.cpp
```diff
@@ -288,12 +288,14 @@ void Transaction::Impl::set_transaction(rpm::solv::GoalPrivate & solved_goal, Go
...
+        groups.emplace_back(std::move(tsgrp));
     }
 
     // Add reason change actions to the transaction
     for (auto & [pkg, reason, group_id] : solved_goal.list_reason_changes()) {
-        packages.emplace_back(pkg, TransactionPackage::Action::REASON_CHANGE, reason, group_id);
+        TransactionPackage tspkg(pkg, TransactionPackage::Action::REASON_CHANGE, reason, group_id);
```
jan-kolarik, line 297: Same as above.

kontura (commented):
Yea, it looks like it was supposed to be `private:` :+1: 

Just a suggestion but I would prefer if this was split into multiple commits.

Also could you add a link to the related JIRA?

jan-kolarik:
To be honest the actual moving of the `set_reason()` into the `private` section is connected with these changes and they cannot be delivered on they own. Or do I understand your intention incorrectly?

kontura (approved):

kontura:
The failing CI is unrelated.
Thanks!

PR #86, opened 2022-10-11
state: Store and load system repo groups metadata

m-blaha:
Comps metadata for each installed group is now stored in the system state and loaded into libsolv pool from there.

### DIFF
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -594,6 +594,8 @@ Transaction::TransactionRunResult Transaction::Impl::run(
         }
 
         // Set correct system state for groups in the transaction
+        auto comps_xml_dir = system_state.get_group_xml_dir();
+        std::filesystem::create_directories(comps_xml_dir);
         for (const auto & tsgroup : groups) {
             auto group = tsgroup.get_group();
             if (transaction_item_action_is_inbound(tsgroup.get_action())) {
@@ -616,6 +618,8 @@ Transaction::TransactionRunResult Transaction::Impl::run(
                     }
                 }
                 system_state.set_group_state(group.get_groupid(), state);
+                // save the current xml group definition
+                group.dump(comps_xml_dir / (group.get_groupid() + ".xml"));
             }
         }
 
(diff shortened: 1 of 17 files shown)


libdnf/repo/solv_repo.cpp
```diff
@@ -277,6 +278,49 @@ void SolvRepo::load_repo_main(const std::string & repomd_fn, const std::string &
...
+            auto & system_state = base->p_impl->get_system_state();
+            auto comps_dir = system_state.get_group_xml_dir();
+            auto installed_groups = system_state.installed_groups();
+            for (const auto & group_id : installed_groups) {
+                auto ext_fn = comps_dir / (group_id + ".xml");
+                fs::File ext_file;
+                try {
+                    ext_file = fs::File(ext_fn, "r", true);
+                } catch (std::filesystem::filesystem_error & e) {
+                    logger.debug(
```
jrohel, line 300: Is this really a "debug" message? Is it logging something that is standard behavior?  It looks more like a warning. We want to warn the user that a loading error has occurred.

libdnf/system/state.cpp
```diff
@@ -315,6 +315,15 @@ std::set<std::string> State::get_package_groups(const std::string & name) {
     }
 }
 
+std::vector<std::string> State::installed_groups() {
```
jan-kolarik, line 318: I would prefer ``get_installed_groups`` here.

jan-kolarik (changes requested):
I left some comments above, they are addressing some minor issues from my point of view. Also a rebase is needed now. Otherwise LGTM.

libdnf/base/transaction.cpp
```diff
@@ -616,6 +618,8 @@ Transaction::TransactionRunResult Transaction::Impl::run(
                     }
                 }
                 system_state.set_group_state(group.get_groupid(), state);
+                // save the current xml group definition
+                group.dump(comps_xml_dir / (group.get_groupid() + ".xml"));
```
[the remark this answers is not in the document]
m-blaha: I remember that python json and xml libraries use `dump()` method name for printing the content of element to string / file. But I like your suggestion and I've added a commit for renaming the methods.

libdnf/system/state.cpp
m-blaha: Renamed.

PR #87, opened 2022-10-12
modules: Add module defaults and make_provides_ready

pkratoch:


libdnf/module/module_metadata.cpp
```diff
@@ -167,4 +167,43 @@ std::pair<std::vector<ModuleItem *>, std::vector<ModuleItem *>> ModuleMetadata::
 }
 
 
+std::map<std::string, std::string> ModuleMetadata::get_default_streams() {
+    std::map<std::string, std::string> default_streams;
+    if (!resulting_module_index) {
```
j-mracek, line 172: Please could you add a comment why resulting_module_index can be absent or something else?

j-mracek: I guess that it should be only absent if there are no modules. But I guess that it can also happened that module defaults were not resolved.

libdnf/module/module_sack.cpp
```diff
@@ -149,6 +155,28 @@ BaseWeakPtr ModuleSack::get_base() const {
     return p_impl->base;
 }
 
+
+void ModuleSack::Impl::module_defaults_resolve() {
+    module_metadata.resolve_added_metadata();
+    module_defaults = module_metadata.get_default_streams();
```
j-mracek, line 161: This is an suggestion - what about to call `resolve_added_metadata()` at beginning of `get_default_streams()`?

libdnf/module/module_sack.cpp
```diff
@@ -149,6 +155,28 @@ BaseWeakPtr ModuleSack::get_base() const {
...
+}
+
+
+const std::string & ModuleSack::get_default_stream(const std::string & name) const {
+    p_impl->module_defaults_resolve();
```
j-mracek, line 166: This might be a performance problem. For each request the current workflow will go to libmodulemd and requests the resolve defaults for all modules and creates a new object in C++ that is used for name search later on (but only once). May be I overlooked something. If the issue is there anyway we do not need to address it right now but it is possible to create a TODO.

j-mracek:
I finished the review and there is only a request for comments and TODO.

libdnf/module/module_sack.cpp
```diff
@@ -295,6 +316,26 @@ void ModuleSack::Impl::module_filtering() {
     // dnf_sack_set_module_includes(sack, includeQuery.getResultPset());*/
 }
 
+
+void ModuleSack::Impl::make_provides_ready() {
```
j-mracek, line 320: I have one additional suggestion related to make provides ready:
I recommend to add in `module_sack.cpp` two additional lines:
In `void ModuleSack::add()` and `void ModuleSack::Impl::add_modules_without_static_context()` to set `provides_ready = false` because additional data will also require to make provide ready.

libdnf/module/module_sack_impl.hpp
```diff
@@ -89,6 +89,7 @@ class ModuleSack::Impl {
 
 inline const std::vector<std::unique_ptr<ModuleItem>> & ModuleSack::Impl::get_modules() {
     add_modules_without_static_context();
+    create_module_solvables();
```
j-mracek, line 92: I am sorry I've got a feeling that it will results that solvables will be created redundantly. I guess we need to distinguish whether solvable was already created or not. May be it is possible to use the same trick like with modules without static context. I propose to create additional tmp vector then the workflow will be as follows:
```modules_without_static -> modules_without_solvable -> fully_initialized_modules```
Modules with static context will go directly to modules_without_solvable.

The second approach could be without tmp vector.
Or solvable can be created directly when static_context is preset. For items without static context, solvable can be created directly when static context is calculated

pkratoch: Thanks. The additional vector has a problem that when determining the static contexts, I'd have to go through both `fully_initialized_modules` and `modules_without_solvable`, when seemed a bit more complicated. So, I found it better to create the solvables when modules are added to the `modules` vector.

j-mracek (approved):
LGTM, feel free to merge it when all tests pass

PR #89, opened 2022-10-12
Add resolve, alldeps to download

inknos:
tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1167

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -43,6 +43,28 @@ void DownloadCommand::set_argument_parser() {
         patterns_to_download_options);
     keys->set_description("List of keys to match");
     keys->set_complete_hook_func([&ctx](const char * arg) { return match_specs(ctx, arg, false, true, false, false); });
+
+    resolve_option = dynamic_cast<libdnf::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
+
+    alldeps_option = dynamic_cast<libdnf::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
+
+    auto resolve = parser.add_new_named_arg("resolve");
+    resolve->set_long_name("resolve");
+    resolve->set_description("Resolve and download needed dependencies");
+    resolve->set_const_value("true");
+    resolve->link_value(resolve_option);
+
+    auto alldeps = parser.add_new_named_arg("alldeps");
+    alldeps->set_long_name("alldeps");
+    alldeps->set_description(
+        "When running with --resolve, download all dependencies (do not exclude already installed ones)");
+    alldeps->set_const_value("true");
+    alldeps->link_value(alldeps_option);
+
+    cmd.register_named_arg(resolve);
+    cmd.register_named_arg(alldeps);
     cmd.register_positional_arg(keys);
 }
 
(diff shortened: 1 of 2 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -71,13 +110,29 @@ void DownloadCommand::run() {
         package_query.resolve_pkg_spec(option->get_value(), {}, true);
         result_pset |= package_query;
     }
+    if (resolve_option->get_value()) {
+        for (auto pkg : result_pset) {
+            goal->add_rpm_install(pkg, settings);
+        }
+        auto transaction = goal->resolve(false);
```
jrohel, line 117: `goal->resolve(false);` -> `goal->resolve();`

dnf5/commands/download/download.cpp
```diff
@@ -56,13 +78,30 @@ void DownloadCommand::configure() {
...
+    if (transaction_problems != libdnf::GoalProblem::NO_PROBLEM) {
+        if (transaction_problems != libdnf::GoalProblem::NOT_FOUND) {
+            throw GoalResolveError(transaction);
+        }
+    }
+}
+
 void DownloadCommand::run() {
     auto & ctx = get_context();
+    auto goal = ctx.get_goal();
```
jrohel, line 101: `auto goal = ctx.get_goal();` -> `auto goal = std::make_unique<libdnf::Goal>(ctx.base);`

We don't want to create a transaction in the context. Instead of `goal` in the context, we will create a local one. We only need local to resolve what to download.
And move this line to the condition `if (resolve_option->get_value()) {`. We don't need a `goal` elsewhere.
And also remove `goal->reset()`. After these changes, it will not be needed because the destructor will be called immediately.

jrohel (changes requested):
* Please use the current API.
* And fix the code logic.

I described the necessary modifications in the review comments.
After making these fixes, the `download` command will still not be complete, but it will be a step forward and I'll mergnu PR.

dnf5/commands/download/download.hpp
```diff
@@ -36,9 +36,13 @@ class DownloadCommand : public Command {
     explicit DownloadCommand(Command & parent) : Command(parent, "download") {}
     void set_argument_parser() override;
     void configure() override;
+    void goal_resolved() override;
```
jrohel, line 39: Do not add this method. It will not be called. Instead, put the code in the `run()` method right after `auto transaction = goal->resolve()`.

dnf5/commands/download/download.cpp
m-blaha: you need to rebase to the newest upstream/main. The `allow_erasing` parameter of `resolve` has been removed.

inknos: Got it, thank you

dnf5/commands/download/download.cpp
```diff
@@ -56,13 +78,19 @@ void DownloadCommand::configure() {
...
+        context.set_load_system_repo(true);
+    } else {
+        context.set_load_system_repo(false);
+    }
     context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
 }
 
 void DownloadCommand::run() {
     auto & ctx = get_context();
+    auto settings = libdnf::GoalJobSettings();
```
jrohel, line 91: * Please move this line to the condition `if (resolve_option->get_value()) {`. It is not used elsewhere.
* C++ is very flexible. Here we are creating an instance of an object on the stack. We can replace this line with `libdnf::GoalJobSettings settings;`. Which I prefer.

An example with a `std::string` object: `auto s = std::string();` can be written as `auto s = std::string{};` and can be replaced with `std::string s;`.

Example with primitive type `int`: `auto a = int();` can be written as `auto a = int{};` and is the same as `int a{};`.

jrohel:
Thank you for the changes. And sorry, I have one more request.

jrohel:
Thanks.

PR #91, opened 2022-10-13
mark: Enhance transaction table and fix changing the reason to current value

m-blaha:


libdnf/base/goal.cpp
```diff
@@ -1202,6 +1202,17 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
     }
     for (const auto & pkg : query) {
+        if (pkg.get_reason() == reason) {
+            // pkg is already installed with correct reason
+            transaction.p_impl->add_resolve_log(
+                GoalAction::REASON_CHANGE,
+                GoalProblem::ALREADY_INSTALLED,
+                settings,
+                pkg.get_nevra(),
+                {pkg.get_nevra()},
```
jrohel, line 1212: How about instead of putting the `nevra` twice in the message, we put the `reason` there.
`{pkg.get_nevra()},` -> `{libdnf::transaction::transaction_item_reason_to_string(reason)},`
and adjust log_event appropriately

libdnf/base/log_event.cpp
```diff
@@ -121,7 +121,12 @@ std::string LogEvent::to_string(
             if (additional_data->size() != 1) {
                 throw std::invalid_argument("Incorrect number of elements for ALREADY_INSTALLED");
             }
-            return ret.append(utils::sformat(_("Package \"{}\" is already installed."), *additional_data->begin()));
+            if (action == GoalAction::REASON_CHANGE) {
+                return ret.append(utils::sformat(
+                    _("Package \"{}\" is already installed with correct reason."), *additional_data->begin()));
```
jrohel, line 126: That's nice. I suggest to improve the report with information about `reason`. A minor modification needs to be made to "goal.cpp", which I also suggested.

`_("Package \"{}\" is already installed with correct reason."), *additional_data->begin()));` -> `_("Package \"{}\" is already installed with reason \"{}\"."), *spec, *additional_data->begin()));`

jrohel (changes requested):
* Nice. I've suggested another improvement to the problem reporting.
* Cannot compile dnf5daemon-client . I think the `get_reason()` method for `DbusPackageWrapper` is missing.

m-blaha:
@jrohel problem reporting has been improved according to your suggestion and I've fixed the compilation problem

PR #92, opened 2022-10-14
libdnf: handle unknown events in the rpm transaction callback

pmatilai:
The repackage callbacks are 15 years gone upstream, they are best treated as unknown events. As for RPMCALLBACK_UNKNOWN, this is a nice little rpm-side WTF: an enum whose value is 0 isn't exactly unknown, so the value doesn't make any sense whatsoever.

Replace these with a default case to warn on all actually unknown events.


jrohel:
Thanks for your PR and clarifying the situation of the obsolete RPM callback types.

PR #94, opened 2022-10-17
Fix test for arch not x86_64

inknos:


m-blaha:
The change seems to me over-complicated. Wouldn't be setting the `x86_64` arch for the base instance configuration enough? (btw that's how the problem was solved in libdnf4).
```
diff --git a/test/libdnf/base_test_case.cpp b/test/libdnf/base_test_case.cpp
index b8518d17..b1e33d9c 100644
--- a/test/libdnf/base_test_case.cpp
+++ b/test/libdnf/base_test_case.cpp
@@ -224,6 +224,8 @@ void BaseTestCase::setUp() {
     base.get_config().installroot().set(libdnf::Option::Priority::RUNTIME, temp->get_path() / "installroot");
     base.get_config().cachedir().set(libdnf::Option::Priority::RUNTIME, temp->get_path() / "cache");
 
+    base.get_vars()->set("arch", "x86_64");
+
     base.setup();
 
     repo_sack = base.get_repo_sack();
```

inknos:
Moved to https://github.com/rpm-software-management/dnf5/pull/95

PR #95, opened 2022-10-17
Force test arch to x86_64

inknos:
Simpler solution than: https://github.com/rpm-software-management/dnf5/pull/94


inknos:
@m-blaha once one of these build pass, I would say it's good to go :)

https://copr.fedorainfracloud.org/coprs/nsella/dnf-test-arch/builds/

m-blaha (approved):

PR #100, opened 2022-10-20
[libdnf] Remove "console_output" table from history database

jrohel:
The "console_output" table is used to store the output of RPM scriptlets in dnf4. This is done after the transaction has ended and the outputs from all scriptlets are concatenated without information about their source and timing. And if the transaction fails, the output of the scriptlets is not saved to the database at all.
We usually need scriptlet output when something goes wrong. And it's useful to know which scriptlet failed and the context.

In libdnf5, the output of RPM scriptlets is stored in the main log file (commit https://github.com/rpm-software-management/dnf5/commit/4873bfec2ebc305bc180a4f8539c33418f319695). The log will also contain other messages - contextual information. It is clear when and which scriptlet generated the given output. In some cases, the scriptlet runs for a long time and prints multiple messages. The time of printing the messages is stored in the log. This is additional information for profiling scriptlets.


j-mracek (approved):
LGTM

PR #101, opened 2022-10-20
Generate initial system state from dnf4 history

m-blaha:


libdnf/system/state.hpp
```diff
@@ -103,6 +103,10 @@ class State {
     /// @since 5.0
     State(const std::filesystem::path & path);
 
+    /// @return True if the State instance does not contain information about packages
+    /// @since 5.0
```
j-mracek, line 114: Is it a test to trigger conversion? May I ask you add a purpose of the method to the description?

j-mracek: If it is a trigger then what about to detect the situation by presence of system state files?

libdnf/system/state.hpp
```diff
@@ -165,6 +169,22 @@ class State {
...
 
+    /// @return The the state for environmental group
+    /// @param id The environmental group id to get the state for.
+    /// @since 5.0
+    EnvironmentState get_environment_state(const std::string & id);
+
+    /// Sets the state for an environmental group id.
+    /// @param id The environmental group id to set the state for.
+    /// @since 5.0
+    void set_environment_state(const std::string & id, const EnvironmentState & environment_state);
```
j-mracek, line 182: May be I will need an update what is the purpose or mechanism of EnvironmentState.

libdnf/system/state.hpp
```diff
@@ -223,6 +225,17 @@ class State {
...
+    /// Reset packages system state to match given values.
+    /// @param installed_packages Vector of tuples <rpm::Nevra nevra, TransactionItemReason reason, std::string repository_id> of currently installed packages
+    /// @param installed_groups Vector of tuples <std::string group_id, TransactionItemReason reason, std::set<std::string> installed_packages> of currently installed groups
+    /// @param installed_environments Vector of tuples <std::string environment_id, std::set<std::string> installed_groups> of currently installed environmental groups
+    /// @since 5.0
+    void reset_packages_states(
+        std::map<std::string, libdnf::system::PackageState> package_states,
+        std::map<std::string, libdnf::system::NevraState> nevra_states,
+        std::map<std::string, libdnf::system::GroupState> group_states,
+        std::map<std::string, libdnf::system::EnvironmentState> environment_states);
```
j-mracek, line 237: I am not sure whether it is intended. Argument `package_states` creates a copy of map and then move it to `this->package_states = std::move(package_states);`. In the last commit there is -
```
            system_state.reset_packages_states(
                std::move(package_states),
                std::move(nevra_states),
                std::move(group_states),
                std::move(environment_states));
        }
```
I suggest that you meant to use `std::map<std::string, libdnf::system::PackageState> && package_states`. Then the code call can be without std::move()
```
            system_state.reset_packages_states(
                package_states,
                nevra_states,
                group_states,
                environment_states);
        }
```
May be it would be also nice to state that methods take ownership of arguments.

m-blaha: You cannot use rvalue reference argument and than use the function with lvalue. But I've changed the `reset_packages_states()` signature to use rvalue references.

j-mracek (approved):
LGTM

PR #103, opened 2022-10-21
[dnf5] Fix: Set dnf5 exit code to ERROR in exception handler

jrohel:


m-blaha:
Do we actually need this outer try / catch block over whole `main()` function? There are two other try/catch blocks inside the function covering argument parsing and actual command running.

jrohel:
@m-blaha Exceptions before parsing the command line are not covered by existing try/catch blocks inside the "main" function.
I can remove the outer try/catch and add a new one inside the "main" function.

m-blaha:
We agreed to merge this change - nested try/catch blocks were already present even before this PR and now missing return value problem is fixed.

PR #97, opened 2022-10-18
Make transaction constructors and setters private

kontura:

PR #106, opened 2022-10-25
Make private not used parts of transaction API

kontura:
This is a follow up for: https://github.com/rpm-software-management/dnf5/pull/97
It uses the same approach to make API private and test it using the `create_getter()` hack.


j-mracek:
LGTM

PR #107, opened 2022-10-26
Initial version of the dnf5 man pages

jan-kolarik:
Created initial version of new `dnf5` man pages. Most of the content is based on the `dnf` man pages.

The main intent was to make the first page shorter, so the user won't be overwhelmed with information at the first look. Each command has its own man page. Also some concepts and terms referenced within man pages were moved into separate pages and described here in more detail.

Commands that are not completed are skipped for now. Also many `#TODO`s are included, so the work is not forgotten in the future.

Tested was both `man` and `HTML` output.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -184,6 +184,33 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %license COPYING.md
 %license gpl-2.0.txt
 %{_mandir}/man8/dnf5.8.*
+%{_mandir}/man8/dnf5-advisory.8.*
+%{_mandir}/man8/dnf5-clean.8.*
+%{_mandir}/man8/dnf5-distro-sync.8.*
+%{_mandir}/man8/dnf5-downgrade.8.*
+%{_mandir}/man8/dnf5-download.8.*
+%{_mandir}/man8/dnf5-environment.8.*
+%{_mandir}/man8/dnf5-group.8.*
+# TODO(jkolarik): history is not ready yet
+# %{_mandir}/man8/dnf5-history.8.*
+%{_mandir}/man8/dnf5-install.8.*
+%{_mandir}/man8/dnf5-makecache.8.*
+%{_mandir}/man8/dnf5-mark.8.*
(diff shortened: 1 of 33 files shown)


doc/commands/history.8.rst
```diff
@@ -0,0 +1,83 @@
...
+========
+
+``dnf5 history <subcommand> [options] [<transaction-spec>]``
+
+
+Description
+===========
+
+The ``history`` command in ``DNF5`` allows the user to view what has happened in past transactions
+and offers several operations above these transactions, like undoing and redoing them. In order to
```
kontura, line 38: The "above" doesn't sound right to me here.
I might be wrong but I would prefer "with" or "on".

doc/commands/repoquery.8.rst
```diff
@@ -0,0 +1,108 @@
...
+
+``--info``
+    | Use the verbose format for the output.
+
+
+Examples
+========
+
+``dnf5 repoquery /etc/koji.conf``
+    | List packages which provides the given file.
```
kontura, line 94: provides -> provide

dnf5.spec
```diff
@@ -184,6 +184,33 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
...
 %{_mandir}/man8/dnf5.8.*
+%{_mandir}/man8/dnf5-advisory.8.*
+%{_mandir}/man8/dnf5-clean.8.*
+%{_mandir}/man8/dnf5-distro-sync.8.*
+%{_mandir}/man8/dnf5-downgrade.8.*
+%{_mandir}/man8/dnf5-download.8.*
+%{_mandir}/man8/dnf5-environment.8.*
+%{_mandir}/man8/dnf5-group.8.*
+# TODO(jkolarik): history is not ready yet
+# %{_mandir}/man8/dnf5-history.8.*
```
m-blaha, line 195: These lines make rpm complain about macro expansion in comment. But we can live with it for a while, I guess.

m-blaha (approved):

PR #110, opened 2022-10-27
Fix Fedora rawhide builds

m-blaha:
There are several issues with Fedora rawhide builds that we've started to see recently. This should hopefully fix them.


kontura:
Thanks!

PR #113, opened 2022-10-31
Prevent macro expansion in spec comments

inknos:
Macros should not expand in comments

https://rpm-software-management.github.io/rpm/manual/spec.html

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -192,12 +192,12 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %{_mandir}/man8/dnf5-environment.8.*
 %{_mandir}/man8/dnf5-group.8.*
 # TODO(jkolarik): history is not ready yet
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Thanks for the fix :slightly_smiling_face:

j-mracek:
LGTM

pmatilai:
This is the traditional way, works and is okay etc. Just FWIW, these days (since 4.15) rpm supports true commenting that doesn't need escaping macros with %dnl, eg:

> -# %{_mandir}/man8/dnf5-history.8.*
+%dnl %{_mandir}/man8/dnf5-history.8.*

The benefit being, it's all in once place and you don't need to muck with the content being commented out, as there may be multiple macros on the line etc.

jan-kolarik:
@pmatilai Thanks for the tip. Next time we'll hopefully remember and use this better way. Now I am merging this as the comments here are to be removed soon.

PR #114, opened 2022-11-01
Fix a couple of problems detected by sanitizers

kontura:


include/libdnf/base/base.hpp
```diff
@@ -123,6 +123,10 @@ class Base {
     /// The files in the directory are read in alphabetical order.
     void load_config_from_dir();
 
+    // Impl has to be the first data member because it contains Pool and that has be destructed last.
+    // See commit: https://github.com/rpm-software-management/dnf5/commit/c8e26cb545aed0d6ca66545d51eda7568efdf232
+    ImplPtr<Impl> p_impl;
```
jrohel, line 128: But `p_impl` is not the first datamember. There is a `base_guard` before. But `base_guard` is used in Impl constructor. So `base_guard` must be before `p_impl`.

jrohel: When I think about it. That's not nice. We can move `base_guard` to `Impl`. Or add a comment that `base_guard` must be the first data member and `p_impl` the second.

kontura: I would prefer to move `base_guard` into the `p_impl` but I didn't manage to make it work without bigger changes.
Therefore I have just updated the comment.

PR #121, opened 2022-11-07
Add obsoletes of microdnf

j-mracek:
DNF5 will replace microdnf. To provide an upgrade path we need to add obsoletes, provide and symlink to ensure replacement and microdnf and compatibility. When user will try to install microdnf, dnf5 will be automatically installed.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -14,6 +14,12 @@ Requires:       libdnf5%{?_isa} = %{version}-%{release}
 Requires:       dnf-data
 Recommends:     bash-completion
 
+# Remove if condition when Fedora 37 is EOL
+%if 0%{?fedora} > 37
+Provides:       microdnf = %{version}-%{release}
+Obsoletes:      microdnf < 4
+%endif
+
 # ========== build options ==========
 
 %bcond_without dnf5daemon_client
@@ -170,6 +176,12 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 
 %files
 %{_bindir}/dnf5
+
+# Remove if condition when Fedora 37 is EOL
+%if 0%{?fedora} > 37
+%{_bindir}/microdnf
+%endif
+
 %{_prefix}/share/dnf5
 %dir %{_sysconfdir}/dnf/dnf5-aliases.d
 %doc %{_sysconfdir}/dnf/dnf5-aliases.d/README
@@ -587,6 +599,10 @@ Core DNF5 plugins that enhance dnf5 with builddep and changelog commands.
 
 #find_lang {name}
 
+# Remove if condition when Fedora 37 is EOL
+%if 0%{?fedora} > 37
+ln -sr %{buildroot}%{_bindir}/dnf5 %{buildroot}%{_bindir}/microdnf
+%endif
 
 %ldconfig_scriptlets
 


j-mracek:
The PR is ready for review but we have to decide when we will ship the change to Fedora 38.

dnf5.spec
```diff
@@ -13,6 +13,8 @@ Source0:        %{url}/archive/%{version}/dnf5-%{version}.tar.gz
 Requires:       libdnf5%{?_isa} = %{version}-%{release}
 Requires:       dnf-data
 Recommends:     bash-completion
+Provides:       microdnf = %{version}-%{release}
+Obsoletes:      microdnf < 4
```
Conan-Kudo, line 17: I think I'd prefer having it as a subpackage instead of a plain `Obsoletes`+`Provides`, like we did for `yum`.

Conan-Kudo (changes requested):

dnf5.spec
j-mracek: Well, dnf4 was and is packaged by multiple ways - Fedora and RHEL  - and the packaging has been modified several times. Do you know that original yum (Fedora 22) required DNF? 

The differences between distributions resulted in mess. DNF5 is replacing microdnf and in near future DNF. It means that we have to repeat the same trick once again and we can generate one additional sub-package (or to reuse the sub-package providing microdnf). I know that this is logical and let say a proper solution from packager community, but it will create optional symlink availability (`dnf`, `microdnf`, `dnf5`). I would like to ensure that all systems with DNF5 will have all symlinks available (`dnf`, `dnf5`, `microdnf`) in default deployment. The decision of symlink available is not exactly user or upstream decision but by distribution - in minimal images no one will install additional package to provide additional symlinks of DNF5.
The same feature could be also achieved by requirement of that new package by dnf5 (and requirement in the new package for DNF5), but then creation of such a package will not make any sense.

Summary - unifying of symling availability (`dnf`, `dnf5`, `microdnf`) is a feature that I would like to delivery to all systems where DNF5 will be present and start to conflict with microdnf and in future with dnf. It is a chap way how to make things working everywhere (hard coded compatibility by upstream and downstream that cannot be messed up by particular deployment) and users will experience less issues.

@Conan-Kudo Please could you provide some reasons why do you think this is not a good idea, what benefits the splitting will provide or so on (facts are critical to understand your preferences)?

j-mracek: Ok, it would mean to name the new sub-package `microdnf`, that somehow we did not plan, because in repository we would like to keep the original `microdnf` package. I afraid that there is no win-win solution.

Conan-Kudo: Rename original microdnf to `microdnf3`.

j-mracek:
I am suggesting to unblock the PR when https://github.com/rpm-software-management/dnf5/pull/243 will be merged

dnf5.spec
m-blaha: @Conan-Kudo Do you have a particular application that uses this check in mind?
Just thinking whether the failing check isn't actually a good thing. The new `dnf5` is highly compatible with old `microdnf` but still   - it's a different application and there are some differences (even on the command line). So maybe applications should take this into account.

Conan-Kudo: Mostly config management tools that try to identify which package manager to use.

j-mracek: Well if they don't want to make difference between `dnf` and `dnf5` then they could use `rpm -q --whatprovides /usr/bin/dnf`. Be honest we cannot make a DNF5 deployment without changes in other packages because DNF5 is not DNF.

Conan-Kudo: Honestly, I'd rather just straight up retire legacy microdnf instead. We don't need both.

m-blaha: We sure will retire old microdnf completely. But for F38 we want users to have at least possibility to install it in case the dnf5 does not work for them (for any reason).

PR #123, opened 2022-11-07
Configuration for loading repository metadata parts

jan-kolarik:
Created `optional_metadata_types` configuration option to define loading specific repository metadata parts. The "optional" word is used as the `primary` and `modules` metadata are always loaded as they are essential.

Each command or client could then also pass additional required types using the newly added `add` and `add_item` methods for configuration list options. They are used for adding a range of values or a single value to the list.


dnf5/context.cpp
```diff
@@ -210,11 +210,12 @@ void Context::print_info(const char * msg) {
 
 
 void Context::update_repo_load_flags_from_specs(const std::vector<std::string> & pkg_specs) {
+    // TODO(jkolarik): unify this behavior with internal libdnf detection of file specs
     // if a pkg_spec contains '/', it's a path and we need to load filelists
     for (auto & spec : pkg_specs) {
         if (spec.find("/") != std::string::npos) {
```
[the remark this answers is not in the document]
jan-kolarik: Yes I am planning to unify the behavior in later PR. Precisely replacing this with the `libdnf::utils::is_file_pattern` you mentioned. I will try to put it also at the right place in the API then.

jan-kolarik:
CI tests will be added after PR review will be done.

dnf5/context.cpp
```diff
@@ -210,17 +210,27 @@ void Context::print_info(const char * msg) {
...
-            available_repos_load_flags = available_repos_load_flags | libdnf::repo::LoadFlags::FILELISTS;
-            break;
+            add_repo_load_flags(libdnf::repo::LoadFlags::FILELISTS);
```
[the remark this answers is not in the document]
j-mracek: I am not sure whether `--setopt=metadata_types=filelists` will remove all other types of metadata from configuration.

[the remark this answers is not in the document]
jan-kolarik: Oh, I didn't realize the `Context` class is in the API. In that case I need to think about it again, because as you wrote there is the ambiguity present. 

So far the `Context` handling was used for commands to put some metadata type to all repo configs which is done there in `Context::load_repos()` using the `RepoQuery`, but it doesn't seem very convenient. Maybe I could create some helper/wrapper method to do this and add it to API, because we can access repo sack through the `Context` API. Then I could drop these `add_repo_load_flags`, `setup_repo_load_flags`, etc. Then really no additional logic dealing with metadata handling will be present.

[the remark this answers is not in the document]
jan-kolarik: > I am not sure whether `--setopt=metadata_types=filelists` will remove all other types of metadata from configuration.

As it is the `append` option it should not and I think I was also testing this manually from CLI, but I will double-check :slightly_smiling_face:

libdnf/conf/config_main.cpp
```diff
@@ -186,6 +187,9 @@ class ConfigMain::Impl {
     OptionBool debug_solver{false};
     OptionStringList installonlypkgs{INSTALLONLYPKGS};
     OptionStringList group_package_types{GROUP_PACKAGE_TYPES};
+    OptionStringSet metadata_types{flags_to_strings(
+        libdnf::repo::LoadFlags::PRIMARY | libdnf::repo::LoadFlags::MODULES | libdnf::repo::LoadFlags::COMPS |
```
j-mracek, line 191: I believe that `libdnf::repo::LoadFlags::PRIMARY | libdnf::repo::LoadFlags::MODULES` should be hardcoded types - not in option.

j-mracek:
I am not sure about `libdnf::repo::LoadFlags` enum. I like it how you defined the convertor and so on, but on other hand I have a feeling that it is an extra layer that have minimal benefit in public space (API). In the original design it was OK, but with string configuration in Conf it may be unneeded redundant. What do you think?

dnf5/context.cpp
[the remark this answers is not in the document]
j-mracek: Context part is not yet in API because there is no devel package yet. And thank you verifying `--setopt`.

libdnf/conf/config_main.cpp
jan-kolarik: I used to have these values "hardcoded" when converting the string list in the `LoadFlags::flags_from_strings()`. I can move it there again, but I think that we cannot really prevent user bypassing these metadata types if they really want to. We'll probably have to override getting the values from the `Option`.

dnf5/context.cpp
```diff
@@ -210,17 +210,27 @@ void Context::print_info(const char * msg) {
...
 
 
-void Context::load_repos(bool load_system, libdnf::repo::LoadFlags flags) {
+void Context::setup_repo_load_flags(const libdnf::repo::RepoWeakPtr & repo) {
+    auto & metadata_types_option = repo->get_config().metadata_types();
+    auto new_flags = flags_to_strings(repo_load_flags);
+    auto configured_flags = metadata_types_option.get_value();
+    new_flags.insert(configured_flags.cbegin(), configured_flags.cend());
+    metadata_types_option.set(libdnf::Option::Priority::RUNTIME, new_flags);
+}
```
j-mracek, line 230: I like the deduplication, but I am not sure whether this is the right place. If configuration Class for metadata_types will handle it it may be make more sense to me.

libdnf/conf/config_main.cpp
jan-kolarik: Oh but when I thinking about it now, we don't have any conditional loading for `PRIMARY` and `MODULES` now. It is always loaded, the `MODULES` could be bypassed only with compile flag. So I think we could just omit these here if it will be more transparent?

jan-kolarik:
> I am not sure about `libdnf::repo::LoadFlags` enum. I like it how you defined the convertor and so on, but on other hand I have a feeling that it is an extra layer that have minimal benefit in public space (API). In the original design it was OK, but with string configuration in Conf it may be unneeded redundant. What do you think?

What about to define string constants for each metadata type and we could pass/compare the known values if we want to. Similar approach is already used for metadata filenames in `repo_downloader.hpp`.

dnf5/context.cpp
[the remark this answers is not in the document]
jan-kolarik: I realized that the appending behavior is currently supported only for CLI bindings. When we want to append a single value to the list from the configuration option I don't see any direct support for this. So this is also why I am handling that in the `Context`. 

Perhaps we could add a support for adding a single value to the OptionStringList instances, but then what about handling the priority?

libdnf/conf/config_main.cpp
j-mracek: his is a good question. We can keep those value there but we have to document that removal of those values has no effect on workflow. But I am not sure what would be best. The requirements is to document the behavior in documentation.

jan-kolarik: I think the best place would be the documentation for configuration options which is not present yet. But I've added a comment about this into our internal JIRA task.

jan-kolarik:
I prepared a next iteration of this PR considering all reported feedback and also updated the description to match the current version.

dnf5/context.cpp
```diff
@@ -209,18 +209,41 @@ void Context::print_info(const char * msg) {
 }
 
 
-void Context::update_repo_load_flags_from_specs(const std::vector<std::string> & pkg_specs) {
+void Context::add_repo_metadata(const std::string & type) {
+    libdnf::repo::RepoQuery repos(base);
+    repos.filter_enabled(true);
```
j-mracek, line 214: I would prefer to modify also disabled repository. If some plugin modify the enabled set after adding load flag it can end up with inconsistency. Proposing to remove line ` repos.filter_enabled(true);`

dnf5/context.cpp
```diff
@@ -209,18 +209,41 @@ void Context::print_info(const char * msg) {
...
+
+    for (const auto & repo : repos) {
+        repo->get_config().metadata_types().add_item(libdnf::Option::Priority::RUNTIME, type);
+    }
+}
+
+
+void Context::add_all_repo_metadata() {
+    libdnf::repo::RepoQuery repos(base);
+    repos.filter_enabled(true);
```
j-mracek, line 225: I would prefer to modify also disabled repository. If some plugin modify the enabled set after adding load flag it can end up with inconsistency. Proposing to remove line `repos.filter_enabled(true);`

libdnf/conf/option_string_list.cpp
```diff
@@ -103,11 +133,24 @@ void OptionStringList::set(Priority priority, const ValueType & value) {
...
-std::string OptionStringList::to_string(const ValueType & value) const {
+template <typename T>
+void OptionStringContainer<T>::add_item(Priority priority, const std::string & item) {
+    assert_not_locked();
+
+    if (priority >= get_priority()) {
+        test_item(item);
+        value.insert(value.end(), item);
+        set_priority(priority);
+    }
```
j-mracek, line 149: I am not sure whether test of priority for append option make sense. Normally priority is used that it ensure consistent result regardless of order. In append option it will somehow mess up with results. Proposing to test priority only for `set_priority(priority);`. What do you think?

```
    test_item(item);
    value.insert(value.end(), item);
    if (priority >= get_priority()) {
        set_priority(priority);
    }
```

jan-kolarik: I see. Now I am thinking about it and the `add_item` is available only through the API = therefore always `Runtime`, right? What about to drop the priority parameter for `add_item` and always set the `Runtime`?

j-mracek:
LGTM

PR #124, opened 2022-11-08
Unify matching file patterns

jan-kolarik:
Move the function for detecting file patterns to the API and use it consistently across the codebase. Also other functions for the `utils_internal.hpp` were moved to the API as I think we can consider them generic.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -37,6 +37,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf/rpm/package_query.hpp>
 #include <libdnf/rpm/package_set.hpp>
 #include <libdnf/rpm/rpm_signature.hpp>
+#include <libdnf/utils/patterns.hpp>
 
 #include <algorithm>
 #include <cctype>
@@ -239,12 +240,10 @@ void Context::print_info(const char * msg) {
 
 
 void Context::update_repo_metadata_from_specs(const std::vector<std::string> & pkg_specs) {
-    // TODO(jkolarik): unify this behavior with internal libdnf detection of file specs
-    // if a pkg_spec contains '/', it's a path and we need to load filelists
     for (auto & spec : pkg_specs) {
-        if (spec.find("/") != std::string::npos) {
+        if (libdnf::utils::is_file_pattern(spec)) {
             base.get_config().optional_metadata_types().add_item(libdnf::METADATA_TYPE_FILELISTS);
-            break;
+            return;
         }
     }
 }
@@ -950,8 +949,8 @@ std::vector<std::string> match_specs(
     base.load_config_from_file();
     base.setup();
 
-    // optimization - if pattern contain '/', disable the search for matching installed and available packages
-    if (pattern.find('/') != std::string::npos) {
+    // optimization - disable the search for matching installed and available packages for file patterns
+    if (libdnf::utils::is_file_pattern(pattern)) {
         installed = available = false;
     }
 
--- a/dnf5/include/dnf5/context.hpp
+++ b/dnf5/include/dnf5/context.hpp
@@ -159,7 +159,7 @@ private:
     Actions action;
 };
 
-/// Downoad packages to destdir. If destdir == nullptr, packages are downloaded to the cache.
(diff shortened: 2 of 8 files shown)


j-mracek:
It looks like that PR contains other commits from another work. Please could you @jan-kolarik take a look?

jan-kolarik:
> It looks like that PR contains other commits from another work. Please could you @jan-kolarik take a look?

Yes, this is why I add the `blocked` label. I will rebase this after merging the dependent PR mentioned in the description.

include/libdnf/utils/convert.hpp
```diff
@@ -17,29 +17,19 @@ You should have received a copy of the GNU Lesser General Public License
...
-    return strpbrk(pattern, "*[?") != nullptr;
-}
-
-/// @brief Test if pattern is file path
-/// Return true if pattern start with "/" or pattern[0] == '*' && pattern[1] == '/'
-static inline bool is_file_pattern(const std::string & pattern) {
-    return pattern[0] == '/' || (pattern[0] == '*' && pattern[1] == '/');
-}
-
+/// @brief Convert an input text to a lowercase version.
```
j-mracek, line 29: I am really sorry, but I don't understand why the file was renamed and moved to the public part? It looks like that it is used by package_query.cpp or did I overlooked something?

dnf5/context.cpp
```diff
@@ -950,8 +949,8 @@ std::vector<std::string> match_specs(
     base.load_config_from_file();
     base.setup();
 
-    // optimization - if pattern contain '/', disable the search for matching installed and available packages
-    if (pattern.find('/') != std::string::npos) {
+    // optimization - disable the search for matching installed and available packages for file patterns
+    if (libdnf::utils::is_file_pattern(pattern)) {
```
j-mracek, line 953: I thing the change is correct, but the description of function is not sufficient. I got wrong impression what it does. Internally it creates another instance of Base() and then it suggest argument name for auto-compleation.

include/libdnf/utils/convert.hpp
jan-kolarik: Basically I wanted to move the `is_file_pattern` and other functions that have the global meaning to the API. Here it was in the `utils_internal.hpp` which by the name tells it is not the API. So I moved the functions and tried to group them to files by their purpose - `convert.hpp` and `patterns.hpp`

dnf5/context.cpp
jan-kolarik: I just changed the comment for the `is_file_pattern` call here to match the current reference. But you probably mean the description of the `match_specs` function? Should I updated the doc string for it in the `context.hpp`?

j-mracek: I am sorry, I wasn't explicit. I meant the description of `match_specs`. If someone will use that method for something else then callback of an argument parser => it will be not nice.

j-mracek: If you can also modify `match_specs` description it would be great.

include/libdnf/utils/convert.hpp
jan-kolarik: I've updated the docstring in the `context.hpp`. But anyway for me it seems this is not the right place for this function. It is sticked to the context file, but is static and related only to the argument parser...

dnf5/context.cpp
jan-kolarik: I've updated the docstring in the context.hpp. But anyway for me it seems this is not the right place for this function. It is sticked to the context file, but is static and related only to the argument parser...

include/libdnf/utils/convert.hpp
j-mracek: I am still not sure here whether I understand it correctly. I propose to make move `convert.hpp` to private part -> `libdnf/utils/convert.hpp`, because I don't see any usage in public space. Or did I overlooked something?

j-mracek: Please check the end of files, there is a missing newline and I believe that it is part of the standard.

jan-kolarik: I see now what was the main concern about. I moved the `convert.hpp` into the private part as you're right there are no API usages (yet). And sorry for the confusion I probably somehow thought you are addressing the `is_file_pattern` function in the original comment.

PR #125, opened 2022-11-08
Improve handling of rich dependencies

j-mracek:
It also requires test in our CI for following arguments
install '(available_package if not_installed)' => expected nothing to do
install '(available_package if package_installed)' => expected install available_package
install '(available_package if not_installed)' not_installed => expected install available_package and `not_installed`

### DIFF
--- a/dnf5-plugins/builddep_plugin/builddep.cpp
+++ b/dnf5-plugins/builddep_plugin/builddep.cpp
@@ -252,8 +252,7 @@ void BuildDepCommand::run() {
     // fill the goal with build dependencies
     auto goal = get_context().get_goal();
     for (const auto & spec : install_specs) {
-        if (spec[0] == '(') {
-            // rich dependencies require different method
+        if (libdnf::rpm::Reldep::is_rich_dependency(spec)) {
             goal->add_provide_install(spec);
         } else {
             goal->add_rpm_install(spec);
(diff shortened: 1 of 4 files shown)


libdnf/base/goal.cpp
```diff
@@ -463,6 +463,15 @@ std::pair<GoalProblem, libdnf::solv::IdQueue> Goal::Impl::add_install_to_goal(
             return {GoalProblem::NO_PROBLEM, result_queue};
         }
     }
+
+    // Correct evaluation of reach deps can be only performed by solver.
```
kontura, line 467: reach -> rich

libdnf/rpm/reldep.cpp
```diff
@@ -85,7 +85,7 @@ ReldepId Reldep::get_reldep_id(
 }
 
 ReldepId Reldep::get_reldep_id(const BaseWeakPtr & base, const std::string & reldep_str, int create) {
-    if (reldep_str[0] == '(') {
+    if (is_rich_dependency(reldep_str)) {
         // Rich dependency
```
jan-kolarik, line 89: I would remove this line now, because the function name `is_rich_dependency` is quite self-explaining.

dnf5-plugins/builddep_plugin/builddep.cpp
```diff
@@ -252,7 +252,7 @@ void BuildDepCommand::run() {
     // fill the goal with build dependencies
     auto goal = get_context().get_goal();
     for (const auto & spec : install_specs) {
-        if (spec[0] == '(') {
+        if (libdnf::rpm::Reldep::is_rich_dependency(spec)) {
             // rich dependencies require different method
```
jan-kolarik, line 256: Maybe we could consider deleting this comment now...

PR #128, opened 2022-11-10
Extend LogEvent with new attribute spec_type

m-blaha:
To correctly print LogEvent messages we need to distinguish what object type the `spec` attibute was matched against.


j-mracek:
LGTM

Issue #122, opened 2022-11-07
Implement additional repoquery options


kontura:
The [dnf4 repoquery command](https://dnf.readthedocs.io/en/latest/command_ref.html#repoquery-command-label) has a lot of options.

There are two categories:
### Select Options
_Together with the \<package-file-spec\> argument, control what packages are displayed in the output._
- [x] \<package-file-spec\>
- [x] -a, --all
- - Proposing to drop it (document as a change). It is a default dnf behavior and it was added as compatibility option with yum

- [x] --arch / --archlist
- - Please change other command that uses the same o similar options (even option can be the same, but schematic of arguments can be different) in DNF and unify it
- [x] --duplicates
- [x] #454
- [x] --available
- [x] --disable-modular-filtering
- [x] --extras
- [x] -f \<file\>, --file \<file\>
- [x] --installed
- [x] --installonly
- [x] --latest-limit \<number\>
- [x] --recent
- - Priority low
- [x] --repo \<repoid\>
- - It is a general option. Investigate why it is explicitly mentioned in repoquery
- - [amatej] The best I can tell the option got imported together with repoquery command into dnf. Perhaps before when repoquery was separate the option was needed.
- [x] --upgrades
- [x] --userinstalled
- [x] --whatdepends \<capability\>[,\<capability\>...]
- [x] --whatconflicts \<capability\>[,\<capability\>...]
- [x] --whatenhances \<capability\>[,\<capability\>...]
- [x] --whatobsoletes \<capability\>[,\<capability\>...]
- [x] --whatprovides \<capability\>[,\<capability\>...]
- [x] --whatrecommends \<capability\>[,\<capability\>...]
- [x] --whatrequires \<capability\>[,\<capability\>...]
- [x] --whatsuggests \<capability\>[,\<capability\>...]
- [x] --whatsupplements \<capability\>[,\<capability\>...]
- [x] --alldeps
- - It is a default behavior of DNF therefore I propose to remove it (document the change). The reason of presence of the option -> the default was changed in history
- [x] --exactdeps
- [x] --srpm
- - Please verify that other commands use --srpm, or --source using the same principle

### Formatting Options
_Set what information is displayed about each package._
- [x] #479
- [x] -l, --list
- - Proposing to rename to more appropriate --files
- [x] -s, --source
- - Please verify that other commands use --srpm, or --source using the same principle
- [x] --changelogs
- [x] --conflicts
- [x] --depends
- [x] --enhances
- [x] --location
[...]

PR #129, opened 2022-11-10
Repoquery: Add --what* and --exactdeps options

kontura:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1177


kontura:
yes, I am working on it.
It is really weird because it passes for me locally on all fedoras but it fails in COPR.

kontura:
@j-mracek I have resolved the issue.

kontura:
Rebased to resolve conflicts.

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -150,6 +227,22 @@ void RepoqueryCommand::load_additional_packages() {
     }
 }
 
+// In case of input being nevras -> resolve them to packages
+static libdnf::rpm::PackageQuery resolve_nevras_to_packges(
```
jrohel, line 231: Do we need `PackageQuery`? Or is `PackageSet` enough?
When I look below, the result is used as a `PackageSet`.

Maybe we can use `PackageSet`:
```
static libdnf::rpm::PackageSet resolve_nevras_to_packges(
    libdnf::Base & base, const std::vector<std::string> & nevra_globs, const libdnf::rpm::PackageQuery & base_query) {
    auto resolved_nevras = libdnf::rpm::PackageSet(base);
    ...
```

jrohel (changes requested):
Thanks for the PR. I added comments.

common/utils/string.hpp
```diff
@@ -78,6 +78,19 @@ inline std::string join(const ContainerT & input, const std::string & delimiter)
 }
 
 
+/// Trim leading and trailing spaces from string `str`.
+inline std::string trim(std::string str) {
```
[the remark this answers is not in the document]
kontura: Ok, I have changed my `trim` to work on the string in place.
I think we can add `trim_copy` once its needed.

libdnf/conf/option_string_list.cpp
```diff
@@ -28,26 +29,34 @@ namespace libdnf {
...
       default_value(default_value),
       value(default_value) {}
 
-OptionStringList::OptionStringList(const ValueType & default_value, std::string regex, bool icase)
+OptionStringList::OptionStringList(
+    const ValueType & default_value, std::string regex, bool icase, const char * delimiters)
     : Option(Priority::DEFAULT),
       regex(std::move(regex)),
       icase(icase),
+      delimiters(delimiters),
```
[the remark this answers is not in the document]
kontura: Ok I have changed it to use the same approach as `OptionBool` regarding the default delimiters.

In repoquery command I have also used `OptionStringList` directly without `StringListOption` from `session.cpp` as I didn't want to add custom delimiter support there.
In general I am not sure if we need the `StringListOption` wrapper, we use it only sometimes and it could be confusing.

There is still the problem with specifying the options multiple times - it is currently not supported but I think I would prefer to work on that in next PRs.

include/libdnf/conf/option_string_list.hpp
```diff
@@ -83,19 +91,26 @@ class OptionStringContainer : public Option {
...
+    const char * get_delimiters() const noexcept;
+
 protected:
     /// Tests new container item value and throws exception if the item value is not allowed.
     void test_item(const std::string & item) const;
 
     std::optional<std::regex> regex_matcher;
     std::string regex;
     bool icase;
+    std::unique_ptr<const char *> delimiters;
```
jrohel, line 113: Are you using `unique_ptr` to store the raw pointer? The problem remains here. Only the pointer is copied, not the string itself. The raw pointer stored in `unique_ptr` points to the original string.

The solution is to use `std::string` instead of `const char*` to store: `std::unique_ptr<std::string> delimiters;`. 

Since it is a private part, we can even simplify it by using `std::optional<std::string> delimiters;` (`OptionBool` comes from libdnf4, it is in C++11 and `std::optional` is from C++17. Therefore `OptionBool` uses `std::unique_ptr`).

Other lines of code will need to be changed as well.
`return delimiters ? *delimiters : get_default_delimiters();` ->
`return delimiters ? delimiters->c_str() : get_default_delimiters();`

And we can also pass `delimiters` as `std::string` in the constructor. Example of solution:
```
template <typename T>
OptionStringContainer<T>::OptionStringContainer(
    const ValueType & default_value, std::string regex, bool icase, std::string delimiters)
    : Option(Priority::DEFAULT),
      regex(std::move(regex)),
      icase(icase),
      delimiters(std::move(delimiters)),
      default_value(default_value),
      value(this->default_value) {
    test(this->default_value);
}

template <typename T>
OptionStringContainer<T>::OptionStringContainer(const OptionStringContainer & src)
    : Option(src),
      regex(src.regex),
      icase(src.icase),
      delimiters(src.delimiters),
      default_value(src.default_value),
      value(src.value) {}

template <typename T>
OptionStringContainer<T> & OptionStringContainer<T>::operator=(const OptionStringContainer & src) {
    assert_not_locked();

    if (this == &src) {
        return *this;
    }
    regex = src.regex;
    icase = src.icase;
    delimiters = src.delimiters;
    default_value = src.default_value;
    value = src.value;
    return *this;
}
```

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -150,6 +260,22 @@ void RepoqueryCommand::load_additional_packages() {
     }
 }
 
+// In case of input being nevras -> resolve them to packages
+static libdnf::rpm::PackageSet resolve_nevras_to_packges(
+    libdnf::Base & base, const std::vector<std::string> & nevra_globs, const libdnf::rpm::PackageQuery & base_query) {
+    auto resolved_nevras_query = libdnf::rpm::PackageQuery(base, libdnf::sack::ExcludeFlags::APPLY_EXCLUDES, true);
```
jrohel, line 266: When I wrote use `PackageSet`, I really meant use. Not just change the return type of the function. As I wrote in the 3 line sample code. Replace this line with:
`auto resolved_nevras = libdnf::rpm::PackageSet(base);`

include/libdnf/conf/option_string_list.hpp
kontura: ooh, I wasn't thinking, my bad.
Updated again hopefully without any major bugs this time.
Thank you.

dnf5/commands/repoquery/repoquery.cpp
kontura: Right, thank you again.

jrohel (approved):

jrohel:
Thank you for PR.

Issue #72, opened 2022-10-02
Progress bars repeat


GabrieleLabanca:
When running `install`, the progress bars repeat themselves. Example output:
```
>>>sudo dnf5 install fuse
Updating and loading repositories:
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[0/0] Fedora 36 - x86_64 - Updates                                                                                                                                                                    100% |  15.6 KiB/s |  14.0 KiB |  00m01s
[...]

PR #130, opened 2022-11-10
[dnf5] Fix: repo metadata download progress bars - repetition, amount of downloaded data

jrohel:
[dnf5] Fix: repeating repository metadata download progress bar - fixes https://github.com/rpm-software-management/dnf5/issues/72
[dnf5] Fix: amount downloaded data in repo metadata download progressbar


m-blaha:
Nice, thanks!
Only one thing - now when the [0/0] numbers are skipped, also the ci tests have to be adjusted (in `ci-dnf-stack/dnf-behave-tests/common/output.py:strip_reposync_dnf5()`)

dnf5/context.cpp
```diff
@@ -93,27 +93,46 @@ class ProgressAndKeyImportRepoCB : public KeyImportRepoCB {
...
+
+            progress_bar->set_state(libdnf::cli::progressbar::ProgressBarState::SUCCESS);
             print_progress_bar();
         }
     }
 
     int progress([[maybe_unused]] double total_to_download, [[maybe_unused]] double downloaded) override {
-        progress_bar.set_total_ticks(static_cast<int64_t>(total_to_download));
-        progress_bar.set_ticks(static_cast<int64_t>(downloaded));
+        libdnf_assert(progress_bar != nullptr, "Called \"proggress\" callback before \"start\" callback");
```
jan-kolarik, line 123: typo `proggress` -> `progress`

jan-kolarik:
Thanks for that! I noticed one small issue which is not related to your work, but to progress bars in general - that 100% percent widget is shown before metadata even starts to download, so I will create a PR follow-up later. Anyway, it is a nice progress since the last state. Still I think we could improve some things - I will create also another follow-up issue where we could discuss that further.

dnf5/context.cpp
jrohel: Thanks. Fixed

PR #158, opened 2022-11-11
Fix showing full percentage progress before metadata download start

jan-kolarik:
There are some uninteresting progress events with zero values coming at the beginning of the metadata download which causes the percent widget to show `100%` before the actual downloading starts.

As there is already implemented showing the `???%`, I assume this was the original intention, so I fixed the incorrectly initialized members in the `ProgressBar` class.

This is a follow-up of https://github.com/rpm-software-management/dnf5/pull/130


jrohel:
Thanks. I know, it's a little PR. But can you please split it into two commits?
- The first commit then fixes the progress_bar implementation of the "progress_bar.hpp" and "progress_bar/cpp" files.
- The second commit will modify the "context.cpp" file of the dnf5 application.

PR #160, opened 2022-11-14
goal: Filter all_installed for requested names

m-blaha:
We do not want to add all installed packages to upgrade job. Only installed packages with matching names are relevant.

https://bugzilla.redhat.com/show_bug.cgi?id=2142257

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1180

### DIFF
--- a/libdnf/base/goal.cpp
+++ b/libdnf/base/goal.cpp
@@ -1160,9 +1160,23 @@ void Goal::Impl::add_up_down_distrosync_to_goal(
     switch (action) {
         case GoalAction::UPGRADE_MINIMAL:
         case GoalAction::UPGRADE:
-            // For a correct upgrade of installonly packages keep only the latest installed packages
-            // Otherwise it will also install not the latest installonly packages
             query.filter_available();
+            // Given that we use libsolv's targeted transactions, we need to ensure that the transaction contains both
+            // the new targeted version and also the current installed version (for the upgraded package). This is
+            // because if it only contained the new version, libsolv would decide to reinstall the package even if it
+            // had just a different buildtime or vendor but the same version
+            // (https://github.com/openSUSE/libsolv/issues/287)
+            //   - Make sure that query contains both the new and installed versions (i.e. add installed versions)
+            //   - However we need to add installed versions of just the packages that are being upgraded. We don't want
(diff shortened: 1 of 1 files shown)


libdnf/base/goal.cpp
```diff
@@ -1164,6 +1164,7 @@ void Goal::Impl::add_up_down_distrosync_to_goal(
             // Otherwise it will also install not the latest installonly packages
             query.filter_available();
             all_installed.filter_latest_evr();
+            all_installed.filter_name(query);
```
j-mracek, line 1167: I think it is a correct solution. The code here is quite delicate therefore I recommend to enhance the description of the issue in commit message (description of reproducer can also help) plus add a reference to DNF5 bug (plus reference to dnf4 bug). I also suggest to enhance documentation in code. In DNF4, the code was modified just recently including a long description what and why it is there (base line 2171). May I ask you for the description enhancement?

m-blaha: Yeah, sure. Will do.
Also I need to investigate why dnf5 with the issue has been passing ci tests for package upgrades - this should also result in a reproducer. Let me block the PR till I have a CI test for it.

j-mracek: I think I have a reproducer. When I try to upgrade a leave package (dnf5) then it works like expected. When I upgrade package that will require upgrade of another installed package, the it writes nothing to do.

m-blaha: Great, thanks!

m-blaha: Uff, it took me a while but I finally have a reproducer. Not only the upgraded package must have a dependency, but also the installed versions must not be available in repos... I'm about to send a PR with ci test.

j-mracek: Perfect

m-blaha: Actually the absence in repos is not required - best=False is. I should have read the comment in dnf4 beforehand :(
Anyway, the test is ready and I've added required descriptions.

kontura: I am sorry, I should have added the comment here as well.

kontura (approved):

j-mracek:
LGTM

PR #163, opened 2022-11-16
Setup SWIG and Sphinx to generate docs for Python bindings

kontura:
Setup SWIG and Sphinx to generate docs for Python bindings

Our current docs pipeline for c++ is:

    c++ code (with docs annotations)
        |
        ▽
    Doxygen gathers the docs into xml files
        |
        ▽
    Sphinx (using a breathe extension) uses xmls to generate html

This adds a pipeline for Python bindings:

    c++ code (with docs annotations)
        |
        ▽
    Swig has a build in parser for Doxygen type of documentation and it
    automatically inserts it when generating bindings (using -doxygen flag).
        |
        ▽
    Sphinx (using a autodoc extension) generates Python docs directly from
    python bindings.

Another option would be to use Doxygen for Python bindings as well.
However Doxygen isn't intended for Python and it generates the binding
xmls improperly resulting in uglier html docs.

Still it is not a perfect solution, the docs string are meant for C++
and the notation can look out of place in Python docs.
We are limited by what Swig generates so functions with multiple
arguments have a signature of `function_name(*args)` (arguments are
described correctly in the documentation it self).
Swig handles overloads by wrapping them into one function so it is
again only described in the docs for the particular functions (ideally
each overload would have its own function).

Resulting example:
image


gotmax23:
This would be great to have. Even the generated docstrings without the Sphinx docs would be helpful.

j-mracek:
LGTM

PR #165, opened 2022-11-23
Disable irrelevant clang-tidy checks

bkrl:
Previously, the clang-tidy configuration enables all checks and then disables a few undesired ones. This enables a lot of checks that are specific to certain kinds of code and aren't relevant for this project. For example, I noticed several errors from checks specific to LLVM libc. It makes more sense to enable only the checks that we want instead of enabling everything and then disabling the checks that we don't want.


kontura:
Thanks!

PR #166, opened 2022-11-23
PGP: Use new librepo PGP API, remove gpgme dependency

jrohel:
Librepo internally uses gpgme to work with PGP keys. Libdnf used the librepo keyring directly via gpgme instead of using the librepo API. It had to, the librepo API was insufficient.

Librepo in version 1.15.0 extended the PGP API.
This commmit uses the extended librepo PGP API and removes libdnf's dependency on gpgme.

PR also adds support for any number of user IDs in a PGP key.

This PR requires new librepo PGP API PR https://github.com/rpm-software-management/librepo/pull/266 .

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -71,7 +71,7 @@ public:
         }
 
         auto tmp_id = id.size() > 8 ? id.substr(id.size() - 8) : id;
-        std::cout << "Importing GPG key 0x" << id << ":\n";
+        std::cout << "Importing PGP key 0x" << id << ":\n";
         std::cout << " Userid     : \"" << user_id << "\"\n";
         std::cout << " Fingerprint: " << fingerprint << "\n";
         std::cout << " From       : " << url << std::endl;
@@ -670,7 +670,7 @@ private:
 std::chrono::time_point<std::chrono::steady_clock> RpmTransCB::prev_print_time = std::chrono::steady_clock::now();
 
 static bool user_confirm_key(libdnf::ConfigMain & config, const libdnf::rpm::KeyInfo & key_info) {
-    std::cout << "Importing GPG key 0x" << key_info.get_short_key_id() << std::endl;
+    std::cout << "Importing PGP key 0x" << key_info.get_short_key_id() << std::endl;
     std::cout << " UserId     : \"" << key_info.get_user_id() << "\"" << std::endl;
     std::cout << " Fingerprint: " << key_info.get_fingerprint() << std::endl;
     std::cout << " From       : " << key_info.get_url() << std::endl;
@@ -779,7 +779,7 @@ bool Context::check_gpg_signatures(const libdnf::base::Transaction & transaction
                             retval = false;
                             break;
                         case ImportRepoKeys::NO_KEYS:
-                            std::cerr << err_msg << "the repository does not have any gpg key configured." << std::endl;
+                            std::cerr << err_msg << "the repository does not have any pgp key configured." << std::endl;
                             retval = false;
                             break;
                         case ImportRepoKeys::IMPORT_FAILED:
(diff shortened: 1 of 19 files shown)


dnf5/context.cpp
```diff
@@ -799,7 +803,7 @@ bool Context::check_gpg_signatures(const libdnf::base::Transaction & transaction
 void Context::download_and_run(libdnf::base::Transaction & transaction) {
     download_packages(transaction, nullptr);
 
-    std::cout << std::endl << "Verifying GPG signatures" << std::endl;
+    std::cout << std::endl << "Verifying PGP signatures" << std::endl;
```
jan-kolarik, line 806: When changing all output from GPG -> PGP, maybe also the line 748 here in the `context.cpp`?

jan-kolarik:
Since we are starting with the `DNF5`, could we also change other occurrences of GPG -> PGP, also in options, etc.? It is quite confusing. Now it seems consistent on the output, but lot of internals and configurations are still referencing GPGs. Or do we differentitate these for some specific usages in the code?

jan-kolarik:
Also another question related to builds. We don't have the new `librepo` release yet, so merging this will result in package build tests failing until that. Are we OK with that or should we wait for the release?

kontura:
> Also another question related to builds. We don't have the new `librepo` release yet, so merging this will result in package build tests failing until that. Are we OK with that or should we wait for the release?

The CI is taking builds from nightly so it should work after tomorrow (librepo 1.15.0 will be available in [nightly repo](https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/)).
We could even run the workflow manually now if you want.

jan-kolarik:
> > Also another question related to builds. We don't have the new `librepo` release yet, so merging this will result in package build tests failing until that. Are we OK with that or should we wait for the release?
> 
> The CI is taking builds from nightly so it should work after tomorrow (librepo 1.15.0 will be available in [nightly repo](https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/)). We could even run the workflow builds manually now if you want.

Thanks for the info :slightly_smiling_face: If it will be tomorrow OK, then it's fine. I've already tried the manual trigger of the workflow and currently it's now working yet.

pmatilai:
> Since we are starting with the DNF5, could we also change other occurrences of GPG -> PGP, also in options, etc.?

PGP is not really any better than GPG, because PGP is the name of the commercial software. Use OpenPGP when referring to the standard.

jrohel:
@pmatilai 
**PGP** (Pretty Good Privacy) **is a program**.
**GnuPG** (GNU Privacy Guard, also known as **GPG**) **is a program**.
And **OpenPGP is the standard**.

What are the keys called? **OpenPGP keys**? Or **PGP keys**? Or **GPG keys**?

When I go through the OpenPGP standard, the Armor Header Line can contain:
**PGP** MESSAGES; **PGP** PUBLIC KEY BLOCK; **PGP** PRIVATE KEY BLOCK; **PGP** MESSAGE, PART X/Y; **PGP** MESSAGES, PART X; **PGP** SIGNATURE

So I chose PGP. But we can discuss where to change PGP to OpenPGP. Because you're right, PGP is a program. In addition, PGP is also a trademark.

pmatilai:
Everything is OpenPGP, including keys and signatures. The ASCII armor has "PGP" as a remnant from its origins in the PGP program, preserved for compatibility I'd think.

jan-kolarik:
I created a new issue for deciding about the naming problem. Here we could just fix the single mentioned occurrence of the GPG in the output from `context.cpp`, but it could be also deferred.

I will wait for tomorrow when new `librepo` snapshot is ready, check the tests and I am for merging this.

PR #167, opened 2022-11-23
Separate pool for comps solvables

m-blaha:


libdnf/base/base_impl.hpp
```diff
@@ -67,6 +81,7 @@ class Base::Impl {
 class InternalBaseUser {
 public:
     static solv::Pool & get_pool(const libdnf::BaseWeakPtr & base) { return base->p_impl->get_pool(); }
+    static solv::Pool & get_comps_pool(const libdnf::BaseWeakPtr & base) { return base->p_impl->get_comps_pool(); }
```
j-mracek, line 84: In DNF4 we used DnfSack for modules and different instance for RPMs. It created a confusion in code what data I am accessing. What about to use a different Pool class for groups?

m-blaha: Hm, both pools use exactly the same data structure, thus they are instances of the same `solv::Pool` class. I do not like adding some `CompsPool` subclass of `Pool` class without any change.
What is different for both pools is contents - and this is expressed by variable / function name.

libdnf/base/base_impl.hpp
```diff
@@ -56,6 +61,15 @@ class Base::Impl {
 
     // Pool as the owner of underlying libsolv data, has to be the first member so that it is destroyed last.
     std::unique_ptr<solv::Pool> pool;
+    // In libsolv the groups and environmental groups are stored as regular
+    // solvables (just with "group:" / "environment:" prefix in their name).
+    // These group solvables than contain hard requirements for included
```
kontura, line 66: Only a tiny detail, I think it should be: `than` -> `then`.

kontura (approved):

j-mracek:
LGTM

PR #175, opened 2022-11-25
modules: Add ModuleQuery

pkratoch:


include/libdnf/module/module_item.hpp
```diff
@@ -52,7 +52,9 @@ struct ModuleItemId {
 // @replaces libdnf:module/ModuleItem.hpp:class:ModuleItem
 class ModuleItem {
 public:
-    bool operator==(const ModuleItem & r) const;
+    bool operator==(const ModuleItem & rhs) const noexcept { return id.id == rhs.id.id; }
```
j-mracek, line 55: I have a question - is id mandatory for ModuleItem costructor?

include/libdnf/module/module_item.hpp
```diff
@@ -165,17 +167,18 @@ class ModuleItem {
     // TODO(jmracek) Read a real decision from the modular solver
     bool is_active() const;
 
-private:
-    friend class ModuleSack;
-    friend class ModuleMetadata;
-    friend ::ModuleTest;
-
     ModuleItem(_ModulemdModuleStream * md_stream, const ModuleSackWeakPtr & module_sack, const std::string & repo_id);
```
kontura, line 170: Should we keep this constructor private?
I believe it is not needed for `ModuleQuery` and I think API users don't have a way to make valid `_ModulemdModuleStream` anyway.

pkratoch: It's not very well visible, because of how the changes are shown, but the constructor is not private.

kontura: Yes, now it is not private, but I think it should be.

pkratoch: Oh, sorry, I misunderstood, you are right

j-mracek:
LGTM

PR #178, opened 2022-11-29
Repoquery: Add --duplicates option

kontura:
Adds `filter_duplicates()` method to `PackageQuery`.

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1187

### DIFF
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -2148,6 +2148,13 @@ static void add_n_first_to_map(
     }
 }
 
+static void add_block_to_map(
+    libdnf::solv::SolvMap & result, libdnf::solv::IdQueue & samename, int start_block, int stop_block) {
+    for (int i = start_block; i < stop_block; ++i) {
+        result.add_unsafe(samename[i]);
+    }
+}
+
 static int latest_cmp(const Id * ap, const Id * bp, libdnf::solv::RpmPool * pool) {
     Solvable * sa = pool->id2solvable(*ap);
     Solvable * sb = pool->id2solvable(*bp);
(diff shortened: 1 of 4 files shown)


libdnf/rpm/package_query.cpp
```diff
@@ -2148,6 +2148,30 @@ static void add_n_first_to_map(
...
+    int stop_block) {
+    Solvable * s_first = nullptr;
+    Solvable * s_second = nullptr;
+    for (int pos_first = start_block; pos_first < stop_block; ++pos_first) {
+        Id id_first = samename[pos_first];
+        s_first = pool.id2solvable(id_first);
+        for (int pos_second = start_block; pos_second < stop_block; ++pos_second) {
+            Id id_second = samename[pos_second];
+            s_second = pool.id2solvable(id_second);
+            if ((s_first->evr == s_second->evr) && (s_first->arch != s_second->arch)) {
```
m-blaha, line 2165: All packages in the block should have the same arch. Why is `&& (s_first->arch != s_second->arch)` needed?

libdnf/rpm/package_query.cpp
```diff
@@ -2148,6 +2148,30 @@ static void add_n_first_to_map(
...
+    libdnf::solv::SolvMap & result,
+    libdnf::solv::IdQueue & samename,
+    int start_block,
+    int stop_block) {
+    Solvable * s_first = nullptr;
+    Solvable * s_second = nullptr;
+    for (int pos_first = start_block; pos_first < stop_block; ++pos_first) {
+        Id id_first = samename[pos_first];
+        s_first = pool.id2solvable(id_first);
+        for (int pos_second = start_block; pos_second < stop_block; ++pos_second) {
```
m-blaha, line 2162: You compare each couple of solvables twice (i.e. pkgA with pkgB and than pkgB with pkgA again). I think you can start the inner cycle with `for (int pos_second = pos_first...`. Or even `pos_second = pos_first + 1` to avoid comparison of solvable with itself.

kontura: Yes, thank you!

libdnf/rpm/package_query.cpp
kontura: It is not, I forgot that the arch check was moved. :+1:  :+1: 

Since thanks to your other suggestion we no longer compare each solvable to it self I think we can drop the condition entirely. If there somehow was a package installed twice with the same `name`, `arch` and even `evr` I think we would want it in the query. Packages with the same `nevra` but different `id` surely are duplicates in general for the `filter_duplices()`.

What do you think?

m-blaha (approved):
Thanks, LGTM!

PR #179, opened 2022-11-29
Add dnf5-devel package

inknos:
Files in dnf5/include are necessary to build plugins. This commit adds header files to cmake installed files and dnf5-devel package in the dnf5 specfile.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -246,6 +246,23 @@ Library for working with a terminal in a command-line package manager.
 %license lgpl-2.1.txt
 %endif
 
+# ========== dnf5-devel ==========
+
+%package -n dnf5-devel
+Summary:        Development files for dnf5
+License:        LGPL-2.1-or-later
+Requires:       dnf5%{?_isa} = %{version}-%{release}
+Requires:       libdnf5-devel%{?_isa} = %{version}-%{release}
+Requires:       libdnf5-cli-devel%{?_isa} = %{version}-%{release}
+
+%description -n dnf5-devel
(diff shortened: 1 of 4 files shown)


m-blaha:
I think this is good to go. There is couple of things that we should consider:
- `include/dnf5/context.hpp` includes some headers from `libdnf-cli/*` and of course from `libdnf/*`. So we should add a dependency on `libdnf5-cli-devel` and `libdnf5-devel`.
- (not in this PR) - there is problem with headers in `dnf5/common/*`, like `bgettext`. Those headers are included by headers in `libdnf5-devel` (e.g. /usr/include/libdnf/common/exception.hpp), but we do not ship them in any package.

inknos:
I will add the dependency

dnf5.spec
```diff
@@ -246,6 +246,24 @@ Library for working with a terminal in a command-line package manager.
...
+Requires:       dnf5%{?_isa} = %{version}-%{release}
+Requires:       libdnf5-devel%{?_isa} = %{version}-%{release}
+Requires:       libdnf5-cli-devel%{?_isa} = %{version}-%{release}
+
+%description -n dnf5-devel
+Develpment files for dnf5.
+
+%files -n dnf5-devel
+%{_includedir}/dnf5/
+%dir %{_libdir}/dnf5
```
m-blaha, line 263: Sorry, I missed this in the first round of review. I think we do not put any content into `%{_libdir}/dnf5` directory. Could you please remove it from %files?

PR #185, opened 2022-11-30
Throw Exception if alldeps used without resolve

inknos:
Add catch condition in main for
libdnf::cli::ArgumentParserConflictingArgumentsError. When exception is caught the help should be printed to show the correct arguments usage.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -19,6 +19,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "download.hpp"
 
+#include "utils/bgettext/bgettext-mark-domain.h"
+
 #include <libdnf/conf/option_string.hpp>
 #include <libdnf/rpm/package.hpp>
 #include <libdnf/rpm/package_query.hpp>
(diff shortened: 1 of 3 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -86,6 +88,8 @@ void DownloadCommand::configure() {
     context.update_repo_load_flags_from_specs(pkg_specs);
     if (resolve_option->get_value() && !alldeps_option->get_value()) {
         context.set_load_system_repo(true);
+    } else if (!resolve_option->get_value() && alldeps_option->get_value()) {
+        throw libdnf::cli::ArgumentParserConflictingArgumentsError(M_("Option alldeps should be used with resolve"));
```
jan-kolarik, line 92: This type of exception and its handling is already used in the parser for handling the actual conflicting behavior. Perhaps we should have new exception like `ArgumentParserMissingRequiredArgumentError` and the related handling methods in the `Argument` class like the `add_conflict_argument()`, `set_conflict_arguments()` etc. for the conflicting purpose?

jan-kolarik: OK since this is a change which could be done in a incremental way, how about to define that new exception and use it now and if we encounter another cases in the future, then we will add some methods to the `Argument` class to handle that...

inknos: yes, I agree. `ArgumentParserMissingRequiredArgumentError` as you proposed looks fine for me. I will implement it and get back to you with an update

jan-kolarik: Or maybe `ArgumentParserMissingDependentArgumentError` ? :slightly_smiling_face:

inknos: they both work since a dependent argument is also required. but since the opposite is not true, I see that `ArgumentParserMissingDependentArgumentError` would be more specific

inknos: @jan-kolarik done

jan-kolarik:
Thanks!

PR #187, opened 2022-12-02
Cast std::tolower() argument to unsigned char to avoid possible UB

bkrl:
`std::tolower()` takes an `int` instead of a `char`, and it's undefined behavior if the `int` value is not representable as an `unsigned char` or equal to `EOF`. Therefore, casting the argument to `unsigned char` is necessary to avoid potential UB if `char` is signed and `ch` is not an ASCII character. See https://en.cppreference.com/w/cpp/string/byte/tolower#Notes.


kontura:
Thank you!

PR #190, opened 2022-12-06
dnfdaemon: Update goal.resolve() return value with missing attributes

m-blaha:
Add missing attributes to transaction items returned by goal.resolve().
- package.from_repo_id
- reason
- ability to accomodate not only packages but also groups, environments and modules
- replaces (packages that are replaced by given transaction item)


inknos:
Honestly I don't have comments on it meaning that I find it consistent and I would then merge it.

PR #192, opened 2022-12-07
Support for specifying set of allowed proxy authentication methods

jrohel:
The "proxy_auth_method" option accepted exactly one value from the enum "none", "basic", "digest", "negotiate", "ntlm", "digest_ie", "ntlm_wb", "any".

It is now possible to enter a combination of several authentication methods, for example "basic, digest".

### DIFF
--- a/libdnf/repo/librepo.cpp
+++ b/libdnf/repo/librepo.cpp
@@ -165,15 +165,20 @@ static void init_remote(LibrepoHandle & handle, const C & config) {
         handle.set_opt(LRO_PROXY, config.proxy().get_value().c_str());
     }
 
-    const std::string proxy_auth_method_str = config.proxy_auth_method().get_value();
-    auto proxy_auth_method = LR_AUTH_ANY;
-    for (auto & auth : PROXYAUTHMETHODS) {
-        if (proxy_auth_method_str == auth.name) {
-            proxy_auth_method = auth.code;
-            break;
+    long proxy_auth_methods = 0;
+    if (config.proxy_auth_method().empty()) {
+        proxy_auth_methods = LR_AUTH_ANY;
+    } else {
+        for (const auto & proxy_auth_method_str : config.proxy_auth_method().get_value()) {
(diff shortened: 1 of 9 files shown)


libdnf/repo/librepo.cpp
```diff
@@ -165,15 +165,20 @@ static void init_remote(LibrepoHandle & handle, const C & config) {
...
-    for (auto & auth : PROXYAUTHMETHODS) {
-        if (proxy_auth_method_str == auth.name) {
-            proxy_auth_method = auth.code;
-            break;
+    long proxy_auth_methods = 0;
+    if (config.proxy_auth_method().empty()) {
+        proxy_auth_methods = LR_AUTH_ANY;
+    } else {
+        for (const auto & proxy_auth_method_str : config.proxy_auth_method().get_value()) {
+            for (auto & auth : PROXYAUTHMETHODS) {
```
jan-kolarik, line 173: Maybe I am missing something, but why don't we have just a const map with the name-code mappings instead of having this double for loop?

jrohel: It is an optimalization. The constexpr array of structures is prepared during compile time. And in this case it contains only 8 elements and usualy none or one element is searched. 

`std::map` will be constructed at runtime. And implements balanced binary search tree. 
So in case of use `std::map` I can replace the internal loop by line: `proxy_auth_methods |= PROXYAUTHMETHODS.at(proxy_auth_method_str);` But the internal code (hidden for programer) will be mutch more complicated and longer.

So that is why I used this aproach. I can use `std::map` if you want.

jan-kolarik: Thanks a lot for an explanation! I didn't realize `std::map` don't have a `constexpr` variant. Therefore your proposed solution is perfectly fine.

jan-kolarik:
LGTM, thanks! Just maybe regarding the tests - it would be probably good to have the functionality tested in unit tests or through the CLI API (CI tests).

PR #196, opened 2022-12-14
Fix: RepoDownloader::get_cache_handle: Don't set callbacks in LibrepoHandle

jrohel:
Before downloading files (packages) from the repository, Librepo itself downloads a metalink/mirrorlist file to a temporary file (if the metalink/mirrorlist url was defined in the configuration) and parses it. An unexpected call to the progress callback occurred while doing so. This fix ensures that the progress callback will not be called.

This fix is not ideal. We should know that librepo is downloading an extra file in the background. For example, in case of network problems, the program may hang for some time and we will not know why. It's just a quick fix for the crash.

The question is, do we even want librepo to download metalink/mirrorlist files itself every time before downloading files from the repository? Why are the metalink/mirrorlist files in the cache?

### DIFF
--- a/libdnf/repo/repo_downloader.cpp
+++ b/libdnf/repo/repo_downloader.cpp
@@ -328,9 +328,11 @@ void RepoDownloader::load_local() try {
 
 /// Returns a librepo handle, set as per the repo options.
 /// Note that destdir is None, and the handle is cached.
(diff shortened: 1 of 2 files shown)


j-mracek:
The patch fixed the issue. Thank you very much for the fast creation of the patch.

PR #197, opened 2022-12-15
[dnf5] Fix: Handling of unknown exceptions

jrohel:
- Fix: Do not pass formatted message as the first parameter to the logger
  fixes "terminate called after throwing an instance of 'fmt::v9::format_error'". 

- Catch only exceptions inherited from std::runtime_error


jan-kolarik:
LGTM :slightly_smiling_face: Thanks for the fix!

PR #199, opened 2022-12-15
modules: Fix ModuleItem copy and move constructor and assignment operator

pkratoch:


libdnf/module/module_item.cpp
```diff
@@ -259,8 +264,36 @@ ModuleItem & ModuleItem::operator=(const ModuleItem & mpkg) {
...
         id = mpkg.id;
+        repo_id = mpkg.repo_id;
+        computed_static_context = mpkg.computed_static_context;
+    }
+    return *this;
+}
+
+
+ModuleItem::ModuleItem(ModuleItem && mpkg) {
+    md_stream = mpkg.md_stream;
```
jrohel, line 277: OK. In general, using the member initializer list is a better approach than having members initialized with default constructors and then assigning them a value in the constructor body. Please can you change it.
```
ModuleItem::ModuleItem(ModuleItem && mpkg)
    : md_stream(mpkg.md_stream),
      module_sack(std::move(mpkg.module_sack)),
      id(std::move(mpkg.id)),
      repo_id(std::move(mpkg.repo_id)),
      computed_static_context(std::move(mpkg.computed_static_context)) {
    mpkg.md_stream = nullptr;
}
```

pkratoch: Thanks, fixed.

jrohel: Thank you.

PR #200, opened 2022-12-15
MultiProgressBar: Take ownership of included ProgressBars

m-blaha:
In current implementation MultiProgressBar only keeps raw pointers to its encapsulated ProgressBar instances.
I think the ownership of progress bars should be transfered to MultiProgressBar and their lifecycles kept tied together.


jrohel (approved):
It seems okay. Thanks for the PR.

Issue #152, opened 2022-11-11
Implement search command


mcurlej:
- [ ] **Command:** ```search```

- [ ] **Aliases:** ```se```
  - drop the alias

- [ ] ```dnf [options] search [--all] <keywords>...```

Search package metadata for keywords. Keywords are matched as case-insensitive substrings, globbing is supported. By default lists packages that match all requested keys (AND operation). Keys are searched in package names and summaries. If the “–all” option is used, lists packages that match at least one of the keys (an OR operation). In addition the keys are searched in the package descriptions and URLs. The result is sorted from the most relevant results to the least.

This command by default does not force a sync of expired metadata. See also [Metadata Synchronization](https://dnf.readthedocs.io/en/latest/command_ref.html#metadata-synchronization-label).

jan-kolarik:
After merging the #202, the CI tests still need to be implemented for the command. Or maybe the existing DNF tests could be adjusted...

frenzymadness:
I'm already testing dnf5 and the missing search command is currently the only big issue I have.

jan-kolarik:
> I'm already testing dnf5 and the missing search command is currently the only big issue I have.

That's great, thanks for the report :slightly_smiling_face: The PR with search command already exists and I would like to have it finished and merged soon, in several days.

PR #202, opened 2022-12-16
Search command: initial version

jan-kolarik:
Introducing a `search` command.

The output is very similar to the existing command in `DNF`. 

For now probably the only functionality missing is the grouping of the results based on the matched patterns. Now only grouping based on the matched metadata keys is present. This could be implemented later if it is found useful for clients.

Related issue link: https://github.com/rpm-software-management/dnf5/issues/152


jan-kolarik:
Example screenshot:

image

dnf5/commands/search/search_processor.cpp
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // For the first pattern we are always adding to the empty list.
+        // If we are using the "--all" option, we want any matches in the result.
+        // Otherwise we use an intersection (AND) with all previous results.
+        if (all_matches.empty() | search_all) {
+            all_matches.update(std::move(pattern_matches));
+        } else {
+            all_matches.intersection(std::move(pattern_matches));
+        }
```
pkratoch, line 150: When the first pattern isn't matched, `all_matches` remains empty, but intersection should still be used for any subsequent pattern. E.g. `dnf5 search foobarbaz dnf` should result in an empty output, but doesn't (provided `foobarbaz` is something non-existent).

Similarly, when an intersection results in empty `all_matches`, intersection is not used for the subsequent pattern. E.g. `dnf5 search glibc nano dnf` should result in an empty output, but doesn't.

jan-kolarik: Oh, I see. I should distinguish going through the first loop instead of checking the empty matches. Thanks!

dnf5/commands/search/search_processor.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+        // For the first pattern we are always adding to the empty list.
+        // If we are using the "--all" option, we want any matches in the result.
+        // Otherwise we use an intersection (AND) with all previous results.
+        if (pattern == patterns.front() || search_all) {
+            all_matches.update(std::move(pattern_matches));
+        } else {
+            all_matches.intersection(std::move(pattern_matches));
+        }
```
pkratoch, line 150: Now, there are different results for e.g. `dnf5 search dnf perl` and `dnf5 search dnf perl dnf` :)

include/libdnf-cli/output/search.hpp
```diff
@@ -0,0 +1,76 @@
...
+    EXACT     ///< When the pattern exactly matches the given key.
+};
+
+/// Defines a type for a key that was matched together with a type of the match.
+using matched_key_pair = std::pair<std::string, SearchKeyMatch>;
+
+/// Comparator used for sorting packages within the results.
+static bool search_package_comparator(const libdnf::rpm::Package & one, const libdnf::rpm::Package & two) {
+    return one.get_full_nevra() < two.get_full_nevra();
+};
```
pkratoch, line 44: This compares the nevra strings, so the packages are not necessarily sorted by version. The result can be e.g.:
```
package-0:1.10-1.noarch
package-0:1.8-1.noarch
package-0:1.9-1.noarch
```
In `include/libdnf/rpm/nevra.hpp`, there are comparators for `Nevra`, but I haven't found a way to get `Nevra` from `Package`. But e.g. `PackageQuery` also implements its own `latest_cmp` for comparing packages by nevra, however, it's not public, so I am wondering if there shouldn't be a shared way to properly compare packages...

doc/commands/search.8.rst
```diff
@@ -0,0 +1,58 @@
...
+``dnf5 search [options] <pattern>...``
+
+
+Description
+===========
+
+The ``search`` command in ``DNF5`` is used for searching packages by matching
+given keywords from the user against various metadata. By default the command
+searches all requested keys in `Name` or `Summary` fields from the RPM
+package metadata.
```
pkratoch, line 37: As opposed to the dnf4 documentation, there is not mentioned case-sensitivity and the fact that "AND operation" is used. The other is shown in the examples and it's not hard to figure it out after one use, but I think it would be still worth mentioning here as well.

doc/commands/search.8.rst
```diff
@@ -0,0 +1,58 @@
...
+
+
+Options
+=======
+
+``--all``
+    | Search patterns also inside `Description` and `URL` fields.
+    | By applying this option the search is also relaxed by including results of any given key matches.
```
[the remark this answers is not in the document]
jan-kolarik: Aha, OK. I will change it. You're right that the expression is kind of vague...

doc/commands/search.8.rst
jan-kolarik: Sure, this should be there, thanks!

dnf5/commands/search/search_processor.cpp
jan-kolarik: OK, this was a stupid mistake. What I really wanted to do was to compare the current iterator to the vector beginning :slightly_smiling_face:

include/libdnf-cli/output/search.hpp
jan-kolarik: Nice, thanks for the reference. I can use that comparer directly, because it's templated and the same `Nevra` methods are available also in the `Package` class.

jan-kolarik:
FYI I used `std::views` in my initial implementation which was fine with `gcc`, but not yet supported with `clang` compiler for now (only in experimental setup). So I switched to a different solution.

libdnf-cli/output/search.cpp
```diff
@@ -0,0 +1,104 @@
...
+        std::cout << bold << "Matched fields: " << reset << construct_keys_string(result.matched_keys) << std::endl;
+
+        for (auto const & package : result.matched_packages.packages) {
+            std::cout << " ";
+            if (results.options.show_duplicates) {
+                std::cout << highlight(package.get_full_nevra());
+            } else {
+                std::cout << highlight(package.get_name()) << "." << package.get_arch();
+            }
+            std::cout << ": " << highlight(package.get_summary()) << std::endl;
```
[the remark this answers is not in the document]
jan-kolarik: I was aiming there for a "user friendliness", but I see the `libsmartcols` is kind of standard and a way to have the output easily parsable. I am for creating the follow-up issue/PR regarding this topic if it's fine also for @pkratoch ...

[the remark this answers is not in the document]
pkratoch: Sure, we can leave it for follow-up issue. May I ask you to file it?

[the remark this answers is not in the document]
jan-kolarik: @pkratoch Done :slightly_smiling_face:

Issue #207, opened 2022-12-19
Write log messages from librepo and libmodulemd to Logger


jrohel:
Log messages from librepo and libmodulemd are now ignored.

The idea is to write these messages to the logger along with the other messages from libdnf.
The logged messages will contain the prefix (domain) of the library they came from.

PR #208, opened 2022-12-19
Write log messages from librepo and libmodulemd to Logger

jrohel:
Log messages from librepo and libmodulemd were ignored.

PR implements writing these messages to the logger along with the other messages from libdnf.
The logged messages contain the prefix (domain) of the library they came from.

Solve issue https://github.com/rpm-software-management/dnf5/issues/207

Note: Logger verbosity is hardcoded to DEBUG. Use configuration in the future.


kontura:
LGTM.
Thanks!

Issue #203, opened 2022-12-16
Simplify usage of callbacks in DNF5


j-mracek:
Multiple methods in DNF requires to use a callback for download operation and RPM transaction. In some cases method in advance will not know whether callback will be required or not (install_spec when remote path to RPM is provided). It would be easier to centralize setting callback to Base Class. In DNF4 CLI download callbacks were stored in Output class progress attribute but this approach was available only for CLI users without possibility to provide customized callbacks.

1. Create a a setter for particular callback type in DNF5
libdnf::repo::DownloadCallbacks
libdnf::rpm::TransactionCallbacks

There is no requirement to have  a different callback for the particular repository.

2. Remove setting callbacks from particular methods - API change
Keeping both approaches in API can create a confusion. Then the method will use callback from the Base Class

j-mracek:
The issue is fixed by https://github.com/rpm-software-management/dnf5/pull/228 and https://github.com/rpm-software-management/dnf5/pull/280.

PR #209, opened 2022-12-21
Centralize download callbacks setting to Base

m-blaha:
This PR introduces centralized setting of download callbacks. Callbacks are not set using `*Downloader.add()` API any more, there is a new `base::set_download_callbacks_factory()` API for setting a callback factory class and `base.create_download_callbacks()` to get a callback using this factory.

Resolves https://github.com/rpm-software-management/dnf5/issues/203

### DIFF
--- a/include/libdnf/base/base.hpp
+++ b/include/libdnf/base/base.hpp
@@ -29,6 +29,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf/logger/log_router.hpp"
 #include "libdnf/module/module_sack.hpp"
 #include "libdnf/plugin/iplugin.hpp"
+#include "libdnf/repo/download_callbacks.hpp"
+#include "libdnf/repo/file_downloader.hpp"
+#include "libdnf/repo/package_downloader.hpp"
 #include "libdnf/repo/repo_sack.hpp"
 #include "libdnf/rpm/package_sack.hpp"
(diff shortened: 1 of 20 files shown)


include/libdnf/base/base.hpp
```diff
@@ -92,6 +93,18 @@ class Base {
...
+    /// all downloads.
+    /// @param download_callbacks_factory   Instance of implementation of DownloadCallbacksFactory interface
+    void set_download_callbacks_factory(
+        std::unique_ptr<libdnf::repo::DownloadCallbacksFactory> && download_callbacks_factory);
+
+    /// Create DownloadCallbacks instance using the DownloadCallbacksFactory
+    /// (see set_download_callbacks_factory() method). In case the factory is
+    /// not set, nullptr is returned.
+    /// @param what String that describes what is being downloaded (URL, package...)
+    std::unique_ptr<libdnf::repo::DownloadCallbacks> create_download_callbacks(const std::string & what);
```
j-mracek, line 106: Do we need it in public part? It looks like we use it only internally.

m-blaha:
There is a problem with generating SWIG bindings which I still need to investigate...

include/libdnf/base/base.hpp
m-blaha: Right. I was just reluctant to add another two friend classes to Base. But I guess it's better to hide as much functionality as we can. I'll move it.

PR #214, opened 2023-01-03
modules: Add ModuleQuery::filter_latest

pkratoch:


include/libdnf/module/module_item.hpp
```diff
@@ -192,7 +192,8 @@ class ModuleItem {
     std::string get_name_stream() const;
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getNameStreamVersion()
     std::string get_name_stream_version() const;
-    std::string get_name_stream_context() const;
+    std::string get_name_stream_staticcontext() const;
```
j-mracek, line 195: May I ask you for a description? Because static_context is a boolean option in metadata, it could be confusing without the description.

pkratoch: Oh, right, I forgot it's named the same. But at least both these methods that operate with the computed static contexts are private.

libdnf/module/module_item.cpp
```diff
@@ -122,6 +122,17 @@ std::string ModuleItem::get_name_stream_context() const {
...
 
+std::string ModuleItem::get_name_stream_staticcontext_arch() const {
+    return libdnf::utils::sformat(
+        "{}:{}:{}:{}",
+        libdnf::utils::string::c_to_str(modulemd_module_stream_get_module_name(md_stream)),
+        libdnf::utils::string::c_to_str(modulemd_module_stream_get_stream_name(md_stream)),
+        computed_static_context.empty() ? libdnf::utils::string::c_to_str(modulemd_module_stream_get_context(md_stream))
+                                        : computed_static_context,
+        libdnf::utils::string::c_to_str(modulemd_module_stream_get_arch(md_stream)));
+}
```
j-mracek, line 133: If you would not care about performance optimization you could shorter the code using `get_name_stream_staticcontext()` call. I guess that the performance difference will be minimal.

pkratoch: Well, there are many methods like this. If we used this approach everywhere, it would result in multiple calls in most of them. And if it was used only sometimes, it would be inconsistent. I don't think it would really improve readability, but of course, it would be shorter. So, I don't know - if you think it would be better, I can change it.

libdnf/module/module_item.cpp
```diff
@@ -122,6 +122,17 @@ std::string ModuleItem::get_name_stream_context() const {
 }
 
 
+std::string ModuleItem::get_name_stream_staticcontext_arch() const {
+    return libdnf::utils::sformat(
```
jrohel, line 126: We should prefer to use `fmt::format` over `libdnf::utils::format`. Since C++20, `fmt::format` does the syntax checking of the format string during compilation. -> And it requires the format string to be a constant expression.
We have `libdnf::utils::sformat` for cases where the format string is not a constant expression, usually when the content of the format string are unknown at compile time.  For example, the format string is the result of translating into the user's language (localization).

pkratoch: So, in this case, `fmt::format` can't be used? Because computed_static_context is not known at compile time?

jrohel: In this case we should use `fmt::format`. The format string "{}:{}:{}:{}" is known at compile time.

Example of incorrect code - The format string requires more arguments than are passed.:
- `fmt::format("{}:{}:{}:{}", "a", "b", "c")` reports an error at compile time.  Which can be used by the IDE.
- `libdnf::utils::sformat("{}:{}:{}:{}", "a", "b", "c")` parses the format string at runtime. We get the exception at runtime.

pkratoch: I see. Ok, I replaced the `libdnf::utils::sformat` with `fmt::format` where possible in the last commit.

libdnf/module/module_item.cpp
```diff
@@ -95,25 +94,25 @@ std::string ModuleItem::get_arch() const {
...
         libdnf::utils::string::c_to_str(modulemd_module_stream_get_module_name(md_stream)),
         libdnf::utils::string::c_to_str(modulemd_module_stream_get_stream_name(md_stream)));
 }
 
 
 std::string ModuleItem::get_name_stream_version() const {
-    return libdnf::utils::sformat(
+    return fmt::format(
         "{}:{}:{}",
         libdnf::utils::string::c_to_str(modulemd_module_stream_get_module_name(md_stream)),
```
j-mracek, line 107: Only note: fmt::format accepts `char *` but it is unable to handle nullptr. I suggest that we could replace libdnf::utils::string::c_to_str by a method that will replace nullptr by a an empty C string. Then the code will not require to allocate memory for temporary std::string

libdnf/module/module_query.cpp
```diff
@@ -100,4 +100,44 @@ void ModuleQuery::filter_arch(const std::vector<std::string> & patterns, libdnf:
...
+    }
+    if (s1->arch != s2->arch) {
+        return s1->arch < s2->arch;
+    }
+    return module_item_1->get_version() > module_item_2->get_version();
+}
+
+
+void ModuleQuery::filter_latest(int limit) {
+    std::vector<const ModuleItem *> same_nsca_vector;
```
j-mracek, line 118: Minor: I believe that the final size of `same_nsca_vector` (size of `same_nsca_vector.size()` == get_data().size() after filling in following for loop) is known, therefore you can reserve the space in the vector.

libdnf/module/module_query.cpp
```diff
@@ -100,4 +100,44 @@ void ModuleQuery::filter_arch(const std::vector<std::string> & patterns, libdnf:
...
+
+    for (auto & module_item : get_data()) {
+        same_nsca_vector.push_back(&module_item);
+    }
+    sort(same_nsca_vector.begin(), same_nsca_vector.end(), latest_cmp);
+
+    std::string nsca;
```
j-mracek, line 125: Why you specify `std::string nsca` here and not in for loop?
`auto nsca = module_item->get_name_stream_staticcontext_arch();`

libdnf/module/module_query.cpp
```diff
@@ -100,4 +100,44 @@ void ModuleQuery::filter_arch(const std::vector<std::string> & patterns, libdnf:
...
+    for (auto module_item : same_nsca_vector) {
+        nsca = module_item->get_name_stream_staticcontext_arch();
+        if (last_nsca != nsca) {
+            last_nsca = nsca;
+            kept_in_query = 0;
+        }
+        if (kept_in_query < limit) {
+            kept_in_query += 1;
+        } else {
+            get_data().erase(*module_item);
```
j-mracek, line 137: Only note - I believe that removal of an item from the set (or vector) is quite expensive operation.

libdnf/module/module_item.cpp
jrohel: @j-mracek Good idea. We can create a new method to replace `libdnf::utils::string::c_to_str` in such cases. 
Example of new method:
```
inline constexpr const char * null_to_empty(const char * str) noexcept {
    return str ? str : "";
}
```
From my point of view, this is a code optimization that we can do additionally (in bulk) in multiple places.

pkratoch: I added a TODO to improve this later.

include/libdnf/module/module_query.hpp
```diff
@@ -133,6 +133,12 @@ class ModuleQuery : public libdnf::sack::Query<ModuleItem> {
     void filter_arch(
         const std::vector<std::string> & patterns, libdnf::sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ);
 
+    /// Group ModuleItems by `name`, `stream`, `context` and `arch`. Then within each group, keep packages that correspond with up to `limit` of latest `version`s in the group.
+    ///
+    /// @param limit            Keep `limit` number `version`s in each group.
+    /// @since 5.0.4
+    void filter_latest(int limit = 1);
```
j-mracek, line 140: I see a problem in limit as int. If I understand the code correctly the code is unable to handle negative values. I checked PackageQueries and the latest filter also used int as limit argument but it supports also negative value.
I see some possibility how to resolve the issue

- document that negative value result in empty query.
  - plus add todo to support negative values
  - not support at all even not in future - add assertion - this is dangerous approach specially if the limit can be specified by user.
- not use int but unsigned int
  - difficult to extend functionality in future
  - different in comparison to PacageQueries

libdnf/module/module_query.cpp
```diff
@@ -100,4 +100,44 @@ void ModuleQuery::filter_arch(const std::vector<std::string> & patterns, libdnf:
...
+        }
+    }
```
j-mracek, line 139: There is a difference in logic in comparison to PackageQuery. If I have set of packages - dnf-1-1.noarch, dnf-2-2.noarch, dnf-2-2.noarch - PackageQuery.filter_latest(2) will return all 3packages, because it returns 2 versions not 2 latest object. Your code with nodejs:master:asssd:1:noarch, nodejs:master:asssd:2:noarch, nodejs:master:asssd:2:noarch, - ModuleQuery.filter_latest(2) will return 2 module items with version 2. I would recommend to use the same logic like PackageQuery.
```
    std::string last_nsca = same_nsca_vector.front()->get_name_stream_staticcontext_arch();
    long long latest version = -1; // invalid value for version to ensure that it cannot be in module item
    int kept_in_query = 0;
    for (auto module_item : same_nsca_vector) {
        std::string nsca = module_item->get_name_stream_staticcontext_arch();
        if (last_nsca != nsca) {
            last_nsca = nsca;
            kept_in_query = 0;
        }
        if (latest version == module_item->get_version()) {  // do nothing => keep it in query, do not increase kept_in_query
        } else if (kept_in_query < limit) {
            kept_in_query += 1;
        } else {
            get_data().erase(*module_item);
        }
```
What do you think?

include/libdnf/module/module_query.hpp
pkratoch: Thank you! I totally forgot about the limit. I added the support for negative values right away, because I think it's quite straightforward. But if you disagree, I can change it back and add a todo.

libdnf/module/module_query.cpp
pkratoch: I changed it to be the same as in PackageQuery. I had to also change the tests a bit (added a new repo with duplicit modulemd documents).

test/data/repos-repomd/repomd-modules-duplicit/repodata/repomd.xml
```diff
@@ -0,0 +1,18 @@
...
+    <open-size>864433</open-size>
+  </data>
+  <data type="modules">
+    <checksum type="sha256">18a868b0227d78b40ae22641390c496c943faafb3d67ea74603e2ab7819615da</checksum>
+    <open-checksum type="sha256">18a868b0227d78b40ae22641390c496c943faafb3d67ea74603e2ab7819615da</open-checksum>
+    <location href="repodata/modules.yaml" />
+    <timestamp>1641802880</timestamp>
+    <size>492</size>
+  </data>
+</repomd>
```
j-mracek, line 18: There is missing new line at the end, but I am not sure whether it is fixable. @pkratoch What do you think?

j-mracek:
LGTM

PR #217, opened 2023-01-04
SWIG: drop the std::optional from the public API

### LINKED ISSUE #186: SWIG: drop std::optional from the public API
As SWIG is not able to automatically or easily transform `std::optional` wrappers into various language bindings, it is better not to use it on the public API neither as the return value or as the method parameter.

Currently used on following places:
- [ ]  `libdnf::advisory::AdvisoryQuery::filter_reference()` methods, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/advisory/advisory_query.hpp#L78)
- [ ] `libdnf::GoalJobSettings` structure, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/base/goal_elements.hpp#L131)
- [ ] `libdnf::Goal::add_rpm_reason_change()` method, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/base/goal.hpp#L249)
- [ ] `libdnf::base::LogEvent` class, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/base/log_event.hpp#L35)
- [ ] `libdnf::base::TransactionPackage::get_reason_change_group_id()` method, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/base/transaction_package.hpp#L72)
- [ ] `libdnf::base::Transaction::run()` method, see [there](https://github.com/rpm-software-management/dnf5/blob/73f53313ad07d49de2c46b87abe13f3c810f6d76/include/libdnf/base/transaction.hpp#L89)

jan-kolarik:
SWIG bindings don't support the `std::optional` type wrappers, so this PR aims to drop all these references away from the public API.

Closes #186.

### DIFF
--- a/include/libdnf/advisory/advisory_query.hpp
+++ b/include/libdnf/advisory/advisory_query.hpp
@@ -29,8 +29,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf/common/sack/query_cmp.hpp"
 #include "libdnf/rpm/package.hpp"
 
-#include <optional>
-
 
 namespace libdnf::advisory {
 
@@ -75,14 +73,15 @@ public:
     /// @param pattern      Pattern to match with reference id.
     /// @param cmp_type     What comparator to use with pattern, allows: EQ, IEXACT, GLOB, IGLOB, CONTAINS, ICONTAINS.
     /// @param type         Possible reference types are: "bugzilla", "cve", "vendor". If none is specified it matches all.
+    void filter_reference(const std::string & pattern, sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ);
     void filter_reference(
-        const std::string & pattern,
-        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
-        const std::optional<std::string> type = {});
+        const std::string & pattern, const std::string & type, sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ);
     void filter_reference(
-        const std::vector<std::string> & pattern,
-        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
-        const std::optional<std::string> type = {});
+        const std::vector<std::string> & patterns, sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ);
+    void filter_reference(
+        const std::vector<std::string> & patterns,
+        const std::string & type,
+        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ);
 
     /// Filter Advisories by severity.
     ///
--- a/include/libdnf/base/goal.hpp
+++ b/include/libdnf/base/goal.hpp
@@ -282,7 +282,7 @@ public:
     void add_rpm_reason_change(
         const std::string & spec,
         const libdnf::transaction::TransactionItemReason reason,
-        const std::optional<std::string> & group_id,
+        const std::string & group_id = {},
         const libdnf::GoalJobSettings & settings = libdnf::GoalJobSettings());
 
     /// Add group install request to the goal. The `spec` will be resolved to groups in the resolve() call.
--- a/include/libdnf/base/goal_elements.hpp
+++ b/include/libdnf/base/goal_elements.hpp
@@ -143,10 +143,16 @@ public:
 
     /// Optionally assign AdvisoryQuery that is used to filter goal target packages (used for upgrade and install)
     void set_advisory_filter(const libdnf::advisory::AdvisoryQuery & filter) { advisory_filter = filter; };
+    const libdnf::advisory::AdvisoryQuery * get_advisory_filter() const {
+        return advisory_filter ? &advisory_filter.value() : nullptr;
+    }
 
     // Which types of group packages are going to be installed with the group.
     // If not set, default is taken from ConfigMain.group_package_types
     void set_group_package_types(const libdnf::comps::PackageType type) { group_package_types = type; }
+    const libdnf::comps::PackageType * get_group_package_types() const {
+        return group_package_types ? &group_package_types.value() : nullptr;
+    }
 
     /// Set whether hints should be reported
     bool report_hint{true};
--- a/include/libdnf/base/goal_elements.hpp
+++ b/include/libdnf/base/goal_elements.hpp
@@ -141,25 +141,25 @@ public:
     /// Return used value for clean_requirements_on_remove
     GoalUsedSetting get_used_clean_requirements_on_remove() const { return used_clean_requirements_on_remove; };
 
-    /// Set whether hints should be reported
-    bool report_hint{true};
-    /// Set strict, AUTO means that it is handled according the the default behavior
-    GoalSetting strict{GoalSetting::AUTO};
-    /// Set best, AUTO means that it is handled according the the default behavior
-    GoalSetting best{GoalSetting::AUTO};
-    /// Set clean_requirements_on_remove, AUTO means that it is handled according the the default behavior
-    GoalSetting clean_requirements_on_remove{GoalSetting::AUTO};
-    /// Define which installed packagies should be modified acording to repoid from which they were installed
-    std::vector<std::string> from_repo_ids;
-    /// Reduce candidates for the operation according repository ids
-    std::vector<std::string> to_repo_ids;
-
     /// Optionally assign AdvisoryQuery that is used to filter goal target packages (used for upgrade and install)
-    std::optional<libdnf::advisory::AdvisoryQuery> advisory_filter;
+    void set_advisory_filter(const libdnf::advisory::AdvisoryQuery & filter) { advisory_filter = filter; };
 
     // Which types of group packages are going to be installed with the group.
     // If not set, default is taken from ConfigMain.group_package_types
-    std::optional<libdnf::comps::PackageType> group_package_types{std::nullopt};
+    void set_group_package_types(const libdnf::comps::PackageType type) { group_package_types = type; }
+
+    /// Set whether hints should be reported
+    bool report_hint{true};
+    /// Set strict, AUTO means that it is handled according to the default behavior
+    GoalSetting strict{GoalSetting::AUTO};
+    /// Set best, AUTO means that it is handled according to the default behavior
+    GoalSetting best{GoalSetting::AUTO};
+    /// Set clean_requirements_on_remove, AUTO means that it is handled according to the default behavior
+    GoalSetting clean_requirements_on_remove{GoalSetting::AUTO};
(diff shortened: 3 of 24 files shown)


jan-kolarik:
Note: CI tests are failing due to API change in binding. This is resolved in related PR https://github.com/rpm-software-management/ci-dnf-stack/pull/1200.

include/libdnf/advisory/advisory_query.hpp
```diff
@@ -78,11 +76,11 @@ class AdvisoryQuery : public AdvisorySet {
     void filter_reference(
         const std::string & pattern,
         sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
-        const std::optional<std::string> type = {});
+        const std::string & type = {});
```
j-mracek, line 79: I am not sure whether this is good approach or not. First of all - do we have a list of valid types or this is a string in metadata that we try to match in field type? From the description it looks like that as a type only three strings are valid.

What other option do we have:

1. Overloaded methods:

```
    void filter_reference(
        const std::string & pattern,
        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
        const std::optional<std::string> type = {});
```
```
    void filter_reference(
        const std::string & pattern,
        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
        const std::optional<std::string> type = {});
        const std::string & type);
```
2. Raw pointer (not nice)
```
    void filter_reference(
        const std::string & pattern,
        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
        const std::optional<std::string> type = {});
        const std::string * type);
```
3. Splitting reference filters ()
```
void filter_reference_name()
void filter_reference_type()
void filter_reference_name_type() - to do both together is good for performance 
```
4. default value
```
   void filter_reference(
        const std::string & pattern,
        sack::QueryCmp cmp_type = libdnf::sack::QueryCmp::EQ,
        const std::optional<std::string> type = {});
        const std::string & type = {});
```
May I ask the community for the feedback?

jan-kolarik: Definitely, I'd be glad to have it discuss in a broader scope. And also to make things consistently when doing any changes in future. Another way, maybe, since we have a several valid values could be defining an enum for this.

include/libdnf/base/goal_elements.hpp
```diff
@@ -137,26 +137,26 @@ struct GoalJobSettings : public ResolveSpecSettings {
...
+    /// Set clean_requirements_on_remove, AUTO means that it is handled according to the default behavior
     GoalSetting clean_requirements_on_remove{GoalSetting::AUTO};
-    /// Define which installed packagies should be modified acording to repoid from which they were installed
+    /// Define which installed packages should be modified according to repoid from which they were installed
     std::vector<std::string> from_repo_ids;
     /// Reduce candidates for the operation according repository ids
     std::vector<std::string> to_repo_ids;
 
-    /// Optionally assign AdvisoryQuery that is used to filter goal target packages (used for upgrade and install)
-    std::optional<libdnf::advisory::AdvisoryQuery> advisory_filter;
```
j-mracek, line 154: Right now we don't need a getter, but I think it would be logical to have it. GoalJobSetting is stored in event log to provide all information related to the job request, therefore there could be a reason to provide a getter. But what could the getter return?

include/libdnf/base/goal.hpp
```diff
@@ -249,7 +249,7 @@ class Goal {
     void add_rpm_reason_change(
         const std::string & spec,
         const libdnf::transaction::TransactionItemReason reason,
-        const std::optional<std::string> & group_id,
+        const std::string & group_id = {},
```
j-mracek, line 252: Shall we have another method for the reason group ?

include/libdnf/base/transaction.hpp
```diff
@@ -86,14 +87,14 @@ class Transaction {
...
-    /// @param user_id UID of the user that started the transaction.
     /// @param comment Any comment to store in the history database along with the transaction.
+    /// @param user_id UID of the user that started the transaction.
     /// @return An enum describing the result of running the transaction.
     TransactionRunResult run(
         std::unique_ptr<libdnf::rpm::TransactionCallbacks> && callbacks,
         const std::string & description,
-        const std::optional<uint32_t> user_id,
-        const std::optional<std::string> comment);
+        const std::string & comment = {},
```
j-mracek, line 96: Here I don't see a difference between empty string and not providing a string, therefore I am fine with the approach. I am even fine if the approach here will be different with a solution for query filters.

include/libdnf/base/transaction.hpp
```diff
@@ -86,14 +87,14 @@ class Transaction {
...
+        const uint32_t user_id = libdnf::INVALID_UID);
```
j-mracek, line 97: Using `libdnf::INVALID_UID` is somehow misleading. Unset vs. unable to detect - shall the code make a difference?

include/libdnf/base/goal_elements.hpp
jan-kolarik: The problem in this case is also that the settings object is being copied to the log, so this needs to be beard in mind and f.e. `unique_ptr` approach is not possible there.

include/libdnf/advisory/advisory_query.hpp
m-blaha: Regarding enum - I'm really not sure that the valid values are defined anywhere. I did a bit of searching and didn't find any standard describing updateinfo (well, I found this opensuse standard - https://en.opensuse.org/openSUSE:Standards_Rpm_Metadata_UpdateInfo which defines only "bugzilla" and "cvs" values). I'm afraid that different distributions can use it's own set of types. And from libsolv POV the reference type looks like an arbitrary string.

In this case I would go with overloaded `filter_reference(pattern, cmp_type)` and `filter_reference(pattern, cmp_type, type)` methods.

kontura: Originally the types were an ENUM but I have removed them in https://github.com/rpm-software-management/libdnf/pull/1365.

I agree with @m-blaha, since we don't have any proper definition of metadata, it can be any string just like type or severity of advisory itself.

jan-kolarik: Thanks for the comments! I also like the overloads approach for this case.

j-mracek: I am just curios, where in DNF4 we require reference type? Or what kind of user case we have for DNF5?

j-mracek: I am asking because there could be multiple types provided, then there could be a vector with strings as an argument.

m-blaha: @j-mracek Yes, there could be third overloaded version accepting the vector. But I consider it out of scope of this PR. The TODO comment will do here.

include/libdnf/base/goal_elements.hpp
jan-kolarik: In the case we want to have a getter here we could use the similar approach as in the commit for `transaction` changes - return raw pointer which would either return pointer to data if assigned, or `nullptr`.

include/libdnf/base/transaction.hpp
jan-kolarik: Another way is to add a setter for `user_id` in the transaction class, but it is used only in this method. I don't like defining another constant for unused UID as I am thinking about it. What about creating an overload - one with user_id, one without it and internally it could call the same private method with `std::optional` parameter?

include/libdnf/base/goal.hpp
jan-kolarik: I suppose we can also use the overload approach here, but in this case, is there anything wrong about having the default empty `group_id`? Is there some usecase that we want to treat empty `group_id` values coming through this call differently?

m-blaha: Empty string shouldn't be any problem. It is only valid and used if the new reason is "GROUP".

include/libdnf/base/transaction.hpp
j-mracek: Overloaded method sounds like a good option here.

include/libdnf/base/goal_elements.hpp
j-mracek: Raw pointer is an option that we agreed on.

include/libdnf/advisory/advisory_query.hpp
j-mracek: I found two different usages of advisory references in our code
```
    if (with_bz->get_value()) {
        advisories.filter_reference("*", libdnf::sack::QueryCmp::IGLOB, {"bugzilla"});
    }
    if (with_cve->get_value()) {
        advisories.filter_reference("*", libdnf::sack::QueryCmp::IGLOB, {"cve"});
    }
```
From the first usage I recommend to add `filter_reference_type` (name may be different). According to usage I recommend vector<string> as an argument for `void filter_reference_type()` because the command requires relation between types `or` rather then `and` - If I am correct advisory has only one type of reference therefore and will always result in empty set (currently implemented with `and` operator).

For testing I would recommend to play with `dnf5 advisory list --with-bz --with-cve`

The second usage uses both types and reference name string and it is correct to have there only one type - string not a vector. What about to have type as a mandatory argument?

j-mracek: @jan-kolarik I am fine with overloaded method. I am sorry for complication. I need to know how we use the method and during investigation I discovered strange behavior. I can create issue to not forget to change behavior of the advisory command. Is it ok?

jan-kolarik: Yes, it sounds even better to me having the issue instead of the TODO. I just wasn't sure if it's not an overkill for this case :)

j-mracek:
LGTM

j-mracek (approved):

j-mracek:
@jan-kolarik It looks like that some test is failing. May I ask you to investigate the issue.

jan-kolarik:
> @jan-kolarik It looks like that some test is failing. May I ask you to investigate the issue.

Yes, these are the API bindings tests, because of the change in the API. There is a PR mentioned in the previous comment fixing this (https://github.com/rpm-software-management/ci-dnf-stack/pull/1200).

j-mracek:
LGTM

j-mracek:
Thank you very much  for your contribution

PR #219, opened 2023-01-04
swig: Fix TransactionItemReason wrappers

### LINKED ISSUE #216: Python Bindings: Memory leak in `Package.get_reason()`
In this gist (https://gist.github.com/gotmax23/9786917434121c55165e512bf8e7f37d), there are two Python scripts that query for `yt-dlp` and print the `reason` attr. One uses dnf and the other uses libdnf5. 

dnf returns either a string or None. libdnf5 returns a SwigPyObject that doesn't allow accessing the value and results in a memory leak. I guess libdnf5 should return None here because it doesn't have access to dnf's history db.

``` console
$ ./reason.py
yt-dlp-2022.11.11-1.fc37.noarch
<class 'str'>
weak-dependency
```

``` console
$ ./reason5.py
yt-dlp-2022.11.11-1.fc37.noarch
<class 'SwigPyObject'>
<Swig Object of type 'libdnf::transaction::TransactionItemReason *' at 0x7fe0c5964270>
swig/python detected a memory leak of type 'libdnf::transaction::TransactionItemReason *', no destructor found
```

jan-kolarik:
Include transaction module prior to dependent modules, otherwise enum wrappers are not ready for these modules.

Resolves #216.

### DIFF
--- a/bindings/libdnf5/comps.i
+++ b/bindings/libdnf5/comps.i
@@ -14,6 +14,7 @@
 
 %import "common.i"
 %import "repo.i"
+%import "transaction.i"
 
 %{
     #include "libdnf/comps/group/package.hpp"
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -14,6 +14,7 @@
 
 %import "common.i"
 %import "conf.i"
+%import "transaction.i"
 
 %exception {
     try {
--- a/test/python3/libdnf5/rpm/test_package_query.py
+++ b/test/python3/libdnf5/rpm/test_package_query.py
@@ -130,3 +130,10 @@ class TestPackageQuery(base_test_case.BaseTestCase):
         self.assertEqual(nevra.get_name(), "pkg")
         self.assertEqual(nevra.get_arch(), "x86_64")
         self.assertFalse(nevra.has_just_name())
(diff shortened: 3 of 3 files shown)


jan-kolarik:
I will also add the CI API test for this.

gotmax23:
This seems to fix the problem. I now get

``` console
$ ./reason.py
yt-dlp-2022.11.11-1.fc37.noarch
<class 'str'>
weak-dependency
$ ./reason5.py
yt-dlp-2022.11.11-1.fc37.noarch
<class 'int'>
4
$ python -c 'from libdnf5.transaction import TransactionItemReason_WEAK_DEPENDENCY; print(TransactionItemReason_WEAK_DEPENDENCY)'
4
```

gotmax23:
In the tests, would it make sense to also assert that `package.get_reason()` equals `libdnf5.transaction.TransactionItemReason_NONE`?

jan-kolarik:
> In the tests, would it make sense to also assert that `package.get_reason()` equals `libdnf5.transaction.TransactionItemReason_NONE`?

Actually it's much better than comparing the value to `0` as the enum definition could change in the future. Thanks for the suggestion, I'll change it immediately.

jrohel:
OK

PR #221, opened 2023-01-09
Fix rawhide build by removing Perl's `-D_FORTIFY_SOURCE=2`

kontura:
Perl's config value `ccflags`: https://perldoc.perl.org/Config#ccflags
is defined during its build. Presumably the current build still uses old
settings `-D_FORTIFY_SOURCE=2`.

Once Perl is recompiled with the new `-D_FORTIFY_SOURCE=3` setting this
commit could be reverted.


j-mracek:
LGTM

PR #230, opened 2023-01-12
conf: Change OptionStringSet implementation to std::set

jan-kolarik:
Because `std::unordered_set` is not supported in some bindings yet (even in new SWIG 4.1), moving to `std::set`. We don't expect lot of stored values in these options, so there should be no performance regression.

Also added a commit with some small refactoring to reference defined `OptionStringSet::ValueType` where it makes sense to avoid possible changes in future when internal container implementation is changed.


jrohel:
OK

PR #226, opened 2023-01-10
Add methods for configuration options to simplify setting values during runtime

jan-kolarik:
Overloads for `set` methods were added in order to simplify usage of configuration option setters for API users.

Related issue: #66.

PR #231, opened 2023-01-12
Add python binding tests for configuration options

jan-kolarik:
Basic python binding tests for configuration options covering some of the #226 and #123 functionality.

Requires: #230


kontura (approved):
Thanks!

Issue #233, opened 2023-01-16
dnf5 does not respect exclude key in .repo files


D4N:
I have added the hashicorp repository to my `/etc/yum.repos.d/` and explicitly excluded the `vagrant` package from that repo so that I get the Fedora package:
```
[hashicorp]
name=Hashicorp Stable - $basearch
baseurl=https://rpm.releases.hashicorp.com/fedora/$releasever/$basearch/stable
enabled=1
gpgcheck=1
gpgkey=https://rpm.releases.hashicorp.com/gpg
exclude=vagrant

[hashicorp-test]
name=Hashicorp Test - $basearch
baseurl=https://rpm.releases.hashicorp.com/fedora/$releasever/$basearch/test
enabled=0
gpgcheck=1
gpgkey=https://rpm.releases.hashicorp.com/gpg
```

However, `dnf5` still tries to update vagrant using the `hashicorp` repository:
```
Package                            Arch    Version                                        Repository            Size
Upgrading:
 vagrant                           x86_64  2.3.4-1                                        hashicorp        249.1 MiB
   replacing vagrant               noarch  2.2.19-5.fc37                                  fedora             2.4 MiB
```

m-blaha:
Thanks for filing this issue!
It looks like we do not have this use case covered with any test. I'll prepare a test scenario.

m-blaha:
Turned out that the tests are present, but were not enabled for dnf5 - https://github.com/rpm-software-management/ci-dnf-stack/pull/1208 enables them.

PR #234, opened 2023-01-16
Fix repository specific excludes

m-blaha:
rq.filter_id(disable_excludes, libdnf::sack::QueryCmp::NOT_GLOB) in case of empty `disable_excludes` filters out all repositories.

Resolves https://github.com/rpm-software-management/dnf5/issues/233

### DIFF
--- a/libdnf/common/sack/match_int64.cpp
+++ b/libdnf/common/sack/match_int64.cpp
@@ -59,11 +59,10 @@ bool match_int64(int64_t value, QueryCmp cmp, int64_t pattern) {
 
 
 bool match_int64(int64_t value, QueryCmp cmp, const std::vector<int64_t> & patterns) {
-    bool result = false;
+    bool result = (cmp & libdnf::sack::QueryCmp::NOT) == libdnf::sack::QueryCmp::NOT;
     for (auto & pattern : patterns) {
-        if (match_int64(value, cmp, pattern)) {
-            result = true;
-            break;
+        if (match_int64(value, cmp, pattern) != result) {
+            return !result;
         }
     }
     return result;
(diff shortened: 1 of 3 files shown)


m-blaha:
There is also another way how to fix it.

What is the desired behavior of `bool match_string(const std::string & value, QueryCmp cmp, const std::vector<std::string> & patterns)` function in case the `patterns` vector is empty and the `cmp` comparator is negative? - e.g. `rq.filter_id([], libdnf::sack::QueryCmp::NOT_GLOB)` in this issue, which leads to `match_string(repo_id, libdnf::sack::QueryCmp::NOT_GLOB, [])`

I believe it should return `true`. Any comments?

https://github.com/rpm-software-management/dnf5/blob/1329442c4c2c58d046823e7a83ce7cf45fd25e92/libdnf/common/sack/match_string.cpp#L88-L97

kontura:
> I believe it should return `true`. Any comments?

It would also make sense to me. It seems like a cleaner solution, especially since we have tests for this. 
If we modify `match_string()` I would also change `match_int64()` in the same way.

m-blaha:
I tried to fix the behavior of `match_string()` function in case one (or both) of operands are vectors and the `cmp` operator is negative. I tried to describe the new behavior both in commit message and in the comments.
The fix for `match_int64()` is still not present, but I'd appreciate a partial review here - to make sure that the logic change makes sense.

kontura:
The PR looks good, but I think that in the commit msg: `match_string("fedora", NOT_GLOB, ["fedora", "testing"]) would return true` is wrong. "fedora" is in the string list and since we use `NOT_GLOB` it would return false no?

m-blaha:
I should have written the commit message more clearly. The part you mentioned is an example of clearly wrong behaviour of current (unfixed by this pach) implementation. Let me try to re-formulate the commit message.

kontura:
I see, thanks, it is much more understandable for me now.

Issue #222, opened 2023-01-09
distro-sync shows a lot of unknown problems


xosevp:
f37 + dnf5-5.0.3-20230109005556.1.g1602a9fa.fc37.x86_64:

Same command with dnf-4.14.0-1.fc37.noarch is clean.

Thanks.

```
# dnf5 --best  --skip-broken distro-sync
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem 1: cannot install the best update candidate for package c-ares-1.17.2-3.fc37.x86_64
  - package c-ares-1.17.2-3.module_f37+15051+531fd4a4.x86_64 is filtered out by exclude filtering
 Problem 2: cannot install the best update candidate for package community-mysql-common-8.0.31-1.fc37.x86_64
  - package community-mysql-common-8.0.31-1.module_f37+15640+52ef8dc9.x86_64 is filtered out by exclude filtering
 Problem 3: cannot install the best update candidate for package ghc-Cabal-3.2.1.0-121.fc37.x86_64
  - package ghc-Cabal-3.6.0.0-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
 Problem 4: cannot install the best update candidate for package ghc-base-4.14.3.0-121.fc37.x86_64
  - package ghc-base-4.16.0.0-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
 Problem 5: cannot install the best update candidate for package ghc-base-devel-4.14.3.0-121.fc37.x86_64
  - package ghc-base-devel-4.16.0.0-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
 Problem 6: cannot install the best update candidate for package ghc-binary-0.8.8.0-121.fc37.x86_64
  - package ghc-binary-0.8.9.0-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
 Problem 7: cannot install the best update candidate for package ghc-bytestring-0.10.12.0-121.fc37.x86_64
  - package ghc-bytestring-0.11.1.0-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
 Problem 8: cannot install the best update candidate for package ghc-compiler-8.10.7-121.fc37.x86_64
  - package ghc-compiler-9.2.1-101.module_f37+14056+25172961.x86_64 is filtered out by exclude filtering
[...]

PR #242, opened 2023-01-20
Make distrosync as a targeted operation

j-mracek:
DNF provides all candidates for transaction to libsolv, therefore we have to force libsolv to use only those candidates for the job solution => to make job TARGETED. If distrosync or upgrade job is not marked as TARGETED explicitly, libsolv mark as TARGETED job without installed package in the set of candidates.

The patch unify upgrade and distrosync flags in DNF5 => using SOLVER_TARGETED.

Resolves: https://github.com/rpm-software-management/dnf5/issues/222


kontura:
It would be great to have a CI test for case.

j-mracek:
> It would be great to have a CI test for case.

Good point

j-mracek:
> 

What about a test - https://github.com/rpm-software-management/ci-dnf-stack/pull/1211

PR #243, opened 2023-01-20
Implementation of autodetecting modular platform ID

jan-kolarik:
Platform ID for modules is automatically detected based on the following values in given order sorted by priority:
1. Use `module_platform_id` configuration option value, if it is explicitly set up.
2. Parse `Provides` value from the latest available package in enabled repositories.
3. Parse `Provides` value from the latest installed package on the local system.
4. Parse `PLATFORM_ID` value from `os-release` files on the local system.

The logic is implemented based on the original [PR](https://github.com/rpm-software-management/libdnf/pull/712) from `libdnf` component.

### DIFF
--- a/libdnf/module/module_item.cpp
+++ b/libdnf/module/module_item.cpp
@@ -329,6 +329,7 @@ std::string ModuleItem::get_yaml() const {
 }
 
 
+/// A common function to create a solvable based on passed parameters.
 static void create_solvable_worker(
     Pool * pool,
     Solvable * solvable,
--- a/libdnf/module/module_item.cpp
+++ b/libdnf/module/module_item.cpp
@@ -329,37 +329,62 @@ std::string ModuleItem::get_yaml() const {
 }
 
 
-void ModuleItem::create_solvable() {
-    Pool * pool = module_sack->p_impl->pool;
-
-    // Create new solvable and store its id
-    id = ModuleItemId(repo_add_solvable(pool_id2repo(pool, Id(module_sack->p_impl->repositories[repo_id]))));
-    Solvable * solvable = pool_id2solvable(pool, id.id);
-
+static void create_solvable_worker(
+    Pool * pool,
+    Solvable * solvable,
+    const std::string & name,
+    const std::string & stream,
+    const std::string & version,
+    const std::string & context,
+    const char * arch) {
     // Name: $name:$stream:$context
-    solvable_set_str(solvable, SOLVABLE_NAME, get_name_stream_staticcontext().c_str());
+    solvable_set_str(solvable, SOLVABLE_NAME, fmt::format("{}:{}:{}", name, stream, context).c_str());
     // Version: $version
-    solvable_set_str(solvable, SOLVABLE_EVR, get_version_str().c_str());
+    solvable_set_str(solvable, SOLVABLE_EVR, version.c_str());
     // TODO(pkratoch): The test can be removed once modules always have arch
     // Arch: $arch (if arch is not defined, set "noarch")
-    auto arch = modulemd_module_stream_get_arch(md_stream);
     solvable_set_str(solvable, SOLVABLE_ARCH, arch ? arch : "noarch");
     // Store original $name:$stream in description
-    solvable_set_str(solvable, SOLVABLE_DESCRIPTION, get_name_stream().c_str());
+    solvable_set_str(solvable, SOLVABLE_DESCRIPTION, fmt::format("{}:{}", name, stream).c_str());
 
     // Create Provides: module($name)
-    std::string provide = fmt::format("module({})", get_name());
-    auto dep_id = pool_str2id(pool, provide.c_str(), 1);
+    auto dep_id = pool_str2id(pool, fmt::format("module({})", name).c_str(), 1);
     solvable_add_deparray(solvable, SOLVABLE_PROVIDES, dep_id, -1);
     // Create Conflicts: module($name)
     solvable_add_deparray(solvable, SOLVABLE_CONFLICTS, dep_id, 0);
     // Create Provides: module($name:$stream)
-    provide = fmt::format("module({})", get_name_stream());
-    dep_id = pool_str2id(pool, provide.c_str(), 1);
+    dep_id = pool_str2id(pool, fmt::format("module({}:{})", name, stream).c_str(), 1);
     solvable_add_deparray(solvable, SOLVABLE_PROVIDES, dep_id, -1);
 
     // TODO(pkratoch): Maybe store original context in summary
     // solvable_set_str(solvable, SOLVABLE_SUMMARY, original_context.c_str());
+}
+
+
+void ModuleItem::create_solvable() {
+    auto pool = module_sack->p_impl->pool;
+
+    // Create new solvable and store its id
+    id = ModuleItemId(repo_add_solvable(pool_id2repo(pool, Id(module_sack->p_impl->repositories[repo_id]))));
+    auto solvable = pool_id2solvable(pool, id.id);
+    auto context = computed_static_context.empty()
+                       ? libdnf::utils::string::c_to_str(modulemd_module_stream_get_context(md_stream))
+                       : computed_static_context;
+    auto arch = modulemd_module_stream_get_arch(md_stream);
+
+    create_solvable_worker(pool, solvable, get_name(), get_stream(), get_version_str(), std::move(context), arch);
+}
+
+
+void ModuleItem::create_platform_solvable(
+    const ModuleSackWeakPtr & module_sack, const std::string & name, const std::string & stream) {
+    auto pool = module_sack->p_impl->pool;
+
+    // TODO(jkolarik): Create constants for known repo ids
+    auto id = repo_add_solvable(repo_create(pool, "@System"));
+    auto solvable = pool_id2solvable(pool, id);
+
+    create_solvable_worker(pool, solvable, name, stream, "0", "00000000", "noarch");
 }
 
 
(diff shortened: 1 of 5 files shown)


libdnf/module/module_item.cpp
```diff
@@ -329,40 +329,65 @@ std::string ModuleItem::get_yaml() const {
...
+                       : computed_static_context;
+    auto arch = modulemd_module_stream_get_arch(md_stream);
+
+    create_solvable_worker(pool, solvable, get_name(), get_stream(), get_version_str(), std::move(context), arch);
+}
+
+
+void ModuleItem::create_platform_solvable(
+    const ModuleSackWeakPtr & module_sack, const std::string & name, const std::string & stream) {
+    auto pool = module_sack->p_impl->pool;
```
j-mracek, line 381: Maybe I overlooked something. It looks like that module_sack is referenced in ModuleSack::Impl only for passing modular pool for `create_platform_solvable`. Is it correct?

j-mracek: The method issupposed to be type ONCE - Is there any protection to not call it multiple times?

jan-kolarik: Yes, but I didn't want to expose the libsolv `Pool`'s headers in the `*.hpp`. Or is it OK?

jan-kolarik: > Maybe I overlooked something. It looks like that module_sack is referenced in ModuleSack::Impl only for passing modular pool for `create_platform_solvable`. Is it correct?

Yes, basically I didn't want to expose `libsolv` headers in our public `*.hpp`.

jan-kolarik: > The method issupposed to be type ONCE - Is there any protection to not call it multiple times?

I tried there to adhere the same workflow as in the DNF. If I understood that correctly, the virtual platform package is there added when doing a modules filtering.

In DNF5 the modules filtering is done "once" at the end of repos loading. And during repos loading also other solvables are added. But as I know there is no true protection to not do that multiple times using the API method `update_and_load_enabled_repos`.

jan-kolarik:
And also CI API tests need to be updated after accepting this.

libdnf/module/module_item.cpp
j-mracek: In DNF4, there was a different situation. Libdnf did not have access to DNF conf and the same approach setting modules was using by PackageKit. In DNF5, modularity is better implemented in workflow, therefore we could define other point for platform loading. The platform detection just before the first module is added to libsolv has advantage for libsolv id to ModuleItem conversion. The first solvable will be always the platform.

libdnf/module/module_sack.cpp
```diff
@@ -478,6 +482,149 @@ std::pair<std::vector<std::vector<std::string>>, ModuleSack::ModuleErrorType> Mo
...
+    auto index = platform_id.find(':');
+    if (index == std::string::npos) {
+        return {};
+    } else {
+        return make_pair(platform_id.substr(0, index), platform_id.substr(index + 1));
+    }
+}
+
+
+static std::pair<std::string, std::string> parse_platform_id_from_file(const std::string & os_release_path) {
```
j-mracek, line 495: May I ask you why the function throw runtime_error? Anyway it is always catch and logged and then the code continue?

jan-kolarik: It was mentioned to terminate on any problem when reading the `os-release` file and then just log the message with the error, because we could have more files from where the platform ID could be detected, so we continue with another attempts. This is basically taken over from `DNF` and I am open for discussion for any better solution. Or is there some specific problem with this approach?

j-mracek: No, there is no a problem with the approach. When I see that construct I always try to understand details. In the function there is an alternative source of exceptions `libdnf::utils::fs::File file{os_release_path, "r"};`, which will require catch anyway. May be could you add a short description of the method that it raises not only `std::runtime_error` but also `std::filesystem::filesystem_error`? Do we expect any other type of error?

libdnf/module/module_sack.cpp
```diff
@@ -478,6 +482,149 @@ std::pair<std::vector<std::vector<std::string>>, ModuleSack::ModuleErrorType> Mo
...
+    // TODO(jkolarik): Create constants for paths
+    std::vector<std::string> os_release_paths{"etc/os-release", "usr/lib/os-release"};
+    auto & installroot = base->get_config().installroot().get_value();
+    for (auto & os_release_path : os_release_paths) {
+        std::string full_path = std::filesystem::canonical(installroot) / os_release_path;
+        std::pair<std::string, std::string> parsed_platform_id;
+
+        try {
+            parsed_platform_id = parse_platform_id_from_file(full_path);
+        } catch (const std::exception & except) {
```
j-mracek, line 612: Very minor: The catch is quite wide . I believe that only two exception are expected.

libdnf/module/module_sack.cpp
```diff
@@ -478,6 +482,149 @@ std::pair<std::vector<std::vector<std::string>>, ModuleSack::ModuleErrorType> Mo
...
+std::pair<std::string, std::string> ModuleSack::Impl::detect_platform_name_and_stream() const {
+    // try to detect platform id from configuration
+    auto & config_platform_option = base->get_config().module_platform_id();
+    if (!config_platform_option.empty()) {
+        auto & config_platform = config_platform_option.get_value();
+        auto parsed_platform_id = parse_platform_id_from_string(config_platform);
+        if (!parsed_platform_id.first.empty() && !parsed_platform_id.second.empty()) {
+            return parsed_platform_id;
+        } else {
+            base->get_logger()->warning("Invalid format of platform ID in configuration: {}", config_platform);
```
j-mracek, line 560: Minor: The message could contain the name of configuration option.

libdnf/module/module_sack.cpp
```diff
@@ -478,6 +482,149 @@ std::pair<std::vector<std::vector<std::string>>, ModuleSack::ModuleErrorType> Mo
...
+        }
+
+        if (!parsed_platform_id.first.empty() && !parsed_platform_id.second.empty()) {
+            return parsed_platform_id;
+        } else {
+            base->get_logger()->warning("Error when parsing platform ID from release path: {}", full_path);
+        }
+    }
+
+    throw std::runtime_error("No valid Platform ID detected");
```
j-mracek, line 624: I am sorry but I have a feeling that terminating DNF here is not a proper action. Platform ID is not an essential part of modularity design. It is a hack from Modular Build System, that we have to handle somehow. Do you have some reason to terminate DNF when platform is not around?

jan-kolarik: No, I don't have any valid reason, it's just my lack of knowledge about how the platform ID is internally handled by modularity and what are all the consequences of not having the virtual platform solvable present.

libdnf/module/module_sack.cpp
```diff
@@ -265,6 +266,9 @@ ModuleSack::Impl::collect_data_for_modular_filtering() {
 void ModuleSack::Impl::module_filtering() {
     auto [include_NEVRAs, exclude_NEVRAs, names, src_names, reldep_name_list] = collect_data_for_modular_filtering();
 
+    auto platform_id = detect_platform_name_and_stream();
+    ModuleItem::create_platform_solvable(module_sack->get_weak_ptr(), platform_id.first, platform_id.second);
```
j-mracek, line 270: I think you are correct. This is the best place to detect platform ID. Because the detection requires all RPM metadata loaded, therefore we cannot call it earlier. I only want to ask you to ensure that the detection will be performed only once. In future we will call modular filtering method multiple times therefore we have to avoid adding platform more than once.

libdnf/module/module_sack.cpp
```diff
@@ -265,6 +266,9 @@ ModuleSack::Impl::collect_data_for_modular_filtering() {
 void ModuleSack::Impl::module_filtering() {
     auto [include_NEVRAs, exclude_NEVRAs, names, src_names, reldep_name_list] = collect_data_for_modular_filtering();
```
j-mracek, line 267: To avoid unnecessary warning and work we have to test first whether there are any modular data in ModuleSack (first line). If it is empty it have to return without doing anything.

j-mracek (changes requested):
I think it looks fine, but there are few issues that I would like to see modified.

libdnf/module/module_sack.cpp
jan-kolarik: Definitely

libdnf/module/module_sack.cpp
jan-kolarik: Agree, better to catch only the things we know about and don't suppress any other potential problems.

libdnf/module/module_sack.cpp
j-mracek: I would suggest to replace it by a warning, info, or debug message. May be debug message would be a better option.

libdnf/module/module_sack.cpp
jan-kolarik: I've changed the exception for parsing problems to the custom one. I don't know about adding the description about internal raising of the filesystem exception. I don't see this approach used in other classes and also this is just a static function in the `.cpp`. But of course if you find it useful, I will add it.

jan-kolarik:
> Note: was not yet properly tested as modules `enable`, `disable`, etc. functionality is missing for now.

Functionality for now successfully tested using the original `DNF` command for enabling a specific module stream and then trying to do a `repoquery` and `install` with DNF5 and checking if seeing modular packages there.

libdnf/module/module_sack.cpp
j-mracek: It would be better to describe what kind of exception the method could raise. It will make life easier for future internal users.

libdnf/module/module_sack.cpp
j-mracek: Report it as debug messages is a very good suggestion.

jan-kolarik: Logging level was changed to debug and regex format was added for the `module_platform_id` option.

libdnf/module/module_sack.cpp
jan-kolarik: Docs were updated.

j-mracek (approved):
LGTM

PR #245, opened 2023-01-23
RepoSack::update_and_load_repos: First load repositories from cache

jrohel:
Work optimization. The order in which the repositories are loaded into the sack is changed. Repositories with a valid cache are sent to the `thread_sack_loader` before downloading other repositories.


m-blaha (approved):
Thank you, LGTM.

Issue #237, opened 2023-01-16
very slow update with local files


xosevp:
It spends a lot of cpu time reading data from the disk.

dnf-4.14.0-1.fc37.noarch is much faster:

```
# echo 1 > /proc/sys/vm/drop_caches  ; time dnf --allowerasing -b upgrade /mnt/mybook_R/fc37/*.rpm
[...]
real    0m7.292s
user    0m2.502s
sys     0m0.449s
```
vs. dnf5-5.0.4-0.20230116005643.5.0.3+17.g1329442c.fc37.x86_64
```
# echo 1 > /proc/sys/vm/drop_caches  ; time dnf5 --best upgrade /mnt/mybook_R/fc37/*.rpm
[...]

Nothing to do.

real    1m3.951s
user    0m30.432s
sys     0m2.079s
```
```
----- very slow FROM HERE ------
# dnf5  upgrade /mnt/mybook_R/fc37/*.rpm
Updating and loading repositories:
Repositories loaded.
Package "firefox.x86_64" is already installed.
Package "firefox.x86_64" is already installed.
Package "firefox.x86_64" is already installed.
Package "firefox-langpacks.x86_64" is already installed.
Package "firefox-langpacks.x86_64" is already installed.
Package "firefox-langpacks.x86_64" is already installed.
Package "firefox-x11.x86_64" is already installed.
Package "firefox-x11.x86_64" is already installed.
Package "firefox-x11.x86_64" is already installed.
Package "iproute.x86_64" is already installed.
Packages for argument 'iproute-devel-6.0.0-1.fc37.x86_64' available, but not installed.
Packages for argument 'iproute-doc-6.0.0-1.fc37.x86_64' available, but not installed.
Package "iproute-tc.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel.x86_64" is already installed.
Package "kernel-core.x86_64" is already installed.
Package "kernel-core.x86_64" is already installed.
Package "kernel-core.x86_64" is already installed.
Package "kernel-core.x86_64" is already installed.
Package "kernel-core.x86_64" is already installed.
[...]

PR #247, opened 2023-01-24
reposack: No checksums for cmdline repo packages

m-blaha:
Calculating checksums of a package in libsolv is pretty expensive and we do not need this feature for cmdline repo packages.

Related: https://github.com/rpm-software-management/dnf5/issues/237

### DIFF
--- a/include/libdnf/repo/repo_sack.hpp
+++ b/include/libdnf/repo/repo_sack.hpp
@@ -88,8 +88,10 @@ public:
 
     /// Add given paths to comdline repository.
     /// @param paths Vector of paths to rpm files to be inserted to cmdline repo. Can contain paths to local files or URLs of remote rpm files. Specifications that are neither file paths, nor URLs are ignored.
+    /// @param calculate_checksum Whether libsolv should calculate and store checksum of added packages. Setting to true significantly reduces performance.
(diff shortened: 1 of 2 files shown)


j-mracek:
It is a simple solution and ok for our current API. But our current API has one disadvantage - the value is hard-coded in RepoSack::add_cmdline_packages(). The user case for checksum calculation is somehow related to RPM-OSTREE (https://github.com/rpm-software-management/libdnf/pull/415). What about to make it configurable using config?
The advantage of configuration is that DNF5 command could modify it if required. Other option would be to modify API - RepoSack::add_cmdline_packages(bool add_checksum = false). What do you think?

m-blaha:
Are you aware of any use case for switching check sums on? To be honest, I do not know exactly what is the advantage of having checksums calculated on libsolv level and why one should need it.
I do not have problems with adding a `add_checksum` parameter to `add_cmdline_packages()`. You are right that a config option would be more flexible. I'm just not convinced that this particular functionality is worth of a new config option.

j-mracek:
LGTM

j-mracek:
Please feel free to merge it when all tests pass. I rescheduled one test, because it failed for no reason - timeout in copr I guess.

m-blaha:
Thanks. Now it passed.

PR #250, opened 2023-01-25
rpm: Correctly order rpm transaction items

m-blaha:
Once the transaction is filled, it must be sorted correctly. Otherwise the transaction (especially the "pre" scriptlets) may end up with unsatisfied dependencies.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2164378


pmatilai (approved):
Yep. Simple but rather elemental :laughing:

kontura:
Thank you!

PR #251, opened 2023-01-25
Add `transaction.test()`

evan-goode:
Allows dry running a transaction using `RPMTRANS_FLAG_TEST` by calling `transaction.verify()`, regardless of what tsflags are set on the base config. So you could run:

```
    libdnf::Goal goal(base);
    auto transaction = goal.resolve();
    auto result = transaction.verify()
    // check result of verification, possibly change the transaction

    transaction.run(...)
```

without creating a whole new `base` with the RPMTRANS_FLAG_TEST flag set.

Example use case: https://bugzilla.redhat.com/show_bug.cgi?id=2109660

### DIFF
--- a/include/libdnf/base/transaction.hpp
+++ b/include/libdnf/base/transaction.hpp
@@ -81,6 +81,14 @@ public:
     /// @return the transaction groups.
     std::vector<libdnf::base::TransactionGroup> & get_transaction_groups() const;
 
+    /// Check the transaction by running it with RPMTRANS_FLAG_TEST. The import
+    /// of any necessary public keys will be requested, and transaction checks
+    /// will be performed, but no changes to the installed package set will be
+    /// made. These checks are performed automatically by run(); it is
+    /// redundant to call test() before calling run().
+    /// @return An enum describing the result of the transaction
+    TransactionRunResult test();
+
     /// Prepare, check and run the transaction. All the transaction metadata
     /// (`description` and `comment`) are stored in the history database.
     ///
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -222,6 +222,10 @@ std::string Transaction::transaction_result_to_string(const TransactionRunResult
     return {};
 }
 
+Transaction::TransactionRunResult Transaction::test() {
+    return p_impl->test();
+}
+
 Transaction::TransactionRunResult Transaction::run(
     std::unique_ptr<libdnf::rpm::TransactionCallbacks> && callbacks,
     const std::string & description,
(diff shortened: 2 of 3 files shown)


pmatilai:
Please call it "test" or "dry-run"  such, but not "verify". Verify is already overloaded as a term (in rpm context) and we do not need more confusion in that area.

glum23:
I'd say when this PR will be merged, execute_transaction function from dnf-behave-tests/dnf/steps/cmd.py should be updated.

test/tutorial/transaction/transaction.cpp
```diff
@@ -76,6 +77,17 @@ class TransactionCallbacks : public libdnf::rpm::TransactionCallbacks {
...
+//
+// `transaction.run` will do all the same checks, so this step is not
+// necessary, but it may be useful in some workflows.
+std::cout << std::endl << "Verifying the transaction" << std::endl;
+libdnf::base::Transaction::TransactionRunResult res = transaction.verify();
+if (res != libdnf::base::Transaction::TransactionRunResult::SUCCESS) {
+    std::cout << std::endl << "Verifying transaction failed!" << std::endl;
+    throw;
+}
+
```
m-blaha, line 90: I'm not sure, whether including the `transaction.verify()` call to the tutorial is a good choice.
The use case for the dry-run is rather marginal and once it gets into the tutorial, I'm afraid that people will start copying it to their projects which is unnecessary in most cases (or even harmful in terms of performance).

evan-goode: Totally, I was not planning to keep it there. Looks like `dnf-behave-tests/dnf/steps/cmd.py` is the place as Eva said.

evan-goode:
Changed to `transaction.test()`.

include/libdnf/base/transaction.hpp
```diff
@@ -80,6 +80,10 @@ class Transaction {
     /// @return the transaction groups.
     std::vector<libdnf::base::TransactionGroup> & get_transaction_groups() const;
 
+    /// Check the transaction by running it with RPMTRANS_FLAG_TEST.
+    /// @return An enum describing the result of the transaction
```
j-mracek, line 84: I have one suggestion - <please could you mention that the test is automatically performed by run() therefore calling test() and run() will be not efficient and not needed.

evan-goode: Done, thanks.

j-mracek:
@evan-goode May I ask you to resolve the merge conflicts?

include/libdnf/base/transaction.hpp
```diff
@@ -81,6 +81,12 @@ class Transaction {
     /// @return the transaction groups.
     std::vector<libdnf::base::TransactionGroup> & get_transaction_groups() const;
 
+    /// Check the transaction by running it with RPMTRANS_FLAG_TEST. This check
+    /// is performed automatically by run(); it is redundant to call test()
+    /// before calling run().
```
j-mracek, line 86: There is one thing that deserves to be documented - import of GPG keys. If I am correct - test in DNF5 and in DNF also imports GPG keys. It is let say incorrect - test is not supposed to do permanent changes. On the other hand - we need that behavior for workflow in system upgrade and offline upgrades.

evan-goode: Thanks, it's a good idea to make that explicit. Is the language correct here?

    /// Check the transaction by running it with RPMTRANS_FLAG_TEST. Packages
    /// will be downloaded, public keys will be installed, and transaction
    /// checks will be performed, but no changes to the package set will be
    /// made. These checks are performed automatically by run(); it is
    /// redundant to call test() before calling run().

j-mracek: I believe that this is incorrect - `Packages will be downloaded`. The method does not download packages.

j-mracek: About `public keys will be installed` - this is not wrong, but it performs complex tasks related to signed rpm keys verification. First of all - the test of keys is only applied when configuration requires it. The workflow - it tries to verify signature using already imported keys in rpm-db. If it fails then it downloads keys, request approval for import (callback), than it imports keys and request the verification of the RPM signature.
`install` vs. `import` keys into RPM DB - I somehow do not prefer `install`.

j-mracek: `to the package set` - wat about to add `installed in the statement - `to the installed package set`.

evan-goode: > I believe that this is incorrect - Packages will be downloaded. The method does not download packages.

Thanks, yes of course, oops!

> About public keys will be installed - this is not wrong...

Changed to "the import of any necessary public keys will be requested".

 > to the package set - what about to add installed in the statement - to the installed package set`.

Done.

j-mracek:
LGTM

j-mracek (approved):

Issue #229, opened 2023-01-11
Problems with DNF5Dbus API


timlau:
I have created some python code to test out the DNF5Dbus API

https://github.com/timlau/testbed/blob/main/dnf5/dbus_client.py

I can't get the resolve method to work

```
/bin/python /home/tim/udv/work/testbed/dnf5/dbus_client.py
11:59:32 INFO  : (dasbus.connection) -  Connecting to the system bus.
11:59:32 INFO  : (dasbus.connection) -  Connecting to the system bus.
11:59:32 DEBUG : (dnf5dbus) -  Open Dnf5Daemon session: /org/rpm/dnf/v0/7285a2ec5311d0ee5a5d7d453ce22581
Searching for 0xFFFF
Found : 1
FOUND: 0xFFFF-0.10-1.fc37.x86_64                repo: fedora
installing
depsolve
==========> Error calling dnf5daemon :
libdnf/./solv/pool.hpp:228: void libdnf::solv::Pool::swap_considered_map(libdnf::solv::SolvMap&): Assertion 'other_considered_map.allocated_size() == 0 || other_considered_map.allocated_size() >= get_nsolvables()' failed: The considered map is smaller than the number of solvables in the pool
11:59:34 DEBUG : (dnf5dbus) -  Close Dnf5Daemon session: /org/rpm/dnf/v0/7285a2ec5311d0ee5a5d7d453ce22581
```

m-blaha:
Similar bugreport: https://bugzilla.redhat.com/show_bug.cgi?id=2184323

PR #252, opened 2023-01-26
dnfdaemon: Fix double loading of system repo

m-blaha:
There are paths in dbus workflows that will cause system repo to be
loaded twice resulting in assertion error:
```
libdnf/./solv/pool.hpp:228: void libdnf::solv::Pool::swap_considered_map(libdnf::solv::SolvMap&): Assertion 'other_considered_map.allocated_size() == 0 || other_considered_map.allocated_size() >= get_nsolvables()' failed: The considered map is smaller than the number of solvables in the pool
```
Resolves: https://github.com/rpm-software-management/dnf5/issues/229

### DIFF
--- a/dnf5daemon-server/session.cpp
+++ b/dnf5daemon-server/session.cpp
@@ -142,14 +142,8 @@ bool Session::wait_for_key_confirmation(const std::string & key_id, sdbus::Signa
 }
 
 void Session::fill_sack() {
-    if (session_configuration_value<bool>("load_system_repo", true)) {
-        get_base()->get_repo_sack()->get_system_repo()->load();
-    }
-
-    if (session_configuration_value<bool>("load_available_repos", true)) {
-        if (!read_all_repos()) {
-            throw std::runtime_error("Cannot load repositories.");
-        }
+    if (!read_all_repos()) {
+        throw std::runtime_error("Cannot load repositories.");
     }
 }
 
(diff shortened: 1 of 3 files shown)


dnf5daemon-server/session.cpp
```diff
@@ -163,20 +157,25 @@ bool Session::read_all_repos() {
...
+            repo->set_callbacks(std::make_unique<DbusRepoCB>(*this));
+        }
+
+        try {
+            base->get_repo_sack()->update_and_load_enabled_repos(session_configuration_value<bool>("load_system_repo", true));
+        } catch (const std::runtime_error & ex) {
+            retval = false;
+        }
+    } else if (session_configuration_value<bool>("load_system_repo", true)) {
+        base->get_repo_sack()->get_system_repo()->load();
```
j-mracek, line 178: Let's discuss the API around loading repositories. To allow multiple points of repository load will create problem with application of modularity rules. We must prevent user to load multiple times any repository including system repository.

dnf5daemon-server/session.cpp
```diff
@@ -163,20 +157,26 @@ bool Session::read_all_repos() {
...
+            repo->set_callbacks(std::make_unique<DbusRepoCB>(*this));
+        }
+
+        try {
+            base->get_repo_sack()->update_and_load_enabled_repos(
+                session_configuration_value<bool>("load_system_repo", true));
+        } catch (const std::runtime_error & ex) {
+            retval = false;
+        }
+    } else if (session_configuration_value<bool>("load_system_repo", true)) {
```
jan-kolarik, line 178: Can we load the `session_configuration_value<bool>("load_available_repo", true)` and `session_configuration_value<bool>("load_system_repo", true)` before the `if` block and then just use it inside? It would improve the readability of the code I'd say and not repeating the load for `load_system_repo`...

m-blaha: Sure, done.

jan-kolarik:
Thanks, looks good!

jan-kolarik (approved):

PR #253, opened 2023-01-26
swig: Fixes for transaction related bindings

jan-kolarik:
Allow accessing methods from `TransactionItem` base class and iterating over transaction history items.


jrohel:
It looks OK.

Issue #249, opened 2023-01-25
DNF4-accepted reposdir=/dev/null doesn't work


praiskup:
`dnf5-5.0.4-2.fc38.x86_64` (Rawhide) is executed like (by Mock):
```
$ /usr/bin/dnf5 '--installroot' '/var/lib/mock/fedora-rawhide-x86_64-bootstrap/root/' '--releasever' '38' install star --config /var/lib/mock/fedora-rawhide-x86_64-bootstrap/root/etc/dnf/dnf.conf
filesystem error: directory iterator cannot open directory: Not a directory [/dev/null]
```

In the config there's something like:
```
[main]
reposdir=/dev/null
```

praiskup:
Blocks: https://github.com/rpm-software-management/mock/pull/1027

PR #256, opened 2023-01-27
Handle std::filesystem::directory_iterator errors

m-blaha:
Setting *dir config options to non-existing or non-accessible directory leads to unnecessary crash.
In case of reposdir also a warning is written to the log.

Resolves: https://github.com/rpm-software-management/dnf5/issues/249

### DIFF
--- a/dnf5/commands/clean/clean.cpp
+++ b/dnf5/commands/clean/clean.cpp
@@ -116,8 +116,9 @@ void CleanCommand::run() {
     auto & ctx = get_context();
     fs::path cachedir{ctx.base.get_config().cachedir().get_value()};
 
+    std::error_code ec;
     libdnf::repo::RepoCache::RemoveStatistics statistics{};
-    for (const auto & dir_entry : std::filesystem::directory_iterator(cachedir)) {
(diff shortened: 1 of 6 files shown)


dnf5/commands/clean/clean.cpp
```diff
@@ -116,8 +116,9 @@ void CleanCommand::run() {
     auto & ctx = get_context();
     fs::path cachedir{ctx.base.get_config().cachedir().get_value()};
 
+    std::error_code ec;
     libdnf::repo::RepoCache::RemoveStatistics statistics{};
-    for (const auto & dir_entry : std::filesystem::directory_iterator(cachedir)) {
+    for (const auto & dir_entry : std::filesystem::directory_iterator(cachedir, ec)) {
```
j-mracek, line 121: Is it intended, that issues are silent?

m-blaha: In most cases yes. But thinking about this particular case - we can print a warning to std::err.

m-blaha:
The CI tests are failing due to syntax error in feature file - https://github.com/rpm-software-management/ci-dnf-stack/pull/1213

PR #260, opened 2023-01-31
modules: Add NSVCAP parsing and ModuleQuery::filter_nsvca

pkratoch:


libdnf/module/module_query.cpp
```diff
@@ -159,4 +160,28 @@ void ModuleQuery::filter_latest(int limit) {
 }
 
 
+void ModuleQuery::filter_nsvca(const Nsvcap & nsvcap, libdnf::sack::QueryCmp cmp) {
+    const std::string & name = nsvcap.get_name();
+    if (name != "") {
```
j-mracek, line 165: I am sorry, I overlooked it. std::string has method empty that could be used instead. The difference is that the code compare the string form instead of asking whether length inside is 0. If I am correct, std::string knows its length therefore it should be cheaper.

`if (!name.empty()) {`

libdnf/module/nsvcap.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+
+static const std::vector<Nsvcap::Form> MODULE_SPEC_FORMS{
+    Nsvcap::Form::N,
+    Nsvcap::Form::NP,
+    Nsvcap::Form::NS,
+    Nsvcap::Form::NSP,
+    Nsvcap::Form::NSV,
+    Nsvcap::Form::NSVP,
+    Nsvcap::Form::NSC,
```
j-mracek, line 74: May I ask you to add additional form - `NSCP`?
Minor - May I also ask you to move NSC and NSCP before `NSV`?

libdnf/module/module_query.cpp
pkratoch: Right, fixed.

libdnf/module/nsvcap.cpp
pkratoch: Added and moved.

j-mracek (approved):
LGTM

libdnf/module/nsvcap.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+const std::vector<Nsvcap::Form> & Nsvcap::get_default_module_spec_forms() {
+    return MODULE_SPEC_FORMS;
+}
+
+
+bool Nsvcap::parse(const std::string nsvcap_str, Nsvcap::Form form) {
+    enum { NAME = 1, STREAM = 2, VERSION = 3, CONTEXT = 4, ARCH = 5, PROFILE = 6, _LAST_ };
+    std::smatch matched_result;
+    auto matched = std::regex_match(nsvcap_str, matched_result, NSVCAP_FORM_REGEX[static_cast<unsigned>(form) - 1]);
```
ppisar, line 97: I worry this std::regex_match() call will suffer from the same issue as I fixed for RPM in #268.

j-mracek: Correct, but creating an issue is sufficient.

pkratoch: I think we can resolve it the same way: https://github.com/rpm-software-management/dnf5/pull/299

PR #261, opened 2023-02-01
Set POOL_FLAG_WHATPROVIDESWITHDISABLED to libsolv pool

j-mracek:
POOL_FLAG_WHATPROVIDESWITHDISABLED ensures that excluded packages are not taken as candidates for solver. This is specific behavior is unique for DNF, because we runs make provides ready on top of all packages including excluded packages.

Resolves: https://github.com/openSUSE/libsolv/issues/518

### DIFF
--- a/libdnf/module/module_sack_impl.hpp
+++ b/libdnf/module/module_sack_impl.hpp
@@ -38,7 +38,11 @@ class ModuleGoalPrivate;
 
 class ModuleSack::Impl {
 public:
-    Impl(const BaseWeakPtr & base) : base(base), module_metadata(base), pool(pool_create()) {}
+    /// Create libsolv pool and set POOL_FLAG_WHATPROVIDESWITHDISABLED to ensure excluded packages are not taken as
+    /// candidates for solver
+    Impl(const BaseWeakPtr & base) : base(base), module_metadata(base), pool(pool_create()) {
+        pool_set_flag(pool, POOL_FLAG_WHATPROVIDESWITHDISABLED, 1);
+    }
     ~Impl() { pool_free(pool); }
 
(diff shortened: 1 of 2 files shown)


kontura:
I think the Resolves link you mentioned should probably be: https://github.com/rpm-software-management/dnf5/issues/222?

If that is the case we need a new test for the issue since https://github.com/rpm-software-management/ci-dnf-stack/pull/1211 seems to work even without the `POOL_FLAG_WHATPROVIDESWITHDISABLED` flag.

j-mracek:
> I think the Resolves link you mentioned should probably be: #222?
> 
> If that is the case we need a new test for the issue since [rpm-software-management/ci-dnf-stack#1211](https://github.com/rpm-software-management/ci-dnf-stack/pull/1211) seems to work even without the `POOL_FLAG_WHATPROVIDESWITHDISABLED` flag.


This patch also remove solver error, but the original patch (added targeted) change behavior according to DNF5 behavior. The difference is if I specify exact version of installed package then you will see the difference. 

installed
a-1-1.noarch

Available
a-1-1.noarch
a-2-2.noarch

command 
dnf5 distro-sync a-1-1.noarch

The current behavior
No action

Without targeted flag:
updrade to a-2-2.noarch

kontura:
I see, both `POOL_FLAG_WHATPROVIDESWITHDISABLED` or setting the job as `targeted` resolve the issue.

I was just thinking how to create a test for this (for when the `POOL_FLAG_WHATPROVIDESWITHDISABLED` is not set). Maybe using some other command than `distro-sync` that is not `targeted`?

kontura (approved):

PR #263, opened 2023-02-03
PackageQuery::filter_unneeded() and wrapper for libsolv Solver class

m-blaha:
- introduces new filter_unneeded for package queries. This is important for implementation of  `autoremove` and `list --autoremove` commands
- also wraps libsolv `Solver` class into `libdnf::solv::Solver`. This enables safer memory management and sharing the debugdata writing code.

### DIFF
--- a/include/libdnf/rpm/package_query.hpp
+++ b/include/libdnf/rpm/package_query.hpp
@@ -611,7 +611,7 @@ public:
     /// @param timestamp Only packages built after this will pass
     void filter_recent(const time_t timestamp);
 
-    /// Keep in the query only user-installed packages.
+    /// Keep in the query only installed packages that are user-installed.
     void filter_userinstalled();
 
     /// Filter unneeded packages. Unneeded packages are those which are installed as
--- a/include/libdnf/rpm/package_query.hpp
+++ b/include/libdnf/rpm/package_query.hpp
@@ -611,6 +611,9 @@ public:
     /// @param timestamp Only packages built after this will pass
     void filter_recent(const time_t timestamp);
 
+    /// Keep in the query only user-installed packages.
+    void filter_userinstalled();
+
     /// Filter unneeded packages. Unneeded packages are those which are installed as
     /// dependencies and are not required by any user-installed package any more.
     void filter_unneeded();
--- a/include/libdnf/rpm/package_query.hpp
+++ b/include/libdnf/rpm/package_query.hpp
@@ -611,6 +611,10 @@ public:
     /// @param timestamp Only packages built after this will pass
     void filter_recent(const time_t timestamp);
 
+    /// Filter unneeded packages. Unneeded packages are those which are installed as
+    /// dependencies and are not required by any user-installed package any more.
+    void filter_unneeded();
+
     // TODO(jmracek) return std::pair<bool, std::unique_ptr<libdnf::rpm::Nevra>>
     // @replaces libdnf/sack/query.hpp:method:std::pair<bool, std::unique_ptr<Nevra>> filterSubject(const char * subject, HyForm * forms, bool icase, bool with_nevra, bool with_provides, bool with_filenames);
     std::pair<bool, libdnf::rpm::Nevra> resolve_pkg_spec(
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -2456,6 +2456,17 @@ void PackageQuery::filter_recent(const time_t timestamp) {
     }
 }
 
+void PackageQuery::filter_userinstalled() {
+    filter_installed();
+    for (const auto & pkg : *this) {
+        auto reason = pkg.get_reason();
+        if (reason == libdnf::transaction::TransactionItemReason::WEAK_DEPENDENCY ||
+            reason == libdnf::transaction::TransactionItemReason::DEPENDENCY) {
+            p_impl->remove_unsafe(pkg.get_id().id);
+        }
+    }
+}
+
 void PackageQuery::filter_unneeded() {
     auto & pool = get_rpm_pool(p_impl->base);
 
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -2470,34 +2470,23 @@ void PackageQuery::filter_userinstalled() {
 void PackageQuery::filter_unneeded() {
     auto & pool = get_rpm_pool(p_impl->base);
 
-    // get a list of all installed packages
     auto * installed_repo = pool->installed;
     if (installed_repo == nullptr) {
         (*p_impl).clear();
         return;
     }
 
-    libdnf::solv::SolvMap installed_solv_map(pool.get_nsolvables());
-    for (int it = installed_repo->start; it < installed_repo->end; ++it) {
-        installed_solv_map.add(it);
-    }
-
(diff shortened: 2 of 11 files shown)


j-mracek:
@m-blaha May I ask you to resolve the merge conflict?

libdnf/solv/solver.hpp
```diff
@@ -0,0 +1,90 @@
...
+extern "C" {
+#include <solv/solver.h>
+}
+
+
+namespace libdnf::solv {
+
+class Solver {
+public:
+    Solver(){};
```
j-mracek, line 37: I am not sure whether we should have default constructor together with constructor requiring pool.
But it is a minor question. Structure is private, and not initialized Solver class is well described.

m-blaha: Yeah, I know. I can try to use unique_ptr<Solver> in GoalPrivate and ModuleGoalPrivate classes to get rid of the the default ctor. Then the uninitialized Solver would be just nullptr.

libdnf/rpm/package_query.cpp
```diff
@@ -2453,6 +2456,56 @@ void PackageQuery::filter_recent(const time_t timestamp) {
...
+    auto & pool = get_rpm_pool(p_impl->base);
+
+    // get a list of all installed packages
+    auto * installed_repo = pool->installed;
+    if (installed_repo == nullptr) {
+        (*p_impl).clear();
+        return;
+    }
+
+    libdnf::solv::SolvMap installed_solv_map(pool.get_nsolvables());
```
j-mracek, line 2469: What about to use a PackageQuery instead? It could replace `installed_solv_map` and `installed_pset`
```
libdnf::rpm::PackageQuery installed{p_impl->base, FLAG};
installed.filter_installed()
```

libdnf/rpm/package_query.cpp
```diff
@@ -2453,6 +2456,56 @@ void PackageQuery::filter_recent(const time_t timestamp) {
...
+    for (int it = installed_repo->start; it < installed_repo->end; ++it) {
+        installed_solv_map.add(it);
```
j-mracek, line 2471: This construct could add non rpm elements to map or not system packages to map - the repository in principle could be interrupted - in future it could create a problems, but not right now.

libdnf/rpm/package_query.cpp
```diff
@@ -2453,6 +2456,56 @@ void PackageQuery::filter_recent(const time_t timestamp) {
...
+    }
+
+    // find user-installed packages and add them to the solver job
+    libdnf::rpm::PackageSet installed_pset{p_impl->base, installed_solv_map};
+    libdnf::solv::IdQueue userinstalled;
+    for (const auto & pkg : installed_pset) {
+        auto reason = pkg.get_reason();
+        // package that is not dep / weak dep is considered user-installed here
+        if (reason != libdnf::transaction::TransactionItemReason::WEAK_DEPENDENCY &&
+            reason != libdnf::transaction::TransactionItemReason::DEPENDENCY) {
```
j-mracek, line 2481: Minor: The similar test is also done by SWDB (swdb.cpp line 479). The question is whether it would be better to have a one point for the decision for both logic?

libdnf/solv/solver.hpp
```diff
@@ -0,0 +1,90 @@
...
+    int set_flag(int flag, int value);
+    int get_decisionlevel(Id p);
+    ::Transaction * create_transaction();
+    unsigned int problem_count();
+    IdQueue findallproblemrules(Id problem);
+    IdQueue allruleinfos(Id rid);
+    IdQueue get_cleandeps();
+    const char * problemruleinfo2str(SolverRuleinfo type, Id source, Id target, Id dep);
+    int describe_decision(Id p, Id * infop);
+    SolverRuleinfo ruleclass(Id rid);
```
j-mracek, line 81: May I ask you to add a description to the methods?

j-mracek (changes requested):
There is only one issue - creation of solv map with installed package. Remaining comments are only minor suggestions.

libdnf/rpm/package_query.cpp
m-blaha: That's a good point, thank you!

libdnf/rpm/package_query.cpp
m-blaha: The question is whether it should be in Swdb. It's not used anywhere (I guess it is a dnf4 heritage) and now, with system state, to resolve userinstalled packages  swdb is not needed.

m-blaha: I'll create a new `filter_userinstalled()` filter in PackageQuery and delete the unused (and I think even broken) code from Swdb.

libdnf/rpm/package_query.cpp
```diff
@@ -2456,6 +2456,17 @@ void PackageQuery::filter_recent(const time_t timestamp) {
     }
 }
 
+void PackageQuery::filter_userinstalled() {
+    filter_installed();
+    for (const auto & pkg : *this) {
+        auto reason = pkg.get_reason();
+        if (reason == libdnf::transaction::TransactionItemReason::WEAK_DEPENDENCY ||
+            reason == libdnf::transaction::TransactionItemReason::DEPENDENCY) {
+            p_impl->remove_unsafe(pkg.get_id().id);
```
j-mracek, line 2465: This is a minor note but interesting - p_impl->remove_unsafe() is in most of cases slower then creating empty PackageSet, add passing candidate and do `and` operator. The trick is the number operations. If I have 1000 packages and 100 is user installed then it requires 900 time operation on bitmap. The other approach requires only 100 modifications and then and operation is very cheap. The first version of DNF5 queries used remove_unsafe() approach and affected performance was significantly affected.

include/libdnf/rpm/package_query.hpp
```diff
@@ -611,6 +611,9 @@ class PackageQuery : public PackageSet {
     /// @param timestamp Only packages built after this will pass
     void filter_recent(const time_t timestamp);
 
+    /// Keep in the query only user-installed packages.
```
j-mracek, line 614: My note will sound not logic but I would suggest to add explicitly that it keeps installed packages. I remember some discussion related to reason and what reason is supposed to be returned for available packages, therefore I would prefer to be explicit here.

What about:
`/// Keep in the query only installed packages that are user-installed.`

libdnf/solv/solver.hpp
```diff
@@ -0,0 +1,120 @@
...
+    void init(::Pool * pool);
+
+    /// drop current `solver` object
+    void reset();
+
+    /// returns true if `solver` exists
+    bool is_initialized() { return solver != nullptr; };
+
+    /// Write solver debug data to given directory
+    /// @param with_transaction Whether transaction data are dumped
```
j-mracek, line 66: I am sorry, I do not understand the description or a use case.

m-blaha: If you mean the with_transaction parameter, it's used in `libdnf/repo/repo_sack.cpp` in this PR.
If true, it sets `TESTCASE_RESULT_TRANSACTION | TESTCASE_RESULT_PROBLEMS` flags in libsolv's testcase_write call.

j-mracek:
I believe that the build fail for rawhide is not related to the PR.

PR #264, opened 2023-02-03
PackageQuery: Add `filter_extras`

m-blaha:
Filter installed "extra" packages - packages which are installed, but not available in any repository.

### DIFF
--- a/include/libdnf/rpm/package_query.hpp
+++ b/include/libdnf/rpm/package_query.hpp
@@ -587,8 +587,18 @@ public:
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
-    /// Filter packages, which are installed but not available in any enabled repository
-    void filter_extras();
+    /// Keep in the query only packages that are installed but not available in any
+    /// enabled repository. Even excluded packages (e.g. using excludepkgs config
+    /// option) are considered as available in repositories for the purpose of extras
+    /// calculation.
+    /// Those installed packages that are only part of non-active modules are also
+    /// considered as extras.
+    /// @param exact_evr If false (default) extras calculation is based only on
+    ///                  `name.arch`. That means package is not in extras if any version
+    ///                  of the package exists in any of the enabled repositories.
+    ///                  If true, filter_extras is more strict and returns each package
+    ///                  which exact NEVRA is not present in any enabled repository.
+    void filter_extras(const bool exact_evr = false);
 
     // TODO(jmracek) return std::pair<bool, std::unique_ptr<libdnf::rpm::Nevra>>
     // @replaces libdnf/sack/query.hpp:method:std::pair<bool, std::unique_ptr<Nevra>> filterSubject(const char * subject, HyForm * forms, bool icase, bool with_nevra, bool with_provides, bool with_filenames);
--- a/include/libdnf/rpm/package_query.hpp
+++ b/include/libdnf/rpm/package_query.hpp
@@ -587,6 +587,9 @@ public:
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
+    /// Filter packages, which are installed but not available in any enabled repository
+    void filter_extras();
+
     // TODO(jmracek) return std::pair<bool, std::unique_ptr<libdnf::rpm::Nevra>>
     // @replaces libdnf/sack/query.hpp:method:std::pair<bool, std::unique_ptr<Nevra>> filterSubject(const char * subject, HyForm * forms, bool icase, bool with_nevra, bool with_provides, bool with_filenames);
     std::pair<bool, libdnf::rpm::Nevra> resolve_pkg_spec(
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -2395,4 +2395,14 @@ void PackageQuery::filter_duplicates() {
 }
 
 
+void PackageQuery::filter_extras() {
+    filter_installed();
+    // Create query with available packages without non-modular excludes.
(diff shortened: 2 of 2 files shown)


j-mracek:
LGTM, but there is one request for a different implementation of extras filter. [Bug 1800910](https://bugzilla.redhat.com/show_bug.cgi?id=1800910) - dnf list extras stopped to shows some packages that should.
I think we should also provide that functionality, but I have no idea, how to delivery it for commandline.

m-blaha:
Hm, yeah. The implementation in the patch uses the same approach as current dnf4 - extras are detected using `name.arch`.
Maybe adding new switch (and then also a filter), e.g. `--extras-versions` which can use algorithm that takes also package versions into account? What do you think?

m-blaha:
I've added a commit which extends `filter_extras()` with optional `with_versions` parameter. If this is set to `true`, the whole NEVRA is used to detect extra packages.

include/libdnf/rpm/package_query.hpp
```diff
@@ -587,8 +587,9 @@ class PackageQuery : public PackageSet {
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
-    /// Filter packages, which are installed but not available in any enabled repository
-    void filter_extras();
+    /// Filter packages, which are installed but not available in any enabled repository.
+    /// @param with_versions If false (default) extras calculation is based only on `name.arch`. That means package is not in extras if any version of the package exists in the repository. If true, filter_extras is more strict and returns each package which exact NEVRA is not present in any repository.
+    void filter_extras(const bool with_versions=false);
```
j-mracek, line 592: Nice, technically it is not a version but EVR.

Just a suggestion. What about to name the variable as `exact_evr` or `exact_EVR`?

include/libdnf/rpm/package_query.hpp
```diff
@@ -587,6 +587,10 @@ class PackageQuery : public PackageSet {
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
+    /// Filter packages, which are installed but not available in any enabled repository.
+    /// @param with_versions If false (default) extras calculation is based only on `name.arch`. That means package is not in extras if any version of the package exists in any of the enabled repositories. If true, filter_extras is more strict and returns each package which exact NEVRA is not present in any enabled repository.
```
j-mracek, line 591: What about to replace `and returns each package which exact NEVRA is not present in any repository.` by `and keeps in query installed packages which exact NEVRA is not present in any repository.`

include/libdnf/rpm/package_query.hpp
```diff
@@ -587,6 +587,10 @@ class PackageQuery : public PackageSet {
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
+    /// Filter packages, which are installed but not available in any enabled repository.
```
j-mracek, line 590: What about to replace `Filter packages` by `Keep packages`?
May I also ask you to describe the behavior with regular excluded packages?

include/libdnf/rpm/package_query.hpp
m-blaha: Naming, the hardest part of the coding...
I agree, your suggested name for parameter is clearer.

include/libdnf/rpm/package_query.hpp
m-blaha: Sure. Done.

include/libdnf/rpm/package_query.hpp
```diff
@@ -587,6 +587,16 @@ class PackageQuery : public PackageSet {
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
+    /// Keep in query only packages that are installed but not available in any enabled
+    /// repository. Also installed packages that are only part of non-active modules
+    /// are considered extras.
```
j-mracek, line 592: Nice improvement, What about to `considered extras` change  to `considered as extras`?

libdnf/rpm/package_query.cpp
```diff
@@ -2395,4 +2395,18 @@ void PackageQuery::filter_duplicates() {
...
+    filter_installed();
+    // Create query with available packages without non-modular excludes.
+    // As extras should be considered also packages in non-active modules.
+    PackageQuery available(p_impl->base, ExcludeFlags::IGNORE_REGULAR_EXCLUDES);
+    available.filter_available();
+    if (exact_evr) {
+        filter_nevra(available, sack::QueryCmp::NEQ);
+    } else {
+        filter_name_arch(available, sack::QueryCmp::NEQ);
+    }
```
j-mracek, line 2408: I am really exited how easily you were able to implemented the filter.

include/libdnf/rpm/package_query.hpp
m-blaha: Done. Sorry for all those mistakes in my English.

include/libdnf/rpm/package_query.hpp
```diff
@@ -587,6 +587,16 @@ class PackageQuery : public PackageSet {
     // TODO(dmach): consider removing the installed packages during the filtering
     void filter_priority();
 
+    /// Keep in query only packages that are installed but not available in any enabled
```
j-mracek, line 590: May I ask you to add a note that filter ignores regular excludes?

j-mracek:
LGTM, feel free to merge it when tests pass.

j-mracek (approved):

include/libdnf/rpm/package_query.hpp
m-blaha: oops. Sorry, I missed this comment. Should be fixed now.

PR #265, opened 2023-02-03
PackageQuery: Add `filter_recent(timestamp)`

m-blaha:
Recent packages are those packages, that were built after the given timestamp.
Needed for `dnf list --recent` implementation.


kontura:
LGTM!

PR #268, opened 2023-02-06
ReldepParser: Reject overlong dependency strings

ppisar:
Passing a very long package name triggered a segfault:

    # dnf5 downgrade $(perl -e 'print q{A} x 13390')
    Repositories loaded.
    Segmentation fault (core dumped)

This is caused by a known bug in GCC std::regex_match().

This patch prevents from the crash by rejecting strings longer than 1 KB. Another solution would be to move to a heap-based regular expression engine, like PCRE2.

<https://bugzilla.redhat.com/show_bug.cgi?id=2164792> <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164>

### DIFF
--- a/libdnf/solv/reldep_parser.cpp
+++ b/libdnf/solv/reldep_parser.cpp
@@ -26,6 +26,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace libdnf::solv {
 
 static const std::regex RELDEP_REGEX("^(\\S*)\\s*(\\S*)?\\s*(\\S*)$");
+#define MAX_RELDEP_LENGTH 2 << 10
 
 static bool set_cmp_type(libdnf::rpm::Reldep::CmpType * cmp_type, std::string cmp_type_string, long int length) {
     if (length == 2) {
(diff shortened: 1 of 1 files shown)


j-mracek:
I am really sorry but I am unable to reproduce the issue with `dnf5-5.0.5-20230207005715.11.g3388f6fc.fc36.x86_64`. Also I don't believe that `dnf5 downgrade` command uses regex at all. @ppisar  May I ask you to verify the reproducer?

ppisar:
It happens both with downgrade and with install. The exact length when the bug triggers varies based how the code is compiled (compiler version, compiler options). I can reproduce it with a code built from the latest git tree (commit 3092cd669d3c4a294244cdc485af1756dfe2cd18), that build requires a longer argument:

~~~~
# /tmp/build/dnf5/dnf5 upgrade $(perl -e 'print q{A} x 53390')
 systemd                                                       100% |   0.0   B/s |   3.0 KiB |  00m00s
Repositories loaded.
Segmentation fault (core dumped)
~~~~

Running it later under debugger and listing outmost 16 frames confirms it's the same crash.

Where can I found dnf5-5.0.5-20230207005715.11.g3388f6fc.fc36.x86_64? Fedora has dnf5-5.0.5-1.fc38.x86_64.

ppisar:
I can see you pushed new commits, I will push a rebased patch.

kontura:
> Where can I found dnf5-5.0.5-20230207005715.11.g3388f6fc.fc36.x86_64? Fedora has dnf5-5.0.5-1.fc38.x86_64.

That should be from our nightly Copr: https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable/

ppisar:
The same crash with unstable dnf5:

~~~~
# dnf5 install $(perl -e 'print q{A} x 53390') 
Repositories loaded.
Segmentation fault (core dumped)
root@fedora-38:~ # rpm -q dnf5
dnf5-5.0.5-20230208005815.21.g3092cd66.fc38.x86_64
~~~~

j-mracek:
I confirm that it is reproducible.

j-mracek (approved):
LGTM

Issue #257, opened 2023-01-27
Missing logic around installroot


j-mracek:
DNF configuration uses several paths that can be related to intallroot or not. As an example lets use `reposdir` configuration option with value `/etc/yum.repos.d/`. If dnf is used in installroot `/tmp/test`, the path is expected to be `/tmp/test/etc/yum.repos.d/`

In DNF4 there is an auto-detection that decide whether path is supposed be modify with `installroot` prefix or not. The mechanism is well described in DNF man pages (option `--installroot`) - see https://dnf.readthedocs.io/en/latest/command_ref.html.
The auto-detection in DNF4 has two weaknesses - it does not work correctly on RHEL and there is no way in API how to enable the auto-detection. Therefore we discussed to replace the auto-detection by two options.

- [ ] Design possibility how to set DNF5 (from commandline) to take configurations from root or installroot path
Solution examples:
  - conflicting options `--installroot=<path>`, `--root=<path>`
  - `--installroot=<path>`, `--use_root`
- [ ] Design API interface to modify paths in configuration class
Example of possibilities
  - Method in base
  - Optional argument of Base::setup()
  - Configuration (bool)

- [ ] Document the new behavior also as a change to DNF4

- [ ] Resolves https://github.com/rpm-software-management/dnf5/issues/248

- [ ] Enable tests in our CI for installroot

evan-goode:
To clarify:
- So `--installroot=/mnt` would look for a configuration file at `/mnt/etc/dnf/dnf.conf`, then error if that file doesn't exist, without falling back to `/etc/dnf/dnf.conf`? And if `--use-root` is _also_ passed, then `/etc/dnf/dnf.conf` is used without looking for `/mnt/etc/dnf/dnf.conf`. Is that what you had in mind?
- What would `--root=<path>` mean?
- Do we need to worry about the case where `installroot` is set in `/etc/dnf/dnf.conf`  and we need to chain load another config file? It might be best if `--installroot` were available only as a command-line option to avoid confusion about this case.
- I kind of like the current behavior in DNF4 where we fall back to the regular paths if the files don't exist in the installroot. Even if the new behavior remained similar to DNF4, we could expose a function in `Base` to search the installroot and modify these paths.

j-mracek:
> To clarify:
> 
[...]

Issue #248, opened 2023-01-25
dnf5 --installroot /the/root doesn't read  /the/root/etc/dnf/dnf.conf


praiskup:
From `man 7 dnf5-installroot`

       ...
       configuration  file  and  reposdir  are searched inside the installroot
       first.  If they are not present, they are taken from the host system.

       Note: When a path is specified within a command line  argument  (--con‐
       fig=CONFIG_FILE_PATH  in case of configuration file and --setopt=repos‐
       dir=REPO_DIR_PATH for reposdir), then this path is always  relative  to
       the host with no exceptions.
       ...

The first paragraph is wrong, but the second is correct;  using a host-relative
path to the correct dnf.conf works-around the problem.

edit: we should fix dnf5 so it reads the config from --installroot, if present

praiskup:
Blocks: https://github.com/rpm-software-management/mock/pull/1027

praiskup:
Just to make it clear;  IMO it is definitely desired that DNF5 reads the configuration
from the --installroot directory (if present).  I am not requesting a documentation fix.

j-mracek:
The issue will be resolved as a part of bigger task. Do you need to resolve your issue first ar can you wait until we will heve the complete solution related usage of installroot.

praiskup:
It would be nice to have this fixed rather sooner than later.  I think we
should have Mock prepared before Fedora switches to DNF5.

I'm not sure what is the ETA for the bigger --installroot task, and how
quickly is this issue in particular fixable.  So intuitively I bet it is
the DNF team who has to decide.  Not sure.

The thing is that having this fixed, I could finish the fixes on Mock side
(related to DNF5 support) and release.  Then DNF team can start
experimenting with mock, etc.  The idea was to prepare an early CI task
for Mock+DNF5, so we don't bring some accidental breakage during the GA
time.

praiskup:
Since #271 is merged, I think we can close this.  Do we have a testing build somewhere?

m-blaha:
We have nightly builds in https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable/ copr repository.

praiskup:
Did you want to write https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/ ?

j-mracek:
Yoy can try to download builds from PR checks - see https://github.com/rpm-software-management/dnf5/pull/271/checks

[...]

PR #271, opened 2023-02-08
Load config and reposdir from installroot

evan-goode:
Loads dnf.conf and reposdir from the installroot, unless --use_root_config is passed. Resolves https://github.com/rpm-software-management/dnf5/issues/257 and https://github.com/rpm-software-management/dnf5/issues/248.

This behavior is different from the current behavior of DNF4. When given an installroot, DNF4 will look for a config file and reposdir in the installroot, but if they are not present it will fall back to the default config file and/or reposdir. In this PR we only consider the config file and reposdir in the installroot, regardless of whether the files are there.

When `--use_root_config` is passed, the setting of installroot will not affect the config file path or reposdir.

The reposdir taken from the installroot is given priority greater than that specified in dnf.conf, but less priority than a reposdir specified by a plugin, so that a plugin could still override the setting.

The config file path can still be overridden with the command-line argument `--config`. Perhaps a similar `--reposdir` argument would be useful too?

The documentation still needs to be updated to reflect the changed behavior and the new `--use_root_config` flag.

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -410,6 +410,13 @@ void RootCommand::set_argument_parser() {
     installroot->link_value(&config.get_installroot_option());
     global_options_group->register_argument(installroot);
 
+    auto use_host_environment = parser.add_new_named_arg("use-host-environment");
+    use_host_environment->set_long_name("use-host-environment");
+    use_host_environment->set_description("use configuration, reposdir, and vars from the host system rather than the installroot");
+    use_host_environment->set_const_value("true");
+    use_host_environment->link_value(&config.use_host_environment());
+    global_options_group->register_argument(use_host_environment);
+
     auto releasever = parser.add_new_named_arg("releasever");
     releasever->set_long_name("releasever");
     releasever->set_has_value(true);
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -410,12 +410,13 @@ void RootCommand::set_argument_parser() {
     installroot->link_value(&config.get_installroot_option());
     global_options_group->register_argument(installroot);
 
-    auto use_host_environment = parser.add_new_named_arg("use-host-environment");
-    use_host_environment->set_long_name("use-host-environment");
-    use_host_environment->set_description("use configuration, reposdir, and vars from the host system rather than the installroot");
-    use_host_environment->set_const_value("true");
-    use_host_environment->link_value(&config.use_host_environment());
-    global_options_group->register_argument(use_host_environment);
+    auto use_host_config = parser.add_new_named_arg("use-host-config");
+    use_host_config->set_long_name("use-host-config");
+    use_host_config->set_description(
+        "use configuration, reposdir, and vars from the host system rather than the installroot");
+    use_host_config->set_const_value("true");
+    use_host_config->link_value(&config.get_use_host_config_option());
+    global_options_group->register_argument(use_host_config);
 
     auto releasever = parser.add_new_named_arg("releasever");
     releasever->set_long_name("releasever");
@@ -639,6 +640,10 @@ int main(int argc, char * argv[]) try {
 
     auto command = context.get_selected_command();
 
+    // Gets set to true when any repository is created from configuration or a
+    // .repo file
+    bool any_repos_from_system_configuration = false;
+
     try {
         command->pre_configure();
 
@@ -674,6 +679,7 @@ int main(int argc, char * argv[]) try {
         base.setup();
 
         base.get_repo_sack()->create_repos_from_system_configuration();
+        any_repos_from_system_configuration = base.get_repo_sack()->size() > 0;
 
         context.apply_repository_setopts();
 
@@ -718,6 +724,15 @@ int main(int argc, char * argv[]) try {
 
             context.download_and_run(*context.get_transaction());
         }
+    } catch (libdnf::cli::GoalResolveError & ex) {
+        if (!any_repos_from_system_configuration && base.get_config().get_installroot_option().get_value() != "/" &&
+            !base.get_config().get_use_host_config_option().get_value()) {
+            std::cout << "No repositories were loaded from the installroot. To use the configuration and repositories "
+                         "of the host system, pass --use-host-config."
+                      << std::endl;
+        }
+        std::cerr << ex.what() << std::endl;
+        return static_cast<int>(libdnf::cli::ExitCode::ERROR);
     } catch (libdnf::cli::ArgumentParserMissingCommandError & ex) {
         // print help if no command is provided
         std::cerr << ex.what() << std::endl;
--- a/doc/misc/installroot.7.rst
+++ b/doc/misc/installroot.7.rst
@@ -30,21 +30,21 @@ relative to where all packages will be installed. Think of it like doing
 ``chroot <root> dnf``, except using ``--installroot`` allows ``DNF5`` to work 
 before the chroot is created.
 
-`cachedir`, `log` files, `releasever`, and `gpgkey` are taken from or stored in 
-the installroot. Gpgkeys are imported into the installroot from a path relative 
-to the host which can be specified in the repository section of configuration files.
+`cachedir`, `log` files, `releasever`, and `gpgkey` are taken from or stored in
+the installroot. GPG keys are imported into the installroot from a path
+relative to the host which can be specified in the repository section of
+configuration files.
 
-`configuration` file and `reposdir` are searched inside the installroot first. 
-If they are not present, they are taken from the host system. 
+`configuration` file, `reposdir`, and `vars` are taken from inside the
+installroot, unless the command-line argument ``--use-host-environment`` is
+passed, in which case the configuration and environment from the host system
+will be used.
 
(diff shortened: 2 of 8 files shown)


praiskup:
> When --use_root_config is passed, the setting of installroot will not affect the config file path or reposdir.

Does it mean that `--installroot <foo> --use_root_config` reverts the effect of this patch, i.e. host configuration will be used?

praiskup:
Can you update the manpage `dnf5-installroot.7`?  That could help us to better understand the change.

j-mracek:
Additional info - the option could be named as `--use-host-environment` and it would be nice to provide a hint when setting in installroot is insufficient - like due to absence of repositories - the command failed then we could suggest to use the new option. The condition should be strict because we don't want to overload user with unnecessary information.

doc/misc/installroot.7.rst
```diff
@@ -30,21 +30,21 @@ relative to where all packages will be installed. Think of it like doing
...
+`configuration` file, `reposdir`, and `vars` are taken from inside the
+installroot, unless the command-line argument ``--use-host-environment`` is
+passed, in which case the configuration and environment from the host system
+will be used.
+
+Note: When a path is specified within a command line argument
+(``--config=CONFIG_FILE_PATH`` in case of `configuration` file,
+``--setopt=reposdir=/path/to/repodir`` for `reposdir` or
+``--setopt=varsdir=/paths/to/varsdir`` for `vars`), then this path is always
+relative to the host with no exceptions.
```
praiskup, line 47: Nice, thank you for the update!

evan-goode: Sure thing! And yes, your above comment is correct.

evan-goode:
> Additional info - the option could be named as `--use-host-environment` and it would be nice to provide a hint when setting in installroot is insufficient - like due to absence of repositories - the command failed then we could suggest to use the new option. The condition should be strict because we don't want to overload user with unnecessary information.

Done. The situation when we want to print a warning is pretty specific, so I'm not sure my implementation is very elegant. Let me know what you think.

dnf5/main.cpp
```diff
@@ -415,6 +415,14 @@ void RootCommand::set_argument_parser() {
     installroot->link_value(&config.installroot());
     global_options_group->register_argument(installroot);
 
+    auto use_host_environment = parser.add_new_named_arg("use-host-environment");
```
praiskup, line 418: Just a small nit:  "host environment" seems like some /home directory was shared, or some set of environment variables are shared.  Since this is about configuration, I'd spell it "use-host-config".

j-mracek: It looks like that the name of the option is the only thing to discuss.

dnf5/main.cpp
```diff
@@ -714,6 +724,15 @@ int main(int argc, char * argv[]) try {
 
             context.download_and_run(*context.get_transaction());
         }
+    } catch (libdnf::cli::GoalResolveError & ex) {
+        if (!any_repos_from_system_configuration && base.get_config().installroot().get_value() != "/" &&
+            !base.get_config().use_host_environment().get_value()) {
+            std::cout << "No repositories were loaded from the installroot. Pass --use-host-environment to use the "
+                         "configuration and repositories of the host system."
+                      << std::endl;
```
j-mracek, line 732: Minor comment: 
```
Pass --use-host-environment to use the configuration and repositories of the host system.
```
What about to rotate the order. Firs - To use configuration and repositories of the host system pass <option>.
But please use your feelings more then my suggestion.

I somehow feel that the original statement is more strong then the second one - first condition then what to do.

dnf5/main.cpp
```diff
@@ -634,6 +642,7 @@ int main(int argc, char * argv[]) try {
     }
 
     auto command = context.get_selected_command();
+    bool any_repos_from_system_configuration = false;
```
j-mracek, line 645: May I ask you for add a description what it means or what it contains. I think it should be `true` when any repository is created from .repo, configuration option - `--repofrompath`, or commandline packages were specified argument `<>.rpm`.

dnf5/main.cpp
```diff
@@ -714,6 +724,15 @@ int main(int argc, char * argv[]) try {
 
             context.download_and_run(*context.get_transaction());
         }
+    } catch (libdnf::cli::GoalResolveError & ex) {
+        if (!any_repos_from_system_configuration && base.get_config().installroot().get_value() != "/" &&
```
j-mracek, line 728: I am not sure whether it is a good place. If I will use `dnf repoquery --installroot <> --releasever 33 something`, I will not get any information`. 

But with `dnf repoquery --installed --installroot <> --releasever 33 something` it would be not nice to get a message.

Ok, it might be not the best place, but good enough - no changes requested

dnf5/main.cpp
evan-goode: I agree, done

dnf5/main.cpp
evan-goode: I agree with @praiskup, I think `--use-host-config` is clearer. Both the configuration file and the reposdir are affected by this flag, so in that sense it is more than just "configuration", but in this case I think it's a better name. "Environment" is just such an overloaded term.

dnf5/main.cpp
evan-goode: Thanks, added a comment here.

If the user specifies a nonexistent RPM on the command line, this variable will not get set to true, and a warning message will still get printed:

```
$ dnf5 --installroot=/mnt install -y ./missing.rpm
No repositories were loaded from the installroot. To use the configuration and repositories of the host system, pass --use-host-environment.
No match for argument: ./missing.rpm
```

but I think that's ok. It will still not print if the file is present and the transaction succeeds.

dnf5/main.cpp
j-mracek: Then please  feel free to rename it. `--use-host-config` and configuration option `use-host-config`.

dnf5/main.cpp
j-mracek: In DNF4 - please take a look into logic of `def _checkEnabledRepo(base, possible_local_files=()):` in `dnf/cli/commands/__init__.py`. The logic there does not report problem with repos when rpms are defined by argument. Related bug - https://bugzilla.redhat.com/show_bug.cgi?id=620802. I an suggesting to add a TODO - implement check from DNF4 in `def _checkEnabledRepo()` and integrate this warning in the method. What do you think?

j-mracek: I am sorry - please add a TODO not here but around line 731.
I have `any_repos_from_system_configuration` part `system_configuration` is confusing, because repository could be created from commandline `--repofrompath`, or by plugin.  What about name - `any_available_repository`

dnf5/main.cpp
jan-kolarik: Looking at our conventions, the option should rather be with underscores `use_host_config`, right?

j-mracek: Correct, my mistake
Then please feel free to rename it. `--use-host-config` and configuration option `use_host_config`.

dnf5/main.cpp
evan-goode: Hmm, I just checked again and I now get the proper error from `dnf5 --installroot=/mnt install -y ./missing.rpm`: 
```
Repositories loaded.
Failed to access RPM "./missing.rpm": No such file or directory
```
Maybe something changed when I rebased.

Anyway, I think we do actually want to show the warning even when RPMs are specified by an argument. Consider the following case:
```
~/git/dnf5 $ sudo ./build/dnf5/dnf5 --installroot=/tmp install -y hello-2.12.1-1.2.x86_64.rpm 
Repositories loaded.
No repositories were loaded from the installroot. To use the configuration and repositories of the host system, pass --use-host-config.
Problem: conflicting requests
  - nothing provides libc.so.6()(64bit) needed by hello-2.12.1-1.2.x86_64
  - nothing provides libc.so.6(GLIBC_2.2.5)(64bit) needed by hello-2.12.1-1.2.x86_64
  - nothing provides libc.so.6(GLIBC_2.3)(64bit) needed by hello-2.12.1-1.2.x86_64
  - nothing provides libc.so.6(GLIBC_2.3.4)(64bit) needed by hello-2.12.1-1.2.x86_64
  - nothing provides libc.so.6(GLIBC_2.34)(64bit) needed by hello-2.12.1-1.2.x86_64
  - nothing provides libc.so.6(GLIBC_2.4)(64bit) needed by hello-2.12.1-1.2.x86_64
```

I specify a valid RPM on the command line, but its dependencies can't be found because no other repositories are loaded. I think it is desirable in this case to show the warning, no?

evan-goode:
Rebased and renamed the flag to `--use-host-config`; and `use_host_config` for the configuration option.

j-mracek:
@evan-goode Please could you check base.cpp? Clang is complaining about line break.

`clang-format -i base.cpp` should fix the issue.

evan-goode:
> @evan-goode Please could you check base.cpp? Clang is complaining about line break.
> 
> `clang-format -i base.cpp` should fix the issue.

Thanks. Got a pre-commit hook set up now.

Just rebased.

j-mracek:
LGTM

PR #272, opened 2023-02-08
nevra: Fix cmp_* comparators

m-blaha:
Current implementation used rpmvercmp() on the whole EVR. But unfortunatelly rpmvercmp does not do any NEVRA parsing, it just compares two strings, which led to incorrect results.

For example these packages are in the current code ordered wrong:

firefox-0:100.0-2.fc36.x86_64
firefox-0:109.0.1-1.fc36.x86_64

the reason is that rmpvercmp() correctly returns that '0:100.0-2.fc36' version string is higher than '0:109.0.1-1.fc36'. But what is true for strings is not true for EVR. We need to compare each part (epoch, version, and release) separately.
For this a new evrcmp() function is introduced.

### DIFF
--- a/include/libdnf/rpm/nevra.hpp
+++ b/include/libdnf/rpm/nevra.hpp
@@ -182,7 +182,26 @@ inline void copy_nevra_attributes(const F & from, T & to) {
 
 
 /// Compare alpha and numeric segments of two versions.
+/// @return 1 if `lhs` < `rhs`, -1 if `lhs` > `rhs`, 0 if they are equal
 int rpmvercmp(const char * lhs, const char * rhs);
+
+
+/// Compare evr part of two objects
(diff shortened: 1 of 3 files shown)


j-mracek:
@m-blaha I am just curious, where do you want to use the nevra_cmp?

m-blaha:
list command and it's behaviour in case no switch was applied (in dnf4 it was --all).
I need to filter available packages and keep only those which are not installed and those with version greater than the installed one.
My first attempt was to use queries, but the performance was poor (still possible, that it can be done better without this new comparator). My current solution is this (uses evrcmp): https://github.com/rpm-software-management/dnf5/blob/76a1d592771204c1956d70ea432d6a33733ae110/dnf5/commands/list/list.cpp#L175-L197

m-blaha:
And if you meant `cmp_nevra`, I do not have usage for it. It was just there and it was broken.
`cmp_naevr` on the other hand is used to sort packages for output.

PR #276, opened 2023-02-09
PackageQuery: filter_nevra to accept more comparators

m-blaha:
Current implementation of filter_nevra(PackageSet) accepts only EQ and NEQ comparators.
This change enhances this filter to accept also GT, GTE, LT, and LTE comparators.

### DIFF
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -215,7 +215,8 @@ inline bool name_compare_icase_lower_id(const std::pair<Id, Solvable *> first, I
     return first.first < id_name;
 }
 
-static inline bool name_arch_compare_lower_solvable(const Solvable * first, const Solvable * second) {
+template <typename T>
+static inline bool name_arch_compare_lower(const Solvable * first, const T * second) {
     if (first->name != second->name) {
         return first->name < second->name;
     }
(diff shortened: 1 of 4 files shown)


libdnf/rpm/package_query.cpp
```diff
@@ -722,30 +735,50 @@ void PackageQuery::filter_nevra(const libdnf::rpm::Nevra & pattern, libdnf::sack
 }
 
 void PackageQuery::filter_nevra(const PackageSet & package_set, libdnf::sack::QueryCmp cmp_type) {
-    if (cmp_type != sack::QueryCmp::EQ && cmp_type != sack::QueryCmp::NEQ) {
-        libdnf_throw_assert_unsupported_query_cmp_type(cmp_type);
+    libdnf_assert_same_base(p_impl->base, package_set.get_base());
+
+    bool cmp_not = (cmp_type & libdnf::sack::QueryCmp::NOT) == libdnf::sack::QueryCmp::NOT;
```
j-mracek, line 740: Minor - to ensure that there is not GT + NOT (or similar) as comperator I would suggest to change the test:
```
bool cmp_not = (cmp_type == libdnf::sack::QueryCmp::NEQ);
if (cmp_not) {
    cmp_type = libdnf::sack::QueryCmp::EQ;
}
```

m-blaha: Actually I plan to use LTE + NOT in list command implementation. Let me improve the description of the function.

m-blaha: I've rephrased the doc string of `filter_nevra(PackageSet)` filter.

PR #278, opened 2023-02-10
modules: Add ModuleQuery::resolve_module_spec

pkratoch:


libdnf/module/module_query.cpp
```diff
@@ -184,4 +185,23 @@ void ModuleQuery::filter_nsvca(const Nsvcap & nsvcap, libdnf::sack::QueryCmp cmp
...
+std::pair<bool, Nsvcap> ModuleQuery::resolve_module_spec(const std::string & module_spec) {
+    libdnf::sack::QueryCmp cmp =
+        libdnf::utils::is_glob_pattern(module_spec.c_str()) ? libdnf::sack::QueryCmp::GLOB : libdnf::sack::QueryCmp::EQ;
+
+    std::vector<Nsvcap> possible_nsvcaps = Nsvcap::parse(module_spec);
+    for (Nsvcap & nsvcap : possible_nsvcaps) {
+        ModuleQuery module_query(*this);
+        module_query.filter_nsvca(nsvcap, cmp);
+        if (!module_query.empty()) {
+            *this = std::move(module_query);
```
pkratoch, line 197: So, I haven't found any standard way to do intersection for sets. But couldn't it be done in this way? Sorry if I'm missing something.

j-mracek: I believe that the solution that you have is the best solution that you could do. The move is the fastest way.

j-mracek (approved):
LGTM

PR #286, opened 2023-02-14
transaction: Change API to run transaction without args

jan-kolarik:
All the `transaction.run()` arguments could be optional, therefore changing the API to pass the arguments with class setters.


jan-kolarik:
Note: failing tests are due to API change and will be resolved in linked PR.

kontura:
Thanks!

PR #293, opened 2023-02-14
Add explicit package version for libdnf5-cli

j-mracek:
The change was suggested by rpminspect. See the description bellow.

Add the indicated explicit Requires to the spec file for the named subpackage. Subpackages depending on shared libraries in another subpackage must carry an explicit
'Requires: SUBPACKAGE_NAME = %{version}-%{release}' in the spec file. VERIFY Subpackage dnf5 on s390x carries
'Requires: libdnf-cli.so.1()(64bit)' which comes from subpackage libdnf5-cli but does not carry an explicit package version requirement. Please add 'Requires: libdnf5-cli = %{version}-%{release}' to the spec file to avoid the need to test interoperability between various combinations of old and new subpackages.


jan-kolarik:
LGTM, failing test in rawhide is unrelated.

Issue #277, opened 2023-02-10
Add a download method for Transaction class


j-mracek:
The new method should simplify the workflow of DNF5 when transaction is created. 

Basically to replace following code 
```
# Add the inbound packages (packages that are being installed on the system)
# to the downloader.
for tspkg in transaction.get_transaction_packages():
    if libdnf5.base.transaction.transaction_item_action_is_inbound(tspkg.get_action()):
        downloader.add(tspkg.get_package())

# Download the packages.
#
# The first argument is `fail_fast`, meaning the download will fail right away
# on a first package download failure. The second argument is `resume`, if
# `true`, the downloader will try to resume downloads of any partially
# downloaded RPMs.
downloader.download(True, True)
```

By 
```
transaction.download()
```

- [x] add the method to Transaction Class
- [x] Modify tutorials to use the method
- [ ] Replace the code in DNF5 part to use the download method

j-mracek:
@evan-goode I am sorry I forget to mention one sub-task.

evan-goode:
https://github.com/rpm-software-management/dnf5/pull/313 has been merged. I think that's all the refactoring we want to do for now.

PR #294, opened 2023-02-14
Add transaction.download()

evan-goode:
`transaction.download()` downloads all inbound packages on the transaction, equivalent to the following:

```
libdnf::repo::PackageDownloader downloader;
for (auto & tspkg : transaction.get_transaction_packages()) {
    if (transaction_item_action_is_inbound(tspkg.get_action())) {
        downloader.add(tspkg.get_package());
    }
}
downloader.download(true, true);
```

Resolves https://github.com/rpm-software-management/dnf5/issues/277

### DIFF
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -32,6 +32,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf/base/base.hpp"
 #include "libdnf/common/exception.hpp"
+#include "libdnf/repo/package_downloader.hpp"
 #include "libdnf/rpm/package_query.hpp"
 
 #include <fmt/format.h>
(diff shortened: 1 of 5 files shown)


libdnf/base/transaction.cpp
```diff
@@ -222,6 +223,16 @@ std::string Transaction::transaction_result_to_string(const TransactionRunResult
     return {};
 }
 
+void Transaction::download() {
+    libdnf::repo::PackageDownloader downloader;
+    for (auto & tspkg : this->get_transaction_packages()) {
+        if (transaction_item_action_is_inbound(tspkg.get_action())) {
```
m-blaha, line 229: We were just discussing this yesterday with @jan-kolarik . I think commandline packages should be skipped here (they should be already downloaded and inserted to commandline repo before goal resolving).
Similarly as they currently are excluded in dnf5 - https://github.com/rpm-software-management/dnf5/blob/97cdbe57c34259a0e0dc87431ac7c3e9a396aba7/dnf5/context.cpp#L281-L293

jan-kolarik: Hey, you were faster! :smile:

jan-kolarik: Also there are usually two usecases, one passing the `dest_dir` to the downloader, one without that using the default `cachedir`. It would be nice to unify that, maybe just add a `dest_dir` parameter to the `download` method with default empty? Just avoid using the `std::optional` on the public API :slightly_smiling_face:

evan-goode: Done and done! `dest_dir` is a good call.

jan-kolarik:
Looks good, thanks! Next step I think is refactoring the existing usages of the `downloader` class to use this new method. But this will better to be done in separate PR.

Issue #143, opened 2022-11-11
Implement list command


mcurlej:
**Command:** ```list```
**Aliases:** ```ls```

Prints lists of packages depending on the packages’ relation to the system. A package is ```installed``` if it is present in the RPMDB, and it is ```available``` if it is not installed but is present in a repository that DNF knows about.

The list command also limits the displayed packages according to specific criteria, e.g. to only those that update an installed package (respecting the repository [priority](https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-priority-label)). The [exclude](https://dnf.readthedocs.io/en/latest/conf_ref.html#exclude-label) option in the configuration file can influence the result, but if the -[-disableexcludes](https://dnf.readthedocs.io/en/latest/command_ref.html#disableexcludes-label) command line option is used, it ensures that all installed packages will be listed.

```dnf [options] list [--all] [<package-file-spec>...]```

Lists all packages, present in the RPMDB, in a repository or both.

```dnf [options] list --installed [<package-file-spec>...]```

Lists installed packages.

```dnf [options] list --available [<package-file-spec>...]```

Lists available packages.

```dnf [options] list --extras [<package-file-spec>...]```

Lists extras, that is packages installed on the system that are not available in any known repository.

```dnf [options] list --obsoletes [<package-file-spec>...]```

List packages installed on the system that are obsoleted by packages in any known repository.

```dnf [options] list --recent [<package-file-spec>...]```

List packages recently added into the repositories.

```dnf [options] list --upgrades [<package-file-spec>...]```

List upgrades available for the installed packages.

```dnf [options] list --autoremove```

List packages which will be removed by the dnf autoremove command.

This command by default does not force a sync of expired metadata. See also [Metadata Synchronization](https://dnf.readthedocs.io/en/latest/command_ref.html#metadata-synchronization-label).

Cyborgscode:
still missing.

m-blaha:
List command was introduced by PR https://github.com/rpm-software-management/dnf5/pull/298

PR #298, opened 2023-02-15
List and info commands

m-blaha:
Fixes issue https://github.com/rpm-software-management/dnf5/issues/143


dnf5/commands/list/list.cpp
```diff
@@ -0,0 +1,284 @@
...
+            base_query |= pkg_query;
+        }
+    }
+
+    // did any package match the criteria?
+    bool package_matched = false;
+    // output table
+    auto sections = create_output();
+
+    switch (pkg_narrow) {
```
j-mracek, line 174: Do we want to expose list and info functionality for DBUS users? Or queries are enough?

j-mracek: Only note: In DNF4 the code was shared with `repository-pkgs list` command, but in DNF5 we will provide the functionality using a different way.

m-blaha: > Do we want to expose list and info functionality for DBUS users? Or queries are enough?

Actually it's currently the other way round. We have `list` functionality on dbus (or something very similar), but we do not support queries there. The problem is that list functionality is "state-less" and always work on top of all packages (or a set limited by pkg_narrow). To implement queries we would need to store query state on the server side - to be able to gradually implement filters.

dnf5/commands/list/list.cpp
```diff
@@ -0,0 +1,284 @@
...
+                    "",  //config.color_list_available_downgrade().get_value(),
+                    config.color_list_available_reinstall().get_value(),
+                    config.color_list_available_upgrade().get_value());
+                package_matched |= sections->add_section("Available packages", available, colorizer);
+            }
+            break;
+        }
+        case PkgNarrow::INSTALLED: {
+            base_query.filter_installed();
+            package_matched |= sections->add_section("Installed packages", base_query);
```
kontura, line 205: I would find it more consistent if we colored the installed and upgradable packages even if their option is used and they are the only packages in output.

dnf5/commands/list/list.cpp
```diff
@@ -0,0 +1,284 @@
...
+                    installed,
+                    "",  //config.color_list_available_install().get_value(),
+                    "",  //config.color_list_available_downgrade().get_value(),
+                    config.color_list_available_reinstall().get_value(),
+                    config.color_list_available_upgrade().get_value());
+                package_matched |= sections->add_section("Available packages", base_query, colorizer);
+            }
+            break;
+        }
+        case PkgNarrow::UPGRADES:
```
kontura, line 229: I am not sure if this is intended but using `dnf5 list pkg --upgrades` also includes src packages.
This is a change when compared to dnf4.

libdnf-cli/output/package_list_sections.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <string>
+
+namespace libdnf::cli::output {
+
+PackageListSections::PackageListSections() {
+    table = scols_new_table();
+    scols_table_enable_noheadings(table, 1);
+    if (libdnf::cli::tty::is_interactive()) {
+        scols_table_enable_colors(table, 1);
+        scols_table_enable_maxout(table, 1);
```
kontura, line 40: In my opinion this setting makes the outputs (the info commands especially) less readable.

But mainly I would prefer if we chose one approach and used it through out dnf5.
Currently I see it is used only in repolist, transaction_table and here.

include/libdnf-cli/output/pkg_colorizer.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /// Compute a color for the package.
+    /// @param package A package for which color is needed.
+    /// @return Escape sequence for the color.
+    std::string get_pkg_color(const libdnf::rpm::Package & package);
+
+private:
+    std::string to_escape(const std::string & color);
+
+    // map N.A of the package to the version
+    std::map<std::string, libdnf::rpm::Package> base_na_version;
```
kontura, line 60: Maybe an [`unordered_map`](https://en.cppreference.com/w/cpp/container/unordered_map) would be more appropriate here? It seems we don't need to have the packages ordered since we do just a single access to retrieve one pkg in `get_pkg_color()`.

dnf5/commands/list/list.cpp
```diff
@@ -0,0 +1,284 @@
...
+            return true;
+        });
+    specs->set_complete_hook_func([&ctx](const char * arg) { return match_specs(ctx, arg, true, true, false, false); });
+    cmd.register_positional_arg(specs);
+
+    auto showduplicates = parser.add_new_named_arg("showduplicates");
+    showduplicates->set_long_name("showduplicates");
+    showduplicates->set_description("Show all versions of the packages, not only the latest ones.");
+    showduplicates->set_const_value("true");
+    showduplicates->link_value(&config.showdupesfromrepos());
```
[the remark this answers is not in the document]
m-blaha: Well, we can drop the config option and keep only commandline option for list / info / search commands.
It shouldn't be a problem given it's not documented in dnf4.
But I'd prefer to do it in separate PR - it will affect also the `search` command.

dnf5/commands/list/list.cpp
m-blaha: Thanks, it's a mistake. I'll fix it.

dnf5/commands/list/list.cpp
m-blaha: Fixed.

libdnf-cli/output/package_list_sections.cpp
m-blaha: I do not have any strong opinion on this. But I agree with the consistency so I've removed this setting.

include/libdnf-cli/output/pkg_colorizer.hpp
m-blaha: Fixed.

libdnf-cli/output/package_list_sections.cpp
kontura: @m-blaha It seems there is now a missing new line at the end of all outputs.

I am not sure how it works but maybe since the option
```c++
scols_table_enable_maxout(table, 1);
```
was filling the window to the end it seemed as the prompt wrapped to a new line? Or perhaps it added a new line automatically.

m-blaha: Oh, I'll check it. Thanks for catching it.

evan-goode:
With DNF 4 you could do `dnf list installed` instead of `dnf list --installed` and it would still work (same with `dnf list available`). Any intention to make that work in DNF 5?

libdnf-cli/output/package_list_sections.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+
+PackageListSections::~PackageListSections() {
+    scols_unref_table(table);
+}
+
+
+void PackageListSections::print() {
```
evan-goode, line 49: There is a missing newline at the end of the output for both `dnf5 list` and `dnf5 info`.

m-blaha:
> With DNF 4 you could do `dnf list installed` instead of `dnf list --installed` and it would still work (same with `dnf list available`). Any intention to make that work in DNF 5?

This "feature" is intentionally dropped and we require using regular `--installed` option instead of this magical positional argument.

libdnf-cli/output/package_list_sections.cpp
m-blaha: Yeah, thanks! My shell hides this from me by automatically adding a newline. I'll add a test (as also @kontura suggests).

libdnf-cli/output/package_list_sections.cpp
m-blaha: I've looked how we deal with command output checking in our CI suite. The thing is, that we ignore any trailing new lines. I guess the reason is to keep the expected outputs in `Then stdout is` step cleaner - without the trailing empty line.
I did some preliminary testing with more strict checks and already discover a couple of issues (like redundant new lines in the end of output). But I'd like to fix this in a separate PR.
If you agree, I'll only fix the missing new line here and file an issue to make our output checks more strict regarding to trailing new lines.

kontura: That sound good, thank you!

m-blaha:
The missing new line is now fixed.

kontura (approved):
Thanks!

libdnf-cli/output/package_list_sections.cpp
evan-goode: Ah yep didn't see his comment :)

PR #299, opened 2023-02-16
modules: Reject overlong nsvcap strings

pkratoch:
There is a known bug in GCC std::regex_match() that exhausts a stack on very long strings.

This patch prevents this by rejecting strings longer than 2 KB. Another solution would be to move to a heap-based regular expression engine, like PCRE2.

<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164>

### DIFF
--- a/common/utils/regex.hpp
+++ b/common/utils/regex.hpp
@@ -1,0 +1,28 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


pkratoch:
Good point. I haven't found any place that would be appropriate for this macro, so I created a file in `common/utils`, but feel free to suggest something better.

ppisar:
A new header file is good. Another option would be adding "add_definitions()" to a CMakeLists.txt. I like the header file approach more because it does not inflate a command line for the compiler.

jan-kolarik:
Looks fine, thank you!

PR #300, opened 2023-02-16
Document set/get vars in python api

inknos:
I believe this should be documented in the api, but not as part of the main workflow.

I propose to keep it commented as optional steps to show how to address getting/setting vars, but not as part of the workflow until we create an example that uses it.

Maybe a specific example in the future?


test/python3/libdnf5/tutorial/session/create_base.py
```diff
@@ -11,6 +11,16 @@
 # configuration from a different location.
 base.load_config_from_file()
 
+# Optionally you can set and get vars
+# vars = base.get_vars().get()
+# vars.set('releasever', '33')
+#
+# Once a variable is set it can be accessed by the getter method
+# vars.get('releasever')
```
j-mracek, line 19: I believe that currently this approach does not work. `vars.get('releasever')` returns something like a pair, but a structure is unknown in python, therefore you cannot use it at all (bug) . What about to describe - `get_priority` and `get_value`? Hope that get priority works

j-mracek: I am sorry - my previous comment was incorrect - get_priority() is not available method for Vars

Get() returns Vars structure that contains
```
    struct Variable {
        std::string value;
        Priority priority;
    };
```

j-mracek: Because get() approach is not working - ok working but not useful - I would recommend to remove example with get() and add todo - provide a way how to get option  priority. What do you think?

j-mracek: @inknos May I ask you what's yours plans?

PR #302, opened 2023-02-16
dnf5: Transaction progress: Fix total number of actions, hide total transaction progress bar

jrohel:
- dnf5: Hides total transaction progress bar - It was showing nonsensical information.
- dnf5: Fixes total number of actions (steps) in transaction progress - It shows [x/max], before fix [x/x+1].

Solves: https://bugzilla.redhat.com/show_bug.cgi?id=2180229

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -190,6 +190,8 @@ public:
         }
     }
 
+    libdnf::cli::progressbar::MultiProgressBar * get_multi_progress_bar() { return &multi_progress_bar; }
+
     void install_progress(
         [[maybe_unused]] const libdnf::rpm::TransactionItem & item,
         uint64_t amount,
@@ -535,7 +537,23 @@ void Context::download_and_run(libdnf::base::Transaction & transaction) {
         cmd_line += argv[i];
     }
 
-    transaction.set_callbacks(std::make_unique<RpmTransCB>());
+    // Compute the total number of transaction actions (number of bars)
+    // Total number of actions = number of packages in the transaction +
+    //                           action of verifying package files if new package files are present in the transaction +
+    //                           action of preparing transaction
+    const auto & trans_packages = transaction.get_transaction_packages();
+    auto num_of_actions = trans_packages.size() + 1;
+    for (auto & trans_pkg : trans_packages) {
+        if (libdnf::transaction::transaction_item_action_is_inbound(trans_pkg.get_action())) {
+            ++num_of_actions;
+            break;
+        }
+    }
+
+    auto callbacks = std::make_unique<RpmTransCB>();
+    callbacks->get_multi_progress_bar()->set_total_num_of_bars(num_of_actions);
+    transaction.set_callbacks(std::move(callbacks));
+
     transaction.set_description(cmd_line);
 
     if (comment) {
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -176,6 +176,10 @@ namespace {
 
 class RpmTransCB : public libdnf::rpm::TransactionCallbacks {
 public:
+    RpmTransCB() {
+        multi_progress_bar.set_total_bar_visible_limit(libdnf::cli::progressbar::MultiProgressBar::NEVER_VISIBLE_LIMIT);
+    }
+
     ~RpmTransCB() {
         if (active_progress_bar &&
             active_progress_bar->get_state() != libdnf::cli::progressbar::ProgressBarState::ERROR) {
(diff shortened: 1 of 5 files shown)


dnf5/context.cpp
```diff
@@ -176,6 +176,8 @@ namespace {
 
 class RpmTransCB : public libdnf::rpm::TransactionCallbacks {
 public:
+    RpmTransCB() { multi_progress_bar.set_total_bar_visible_limit(static_cast<std::size_t>(-1)); }
```
jan-kolarik, line 179: This is kinda not transparent at the first sight, it took me some time to understand what it is actually doing and that -1 is transformed to `SIZE_MAX`... Could you please define some local constant for this purpose like `NEVER_VISIBLE_LIMIT` or something like that?

include/libdnf/base/transaction.hpp
```diff
@@ -79,6 +79,9 @@ class Transaction {
     // TODO(jrohel): Return reference instead of copy?
     std::vector<libdnf::base::TransactionPackage> get_transaction_packages() const;
 
+    /// @return the number of transaction packages.
+    std::size_t get_num_of_transaction_packages() const;
```
jan-kolarik, line 83: Just a note / minor thing: if we would use the reference return value for `get_transaction_packages()` then this won't be needed I guess. 

Also I'd prefer not using abbreviations in the method name as it will be on public API, f.e. like `get_transaction_packages_count()`, what do you think?

libdnf-cli/progressbar/multi_progress_bar.cpp
```diff
@@ -60,15 +60,38 @@ void MultiProgressBar::add_bar(std::unique_ptr<ProgressBar> && bar) {
     }
 
     bars_all.push_back(std::move(bar));
-    total.set_total(static_cast<int>(bars_all.size()));
+    if (total.get_total() < static_cast<int32_t>(bars_all.size())) {
+        total.set_total(static_cast<int32_t>(bars_all.size()));
```
jan-kolarik, line 64: Minor: repeated expression `static_cast<int32_t>(bars_all.size())`.

jan-kolarik:
Today I was using `dnf5` in container rebased to this PR and I've found the following bug.

First I've installed the `nudoku` package:
```
[root@b97345e505af dnf5]# dnf5 install nudoku
Repositories loaded.
Package                  Arch    Version                  Repository        Size
Installing:                                                                     
 nudoku                  x86_64  2.1.0-4.fc36             fedora        78.4 KiB

Transaction Summary:
 Installing:        1 packages

Is this ok [y/N]: y
[1/1] nudoku-0:2.1.0-4.fc36.x86_64      100% |   0.0   B/s |   0.0   B |  00m00s
>>> Already downloaded
--------------------------------------------------------------------------------
[1/1] Total                             100% |   0.0   B/s |   0.0   B |  00m00s
Verifying PGP signatures

Running transaction
[1/3] Verify package files              100% |   0.0   B/s |   1.0   B |  00m00s
[2/3] Prepare transaction               100% |  21.0   B/s |   1.0   B |  00m00s
[3/3] Installing nudoku-0:2.1.0-4.fc36. 100% | 540.9 KiB/s |  80.6 KiB |  00m00s
>>> Running trigger-install scriptlet: glibc-common-0:2.35-15.fc36.x86_64
>>> Stop trigger-install scriptlet: glibc-common-0:2.35-15.fc36.x86_64
```
And then I've removed it, but incorrect number of steps was shown in progress bar there:
```
[root@b97345e505af dnf5]# dnf5 remove nudoku
Package                  Arch    Version                  Repository        Size
Removing:                                                                       
 nudoku                  x86_64  2.1.0-4.fc36             fedora        78.4 KiB

Transaction Summary:
 Removing:          1 packages

Is this ok [y/N]: y

Verifying PGP signatures

Running transaction
[1/3] Prepare transaction               100% |  31.0   B/s |   1.0   B |  00m00s
[2/3] Erasing nudoku-0:2.1.0-4.fc36.x86 100% | 112.0   B/s |  14.0   B |  00m00s
[root@b97345e505af dnf5]#
```
Tried that repeatedly with the same results.

dnf5/context.cpp
jrohel: > This is kinda not transparent at the first sight, it took me some time to understand what it is actually doing and that -1 is transformed to SIZE_MAX

Yes. It is a portable method to find the maximum value of `std::size_t`. It is guaranteed to work with any unsigned type. But I agree that it is more transparent and user-friendly to define a constant for this purpose.

include/libdnf/base/transaction.hpp
jrohel: >Just a note / minor thing: if we would use the reference return value for get_transaction_packages() then this won't be needed I guess.

Yes. I'd rather return a constant reference instead of an expensive copy. The user can then make a copy only when needed. But here is some background. We should discuss. But not as part of this PR.

Thanks. `get_transaction_packages_count()` is exactly the name I was looking for.

libdnf-cli/progressbar/multi_progress_bar.cpp
jrohel: It is minor. But I will do change.

jan-kolarik (approved):
Thanks, looks great and the issue seems to be resolved.

PR #308, opened 2023-02-20
swig: Throw Python exception on undefined var

### LINKED ISSUE #292: Undefined substitution variables kill the Python process
When I try to get a value of an undefined substitution variable, it kills the Python process.

```
base = libdnf5.base.Base()
vars = base.get_vars()
vars.get_value("undefined")
```

The example above fails with the exit code 134 and the error messages below.

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

Please, raise an exception instead.

jan-kolarik:
Throw a Python exception instead of C++ one when undefined variable is being accessed.

Closes #292.

### DIFF
--- a/bindings/libdnf5/conf.i
+++ b/bindings/libdnf5/conf.i
@@ -17,6 +17,8 @@
 %exception {
     try {
         $action
+    } catch (const std::out_of_range & e) {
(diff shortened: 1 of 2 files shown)


j-mracek (approved):
LGTM

Issue #305, opened 2023-02-17
Error thrown when loading a non-existent config file


evan-goode:
DNF5 throws an error when it tries to load a config file that doesn't exist:
```
$ mv /etc/dnf/dnf.conf /etc/dnf/dnf.conf.old
$ dnf5 -y install hello
filesystem error: cannot open file: No such file or directory [/etc/dnf/dnf.conf]
```

DNF4, however, ignores the missing file and works fine.

j-mracek:
This is an issue. DNF5 must be able to work without presence of config. There is one exception in DNF4 - when user specify config from commandline - `sudo dnf install --config /asas/asas.conf sss` then it complains `Config error: Config file "/asas/asas.conf" does not exist`.

Also check the behavior of DNF4 when you do not have a permissions for `dnf.conf`.

j-mracek:
The implementation in DNF4 was not optimal:

```
        filename = conf._get_value('config_file_path')
        if (conf._get_priority('config_file_path') == dnf.conf.PRIO_COMMANDLINE) and \
                not os.path.isfile(filename):
            raise dnf.exceptions.ConfigError(_('Config file "{}" does not exist').format(filename))

        # read config
        conf.read(priority=dnf.conf.PRIO_MAINCONFIG)
```
It tested the existence of the path and config path priority and then it called conf.read().
DNF4 API did not read the configuration by default - it required additional call.

PR #312, opened 2023-02-21
Ignore inaccessible config file unless its path was specified as --config=...

evan-goode:
Ignore a missing or otherwise inaccessible `/etc/dnf/dnf.conf` instead of throwing an error.

If the user specifies the path to the config file on the command line, like `dnf --config=/path/to/dnf.conf`, then an error will still be thrown and displayed.

Similarly, if the config file is accessible but malformed/invalid, an error will still be thrown.

Resolves https://github.com/rpm-software-management/dnf5/issues/305

### DIFF
--- a/libdnf/repo/repo_sack.cpp
+++ b/libdnf/repo/repo_sack.cpp
@@ -352,7 +352,20 @@ void RepoSack::create_repos_from_file(const std::string & path) {
 }
 
 void RepoSack::create_repos_from_config_file() {
-    create_repos_from_file(std::filesystem::path(base->get_config().config_file_path().get_value()));
+    const auto & path = base->get_config().config_file_path().get_value();
+    try {
+        create_repos_from_file(std::filesystem::path(path));
+    } catch (const libdnf::MissingConfigError & e) {
+        // Ignore the missing config file unless user specified it via --config=...
+        if (base->get_config().config_file_path().get_priority() >= libdnf::Option::Priority::COMMANDLINE) {
+            throw;
+        }
+    } catch (const libdnf::InaccessibleConfigError & e) {
+        // Ignore the inaccessible config file unless user specified it via --config=...
+        if (base->get_config().config_file_path().get_priority() >= libdnf::Option::Priority::COMMANDLINE) {
+            throw;
+        }
+    }
 }
 
 void RepoSack::create_repos_from_dir(const std::string & dir_path) {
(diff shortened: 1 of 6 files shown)


evan-goode:
Moved the new error definitions to `include/libdnf/common/exception.hpp`; they wouldn't compile in `base.hpp` for some reason.

libdnf/repo/repo_sack.cpp
```diff
@@ -351,8 +351,20 @@ void RepoSack::create_repos_from_file(const std::string & path) {
...
 }
 
-void RepoSack::create_repos_from_config_file() {
-    create_repos_from_file(std::filesystem::path(base->get_config().config_file_path().get_value()));
+void RepoSack::create_repos_from_config_file() try {
+    const auto & path = base->get_config().config_file_path().get_value();
+    try {
+        create_repos_from_file(std::filesystem::path(path));
+    } catch (const std::filesystem::filesystem_error & e) {
+        std::throw_with_nested(InaccessibleConfigError(M_("Unable to access configuration file \"{}\""), path));
```
m-blaha, line 359: Do we need these nested try/catch blocks? I mean, couldn't the `if (base->get_config().config_file_path().get_priority() >= libdnf::Option::Priority::COMMANDLINE)` condition be right here?

include/libdnf/common/exception.hpp
```diff
@@ -160,6 +160,24 @@ class SystemError : public Error {
...
+};
+
+
+/// Couldn't parse configuration file
+class InvalidConfigError : public Error {
+public:
+    using Error::Error;
+    const char * get_domain_name() const noexcept override { return "libdnf::base"; }
+    const char * get_name() const noexcept override { return "InvalidConfigError"; }
+};
```
m-blaha, line 178: It seems to me that these specific errors do not fit in this header file where more general errors are defined.
`include/libdnf/conf/config_parser.hpp` might be a better fit for them, what do you think?
After all, config parser is the place where these exceptions has their roots. Maybe, they can be even handled / thrown directly in the `ConfigParser::read()`?

jan-kolarik: In case of adding the exceptions into some header which is already included in SWIG, like the `include/libdnf/conf/config_parser.hpp`, make sure to ignore these exceptions in related `*.i` file, otherwise it won't compile, e.g. in `conf.i` the following should do the trick:

```
%ignore libdnf::InaccessibleConfigError;
%ignore libdnf::InvalidConfigError;
```

evan-goode:
I moved the error declarations to `include/libdnf/conf/config_parser.hpp`, and they are now thrown from `ConfigParser::read`, per @m-blaha's suggestion.

m-blaha:
Thank you!

PR #313, opened 2023-02-21
Prefer transaction.download() over directly using PackageDownloader

evan-goode:
Use the new [transaction.download()](https://github.com/rpm-software-management/dnf5/pull/294) in a couple places instead of using PackageDownloader directly. 

Not sure whether it's worth it to use it [here](https://github.com/rpm-software-management/dnf5/blob/adfdfb0535b62b96a8a4015dedbae663483f2ba9/dnf5daemon-server/services/goal/goal.cpp#L157) since download callbacks need to be set up.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -270,26 +270,8 @@ void download_packages(const std::vector<libdnf::rpm::Package> & packages, const
     }
 
     std::cout << "Downloading Packages:" << std::endl;
-    try {
-        downloader.download(true, true);
-    } catch (const std::runtime_error & ex) {
-        std::cout << "Exception: " << ex.what() << std::endl;
-    }
+    downloader.download(true, true);
     std::cout << std::endl;
-}
-
-void download_packages(libdnf::base::Transaction & transaction, const char * dest_dir) {
-    std::vector<libdnf::rpm::Package> downloads;
-    for (auto & tspkg : transaction.get_transaction_packages()) {
-        if (transaction_item_action_is_inbound(tspkg.get_action()) &&
-            tspkg.get_package().get_repo()->get_type() != libdnf::repo::Repo::Type::COMMANDLINE) {
-            downloads.push_back(tspkg.get_package());
-        }
-    }
-
-    if (!downloads.empty()) {
-        download_packages(downloads, dest_dir);
-    }
 }
 
 namespace {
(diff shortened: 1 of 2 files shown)


dnf5/context.cpp
```diff
@@ -634,7 +620,11 @@ bool Context::check_gpg_signatures(const libdnf::base::Transaction & transaction
 
 
 void Context::download_and_run(libdnf::base::Transaction & transaction) {
-    download_packages(transaction, nullptr);
+    try {
+        transaction.download();
+    } catch (const std::runtime_error & ex) {
```
j-mracek, line 625: What kind of exception do you excerpt to catch?

j-mracek: I see that there is a set of various errors that it could trow.

dnf5/context.cpp
```diff
@@ -634,7 +620,11 @@ bool Context::check_gpg_signatures(const libdnf::base::Transaction & transaction
...
+        std::cout << "Exception: " << ex.what() << std::endl;
+    }
```
j-mracek, line 627: At this state you cannot continue with dnf5. The process must be terminated (nicely not with a core dump).

j-mracek: There is a bug report for that - [Bug 2149377](https://bugzilla.redhat.com/show_bug.cgi?id=2149377) - [dnf5] dnf5 says not downloaded packages have invalid signatures

evan-goode: Yeah, here I'm copying the existing error handling behavior from `download_packages`. But you're right, I'm not sure why the error is caught and logged there.

I will take this out.

j-mracek:
@m-blaha May I ask you for your opinion for part related to Daemon?

dnf5/commands/download/download.cpp
```diff
@@ -122,22 +122,15 @@ void DownloadCommand::run() {
...
             }
         }
-        download_pkgs.reserve(transaction.get_transaction_packages().size());
-        for (auto & tspkg : transaction.get_transaction_packages()) {
-            if (transaction_item_action_is_inbound(tspkg.get_action()) &&
-                tspkg.get_package().get_repo()->get_type() != libdnf::repo::Repo::Type::COMMANDLINE) {
-                download_pkgs.push_back(tspkg.get_package());
-            }
-        }
+        transaction.download();
```
j-mracek, line 125: This modification makes failing test for download command. The problem is that the original implementation  downloaded packages into current directory, but the new implementation downloads them to cache of repository. There are two possibilities - to keep the original code or to change `destdir` for each repository to current directory ".".

evan-goode: Oh I didn't see that, thanks. I think it will be best for now to keep the existing implementation. Earlier today @jan-kolarik suggested changing the API to add a setter for specifying a transaction's download destination, rather than having the destination dir as an argument to `transaction.download()`. So I think he's planning on changing things soon, might as well keep it as-is for now.

evan-goode:
Looks like we're back to passing integration tests.

PR #314, opened 2023-02-22
swig: Drop unused Python-specific typemaps

jan-kolarik:
As `std::optional` was dropped from the public API, these typemaps for Python are not needed anymore.


kontura:
Thank you!

Issue #269, opened 2023-02-07
Future of configuration options `strict` and `skip_broken`


j-mracek:
Both options are somehow related from CLI `strict` = not `skip_broken` This is not valid for API or `microdnf` where `skip_broken` is ignored.

### Background
`YUM 3` has option `skip_broken`. DNF split the option into two => `strict` and `best`, but skip_broken remained as a compatibility option with `YUM`. `strict` still means two things => skip_broken (solver), `skip_unavailable` not in any repository and not installed.

### Possible solutions
We could split `strict` option, and remove or mark deprecated one of the options.

- create new configuration option `skip_unavailable`
  - it will help us to modify behavior of group - strict behavior with one unified switch
  - the name is already used by `builddep` command

- mark `strict` as deprecated
  - align `strict` = `not skip_broken`
  - or remove `strict`

- mark `skip_broken` as deprecated
  - align `strict` = `not skip_broken`
  - or remove `skip_broken`

pkratoch:
With `strict`, I need to remember in what way it's strict, while `skip_broken` and `skip_unavailable` are more descriptive. It'd be also more straightforward to configure dnf with these two options rather than having to think how to set some combination of `strict` and another sub-option.

m-blaha:
Maybe more examples and use cases of what currently is / is not possible can help here.
Let me summarize how I understand the problem (and please correct me if I'm mistaken).

- first problem is that we have `strict` config option (defaulting to true), but instead of having `--strict` and `--no-strict` command line options, there is `--skip-broken` which sets `strict=false`. This is confusing.
- `strict` means skip both - unavailable packages, and packages with broken dependencies. The problem is that there is no way how to skip only unavailable (group command use case) and error out on broken dependencies.

For me the best solution would be

- keep config options and command line options the same
- introduce two new config/commandline options - `skip_broken` and `skip_unavailable`, both default to false
- deprecate `strict` config option and temporarily keep the dnf4 meaning for compatibility -
    - strict=true means skip_broken=false and skip_unavailable=false
    - strict=false means skip_broken=true and skip_unavailable=true

jan-kolarik:
I agree with previous comments. I found it better having two options (`skip_broken` and `skip_unavailable`) doing only that one thing stated in their name than one (`strict`) doing more actions in the background. Deprecating the `strict`, but keeping it there for a while sounds for me like a good way to do the transition more smoothly.

kontura:
[...]

PR #316, opened 2023-02-22
Strict config option deprecation

m-blaha:
The PR introduces new config options

- `skip_broken` - for solver to enable resolving problems by skipping problematic packages
- `skip_unavailable` - to not fail when requested package is not available

Deprecated `strict` option (if explicitely set) is used as a default value for these new options.

Also all occurrences of `strict` usage in the codebase are replaced by `skip_broken` / `skip_unavailable`.

Fixes https://github.com/rpm-software-management/dnf5/issues/269

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -264,20 +264,6 @@ void RootCommand::set_argument_parser() {
         global_options_group->register_argument(exclude);
     }
 
-    auto skip_broken = parser.add_new_named_arg("skip-broken");
-    skip_broken->set_long_name("skip-broken");
-    skip_broken->set_description("resolve depsolve problems by skipping packages");
-    skip_broken->set_const_value("true");
-    skip_broken->link_value(&config.skip_broken());
-    global_options_group->register_argument(skip_broken);
-
-    auto skip_unavailable = parser.add_new_named_arg("skip-unavailable");
-    skip_unavailable->set_long_name("skip-unavailable");
-    skip_unavailable->set_description("allow skipping unavailable packages");
-    skip_unavailable->set_const_value("true");
-    skip_unavailable->link_value(&config.skip_unavailable());
-    global_options_group->register_argument(skip_unavailable);
-
     auto enable_repo_ids = parser.add_new_named_arg("enable-repo");
     enable_repo_ids->set_long_name("enable-repo");
     enable_repo_ids->set_has_value(true);
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -271,6 +271,13 @@ void RootCommand::set_argument_parser() {
     skip_broken->link_value(&config.skip_broken());
     global_options_group->register_argument(skip_broken);
 
+    auto skip_unavailable = parser.add_new_named_arg("skip-unavailable");
+    skip_unavailable->set_long_name("skip-unavailable");
+    skip_unavailable->set_description("allow skipping unavailable packages");
+    skip_unavailable->set_const_value("true");
+    skip_unavailable->link_value(&config.skip_unavailable());
+    global_options_group->register_argument(skip_unavailable);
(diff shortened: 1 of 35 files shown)


m-blaha:
The failing CI test is expected - it uses `--skip-broken` option which before this patch used to set `strict` config option. Now with separated `skip_broken` and `skip_unavailable` options the test needs to be adapted: https://github.com/rpm-software-management/ci-dnf-stack/pull/1227

libdnf/base/goal.cpp
```diff
@@ -496,10 +496,16 @@ GoalProblem Goal::Impl::add_group_specs_to_goal(base::Transaction & transaction)
     std::vector<std::tuple<std::string, transaction::TransactionItemReason, comps::GroupQuery, GoalJobSettings>>
         groups_remove, groups_install;
     for (auto & [action, reason, spec, settings] : group_specs) {
-        bool strict = settings.resolve_strict(cfg_main);
+        bool skip_unavailable = settings.resolve_skip_unavailable(cfg_main);
         if (action != GoalAction::INSTALL && action != GoalAction::REMOVE) {
             transaction.p_impl->add_resolve_log(
```
j-mracek, line 501: Because calls of add_resolve_logs() uses a different value for the last bool argument (skip_broken, skip_unavailable, ...) - I would recommend to add a description to that private method and may be not call the argument strict. What do you think?

dnf5/main.cpp
```diff
@@ -267,10 +267,17 @@ void RootCommand::set_argument_parser() {
     auto skip_broken = parser.add_new_named_arg("skip-broken");
     skip_broken->set_long_name("skip-broken");
     skip_broken->set_description("resolve depsolve problems by skipping packages");
-    skip_broken->set_const_value("false");
-    skip_broken->link_value(&config.strict());
+    skip_broken->set_const_value("true");
+    skip_broken->link_value(&config.skip_broken());
```
j-mracek, line 271: I think that description of both option is on incorrect place. Those options are applicable only for a set of transaction commands - for upgrade does not make sense to allow skip-broken and so on. What about to use a similar approach like we have for security options `--security`

m-blaha: That's a good point. I wasn't completely sure about making them global options. I'll move them to specific commands.

libdnf/base/goal.cpp
m-blaha: The only meaning of this argument is to distinguish whether the `error` or `warning` level is gonna be logged.
So I was thinking that `strict` in this case kind of make sense. How about calling it `is_error`?

dnf5/main.cpp
m-blaha: The change (move from global options to specific commands) was a bit bigger than expected. I added it as 3 new commits to this PR - basically to make reviewer's life easier (and my also :)

libdnf/base/goal.cpp
j-mracek: What about to use: logger::Level instead? It will be mor generic and it will state what it does.

enum class Level : unsigned int { CRITICAL, ERROR, WARNING, NOTICE, INFO, DEBUG, TRACE };

dnf5/main.cpp
j-mracek: Feel free to create an issue for dnf5 project. I am fine with not implementing what I suggested in this PR.

libdnf/base/goal.cpp
m-blaha: Makes sense. Although bool option was easier to use.

m-blaha: I've added a commit changing the `bool strict` argument of logging functions to `libdnf::Logger::Level log_level`.

j-mracek:
LGTM

PR #318, opened 2023-02-24
Add --queryformat, --requires, --provides, --conflicts .. options

kontura:
For https://github.com/rpm-software-management/dnf5/issues/122

Implement `--queryformat` and `--requires`, `--provides`, ... options.


libdnf-cli/output/repoqueryformat.cpp
```diff
@@ -223,4 +225,33 @@ void print_pkg_set_with_format(
...
+void print_pkg_attr_uniq_sorted(
+    std::FILE * target, const libdnf::rpm::PackageSet & pkgs, const std::string & getter_name) {
+    auto getter = NAME_TO_GETTER.find(getter_name);
+    if (getter == NAME_TO_GETTER.end()) {
+        throw RuntimeError(M_("package getter: %s not available"), getter_name);
+    }
+    std::set<std::string> output;
+    for (auto package : pkgs) {
+        std::visit(
+            [&output, &package](auto && getter_func) {
```
m-blaha, line 238: Why is rvalue reference needed here? Wouldn't `(const auto & getter_func)` do?

m-blaha:
Did you investigate the rawhide build error? Is it something that can be fixed on our side?

libdnf-cli/output/repoqueryformat.cpp
kontura: Right, updated, thanks. :+1:

PR #321, opened 2023-02-27
Document --strict deprecation

m-blaha:
It's a difference between dnf4 and dnf5 and we should forget to mention it in the docs.

### DIFF
--- a/CHANGES-DNF-5.md
+++ b/CHANGES-DNF-5.md
@@ -49,6 +49,15 @@ options and implemented only for related commands
 Renaming boolean options to format `--<option>`, and `--no-<option>`
 `--nobest` -> `--no-best`. Proposing to keep `--nobest` as a deprecated option.
 
+### strict configuration option deprecation
+`strict` config option is now deprecated. The problem with this option is that it does two things:
+
+1. if disabled it allows the solver to skip uninstallable packages to resolve depsolv problems
+2. if disabled it allows dnf to skip unavailable packages (this is for `install` command mostly)
+
+The functionality is now split to two config options - `skip_broken` for the uninstallable packages and `skip_unavailable` for packages not present in repositories.
+Together with these new config options there are also corresponding command line options `--skip-broken` and `--skip-unavailable` for commands where it makes sense.
+
 
 ### Upgrade command
 - New dnf5 option `--minimal` (`upgrade-minimal` command still exists as a compatibility alias for `upgrade --minimal`).


CHANGES-DNF-5.md
```diff
@@ -49,6 +49,14 @@ options and implemented only for related commands
 Renaming boolean options to format `--<option>`, and `--no-<option>`
 `--nobest` -> `--no-best`. Proposing to keep `--nobest` as a deprecated option.
 
+### --strict deprecation
+`--strict` option is now deprecated. The problem with this option is that it does two things:
+
+1. it allows the solver to skip uninstallable packages to resolve depsolv problems
+2. it allowd dnf to skip unavailable packages (this is for `install` command mostly)
```
kontura, line 56: allowd -> allowed

CHANGES-DNF-5.md
```diff
@@ -49,6 +49,14 @@ options and implemented only for related commands
 Renaming boolean options to format `--<option>`, and `--no-<option>`
 `--nobest` -> `--no-best`. Proposing to keep `--nobest` as a deprecated option.
 
+### --strict deprecation
+`--strict` option is now deprecated. The problem with this option is that it does two things:
```
kontura, line 53: If I understand correctly I would maybe reword it a bit:
- isn't the `strict` option a config option? I can't find any `--strict` switch
- I would say if the strict option was disabled it allows those two things, currently to me the text suggests that if the option was enabled it allows those things.

kontura (changes requested):

CHANGES-DNF-5.md
```diff
@@ -49,6 +49,14 @@ options and implemented only for related commands
...
+
+1. it allows the solver to skip uninstallable packages to resolve depsolv problems
+2. it allowd dnf to skip unavailable packages (this is for `install` command mostly)
+
+The functionality is now splitted to two options - `--skip-broken` for the uninstallable packages and `--skip-unavailable` for packages not present in repositories.
```
kontura, line 58: I think that while `splitted` is a word just `split` would [be better](https://www.usingenglish.com/reference/irregular-verbs/split.html).

CHANGES-DNF-5.md
m-blaha: Thanks for suggestions. I've tried to make it more clear.

kontura (approved):
LGTM

PR #322, opened 2023-02-27
SWIG: Add advisory bindings

jan-kolarik:
Generate bindings for `Advisory` objects.

This is needed for example for Ansible module, to apply security filters in `GoalJobSettings`.


kontura:
Thanks!

PR #325, opened 2023-02-27
Fix spec bconds and cmake options

dmach:
There's also a test running in GitHub Actions, but it takes over 40 minutes.
It's started only on changes in CMakeLists.txt and spec.
I'm not sure whether to keep it on by default.


ppisar:
Does libsolv provide a pkg-config file? If it does, then you should use "pkgconfig --static --libs libsolv" way instead of hard-coding libsolv dependencies into DNF5. Also dnf5.spec should build-require libsolv-static instead of repeating transitive dependencies of libsolv.

bindings/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(SWIG_COMPILE_OPTIONS
     -Wno-missing-declarations
     -Wno-missing-field-initializers
     -Wno-sign-compare
-    -Wno-sometimes-uninitialized
+#    -Wno-sometimes-uninitialized
```
kontura, line 32: This should be no longer needed, the problem was fixed in glib.

j-mracek:
@dmach There is unanswered question from ppisar. May I ask you for your plans with the PR?

inknos:
rebased here to test https://github.com/rpm-software-management/dnf5/pull/876

PR #328, opened 2023-03-03
GoalPrivate: Copy ctor to set user_installed_packages

m-blaha:
The copy of the solved GoalPrivate is used to detect skipped jobs or usage of non-best candidate in the solution.
Without user_installed_packages set the results of the copy resolving are invalid.


j-mracek:
LGTM, please feel free to merge it when tests pass

PR #331, opened 2023-03-03
SWIG: Add rpm_signature bindings

jan-kolarik:
Add SWIG bindings for `RpmSignature` class to allow checking package signatures for API users.


kontura:
Thanks!

PR #333, opened 2023-03-06
logger: Add support for logging in API scripts

jan-kolarik:
Create helper method for easily creating a file logger based on configured file path. The logic was separated from the `main` to its own file in order to be usable for API users.

Also added bindings for `GlobalLogger` class to setup libraries logging in API scripts.


m-blaha (approved):
Thank you, looks good to me.

PR #335, opened 2023-03-09
Disable parallel running of some unit tests

jrohel:
It's because of librpm. Testing CLI utilities can run in parallel.

Fixes an issue that often appeared in unit tests in the COPR build when using the multi-threaded version of the `RepoSack::update_and_load_repos`:
```
TRACE [rpm] entering chroot /tmp/libdnf5_unittest.wkCbFF/installroot/
ERROR [rpm] Unable to change root directory: Operation not permitted
```


j-mracek:
LGTM

Issue #343, opened 2023-03-13
Linters for the repo


jpodivin:
This is an extremely minor issue. But I believe it would be much easier to maintain contribution standards, as described in[0], if the project were to utilize some linters.

In Openstack we've had a very good experience with plain pre-commit[1], which is lightweight and extremely modular.
There would be a relatively significant initial patch involved, as all the rule issues would have to be resolved, but besides that it should be mostly self-sustaining.

[0] https://dnf5.readthedocs.io/en/latest/contributing/index.html
[1] https://pre-commit.com/

m-blaha:
I concur. @inknos pointed pre-commit out to me some time ago, I've been testing it for several weeks and it is really helpful.

jpodivin:
I'll propose a minimum set of hooks and we can later expand as needed. It shouldn't get too complex as otherwise nobody will run them.

jan-kolarik:
I also support this idea. As it is not currently a part of the commit process, it is sometimes annoying to find out having an extra space at the end of the line only after CI tests are initiated...

PR #344, opened 2023-03-13
Basic pre-commit configuration

jpodivin:
#343 

Pre-commit hooks are now included to check for common style violations. These checks include remains of merge conflict delimiters, trailing whitespaces, pep8 violations and clang-format.


inknos:
hey, thanks for the PR. I am using pre-commit for a while now and in the .pre-commit file I test rpmlint on `dnf5.spec`. could you add it? https://github.com/inknos/dnf5/blob/pre-commit/.pre-commit-config.yaml#L17

gotmax23:
It's probably a good idea to add a CI job to run the pre-commit hooks as well.

jpodivin:
@inknos I'm having trouble testing this locally.
When I try to run a build I'm getting multiple:

`collect2: error: ld returned 1 exit status`

test/libdnf/iniparser/test_iniparser.cpp
```diff
@@ -127,15 +127,15 @@ void IniparserTest::test_iniparser2() {
 # Test comment2
 [section1]
 key1 = value1
-key2 = two line  
+key2 = two line
```
jpodivin, line 130: Regarding this test. It's throwing errors for obvious reasons. I can either normalize the string to not trigger linters but then the assertions will have to be changed as well.

inknos:
>  I'm having trouble testing this locally.
Testing now

> Edit: However, RPM lint is not released as pre-commit hook yet. Is it? So we would first need take care of that. Otherwise it would have to be installed manually, and I don't think that would be the most resilient setup.

It is not released and yes, it needs to be installed manually now. I am not sure on what could be the best way to add it as a requirement

test/libdnf/iniparser/test_iniparser.cpp
inknos: I would exclude the tests that involve text parsing directly by instructing pre-commit to not perform the checks on such files

jpodivin: ok. That should let the checks pass.

inknos: all is passing locally now.waiting for the build and tests to be green in GH

inknos:
I would merge it and open an issue to include rpmlint in the pre-commit file, another issue to replace clang-format CI check with pre-commit. Thank you!

Issue #339, opened 2023-03-10
Do we need to add obsoletes for install action?


m-blaha:
[Bug 2176263](https://bugzilla.redhat.com/show_bug.cgi?id=2176263) - Attempting to run "dnf install systemd-udev" no longer installs systemd-udev on F37 container

This bug demonstrates really strange behavior. Let me try to describe it here:

Let's have a package **systemd-udev-1.0**.
Later, the maintainers decide to deliver a feature from this package as a separate subpackage using obsoletes:

Now we have
**systemd-udev-2.0**
  Obsoletes: systemd-udev < 2   (I would argue this is not needed since the name has not changed)

**systemd-boot-unsigned-2.0**
  Obsoletes: systemd-udev < 2  (important part is that there is only obsolete, package does not provide systemd-udev)

If you try to run `dnf5 install systemd-udev` with these three packages available, the result is kind of unexpected:
```
# dnf5 install systemd-udev
============================================================================================
 Package                      Architecture  Version              Repository            Size
============================================================================================
Installing:
 systemd-boot-unsigned        noarch        2.0-1.fc29           split-package        6.0 k

Transaction Summary
============================================================================================
Install  1 Package
```

The package `systemd-boot-unsigned` does not provide systemd-udev, only obsoletes the package.
I dumped the debugsolver data and this is interesting part of this transaction:
```
job install oneof systemd-boot-unsigned-2.0-1.fc29.noarch@split-package
                  systemd-udev-1.0-1.fc29.noarch@split-package
                  systemd-udev-2.0-1.fc29.noarch@split-package [forcebest,setevr,setarch]
```

So libdnf actually allows libsolv to install `systemd-boot-unsigned`.

Here is a simple test for the CI - https://github.com/rpm-software-management/ci-dnf-stack/commit/3301e50fafff3ab8f61ea2a3a785ff7bd3f859ed

So the question is - why we add obsoletes (without provides) to the install job? I tried to dig in dnf git history to find why the feature was added (because you can observe exactly the same behavior in dnf4), but did not find any reason for it.

j-mracek:
This is very complicated topic. The original reason of adding obsoletes was that user complains that after installation of the latest package package is replaced and remove by the first upgrade. They expect that obsoletes will be applied in the first place. Obsoletes are not related to what obsoleters provides.

m-blaha:
I did a bit more digging and found this - https://github.com/rpm-software-management/dnf/pull/1708
[...]

PR #354, opened 2023-03-15
goal: Better handling of obsoleted pkg installation

m-blaha:
In situation where a package exists in multiple versions and some older version is being obsoleted, any of obsoleters was considered a valid solution.

The result could be really misleading. For example let's have this package set:

systemd-udev-1.0

systemd-udev-2.0
Obsoletes: systemd-udev < 2

systemd-boot-unsigned-2.0
Obsoletes: systemd-udev < 2

In this case `dnf install systemd-udev` may lead to installation of systemd-boot-unsigned which is probably not what the user expected. The reason is the split in the upgrade-path created by obsolete and both branches - systemd-udev-2.0 and systemd-boot-unsigned-2.0 are considered valid.

With this patch install command takes into account only obsoleters of the best version of the package so the `dnf install systemd-udev` results in correct installation of systemd-udev-2.0 package.

Resolves: https://github.com/rpm-software-management/dnf5/issues/339
Requires tests adjustment: https://github.com/rpm-software-management/ci-dnf-stack/pull/1236

### DIFF
--- a/libdnf/base/goal.cpp
+++ b/libdnf/base/goal.cpp
@@ -45,8 +45,25 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace {
 
 void add_obsoletes_to_data(const libdnf::rpm::PackageQuery & base_query, libdnf::rpm::PackageSet & data) {
+    libdnf::rpm::PackageQuery data_query(data);
+
+    // In case there is an installed package in the `data` behave consistently
+    // with upgrade and add all the obsoleters.
+    libdnf::rpm::PackageQuery installed_data(data_query);
+    installed_data.filter_installed();
+
+    if (installed_data.empty()) {
+        // If there is no installed package in the `data`, add only obsoleters
+        // of the latest versions.  This should prevent unexpected results in
+        // case a package has multiple versions and some older version is being
+        // obsoleted.
+        // See also https://bugzilla.redhat.com/show_bug.cgi?id=2176263
+        data_query.filter_priority();
+        data_query.filter_latest_evr();
+    }
+
     libdnf::rpm::PackageQuery obsoletes_query(base_query);
-    obsoletes_query.filter_obsoletes(data);
+    obsoletes_query.filter_obsoletes(data_query);
     data |= obsoletes_query;
 }
 
--- a/libdnf/rpm/package_query.cpp
+++ b/libdnf/rpm/package_query.cpp
@@ -267,6 +267,11 @@ PackageQuery::PackageQuery(const BaseWeakPtr & base, ExcludeFlags flags, bool em
 PackageQuery::PackageQuery(libdnf::Base & base, ExcludeFlags flags, bool empty)
     : PackageQuery(base.get_weak_ptr(), flags, empty) {}
 
+PackageQuery::PackageQuery(const PackageSet & pkgset, ExcludeFlags flags)
+    : PackageQuery(pkgset.get_base(), flags, true) {
+    update(pkgset);
+}
+
 PackageQuery::PackageQuery(const PackageQuery & src) : PackageSet(src), p_pq_impl(new PQImpl(*src.p_pq_impl)) {}
 
 PackageQuery::PackageQuery(PackageQuery && src) noexcept = default;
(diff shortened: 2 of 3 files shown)


m-blaha:
Test https://github.com/rpm-software-management/ci-dnf-stack/pull/1235

m-blaha:
Oh, the test for this issue is already present so I'll only change it according to the changed behavior.

libdnf/base/goal.cpp
```diff
@@ -45,8 +45,17 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
 
 void add_obsoletes_to_data(const libdnf::rpm::PackageQuery & base_query, libdnf::rpm::PackageSet & data) {
+    // Add only for obsoleters of the latest versions.
+    // This should prevent unexpected results in case a package has multiple versions
+    // and some older version is being obsoleted.
+    // See also https://bugzilla.redhat.com/show_bug.cgi?id=2176263
+    libdnf::rpm::PackageQuery data_query(base_query.get_base(), libdnf::sack::ExcludeFlags::APPLY_EXCLUDES, true);
+    data_query.update(data);
+    data_query.filter_priority();
+    data_query.filter_latest_evr();
```
j-mracek, line 55: I am thinking about two edge cases where we might have a problem. Both cases are related to situation when package name is installed.
1. When vendor change is not allowed (configuration allow_vendor_change)
2. We will have a different set of candidates in comparison to upgrade request with the same argument

I have a suggestion - if `data` contains installed package than use the original approach. What do you think?

m-blaha: I think we should have test scenarios for these situations. I'll try to prepare them - it will definitely help me wrap my head around it.

libdnf/base/goal.cpp
```diff
@@ -45,8 +45,26 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace {
 
 void add_obsoletes_to_data(const libdnf::rpm::PackageQuery & base_query, libdnf::rpm::PackageSet & data) {
+    libdnf::rpm::PackageQuery data_query(base_query.get_base(), libdnf::sack::ExcludeFlags::APPLY_EXCLUDES, true);
```
j-mracek, line 48: This is not a problem of the PR - The creation of an empty query was cryptic to me.

j-mracek: What about to have query constructor with packageset and flags?

j-mracek:
It looks like that one test is failing
```
   @dnf5 @bz1902279
  Scenario: Install of obsoleting package, even though higher version than obsoleted present  # dnf/obsoletes.feature:34
    Given I use repository "dnf-ci-obsoletes"                                                 # dnf/steps/repo.py:182
    When I execute dnf with args "install PackageE"                                           # dnf/steps/cmd.py:106
    Then the exit code is 0                                                                   # common/output.py:57
    And Transaction is following                                                              # dnf/steps/transaction.py:200
      | Action  | Package                           |
      | install | PackageA-Obsoleter-0:1.0-1.x86_64 |
      Assertion Failed: [rpmdb] Package PackageA-Obsoleter-0:1.0-1.x86_64 not 'install'; Possible candidates: PackageE-0:3.0-1.x86_64
      Captured stdout:
```

libdnf/base/goal.cpp
m-blaha: Yup. The reason is to by-pass the missing constructor.

libdnf/rpm/package_query.cpp
```diff
@@ -267,6 +267,11 @@ PackageQuery::PackageQuery(const BaseWeakPtr & base, ExcludeFlags flags, bool em
 PackageQuery::PackageQuery(libdnf::Base & base, ExcludeFlags flags, bool empty)
     : PackageQuery(base.get_weak_ptr(), flags, empty) {}
 
+PackageQuery::PackageQuery(const PackageSet & pkgset, ExcludeFlags flags)
+    : PackageQuery(pkgset.get_base(), flags, true) {
+    update(pkgset);
```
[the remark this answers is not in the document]
j-mracek: Please the decision is up to you.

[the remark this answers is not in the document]
m-blaha: On the other hand your solution introduces code duplicity (it's basically a copy of current PackageQuery ctor).

I don't know, the difference in performance is one SolvMap `|` operation. Maybe if this constructor is massively used, then even this small performance increase may make a difference. But now there is only one place where this constructor is actually used.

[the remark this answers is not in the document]
m-blaha: Or would it be better from your point of view to use directly something like `*p_impl |= *pkgset.p_impl` instead of `update()`?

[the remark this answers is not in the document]
j-mracek: There is no performance difference between both solutions. Creation of Map and copy is cheap. And I agree that there is an overlap with other code.

Issue #350, opened 2023-03-14
Replace clang github check with pre-commit check


inknos:
Follow up to https://github.com/rpm-software-management/dnf5/pull/344. Pre-commit now includes clang and other sanity checks.

inknos:
https://github.com/rpm-software-management/dnf5/pull/362

PR #362, opened 2023-03-16
Add pre-commit to github actions and remove clang

inknos:
The aim is to have a clear output and replace clang-format actions with pre-commit. I will push some bad formatted files to test it so be aware not to merge it with bad files!

Resolves: https://github.com/rpm-software-management/dnf5/issues/350


m-blaha:
From the log it doesn't look like `python3-rpm` got installed.
Apt does not automatically fetch repository metadata so maybe running `apt-get update` before the installation should help. Also you should add `-y` switch to the install command - `apt-get install python3-rpm -y`

m-blaha:
I might be mistaken, but from the logs and experiments with ubuntu:latest container it looks like:

- ubuntu:latest is distributed with python3.10 and this version is used also by python3-rpm for rpm bindings.
- pre-commit uses virtual environment with python 3.11, therefore it cannot use pre-installed system python3-rpm and fails with `ImportError: Failed to import system RPM module. Make sure RPM Python bindings are installed on your system.`

Is it possible to run pre-commit with python 3.10?

m-blaha:
Also on https://github.com/marketplace/actions/pre-commit page they say that the action is in maintenance only mode and we should use pre-comit.ci

jpodivin:
@m-blaha We can run pre-commit with different python version. I only requires writing our own action. That is relatively simple to do with a `run` step.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - run: sudo apt-get install -y python3-rpm
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.10'
+    - uses: pre-commit/action@v3.0.0
```
jpodivin, line 19: I would suggest this instead:

```
- name: Install pre-commit
  run: pip install pre-commit
- name: Run pre-commit on entire codebase
  run: pre-commit run -a
```
It does look a bit clunky, but it would remove dependency on an unmaintained action.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

inknos: I like this, I fixed it.
I also removed clang now because the output is quite similar and fixed the pre commit issues. I believe it's good to go now. @m-blaha

inknos:
@m-blaha I did bet on the specfile issues to be solved, but they are not. let me investigate them

jpodivin:
The container seems to lack some tools. Or at very least it behaves that way. Isn't it based on RHEL by any chance?

inknos:
> The container seems to lack some tools. Or at very least it behaves that way. Isn't it based on RHEL by any chance?
> 
> https://github.com/rpm-software-management/dnf5/actions/runs/4447461711/jobs/7809033212#step:3:5

I am playing with the containers now. I think it can be simpler if we just build a container for our needs like we did for clang hooks

inknos:
The ci container was updated to run pre-commit hooks https://github.com/rpm-software-management/ci-dnf-stack/pull/1238

m-blaha:
Thank you!

PR #370, opened 2023-03-17
dnfdaemon: Handle errors when transaction is not created

m-blaha:
Client can call `goal.do_transaction()` before the transaction is created by calling `goal.resolve()`.
This patch handles such error by returning dbus error reply instead of dnf5daemon-server segfault.


jan-kolarik (approved):
LGTM

PR #372, opened 2023-03-17
Set a minimal sqlite version

ppisar:
Installing a package on a system with sqlite-3.34.1 failed with:

    Running transaction
    SQL statement compilation failed: "
	INSERT INTO
	    "pkg_name" (
		"name"
	    )
	VALUES
	    (?)
	ON CONFLICT DO NOTHING
	RETURNING "id"
    ": (1) - SQL logic error

The cause is that libdnf5 uses "RETURNING" clause in sqlite queries. A support for this clause was added in sqlite-3.35.0. (Figured out experimentally and by reading sqlite sources. Sqlite news file only mentions "adds a number of new language features".)

This patch constrains the minial sqlite version to 3.35.0.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -63,6 +63,7 @@ Obsoletes:      microdnf < 4
 %global libmodulemd_version 2.5.0
 %global librepo_version 1.15.0
 %global libsolv_version 0.7.21
+%global sqlite_version 3.35.0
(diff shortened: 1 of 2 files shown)


inknos:
LGTM

ppisar:
Fix supplied in #374.

PR #377, opened 2023-03-20
transaction: Apply keepcache option after successful run (RhBug:2176384)

jan-kolarik:
Currently, before running the actual transaction, all inbound packages are downloaded or copied to a subdirectory of the configured `cachedir` directory. This applies to all packages from remote or command-line repositories. Packages from repositories with a local xml:base are not cached at all.

In this commit, the `keepcache` option is implemented so that if it is set to `False`, all cached packages under the `cachedir` directory are removed after a next successful transaction execution. Otherwise, they are kept there. For the packages' removal, the same logic as in the `clean packages` command is used.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2176384.


jan-kolarik:
Closed in favor of the new https://github.com/rpm-software-management/dnf5/pull/459.

PR #379, opened 2023-03-20
Remove showdupesfromrepos config option

m-blaha:
- the option is not documented even for dnf4
- the name is kind of cryptic

For commands that use it (list, info, search), the functionality is replaced by --showduplicates command line option.

### DIFF
--- a/dnf5/commands/list/list.cpp
+++ b/dnf5/commands/list/list.cpp
@@ -41,7 +41,6 @@ void ListCommand::set_parent_command() {
 void ListCommand::set_argument_parser() {
     auto & ctx = get_context();
     auto & parser = ctx.get_argument_parser();
-    auto & config = ctx.base.get_config();
 
     auto & cmd = *get_argument_parser_command();
     cmd.set_description("Lists packages depending on the packages' relation to the system");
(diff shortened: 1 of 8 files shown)


dnf5/commands/list/list.cpp
```diff
@@ -58,12 +57,8 @@ void ListCommand::set_argument_parser() {
...
     cmd.register_positional_arg(specs);
 
-    auto showduplicates = parser.add_new_named_arg("showduplicates");
-    showduplicates->set_long_name("showduplicates");
-    showduplicates->set_description("Show all versions of the packages, not only the latest ones.");
-    showduplicates->set_const_value("true");
-    showduplicates->link_value(&config.get_showdupesfromrepos_option());
-    cmd.register_named_arg(showduplicates);
+    show_duplicates = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this, "showduplicates", '\n', "Show all versions of the packages, not only the latest ones.", false);
```
jan-kolarik, line 61: I am just wondering, why is the short name for the command == '\n'. Does it have some special hidden meaning?

m-blaha: To be honest, I've copy/pasted it from somewhere. But good question, I'll investigate.

m-blaha: Well, it should be `\0`

m-blaha: So it looks like it wasn't copy/paste, but my invention. Thanks for spotting this!. Fixed.

jan-kolarik (approved):
LGTM, thanks!

PR #387, opened 2023-03-22
modules: Change State to set and get the whole ModuleState

pkratoch:
This is the same as for groups and environments and it allows to set a new state (instead of just modifying existing ones).


libdnf/module/module_sack.cpp
```diff
@@ -668,15 +668,17 @@ ModuleSack::resolve_active_module_items() {
     p_impl->excludes.reset(new libdnf::solv::SolvMap(p_impl->pool->nsolvables));
 
     auto system_state = p_impl->base->p_impl->get_system_state();
+    system::ModuleState module_state;
```
j-mracek, line 671: I am sorry but I see the naming little bit confusing. ModuleState is a structure and also enum. I know that this is in different name space, but still confusing for me. Or I am completely lost.

pkratoch: I agree. Though it didn't originate here, we can discuss renaming one of them. It seems to me, that from a naming perspective it would be better to rename the enum (e.g. to ModuleEnabled or something like that?), but on the other hand, the enum is on the API, unlike the structure. But for the structure, I don't have any ideas for a good, descriptive name (maybe just to make it somehow longer, e.g. ModuleSystemState, but I don't find that very helpful). Any thoughts?

j-mracek:
LGTM

PR #394, opened 2023-03-23
Correctly load repos from installroot config file

evan-goode:
Fixes a bungle in https://github.com/rpm-software-management/dnf5/pull/271; `create_repos_from_config_file` was not reading the config file from the installroot.

Resolves https://github.com/rpm-software-management/dnf5/issues/248

### DIFF
--- a/include/libdnf/base/base.hpp
+++ b/include/libdnf/base/base.hpp
@@ -35,6 +35,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf/rpm/package_sack.hpp"
 #include "libdnf/transaction/transaction_history.hpp"
 
+#include <functional>
 #include <map>
 
 
@@ -70,6 +71,9 @@ public:
 
     /// Returns a pointer to a locked "Base" instance or "nullptr" if no instance is locked.
     static Base * get_locked_base() noexcept;
+
+    /// Call a function that loads the config file, catching errors appropriately
+    void with_config_file_path(std::function<void(const std::string &)> func);
 
     /// Loads main configuration from file defined by the current configuration.
     void load_config_from_file();
--- a/libdnf/base/base.cpp
+++ b/libdnf/base/base.cpp
@@ -89,7 +89,7 @@ void Base::load_config_from_file(const std::string & path) {
     config.load_from_parser(parser, "main", vars, *get_logger());
 }
 
-void Base::load_config_from_file() {
+void Base::with_config_file_path(std::function<void(const std::string &)> func) {
     std::filesystem::path conf_path{config.get_config_file_path_option().get_value()};
     const auto & conf_path_priority = config.get_config_file_path_option().get_priority();
     const auto & use_host_config = config.get_use_host_config_option().get_value();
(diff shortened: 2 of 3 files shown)


libdnf/base/base.cpp
```diff
@@ -92,7 +92,8 @@ void Base::load_config_from_file(const std::string & path) {
 void Base::load_config_from_file() {
     std::filesystem::path conf_path{config.get_config_file_path_option().get_value()};
     const auto & conf_path_priority = config.get_config_file_path_option().get_priority();
-    if (!config.get_use_host_config_option().get_value() && conf_path_priority < Option::Priority::COMMANDLINE) {
+    const auto & use_host_config = config.get_use_host_config_option().get_value();
```
jan-kolarik, line 95: Just to note to all the `const auto &` variables - I believe the `const` keyword is not needed there, it is automatically deduced.

libdnf/repo/repo_sack.cpp
```diff
@@ -352,17 +352,23 @@ void RepoSack::create_repos_from_file(const std::string & path) {
 }
 
 void RepoSack::create_repos_from_config_file() {
-    const auto & path = base->get_config().get_config_file_path_option().get_value();
+    std::filesystem::path conf_path{base->get_config().get_config_file_path_option().get_value()};
```
jan-kolarik, line 355: Hmm, now there is the exactly same logic on two places - in `RepoSack::create_repos_from_config_file` and in `Base::load_config_from_file` except internally calling one different method `create_repos_from_file` vs `load_config_from_file` ... Maybe we could refactor it by using one method with logic and just passing the pointer to different internal one as an argument? Maybe it's a bit overkill, but on the other hand, I feel like when there will be some change in the logic in the future we might forget about the other place.

jan-kolarik:
I'd also supply a test usecase to the CI stack, so we have this issue covered...

libdnf/repo/repo_sack.cpp
evan-goode: Yeah, it would be nice not to duplicate the logic. Just pushed a commit that does it this way. It's a little awkward, but yes, this way the try/catch doesn't have to be repeated.

A good middle ground could be a `base.config_path()` that returns the correct path to the config file, possibly one inside the installroot, and then only the try/catch would be duplicated.

include/libdnf/base/base.hpp
```diff
@@ -71,6 +72,10 @@ class Base {
     /// Returns a pointer to a locked "Base" instance or "nullptr" if no instance is locked.
     static Base * get_locked_base() noexcept;
 
+    /// Call a function that loads the config file, catching errors appropriately
+    template <typename T>
+    T with_config_file_path(std::function<T(const std::string &)> func);
```
jan-kolarik, line 77: Is making the method templated needed here? Both current usages are void functions with the same signature, do you think of any different use case now? Because, IMHO, it distracts the user and adds some unnecessary complexity there...

evan-goode: Sure, I'll take out the generic and just make it take a `void` function.

libdnf/base/base.cpp
```diff
@@ -32,6 +32,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <atomic>
 #include <cstdlib>
 #include <filesystem>
+#include <functional>
```
jan-kolarik, line 35: Sorry, just last thing, can you take this out? We are including it already in the header...

evan-goode: Sure!

jan-kolarik (approved):
Thanks for the fix!

PR #395, opened 2023-03-24
pep8 fix in the tests

m-blaha:
pre-commit starts failing on this.


jan-kolarik:
OK you were faster, thanks :)

PR #396, opened 2023-03-27
Add Python API test for resolving group transaction

jan-kolarik:
This is intended to cover the use case from #392.

### DIFF
--- a/test/python3/libdnf5/comps/test_group.py
+++ b/test/python3/libdnf5/comps/test_group.py
@@ -16,8 +16,6 @@
 # along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 
-import unittest
-
 import libdnf5
 
 import base_test_case
@@ -27,7 +25,7 @@ class TestGroup(base_test_case.BaseTestCase):
     def test_group(self):
         self.add_repo_repomd("repomd-comps-core")
         q_core = libdnf5.comps.GroupQuery(self.base)
-        core = q_core.get()
+        _ = q_core.get()
 
     def test_group_query_without_setup(self):
         # Create a new Base object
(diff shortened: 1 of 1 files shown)


test/python3/libdnf5/comps/test_group.py
```diff
@@ -27,11 +25,36 @@ class TestGroup(base_test_case.BaseTestCase):
...
+        goal_install = libdnf5.base.Goal(self.base)
+        goal_install.add_install("@core")
+
+        transaction_install = goal_install.resolve()
+        self.assertEqual(transaction_install.get_problems(),
+                         libdnf5.base.GoalProblem_NO_PROBLEM)
+        transaction_install.run()
+        self.assertFalse(transaction_install.get_transaction_problems())
+
+        self.repo_sack.update_and_load_enabled_repos(True)
```
jan-kolarik, line 51: This is a thing I was struggling with, because without reloading the sack, the `goal_remove` object then don't know about the group installed state. I was wondering whether we could update the objects state automatically as we use the same one `base` throughout the script?

kontura: I agree, that is odd. We should either handle that or at least document it. If I understand correctly then basically running a transaction invalidates the sack? Or is it just for groups/modules?

jan-kolarik: Yes, it's a general thing.

jan-kolarik: There is a fresh [issue](https://github.com/rpm-software-management/dnf5/issues/412) related to this. If there is really no working way how to reset the sack I think it's a bug, because users cannot really do more transactions within one API script.

jan-kolarik: So I would postpone this until the issue is resolved, because the current solution calling the `update_and_load_enabled_repos` twice is a bad way how to solve this and also probably unwanted - it should be either documented as you said, or fixed to be callable only once.

jan-kolarik:
Related issue: https://github.com/rpm-software-management/dnf5/issues/439.

Issue #301, opened 2023-02-16
Move a content of CHANGES-DNF-5.md to DNF5 documentation


j-mracek:
It will increase the visibility of the changes between DNF4 and DNF5

PR #398, opened 2023-03-27
Move description of DNF5 changes to doc

j-mracek:
It increases visibility of changes in DNF5, because documentation is visible on multiple places.

Resolves: https://github.com/rpm-software-management/dnf5/issues/301


jan-kolarik:
Thanks!

PR #400, opened 2023-03-27
RepoSack: remember the repo file path

praiskup:
The *.repo file name will be needed to implement the 'dnf5 copr' logic in the future.


jan-kolarik:
Thanks for the fix!

jan-kolarik (approved):

PR #401, opened 2023-03-27
modules: Add ModuleDB to store the runtime states of all modules

pkratoch:
Contains also the same commit as in https://github.com/rpm-software-management/dnf5/pull/387 , because it requires it.

### DIFF
--- a/libdnf/module/module_db.cpp
+++ b/libdnf/module/module_db.cpp
@@ -1,0 +1,302 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "module/module_db.hpp"
+
+#include "base/base_impl.hpp"
+#include "utils/bgettext/bgettext-mark-domain.h"
+
+#include "libdnf/module/module_errors.hpp"
+#include "libdnf/module/module_query.hpp"
+
+#include <modulemd-2.0/modulemd.h>
+
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <set>
+#include <string>
+
+namespace libdnf::module {
+
+
+void ModuleDB::initialize() {
+    if (initialized) {
+        return;
+    }
+
+    for (const auto & state : base->p_impl->get_system_state().get_module_states()) {
+        runtime_module_states.emplace(
+            state.first, RuntimeModuleState({.original = state.second, .changed = state.second}));
+    }
+
+    for (const auto & module_item : base->module_sack.get_modules()) {
+        auto module_name = module_item->get_name();
+        if (runtime_module_states.find(module_name) == runtime_module_states.end()) {
+            runtime_module_states.emplace(module_name, RuntimeModuleState());
+        }
+    }
+
+    initialized = true;
+}
+
+
+void ModuleDB::save() {
+    for (auto & item : runtime_module_states) {
+        base->p_impl->get_system_state().set_module_state(item.first, item.second.changed);
+    }
+}
+
+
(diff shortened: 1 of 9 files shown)


libdnf/module/module_db.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <vector>
+
+namespace libdnf::module {
+
+
+class ModuleDB {
+public:
+    ModuleDB(libdnf::system::State & system_state, ModuleSackWeakPtr module_sack)
+        : system_state(system_state),
+          module_sack(module_sack) {}
```
j-mracek, line 38: I have a proposal to resolve the init issue. What about to replace the arguments by Base argument (weak_ptr)?
Correct me if I am wrong, Module_sack is only used to get Base. And the system state could be received from Base. What do you think?

pkratoch: Oh, right. My idea was to access it via `ModuleSack`, but you are right it could be replaced by `Base` weak pointer. I'll make the change.

libdnf/module/module_db.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace libdnf::module {
+
+
+class ModuleDB {
+public:
+    ModuleDB(BaseWeakPtr base) : base(base) {}
```
kontura, line 36: We could take a const reference `ModuleDB(const BaseWeakPtr & base)` instead of a copy.

libdnf/module/module_db.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+namespace libdnf::module {
+
+
+class ModuleDB {
+public:
+    ModuleDB(const BaseWeakPtr & base) : base(base) {}
+
+    /// Update the `system::State` with all the changes.
+    void save_everything_to_system_state();
```
j-mracek, line 39: Note - very descriptive name, but may be it is too descriptive - it is an opinion with no evidence

libdnf/module/module_db.cpp
```diff
@@ -0,0 +1,346 @@
...
+        return runtime_module_states.at(module_name);
+    } catch (std::out_of_range &) {
+        throw NoModuleError(M_("No such module: {}"), module_name);
+    }
+}
+
+
+ModuleDB::RuntimeModuleState & ModuleDB::get_runtime_module_state(const std::string & module_name) {
+    try {
+        return runtime_module_states.at(module_name);
```
j-mracek, line 82: May I ask you what is an advantage of the approach in comparison to `find()` method?

libdnf/module/module_db.cpp
```diff
@@ -0,0 +1,346 @@
...
+    } catch (std::out_of_range &) {
+    }
+
+    add_runtime_module_state(module_name);
```
j-mracek, line 86: The methods `get_runtime_module_state` calls 2x `runtime_module_states.at(module_name)` and 1X `runtime_module_states.find(module_name)` . It casts not much but it is not optimal.

pkratoch: Hm, right now, there are 3 calls the first time the state is accessed, 1 call afterwards. I could either change it to 2 calls always, or change the `add_runtime_module_state` to return the object it just created or that existed before (then the `add_runtime_module_state` could be called directly, but then it probably should be renamed to not be confusing). What do you think?

libdnf/module/module_db.hpp
j-mracek: I will use example - please do not take it as a suggestion for a new name.

`save_everything_to_system_state()` vs. `commit()`
The first description says not only what it does, but also where the information will be stored. In case that we will change internal implementation, it can happen that the method will have a different implementation but the name will keep original description. The longer name also is not in favor for some programmers (personal opinion in C++).

The more generic name is more robust for future changes. Then it requires only a modification of the method description, when the implementation is changed.

pkratoch: Ok, you convinced me. I renamed it to `save`, but we can discuss any other name if you want.

libdnf/module/module_db.cpp
pkratoch: `find` is better in this case. I changed it.

libdnf/module/module_db.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+
+    /// Get the runtime module state for given module name.
+    const RuntimeModuleState & get_runtime_module_state(const std::string & module_name) const;
+    RuntimeModuleState & get_runtime_module_state(const std::string & module_name);
+
+    /// Get all module names for modules that newly have the given status.
+    std::vector<std::string> get_all_changed_modules(const ModuleStatus new_status) const;
+    /// Get all module names and changes in their profiles if there are any.
+    std::map<std::string, std::vector<std::string>> get_all_changed_profiles(bool installed = true) const;
```
j-mracek, line 95: I am sorry but installed argument is not described.

pkratoch: Fixed.

j-mracek:
LGTM

PR #403, opened 2023-03-27
Command::get_parent_command(): return the real parent

praiskup:
Previously this method just returned the pointer to itself.  The future 'dnf5 copr' will depend on this functionality in the future.


include/libdnf-cli/session.hpp
```diff
@@ -128,7 +128,7 @@ class Command : public libdnf::cli::ArgumentParserUserData {
     ///         The returned pointer must **not** be freed manually.
     /// @since 5.0
     Command * get_parent_command() const noexcept {
-        return static_cast<Command *>(argument_parser_command->get_user_data());
+        return static_cast<Command *>(argument_parser_command->get_parent()->get_user_data());
```
jan-kolarik, line 131: For root command, this will throw an exception, because `get_parent()` is `nullptr`.

praiskup: Indeed, good catch.

PR #404, opened 2023-03-28
API for GPG signature checking of transaction packages

### LINKED ISSUE #386: Expose GPG signatures checking and recovery process to API
Currently we have an API [method](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf/rpm/rpm_signature.hpp) for checking packages GPG signatures which returns an error code. In the context [class](https://github.com/rpm-software-management/dnf5/blob/dd990eb5209e5163bd36b5021af1e9195f814afc/dnf5/context.cpp#L548) we are translating these error codes into user messages and there is also a recovery process attempt regarding importing additional repo keys. This whole process could be exposed to API in order to be usable for our clients like Ansible.

jan-kolarik:
Includes handling of recovery process when new keys are automatically tried to be imported. Also some refactoring was done with intention for a better code readability + switching to translatable strings.

Closes #386.

### DIFF
--- a/include/libdnf/base/transaction.hpp
+++ b/include/libdnf/base/transaction.hpp
@@ -31,6 +31,11 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <optional>
 
 
+namespace libdnf::rpm {
+class KeyInfo;
+}  // namespace libdnf::rpm
+
+
 namespace libdnf::base {
 
 class TransactionGroup;
@@ -132,6 +137,30 @@ public:
 
     /// Retrieve list of problems that occurred during transaction run attempt
     std::vector<std::string> get_transaction_problems() const noexcept;
+
+    /// @brief Check signatures of packages in the resolved transaction.
+    ///
+    /// @param import_keys When the parameter is on, problems related to missing GPG keys are
+    /// tried to be automatically recovered by importing the missing ones.
+    ///
+    /// @return True if all packages have correct signatures or checking is turned off with `gpgcheck` option,
+    /// otherwise false. More info about occurred problems can be retrieved using the `get_gpg_signature_problems`
+    /// method.
+    bool check_gpg_signatures(bool import_keys = true);
+
+    /// @brief Check signatures of packages in the resolved transaction.
+    ///
+    /// @param import_confirm_func This function is used for deciding whether to import the given GPG key
+    /// when a problem related to missing keys occurs.
+    ///
+    /// @return True if all packages have correct signatures or checking is turned off with `gpgcheck` option,
+    /// otherwise false. More info about occurred problems can be retrieved using the `get_gpg_signature_problems`
+    /// method.
+    // TODO(jkolarik): To be reworked as we don't want std::function exposed on the public API
+    bool check_gpg_signatures(std::function<bool(const libdnf::rpm::KeyInfo &)> & import_confirm_func);
+
+    /// Retrieve a list of the problems that occurred during `check_gpg_signatures` procedure.
+    std::vector<std::string> get_gpg_signature_problems() const noexcept;
 
 private:
     friend class TransactionGroup;
--- a/include/libdnf/rpm/rpm_signature.hpp
+++ b/include/libdnf/rpm/rpm_signature.hpp
@@ -99,6 +99,9 @@ public:
     /// @param key_url: URL of the public key to be imported.
     std::vector<KeyInfo> parse_key_file(const std::string & key_url);
 
+    /// Return string representation of the CheckResult enum
+    static std::string check_result_to_string(CheckResult result);
+
 private:
     BaseWeakPtr base;
 };
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -34,6 +34,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf/common/exception.hpp"
 #include "libdnf/repo/package_downloader.hpp"
 #include "libdnf/rpm/package_query.hpp"
+#include "libdnf/utils/format.hpp"
 
 #include <fmt/format.h>
 #include <unistd.h>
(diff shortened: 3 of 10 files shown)


jan-kolarik:
Freshly rebased.

test/libdnf/base/test_transaction.hpp
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
```
dcantrell, line 6: See other comment about license boilerplates.

dcantrell (changes requested):
I do not see any major problems.  I put some comments in there.  Most of mine are about consistency calling them GPG or PGP keys and cleaning up some of the messages.

libdnf/base/transaction.cpp
```diff
@@ -87,10 +95,33 @@ Transaction::Impl::~Impl() {
...
 
+Transaction::Impl::Impl(Transaction & transaction, const BaseWeakPtr & base)
+    : transaction(&transaction),
+      base(base),
+      rpm_signature(base) {}
+
+Transaction::Impl::Impl(Transaction & transaction, const Impl & src)
+    : transaction(&transaction),
+      base(src.base),
+      libsolv_transaction(src.libsolv_transaction ? transaction_create_clone(src.libsolv_transaction) : nullptr),
```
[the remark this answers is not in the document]
jan-kolarik: This particular block is only copied from `.hpp` here. I agree with you, also don't like overusing ternary operators for anything other than very simple expressions. But looking into `libsolv`, it seems that any such checking for `NULL` inputs in other `extern` functions are not present, so I doubt the maintainer would fix this specific case, but anyway I will give it a go :slightly_smiling_face:

test/libdnf/base/test_transaction.cpp
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
```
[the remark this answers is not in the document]
jan-kolarik: As this is a general issue in our codebase I've created an [issue](https://github.com/rpm-software-management/dnf5/issues/410) for this.

libdnf/base/transaction.cpp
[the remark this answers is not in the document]
jan-kolarik: Hmm I was checking again through the `libsolv` codebase and it seems the failing hard and early for any `NULL` input parameter is a standard there, so I guess I'd leave it as it is. I don't think this can be considered as a bad or good way of handling this in `libsolv` and if it is treated in a consistent way, I would respect it.

include/libdnf/rpm/rpm_signature.hpp
```diff
@@ -99,6 +99,9 @@ class RpmSignature {
     /// @param key_url: URL of the public key to be imported.
     std::vector<KeyInfo> parse_key_file(const std::string & key_url);
 
+    /// Return string representation of the CheckResult enum
+    static std::string check_result_to_string(const CheckResult result);
```
jrohel, line 103: Parameter "result" is passed by value. const-qualification of parameters only has an effect in function definitions. The header file is for the user. It is unnecessary to tell him that the copy inside the implementation is a constant.

include/libdnf/base/transaction.hpp
```diff
@@ -130,6 +135,29 @@ class Transaction {
...
+
+    /// @brief Check signatures of packages in the resolved transaction.
+    ///
+    /// @param import_confirm_func This function is used for deciding whether to import the given GPG key
+    /// when a problem related to missing keys occurs.
+    ///
+    /// @return True if all packages have correct signatures or checking is turned off with `gpgcheck` option,
+    /// otherwise false. More info about occurred problems can be retrieved using the `get_gpg_signature_problems`
+    /// method.
+    bool check_gpg_signatures(std::function<bool(const libdnf::rpm::KeyInfo &)> & import_confirm_func);
```
jrohel, line 156: `std::function` on public API? Probably a new complication for SWIG binding. Especially if we want to be able to define a callback in the target language.

But there is a solution. The target language can override virtual methods of a wrapped C++ class. Code can 'call up' from C++ into the target language by simply calling a virtual method overridden in a derived class in the target language.

And we use this approach. `libdnf::repo::DownloadCallbacks` for "download'' callbacks and `libdnf::repo::RepoCallbacks` for "import repo keys" callbacks. Previously, the `RepoCallbacks` class was also used for download repository metadata callbacks. I consolidated the download callbacks and today the `RepoCallbacks` class is only used for "import repo key" callback (virtual method `repokey_import`. I wanted to rename the class (to `RepoKeyImportCallbacks` for example).

In "dnf5/context.cpp", the `user_confirm_key` function is basically a copy of the `KeyImportRepoCB::user_confirm_key` method.
**So, the solution is to unify the code and use the class method everywhere.**

I am considering whether to merge this PR as it is. It's a step forward. And perform the unification in the next PR.

jan-kolarik: I know about that and it wasn't intended to be used from SWIG bindings. It is there basically just because of that `Context` usage. I was considering a `friend class` and private method for this purpose, but the `Context` class is outside the library scope, so I didn't think of this as a good idea.

Your proposal looks good, if it's OK, I will do it in another PR and put a TODO comment there for now.

test/libdnf/base/test_transaction.hpp
jan-kolarik: TBD in separate issue.

test/python3/libdnf5/base/test_transaction.py
```diff
@@ -0,0 +1,44 @@
+# Copyright Contributors to the libdnf project.
+#
+# This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+#
+# Libdnf is free software: you can redistribute it and/or modify
```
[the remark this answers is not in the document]
jan-kolarik: TBD in separate issue.

include/libdnf/base/transaction.hpp
jrohel: @jan-kolarik  OK. I will merge it. A TODO comment is enough for now. You can continue in the next PR.

Issue #410, opened 2023-03-29
Replace license boilerplates in source files


jan-kolarik:
Based on the Fedora licensing improvements from the last year, we should drop the long license headers in the source code and just add the matching SPDX identifier instead like:
```
SPDX-License-Identifier: GPL-2.0-or-later
```
See the announcement [here](https://communityblog.fedoraproject.org/important-changes-to-software-license-information-in-fedora-packages-spdx-and-more/) and the docs [here](https://docs.fedoraproject.org/en-US/legal/update-existing-packages/).

m-blaha:
FWIW there was a discussion on this topic some time ago: https://github.com/rpm-software-management/libdnf/pull/1259

evan-goode:
If possible, I would like to replace all license headers in the repository with SPDX headers that are consistent with https://github.com/rpm-software-management/dnf5/blob/main/COPYING.md. So most files would have:

```
// Copyright Contributors to the DNF5 project
// SPDX-License-Identifier: GPL-2.0-or-later
```

and files under the 'libdnf5', 'libdnf5-cli', 'libdnf5-plugins', 'include' and 'bindings' subdirectories are dual-licensed, so they'd be:

```
// Copyright Contributors to the DNF5 project
// SPDX-License-Identifier: (GPL-2.0-or-later or LGPL-2.1-or-later)
```

But the current situation of the headers is messy. All the copyright headers refer to the project as "Libdnf", for historical reasons, I'm sure. Some headers say "Copyright (C) 2022 Red Hat, Inc." and some say "Copyright Contributors to the libdnf project." Some files under `libdnf5/` are LGPL-2.0-or-later instead of LGPL-2.1-or-later like COPYING.md says.

So if we want to clean this up and switch to SPDX license identifiers, we should proceed carefully. For example, we may need to word the new header as "Copyright Contributors to the DNF5 project and Contributors to the Libdnf project" since many DNF5 files are derived from Libdnf, and in some ways they are two separate projects.

evan-goode:
Talking with @dcantrell, we should:

* Keep any existing `Copyright` lines, such as

    ```
    Copyright Contributors to the libdnf project.
    Copyright (C) 2021 Red Hat, Inc.
    ````
* Add the line `Copyright Contributors to the dnf5 project` (the language most preferred by RH legal is "Copyright DNF5 Authors", but we already have "Contributors to the dnf5 project" in some files and this is fine)
[...]

PR #414, opened 2023-03-30
dnf5 copr plugin

praiskup:
Currently work-in-progress, just the boilerplate plugin code, and a basic 'copr list' functionality for listing enabled copr projects.


dnf5-plugins/CMakeLists.txt
```diff
@@ -6,3 +6,4 @@ include_directories("${PROJECT_SOURCE_DIR}/dnf5/include/")
 
 add_subdirectory("changelog_plugin")
 add_subdirectory("builddep_plugin")
+add_subdirectory("copr_plugin")
```
praiskup, line 9: Does it make sense to have it here (dnf5-plugins)?  Or do you think we could make it a default dnf5 command (non-plugin)?  We don't expect any additional third-party dependencies that would enlarge the default minimal installations.

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,286 @@
+/// TODO: copyright headers mention libdnf?
```
praiskup, line 1: I can see most of the files say: `Copyright Contributors to the libdnf project.`.  Is it OK to spell it DNF5 project nowadays?

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include <regex>
+
+
+namespace dnf5 {
+
+using namespace libdnf::cli;
+
+#define dnf5_command "dnf5"
+
+#define warning(fmt, ...)                  \
```
praiskup, line 19: I failed to find a built-in warning utility, is there some?

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    // Check that the repo filename matches the repo ID
+    if (repo_id_to_copr_file(repo_id) != basename(dnfRepo->get_repo_file_path())) {
+        warning(_("Repo id '%s' doesn't belong to '%s' repofile"), repo_id.c_str(), repo_file_path);
+    }
+}
+
+static bool is_copr_repofile(const std::string & repofile) {
+    auto base = basename(repofile);
```
jrohel, line 251: I prefer using the existing `filename` method rather than reimplementing it.
`auto filename = repofile.filename();`

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,285 @@
...
+        return;
+    }
+
+    // Check that the repo filename matches the repo ID
+    if (repo_id_to_copr_file(repo_id) != basename(dnfRepo->get_repo_file_path())) {
+        warning(_("Repo id '%s' doesn't belong to '%s' repofile"), repo_id.c_str(), repo_file_path);
+    }
+}
+
+static bool is_copr_repofile(const std::string & repofile) {
```
jrohel, line 250: `is_copr_repofile(const std::filesystem::path & repofile)`

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,285 @@
...
+    auto base = basename(repofile);
+    return fnmatch("_copr:*", base.c_str(), 0) == 0;
```
jrohel, line 252: Why are you using regex in some places and fnmatch in others? Is using fnmatch an optimization? What about:
`return filename.native().starts_with("_copr:");`

dnf5-plugins/CMakeLists.txt
jrohel: @praiskup Good question. I don't want to answer yes or no now. But something to think about.
For example, the `install` command is a general command. Usable by everyone regardless of Linux distribution.
What about the `copr` command? Is this a general command? Or is it specific? For example, for users of a specific Linux distribution (eg Fedora). What will be the benefit of having it installed everywhere as a default command?

dnf5-plugins/copr_plugin/copr.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+        warn_weird_copr_repo(repo);
+        copr_repos[repo_file_path].add_dnf_repo(repo);
+    }
+
+    for (const auto & [key, copr_repo] : copr_repos) {
+        if (hubspec != "" && copr_repo.id.rfind(hubspec + "/", 0) != 0)
+            continue;
+        cb(copr_repo);
+    }
```
jrohel, line 275: You can use C++20. `starts_with` is much more readable.
Example:
```
    for (const auto & [key, copr_repo] : copr_repos) {
        if (hubspec.empty() || copr_repo.id.starts_with(hubspec + "/")) {
            cb(copr_repo);
        }
    }
```

dnf5-plugins/copr_plugin/copr.cpp
jan-kolarik: I'd suggest to use the license header from some of the existing file even if it's not looking right. There is a plan to do some batch renaming/change regarding these boilerplates [here](https://github.com/rpm-software-management/dnf5/issues/410).

dnf5-plugins/CMakeLists.txt
praiskup: It is mostly usable by anyone I think (but my perspective might be different). Copr provides a wide variety of chroots potentially consumable on any RPM distribution (Mageia/OpenSUSE/Fedora/...).  In some situations it is viable to enable "fedora-rawhide-x86_64" on "any-x86_64" host, because the packages are noarch and installable anywhere.

Benfit is clear;  closer integration with Copr (e.g. in `Containerfile` there's no need to `dnf5 install -y dnf5-plugins` first, you go directly to `dnf5 copr enable alice/project`).

dnf5-plugins/copr_plugin/copr.cpp
praiskup: Nice, thank you for the tip.

praiskup:
Thank you for all the tips!  I'll update the PR with your suggestions later this week (including commands enable/disable/remove).

dnf5-plugins/copr_plugin/copr.cpp
jrohel: @praiskup 
> I failed to find a built-in warning utility, is there some?

No. But I wanted to add. Thanks for the reminder.

In general, try to use type-safe code.
- C++ templates instead of macros
- `libdnf::utils::sformat` instead of `printf`

NOTE: `libdnf::utils::sformat` has a format string similar to Python formating. See https://en.cppreference.com/w/cpp/utility/format/format. Don't use `std::format`, in Fedora 37 is GCC12 and support for `std::format` starts from GCC13.

Example:
```
template <typename... Args>
void warning(const char * format, Args &&... args) {
    std::cerr << "WARNING: " + libdnf::utils::sformat(format, std::forward<Args>(args)...) << std::endl;
}
```

praiskup:
There's still a lot of stuff missing, but let's have it pushed here in case someone wants to take an early look.

dnf5-plugins/copr_plugin/copr.cpp
praiskup: ok

dnf5-plugins/copr_plugin/json.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "tempfile.hpp"
+
+#include "libdnf/repo/file_downloader.hpp"
+
+#include <fstream>
+
+Json::Json(libdnf::Base & base, const std::string & url) {
+    auto temp_file = TempFile();
+    libdnf::repo::FileDownloader downloader(base);
```
praiskup, line 30: Do we have a better "download" support in DNF5?  I dislike the way the progress bar touches stdout.

dnf5-plugins/copr_plugin/tempfile.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef DNF5_COMMANDS_COPR_TEMPFILE_HPP
+#define DNF5_COMMANDS_COPR_TEMPFILE_HPP
+
+#include <unistd.h>
+
+#include <filesystem>
+
+/// TODO: make copr plugin work with libdnf::utils::fs::TempFile (not a libdnf API ATM)
+class TempFile {
```
praiskup, line 28: Ideas how to better do this?

dnf5-plugins/copr_plugin/json.hpp
```diff
@@ -0,0 +1,45 @@
...
+#define DNF5_COMMANDS_COPR_JSON_HPP
+
+#include "libdnf/base/base.hpp"
+
+#include <json.h>
+
+class Json {
+private:
+    bool cleanup;
+    struct json_object * root;
```
praiskup, line 30: I asked on IRC without any answer (praiskup> Hi, is there some tooling in dnf5 (usable from a plugin) that would allow me to download a json file into memory and parse?)

This json-c wrapper seemed pretty trivial to implement, so I did it (without thinking too much about the API).  Lemme know if it deserves some tweaks to make it more useful.

dnf5-plugins/CMakeLists.txt
j-mracek: This is really tricky.  The most important part is related to containers and minimal images. It provides functionality for managing repositories not delivered by RPMs. I can use copr repositories without copr plugin, therefore I believe that it is not an essential functionality, but enhancement. Somehow I have a feeling that it would be good to keep it as a plugin.

m-blaha:
It looks like the default hub is not being correctly used:
```
❯ ./dnf5/dnf5 copr enable phracek/PyCharm
 https:///api_3/rpmrepo/phracek/PyCharm/fedora-37/                  ???% |   0.0   B/s |   0.0   B |  00m16s
>>> Curl error (6): Couldn't resolve host name for https://api_3/rpmrepo/phracek/PyCharm/fedora-37/ [Could n
>>> Curl error (6): Couldn't resolve host name for https://api_3/rpmrepo/phracek/PyCharm/fedora-37/ [Could n
>>> Curl error (6): Couldn't resolve host name for https://api_3/rpmrepo/phracek/PyCharm/fedora-37/ [Could n
>>> Curl error (6): Couldn't resolve host name for https://api_3/rpmrepo/phracek/PyCharm/fedora-37/ [Could n
>>> Curl error (6): Couldn't resolve host name for https://api_3/rpmrepo/phracek/PyCharm/fedora-37/ [Could nFailed to download files
```

Also there is missing new line between the progress bar and `Failed to download files` message.

m-blaha:
I noticed a significant increase in the execution time between dnf4 and dnf5:

**dnf4:**
```
❯ time dnf copr list
copr.fedorainfracloud.org/atim/starship
copr.fedorainfracloud.org/jstanek/centos-packager
copr.fedorainfracloud.org/mblaha/clang-format-13 (disabled)
copr.fedorainfracloud.org/mblaha/distro (disabled)
copr.fedorainfracloud.org/mblaha/rpm (disabled)
copr.fedorainfracloud.org/mblaha/rpm-gitoverlay (disabled)
copr.fedorainfracloud.org/mblaha/sdbus-cpp
copr.fedorainfracloud.org/mikelo2/PyDrive2 (disabled)
copr.fedorainfracloud.org/nsella/thinbox (disabled)
copr.fedorainfracloud.org/phracek/PyCharm (disabled)
copr.fedorainfracloud.org/pmatilai/rpm-snapshot (disabled)
copr.fedorainfracloud.org/psss/tmt
copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf-nightly-sanitizers (disabled)
copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf5-unstable
copr.fedorainfracloud.org/rpmsoftwaremanagement/rpm-gitoverlay
copr.fedorainfracloud.org/rpmsoftwaremanagement/test-utils
copr.fedorainfracloud.org/thozza/bughunting (disabled)
copr.fedorainfracloud.org/group_mock/mock (disabled) *
copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf-nightly *
* These coprs have repo file with an old format that contains no information about Copr hub - the default one was assumed. Re-enable the project to fix this.

real    0m0.867s
user    0m0.526s
sys     0m0.072s
```

**dnf5:**
```
❯ time ./dnf5/dnf5 copr list
WARNING: Copr repofile '/etc/yum.repos.d/_copr_@mock-mock.repo' has an old format, fix by re-enabling it $ dnf5 copr enable @mock/mock
WARNING: Repo id 'mblaha-rpm-gitoverlay' in '/etc/yum.repos.d/_copr:copr.fedorainfracloud.org:mblaha:rpm-gitoverlay.repo' is suspicious
WARNING: Copr repofile '/etc/yum.repos.d/_copr_rpmsoftwaremanagement-dnf-nightly.repo' has an old format, fix by re-enabling it $ dnf5 copr enable rpmsoftwaremanagement/dnf-nightly
copr.fedorainfracloud.org/atim/starship
copr.fedorainfracloud.org/jstanek/centos-packager
copr.fedorainfracloud.org/mblaha/clang-format-13 (disabled)
copr.fedorainfracloud.org/mblaha/distro (disabled)
 (disabled)
copr.fedorainfracloud.org/mblaha/rpm (disabled)
copr.fedorainfracloud.org/mblaha/sdbus-cpp
copr.fedorainfracloud.org/mikelo2/PyDrive2 (disabled)
copr.fedorainfracloud.org/nsella/thinbox (disabled)
copr.fedorainfracloud.org/phracek/PyCharm (disabled)
copr.fedorainfracloud.org/pmatilai/rpm-snapshot (disabled)
copr.fedorainfracloud.org/psss/tmt
copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf-nightly-sanitizers (disabled)
copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf5-unstable
copr.fedorainfracloud.org/rpmsoftwaremanagement/rpm-gitoverlay
copr.fedorainfracloud.org/rpmsoftwaremanagement/test-utils
copr.fedorainfracloud.org/thozza/bughunting (disabled)

real    0m8.996s
user    0m7.228s
sys     0m0.045s
```

dnf5-plugins/copr_plugin/json.hpp
m-blaha: I would prefer to use some external C/C++ library over having to maintain our own implementation.
But I haven't done any thorough investigation so I do not know what is available. Maybe `rapidjson-devel` package which seems to be part of RHEL buildroot?

praiskup:
> It looks like the default hub is not being correctly used for `enable` subcommand:

Hmmm, I'm going to take a look...

> Also there is missing new line between the progress bar and `Failed to download files` message.

Going to add there a newline then, if it is "as easy as that" :-).

> I noticed a significant increase in the execution time between dnf4 and dnf5:

I suppose this happens because of some "magic" in the ./dnf5/dnf5 library linking, this never
happens with the /usr/bin/dnf5 library:
`sudo --preserve-env=DNF5_PLUGINS_DIR --preserve-env=LD_LIBRARY_PATH dnf5 copr list`
My experience is quite the opposite (dnf5 is much faster).

m-blaha:
> > I noticed a significant increase in the execution time between dnf4 and dnf5:
> 
> I suppose this happens because of some "magic" in the ./dnf5/dnf5 library linking, this never happens with the /usr/bin/dnf5 library: `sudo --preserve-env=DNF5_PLUGINS_DIR --preserve-env=LD_LIBRARY_PATH dnf5 copr list` My experience is quite the opposite (dnf5 is much faster).

OK, sorry for the confusion. I looked into it a bit and the part that takes so long is system state loading. So nothing's wrong with your PR.

dnf5-plugins/copr_plugin/json.hpp
praiskup: I'd probably prefer keeping this trivial 50-liner (when already done), because additional dep might easily become a valid reason to not move the "copr" plugin into the set of built-in commands in the future.  Just my preference.

praiskup:
>  Also there is missing new line between the progress bar and Failed to download files message.

I'm rather asking explicitly, do you really want me to touch this libdnf code?  What I do is just:

https://github.com/rpm-software-management/dnf5/blob/5ca509c77f78afc3f38de4d3dd16500989aec40d/dnf5-plugins/copr_plugin/download_file.cpp#L27-L29

So it probably means we have to touch the libdnf's downloader logic.

m-blaha:
> > Also there is missing new line between the progress bar and Failed to download files message.
> 
> I'm rather asking explicitly, do you really want me to touch this libdnf code? What I do is just:
> 
> https://github.com/rpm-software-management/dnf5/blob/5ca509c77f78afc3f38de4d3dd16500989aec40d/dnf5-plugins/copr_plugin/download_file.cpp#L27-L29
> 
> So it probably means we have to touch the libdnf's downloader logic.

The fix will be in dnf5/main.cpp, so the issue is unrelated to this PR. I'm going to provide a patch later.

dnf5-plugins/CMakeLists.txt
mcurlej: From my POW it is better to have all non-dnf related features in a plugin. It makes things more granular and with a direct way to distinguish which is a crucial part of DNF and what is not. Also if ever COPR would be discontinued this will not affect DNF in any way. DNF is a package manager and a library first and should be build system agnostic.

praiskup:
Updated

dnf5-plugins/CMakeLists.txt
praiskup: Accepted, let's have it as plugin (and perhaps revamp the idea in the future?).

Just a small correction, Copr is not just a build system, it is a place where projects grow.
Especially from this plugin perspective where we solve "how to consume project content"
(definitely not "how to build").  The functionality is close to the configmanager logic.

Copr is not just Fedora Copr;  there are other instances out there already - and more
deployments are likely to appear because deployment is easier nowadays.

m-blaha:
Thank you! Let's test in on users :)

PR #415, opened 2023-03-30
doc: Add info about package spec expressions (RhBug:2160420)

jan-kolarik:
Provide documentation regarding supported dependency expressions when passing package specs.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2160420


doc/misc/specs.7.rst
```diff
@@ -132,6 +132,19 @@ against NEVRAs and file provides.
 ``<package-name-spec>`` is matched against NEVRAs only.
 
 
+Dependency expressions
+======================
+
+When passing a ``<package-spec>``, the user can provide additional restriction
```
ppisar, line 138: Since the paragraph extends <package-spec>, I would keep the paragraph in Packages section, just after first paragraph "Many  commands…file provides.

doc/misc/specs.7.rst
```diff
@@ -132,6 +132,19 @@ against NEVRAs and file provides.
...
+rules for matching the arguments. Basic version comparisons, such as
+``<package-spec> >= <version>``, can be used for this purpose.
```
ppisar, line 140: I would name all operators (=, >=, <=). That's to tackle another problem in which people write == because DNF4 interprets it as =.

doc/misc/specs.7.rst
```diff
@@ -125,6 +125,15 @@ Many commands take a ``<package-spec>`` parameter that selects a package for
 the operation. The ``<package-spec>`` argument is matched against package
 NEVRAs, provides and file provides.
 
+When passing a ``<package-spec>``, the user can provide additional restriction
+rules for matching the arguments. Basic version comparisons can be used for this
+purpose (=, >, <, >=, <=), like this ``<package-spec> >= <version>``.
```
ppisar, line 130: Now it looks like "NEVRA >= VERSION" would be acceptable. What about rephrasing it like "When \<package-spec\> is a package name or a provide, the user can…"?

Maybe we should also specify that \<version\> here is actually "[EPOCH:]VERSION[-RELEASE]" as specified in NEVRA Matching section.

ppisar:
Now it looks good for me. Thanks for improving the documentation.

PR #421, opened 2023-03-31
[dnf5] Implement new argument "--show-new-leaves"

jrohel:
Shows newly installed leaf packages and packages that became leaves after a transaction.

I intend that the implementation of the `--show-new-leaves` argument in DNF5 will replace the functionality of the "show-leaves" plugin that is in the old DNF.

Unlike the plugin in the old DNF, this code calculates the new state (the list of new leaf packages) before the transaction is executed. The user will see the list even before confirming the transaction.

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -61,6 +61,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/logger/global_logger.hpp>
 #include <libdnf5/logger/memory_buffer_logger.hpp>
 #include <libdnf5/repo/repo_cache.hpp>
+#include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 #include <libdnf5/version.hpp>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


dnf5/main.cpp
```diff
@@ -825,6 +841,47 @@ static void dump_variables(Context & context) {
...
+
+    pkg_query.filter_leaves();
+    pkg_query.difference(pre_trans_leaves_query);
+
+    std::set<std::string> pre_trans_leaves_na;
+    std::set<std::string> new_leaves_na;
+    for (const auto package : pre_trans_leaves_query) {
+        pre_trans_leaves_na.insert(package.get_na());
+    }
+    for (const auto package : pkg_query) {
```
jan-kolarik, line 868: Maybe rather loop using `const auto &`, also in the for cycle above?

jrohel: `PackageQuery` uses a bitmap internally. The package is created during the dereference of the iterator and is returned by value. So using `&` reference after `auto` does not save anything. It will be a reference to the object that will be created anyway. I wanted to emphasize that the Package objects are being created and intentionally did not put `&`.

But ok, I'll add `&`. In this case, the compiler should generate the same code after optimization. So it will be a cosmetic change.

jan-kolarik: Thanks for the explanation. So it makes better sense to omit the `&` there, but as of now, we already have a lot of `const auto &` usages iterating through the package queries in our codebase, so at least we are keeping it consistent :slightly_smiling_face:

jan-kolarik (approved):

PR #426, opened 2023-04-03
Hide swig/python detected a memory leak message

kontura:
Since swig-4.1 (from f38) message warning about memory leak can be
printed when libdnf5 python api is used. It is reproducible with just:
```
import libdnf5.base

def asd():
    pass

base = libdnf5.base.Base()
```
But also happens in our ci-dnf-stack python API tests, therefore it is
blocking our switch to f38 as a base for CI: https://github.com/rpm-software-management/ci-dnf-stack/pull/1226

Currenly it happens due to a bug in SWIG therefore we wan't to silent
the message.
Unfortunately this might hide some real problems.

### DIFF
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -245,3 +245,6 @@ del ClassName##__iter__
 %template(PreserveOrderMapStringPreserveOrderMapStringString) libdnf::PreserveOrderMap<std::string, libdnf::PreserveOrderMap<std::string, std::string>>;
 
 %exception;  // beware this resets all exception handlers if you import this file after defining any
+
+// Base weak ptr is used across the codebase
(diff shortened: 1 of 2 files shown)


bindings/libdnf5/advisory.i
```diff
@@ -39,6 +39,7 @@
 
 #define CV __perl_CV
 
+%include "libdnf/base/base_weak.hpp"
```
jan-kolarik, line 42: Maybe it could be added into the `common.i` file and not including it in each `*.i`?

kontura: That sounds good, I have updated the PR.
Thanks.

jan-kolarik (approved):
Thanks!

PR #436, opened 2023-04-04
Use smartcols for search output

inknos:
To benefit from machine-readable output (which is something we might want in the future) from libsmartcols, the search command should format the output using the same logic as other DNF5 command.
I believe we should at least unify the output style as much as possible and don't rely on two different libraries (libsmartcols, iostream) to achieve the same goal


j-mracek:
@inknos May I ask you for a better description what PR should implement or what kind of issue it resolves? I am sorry, I was unable to get it from brief look and the description was not helpful to me (may be not enough coffee). Thank you.

inknos:
@j-mracek I updated the description

inknos:
closing and tracking here https://github.com/rpm-software-management/dnf5/issues/674

PR #444, opened 2023-04-11
dnfdaemon: Fix download callbacks

m-blaha:
This patch adapts dnfdaemon callbacks to unified repo::DownloadCallbacks introduced in commit d24cad69.
The new implementation is much simpler and resolves several segfaults in the dnf5daemon-server.

- specialized RepoCB and PackageDownloadCB are replaced with shared DownloadCB
- since repository and package downloads now share the same callbacks, signals are moved from REPO to BASE interface
- added new "download_add_new" signal for the new download callback "add_new_download"

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2184276

### DIFF
--- a/dnf5daemon-client/callbacks.cpp
+++ b/dnf5daemon-client/callbacks.cpp
@@ -43,156 +43,107 @@ bool DbusCallback::signature_valid(sdbus::Signal & signal) {
 }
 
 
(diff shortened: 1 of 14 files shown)


martinpitt:
Any chance that this could be reviewed soon? Thanks!

Issue #317, opened 2023-02-23
`dnf5 install --help` does not provide help


j-mracek:
```
dnf5 install --help
Missing positional argument "specs" for command "install"
```
To receive a help for the command it requires provide argument

rafagomes:
Hello @j-mracek, my name is Rafa, and I'm interested in contributing to the Fedora project. I have experience with Python, Javascript, and C++ (I want to improve that and help in any way I can). I would love to work on this issue. Can anyone guide me on how to get started with this issue? Or can I check what has been done to the group and others in the code and follow the same?

Any help would be appreciated. I don't want to start breaking crystal glasses. :)

j-mracek:
Hello @rafagomes, thank you for the interest in DNF5 project. I will ask my colleague @jrohel to provide the initial guidance.

rafagomes:
Thank you @j-mracek.

rafagomes:
Thank you @jrohel 

I'll open a PR asap. <3

praiskup:
IMO we need to take a look at [this part](https://github.com/rpm-software-management/dnf5/blob/99f6bc40f01a08f4f532c38b348ef33e1cd4e06e/dnf5/main.cpp#L673-L695).

The `--help` argument is handled if and only if no exception was thrown in `parse()`.  So should there be some fix in argument parser logic that will avoid throwing the exception with `--help` specified?

rafagomes:
@praiskup That is what I'm planning to do. I'll try to start this today...
Maybe we need to split responsibilities... if --help/-h is passed, we don't go checking unnecessary expectations.

praiskup:
This is IMO a high-priority issue (it is not possible to get the --help output at all in some cases).
Therefore please consider (a short-term work-around?) in #445.

rafagomes:
Well, I was trying to make my first contribution. I'll move on to another project from Fedora then... Two things I disagree with and made me frustrated and I think is fair to give my feedback:
- I asked a question and waited for 2,5weeks to get a reply;
- When I finally have some direction I receive a "short-term workaround" for something that is not even launched.
- A PR is made in front of me. Well, I tried...

Good luck all <3

praiskup:
@rafagomes sorry to hear that, by no means #445 was meant to **solve** this issue (that is just a related PR, see the commit message).  I am just a contributor here, not the DNF maintainer (I'm working on #414, and this issue is rather a painful thing, really).

[...]

PR #445, opened 2023-04-11
dnf5: print --help output in case of _any_ ArgumentParserError

praiskup:
Especially for positional arguments, this is extremely helpful for users because there's no other way to print out the --help output (argument parser just swallows the --help, and throws a "missing argument" exception).

Relates: #317

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -672,8 +672,8 @@ int main(int argc, char * argv[]) try {
     {
         try {
             context.get_argument_parser().parse(argc, argv);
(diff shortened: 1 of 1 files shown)


jrohel:
@praiskup I know from your post https://github.com/rpm-software-management/dnf5/issues/317#issuecomment-1504108191 that this was intended as short-term work-around. Since there was a new contributor here, I waited for his PR and didn't create my own. But it turned out differently, so I'd rather create my own PR https://github.com/rpm-software-management/dnf5/pull/450 than merge a short-term work-around. Sorry, I'm closing this PR.

Issue #281, opened 2023-02-12
[RFE] Expose detect_releasever() functionality in Python API


gotmax23:
The dnf API allows retrieving the system release with `dnf.rpm.detect_releasever()`. libdnf5 has this same functionality in https://github.com/rpm-software-management/dnf5/blob/836d5dbfb061843582b141dbe0ee144def92e55e/libdnf/conf/vars.cpp#L132, but it is not exposed via the (Python) API. Can this be made available?

gotmax23:
A Python function to detect the basearch should also be exposed.

j-mracek:
By the way you can get `releasever` value from vars

```
import libdnf5

base = libdnf5.base.Base()

if base.get_vars().contains("releasever"):
    print("releasever_1", base.get_vars().get_value("releasever"))

base.setup()

if base.get_vars().contains("releasever"):
    print("releasever_2", base.get_vars().get_value("releasever"))
```

gotmax23:
fedrq has a "smartcache" feature where it changes the cachedir based on the value of the system's releasever [^1]. It sets `cachedir` before calling `base.setup()`, so it can't call `base.get_vars().get_value("releasever")`, as `base.setup()` is what populates the Vars object.

Currently, fedrq's libdnf5 backend contains a slightly modified version of the `get_releasever()` function from dnf (https://git.sr.ht/~gotmax23/fedrq/tree/15a8946340c3eb791fe67252fdea120556f6cd37/item/src/fedrq/backends/libdnf5/backend.py#L805) which I'd very much like to drop.

[^1]: See `smartcache` in https://fedrq.gtmx.me/fedrq5/#options. https://git.sr.ht/~gotmax23/fedrq/tree/15a8946340c3eb791fe67252fdea120556f6cd37/item/src/fedrq/config.py#L260-267 is the relevant code.

j-mracek:
Thank you for additional information. It was exactly what I need to know.

gotmax23:
Thanks for adding this! fedrq [now uses](https://git.sr.ht/~gotmax23/fedrq/commit/f88b8f5813b2e5b8f411c6865b3bb3210c61d525) the new interface when available.

PR #448, opened 2023-04-12
Add detect_release to public API

j-mracek:
There is a use case where they need to detect releasever before setting cachedir and calling base.setup()

Resolves: https://github.com/rpm-software-management/dnf5/issues/281


j-mracek:
I think that I added blocked duet to usage of unique_ptr. We will investigate whether we can increase minimal swig version to use inbuilt wrapper around.

PR #450, opened 2023-04-12
[dnf5] Fix help in case argument parser detect error

jrohel:
- if the user wants help (--help/-h argument is present) -> print help and only help without errors and the application ends with the return code libdnf::cli::ExitCode::SUCCESS.
- if user doesn't want help (--help/-h argument is not present) and argument parser detects missing/bad argument -> print an error message with a note about '--help' and application exits with return code libdnf::cli::ExitCode::ARGPARSER_ERROR

Solves: https://github.com/rpm-software-management/dnf5/issues/317

Requires modification of CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1247

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -670,27 +670,29 @@ int main(int argc, char * argv[]) try {
 
     // Parse command line arguments
     {
+        auto & arg_parser = context.get_argument_parser();
         try {
-            context.get_argument_parser().parse(argc, argv);
-        } catch (libdnf::cli::ArgumentParserUnknownArgumentError & ex) {
-            // print help if an unknown command is provided
-            std::cerr << ex.what() << std::endl;
-            context.get_argument_parser().get_selected_command()->help();
-            return static_cast<int>(libdnf::cli::ExitCode::ARGPARSER_ERROR);
-        } catch (const std::exception & ex) {
-            std::cout << ex.what() << std::endl;
+            arg_parser.parse(argc, argv);
+        } catch (libdnf::cli::ArgumentParserError & ex) {
+            // Error during parsing arguments. Try to find "--help"/"-h".
+            for (int idx = 1; idx < argc; ++idx) {
+                if (strcmp(argv[idx], "-h") == 0 || strcmp(argv[idx], "--help") == 0) {
+                    arg_parser.get_selected_command()->help();
+                    return static_cast<int>(libdnf::cli::ExitCode::SUCCESS);
+                }
+            }
+            std::cerr << ex.what() << _(". Add \"--help\" for more information about the arguments.") << std::endl;
             return static_cast<int>(libdnf::cli::ExitCode::ARGPARSER_ERROR);
         }
 
-        auto & arg_parser = context.get_argument_parser();
+        // print help of the selected command if --help was used
+        if (arg_parser.get_named_arg("help", false).get_parse_count() > 0) {
+            arg_parser.get_selected_command()->help();
+            return static_cast<int>(libdnf::cli::ExitCode::SUCCESS);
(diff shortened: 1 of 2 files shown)


jrohel:
This PR is OK. According to the output, there is a bug in the CI tests. The "alldeps" option is really missing in the last command.

Last Command: dnf5 -y --installroot=/tmp/dnf_ci_installroot_ze5yse_k --releasever=29 --setopt=module_platform_id=platform:f29 --setopt=reposdir=/tmp/dnf_ci_installroot_ze5yse_k/etc/yum.repos.d --setopt=config_file_path=/tmp/dnf_ci_installroot_ze5yse_k/etc/dnf/dnf.conf --setopt=cachedir=/tmp/dnf_ci_installroot_ze5yse_k/var/cache/**dnf download abcde --alldeps**

Last Command stderr:
**Option alldeps should be used with resolve**

m-blaha:
I think that's what the scenario tests - `--alldeps` without `--resolve` print error message.
But the real problem is different exit code. Command exited with `1` but `2` (we use this value for argparsing errors) was expected.

dnf5/main.cpp
```diff
@@ -784,18 +786,6 @@ int main(int argc, char * argv[]) try {
...
-        std::cerr << ex.what() << std::endl;
-        context.get_argument_parser().get_selected_command()->help();
-        return static_cast<int>(libdnf::cli::ExitCode::ARGPARSER_ERROR);
-    } catch (libdnf::cli::ArgumentParserMissingDependentArgumentError & ex) {
-        std::cerr << ex.what() << std::endl;
-        context.get_argument_parser().get_selected_command()->help();
-        return static_cast<int>(libdnf::cli::ExitCode::ARGPARSER_ERROR);
-    } catch (libdnf::cli::ArgumentParserError & ex) {
-        std::cerr << ex.what() << std::endl;
-        return static_cast<int>(libdnf::cli::ExitCode::ARGPARSER_ERROR);
```
praiskup, line 798: Removing the `catch` statements evokes that ArgumentParserError may actually never be raised in the corresponding `try` statement.  Does it mean that we should never ever `throw` it in `command->run()` etc.?

jrohel: @praiskup  I thought about it. There is at least one case where we throw an exception in command->configure (a problem with the download command). So I will modify the PR.

jrohel: @m-blaha Good point. Thanks.

jrohel:
CI tests fails again:
1. The original code in some cases writes the argument parser error to stdout instead of stderr. I think it was a bug. Now all arguments parser errors are written to stderr. (`search` command)
2. A test using an exact match cannot pass because a note about the `--help` argument is added. (`download` command)

dnf5/main.cpp
```diff
@@ -670,29 +670,31 @@ int main(int argc, char * argv[]) try {
...
+            arg_parser.parse(argc, argv);
+        } catch (libdnf::cli::ArgumentParserError & ex) {
+            // Error during parsing arguments. Try to find "--help"/"-h".
+            for (int idx = 1; idx < argc; ++idx) {
+                if (strcmp(argv[idx], "-h") == 0 || strcmp(argv[idx], "--help") == 0) {
+                    arg_parser.get_selected_command()->help();
+                    return static_cast<int>(libdnf::cli::ExitCode::SUCCESS);
+                }
+            }
+            std::cerr << ex.what() << ". Add \"--help\" for more information about the arguments." << std::endl;
```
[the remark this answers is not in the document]
jrohel: I added. This reminded me to look at localization in dnf5 in general.

PR #455, opened 2023-04-14
Implement repoquery --unneeded option

m-blaha:
Resolves: https://github.com/rpm-software-management/dnf5/issues/454


kontura:
Thanks!

Issue #132, opened 2022-11-11
Implement autoremove command


mcurlej:
**Command:** ```autoremove```

Aliases for [explicit NEVRA matching](https://dnf.readthedocs.io/en/latest/command_ref.html#specifying-nevra-matching-explicitly-label): ```autoremove-n```, ```autoremove-na```, ```autoremove-nevra```

```dnf [options] autoremove```

Removes all “leaf” packages from the system that were originally installed as dependencies of user-installed packages, but which are no longer required by any such package.

Packages listed in [installonlypkgs](https://dnf.readthedocs.io/en/latest/conf_ref.html#installonlypkgs-label) are never automatically removed by this command.

```dnf [options] autoremove <spec>...```

**The team decided to drop <spec> feature support**

This is an alias for the [Remove Command](https://dnf.readthedocs.io/en/latest/command_ref.html#remove-command-label) command with clean_requirements_on_remove set to ```True```. It removes the specified packages from the system along with any packages depending on the packages being removed. Each ```<spec>``` can be either a ```<package-spec>```, which specifies a package directly, or a ```@<group-spec>```, which specifies an (environment) group which contains it. It also removes any dependencies that are no longer needed.

There are also a few specific autoremove commands ```autoremove-n```, ```autoremove-na``` and ```autoremove-nevra``` that allow the specification of an exact argument in the NEVRA (name-epoch:version-release.architecture) format.

This command by default does not force a sync of expired metadata. See also [Metadata Synchronization](https://dnf.readthedocs.io/en/latest/command_ref.html#metadata-synchronization-label).

- [x] implement autoremove command

  - priority medium
  - Suggesting to drop a support of arguments, because it is default behavior of `remove` command anyway
    - ```dnf [options] autoremove``` syntax only

- [x] ```autoremove-n```, ```autoremove-na``` and ```autoremove-nevra```
- - priority low
- - document as not supported by DNF5. First we need to find a way how to replace the functionality

PR #456, opened 2023-04-14
dnf5: Add autoremove command

m-blaha:
Implements `dnf5 autoremove` command. According to https://github.com/rpm-software-management/dnf5/issues/132 the <spec> arguments support is dropped, explicit NEVRA matching is currently not supported.

For: https://github.com/rpm-software-management/dnf5/issues/132


dnf5/commands/autoremove/autoremove.cpp
```diff
@@ -0,0 +1,57 @@
...
+    context.set_load_system_repo(true);
+    context.set_load_available_repos(Context::LoadAvailableRepos::NONE);
+}
+
+void AutoremoveCommand::run() {
+    auto & ctx = get_context();
+    libdnf::rpm::PackageQuery unneeded(ctx.base);
+    unneeded.filter_unneeded();
+    auto goal = get_context().get_goal();
+    goal->set_allow_erasing(true);
```
j-mracek, line 51: I am curious, why the command requires `allow_erasing == true`?

I am suggesting that it is a left over from handling of spec code from DNF4. Or did I overlook something?
In principle unneeded packages are not supposed to remove any dependent package.
The question is how it works with excludes when I exclude top package from dependency tree.

m-blaha: I was thinking about a chain of dependencies:
```
# dnf5 install Pkg
[Pkg - userinstalled] ---requires--> [Unneeded - dependency] ---requires--> [UnneededDep]

# dnf5 upgrade Pkg && new version of Pkg dropped dependency on Unneeded

# dnf5 autoremove
```

Here hopefully `Unneeded` package should be removed with all it's dependencies (`UnneededDep`). I think to enable removal of unused dependencies this flag is needed. Or am I mistaken?

j-mracek: From DNF4:

```
        if any([self.opts.grp_specs, self.opts.pkg_specs, self.opts.filenames]):
            self.base.conf.clean_requirements_on_remove = True
            demands.allow_erasing = True
            # disable all available repos to delete whole dependency tree
            # instead of replacing removable package with available packages
            demands.available_repos = False
        else:
            demands.available_repos = True
            demands.fresh_metadata = False
```
As you can see, `demands.allow_erasing = True` is only set when argument (spec) is provided.
I believe that line `goal->set_allow_erasing(true);` could be removed without any issue. I also believe that unneeded query does not return only unneeded leave packages but also unneeded dependencies of such a packages.

m-blaha: OK, I've done some tests and it seems you are right :) I've removed the flag.

j-mracek:
LGTM

j-mracek (approved):

PR #459, opened 2023-04-14
Implement keepcache functionality (RhBug:2176384)

jan-kolarik:
This is a follow-up to previous "simple" solution from https://github.com/rpm-software-management/dnf5/pull/377.

Current behavior should be more robust compared to the original PR. Now all package files that were downloaded when user requested the `keepcache=False` mode are tracked in the special TOML file and they are removed after the next successful transaction run.

Download command has an exception with overridden behavior to always keep downloaded packages on the system as this is matching the existing DNF behavior.

The list of the packages is created before the actual download begins, so then if any error is encountered and any data has been already downloaded, they are still removed after the next transaction.

Additionally, when user runs the `dnf5 clean all` command, all metadata downloaded under the `cachedir`, including the file tracking downloaded packages, are removed.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2176384.
CI tests and fixes: https://github.com/rpm-software-management/ci-dnf-stack/pull/1239

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -23,9 +23,12 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "utils/bgettext/bgettext-mark-domain.h"
 
 #include <libdnf/conf/option_string.hpp>
+#include <libdnf/repo/package_downloader.hpp>
 #include <libdnf/rpm/package.hpp>
 #include <libdnf/rpm/package_query.hpp>
 #include <libdnf/rpm/package_set.hpp>
+
+#include <iostream>
 
 
 namespace dnf5 {
@@ -146,7 +149,16 @@ void DownloadCommand::run() {
     }
 
     if (!download_pkgs.empty()) {
-        download_packages(download_pkgs, ctx.base.get_config().get_destdir_option().get_value().c_str());
+        libdnf::repo::PackageDownloader downloader(ctx.base);
+        auto destination = ctx.base.get_config().get_destdir_option().get_value().c_str();
+
+        for (auto & pkg : download_pkgs) {
+            downloader.add(pkg, destination);
+        }
+
+        std::cout << "Downloading Packages:" << std::endl;
+        downloader.download(true, true);
+        std::cout << std::endl;
     }
 }
 
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -152,6 +152,9 @@ void DownloadCommand::run() {
         libdnf::repo::PackageDownloader downloader(ctx.base);
         auto destination = ctx.base.get_config().get_destdir_option().get_value().c_str();
 
+        // for download command, we don't want to mark the packages for removal
+        downloader.force_keep_packages(true);
+
         for (auto & pkg : download_pkgs) {
             downloader.add(pkg, destination);
         }
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -150,13 +150,12 @@ void DownloadCommand::run() {
 
     if (!download_pkgs.empty()) {
         libdnf::repo::PackageDownloader downloader(ctx.base);
-        auto destination = ctx.base.get_config().get_destdir_option().get_value().c_str();
 
         // for download command, we don't want to mark the packages for removal
         downloader.force_keep_packages(true);
 
         for (auto & pkg : download_pkgs) {
-            downloader.add(pkg, destination);
+            downloader.add(pkg);
         }
 
         std::cout << "Downloading Packages:" << std::endl;
--- a/include/libdnf/repo/package_downloader.hpp
+++ b/include/libdnf/repo/package_downloader.hpp
@@ -36,7 +36,8 @@ class PackageDownloadError : public Error {
 
 class PackageDownloader {
 public:
-    PackageDownloader();
+    explicit PackageDownloader(const libdnf::BaseWeakPtr & base);
+    explicit PackageDownloader(libdnf::Base & base);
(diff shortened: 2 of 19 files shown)


libdnf/repo/package_downloader.cpp
```diff
@@ -90,12 +92,16 @@ static int mirror_failure_callback(void * data, const char * msg, const char * u
 
 
 class PackageDownloader::Impl {
+public:
+    Impl(const BaseWeakPtr & base) : base(base) {}
+    BaseWeakPtr base;
```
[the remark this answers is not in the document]
jan-kolarik: This is based on the `file_downloader.cpp` behavior as it is a very similar context. But is there any actual problem with that? It is available only in this `.cpp` file if I am not mistaken ...

dnf5/commands/download/download.cpp
```diff
@@ -152,7 +155,18 @@ void DownloadCommand::run() {
...
     if (!download_pkgs.empty()) {
-        download_packages(download_pkgs, destdir_option->get_value().c_str());
+        libdnf::repo::PackageDownloader downloader(ctx.base);
+
+        // for download command, we don't want to mark the packages for removal
+        downloader.force_caching(true);
+
+        for (auto & pkg : download_pkgs) {
+            downloader.add(pkg, destdir_option->get_value().c_str());
+        }
```
j-mracek, line 165: I think it would be better to use configuration value directly - see https://github.com/rpm-software-management/dnf5/pull/521

jan-kolarik:
We've also discussed including changes related to `*Downloader` classes to refactor usages related to `fail_fast` and `resume` arguments, but I consider it already out of scope of this PR. Therefore a separate issue was created for this: https://github.com/rpm-software-management/dnf5/issues/528.

libdnf/repo/temp_files_memory.cpp
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright Contributors to the libdnf project.
```
j-mracek, line 2: What about to use `dnf5 project`

libdnf/repo/temp_files_memory.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
```
j-mracek, line 4: What about to modify the link to `dnf5`?

libdnf/repo/temp_files_memory.cpp
jan-kolarik: This was already discussed in some other PR, these headers are same across the codebase and should be replaced in a batch. Referenced in a separate issue: https://github.com/rpm-software-management/dnf5/issues/410

libdnf/repo/temp_files_memory.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+void TempFilesMemory::add_files(const std::vector<std::string> & paths) {
+    auto files = get_files();
+    files.insert(files.end(), paths.begin(), paths.end());
+
+    // sort and deduplicate resulting vector
+    std::sort(files.begin(), files.end());
+    files.erase(std::unique(files.begin(), files.end()), files.end());
+
+    utils::fs::File(full_memory_path, "w").write(toml::format(toml::value({{FILE_PATHS_TOML_KEY, files}})));
```
j-mracek: I guess that the line could generate various number of errors. Is it covered somehow. 
Is the operation safe when multiple instances of DNF will be modifying the file?

j-mracek: > I guess that the line could generate various number of errors. Is it covered somehow.

It is OK, the exception part is well described in method. Perfect job.

jan-kolarik: I think you are not referring to the latest changes in the file which is weird as the PR is already updated for some time... Anyway, this should be handled by this commit: https://github.com/rpm-software-management/dnf5/pull/459/commits/709f245cfa2189bd06af0b3bb1771f19e475e9f2.

j-mracek: I am sorry, I just realized that in the next commit

dnf5/commands/download/download.cpp
```diff
@@ -152,6 +152,9 @@ void DownloadCommand::run() {
         libdnf::repo::PackageDownloader downloader(ctx.base);
         auto destination = ctx.base.get_config().get_destdir_option().get_value().c_str();
 
+        // for download command, we don't want to mark the packages for removal
+        downloader.force_keep_packages(true);
```
j-mracek, line 156: This is not a mistake, but I think it is redundand. I guess that the download command always use destdir (default value is '"."').

include/libdnf/repo/package_downloader.hpp
```diff
@@ -36,7 +36,8 @@ class PackageDownloadError : public Error {
 
 class PackageDownloader {
 public:
-    PackageDownloader();
+    explicit PackageDownloader(const libdnf::BaseWeakPtr & base);
+    explicit PackageDownloader(libdnf::Base & base);
```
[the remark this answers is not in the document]
jan-kolarik: I have discussed this with Nicola and I've updated the `CHANGELOG.md` regarding this. The changelogs from commit messages are not used now.

jan-kolarik:
@j-mracek PR is ready for the next round of review now.

j-mracek:
LGTM

j-mracek (approved):

jan-kolarik:
The recently added `Keep reason for installonly packages` test from `installonly.feature` was failing due to non-existing `cachedir` during the `PackageDownloader::download()` call. I fixed this use case, hopefully the tests should be OK now.

PR #460, opened 2023-04-14
[dnf5] Support for "--userinstalled" argument in repoquery

jrohel:
Part of https://github.com/rpm-software-management/dnf5/issues/122

### DIFF
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -54,6 +54,9 @@ void RepoqueryCommand::set_argument_parser() {
     installed_option = dynamic_cast<libdnf::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
 
+    userinstalled_option = dynamic_cast<libdnf::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
+
     leaves_option = dynamic_cast<libdnf::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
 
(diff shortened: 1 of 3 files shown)


j-mracek:
LGTM

jrohel:
@j-mracek I made one more change. But it is only a modification of the text in the documentation.

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -79,6 +82,13 @@ void RepoqueryCommand::set_argument_parser() {
     installed->set_const_value("true");
     installed->link_value(installed_option);
 
+    auto userinstalled = parser.add_new_named_arg("userinstalled");
+    userinstalled->set_long_name("userinstalled");
+    userinstalled->set_description("Limit to packages that were installed by user.");
```
j-mracek, line 87: I am sorry, but I would woild like to modify the description. I checked the implementation and it correctly returns all packages that are not installed as dependency or weak dependency. It means everything what was installed by user request, or indirectly as a part of module profile or comps group. Additionally it returns as user installed packages packages with unknown reason.

I think that it is correct, then the description and may be even name of the option does not fit.

jrohel: @j-mracek I changed the description and the documentation.

doc/commands/repoquery.8.rst
```diff
@@ -78,6 +78,11 @@ Options
 ``--installed``
     | Display only installed packages.
 
+``--userinstalled``
+    | Limit to packages that are not installed as dependencies or weak dependencies.
+    | This means limit to packages that were installed at the user request or indirectly as a part of a module profile or comps group. Additionally it returns packages with unknown reason.
```
j-mracek, line 83: I am thinking about modiffication of - `Additionally it returns packages with unknown reason.`
What about ``Additionally it returns packages installed by a different package manager due to missing reason``. But it is still not perfect.

j-mracek: Let's keep the original statement

j-mracek:
LGTM, feel free to merge it when all tests pass

PR #466, opened 2023-04-18
Fix packit configuration

inknos:


.packit.yaml
```diff
@@ -0,0 +1,24 @@
+files_to_sync:
+  - dnf5.spec
+
+jobs:
+  - job: propose_downstream
+    trigger: release
+    # Use the stage instance once it works in downstream.
```
jan-kolarik, line 7: Sorry, but I don't understand the comment. Could you please clarify?

.packit.yaml
```diff
@@ -0,0 +1,24 @@
...
+    dist_git_branches:
+      - fedora-all
```
jan-kolarik, line 9: What is the meaning of the `fedora-all` here?

.packit.yaml
inknos: /me copypasting with no coffee today. sorry, I it's fixed now

.packit.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - dnf5.spec
+
+jobs:
+  - job: propose_downstream
+    trigger: release
+    dist_git_branches:
+        - fedora-rawhide
+        - fedora-38
+  - job: koji_build
+    trigger: commit
```
jan-kolarik, line 11: Is it like after each commit the build is triggered?

inknos: as I understood it, after a commit in dist-git, yes
https://packit.dev/docs/fedora-releases-guide/#koji-build-job

jan-kolarik: I am just wondering, because compared to our `ci.yml` is seems like those jobs there are independent and I was rather thinking about this process being in a single pipeline.

jan-kolarik: Oh, right. I was a bit confused if it is triggered by a commit in the dnf5 repo...

jan-kolarik (approved):

Issue #416, opened 2023-03-30
dnf5 should provide/obsolete dnf and yum


j-mracek:
- [x] Add dnf obsolete, dnf symlink and dnf provide
  - [x] Add yum obsolete, yum symlink and yum provide
    - [x] requires changes in yum-utils
  - [x] modify DNF protection in DNF package (Fedora 37+) to allow removal during upgrade to Fedora 39

evan-goode:
> Add dnf obsolete, dnf symlink and dnf provide
> Add yum obsolete, yum symlink and yum provide

Addressed in https://github.com/rpm-software-management/dnf5/pull/469

> Modify DNF protection in DNF package (Fedora 37+) to allow removal during upgrade to Fedora 39

Addressed in https://github.com/rpm-software-management/dnf/pull/1924

evan-goode:
>  requires changes in yum-utils

See https://github.com/rpm-software-management/dnf-plugins-core/pull/488

mcurlej:
Can we have a better issue name? Thanks... :)

Issue #417, opened 2023-03-30
Move ownership of /etc/dnf/dnf.conf to libdnf5


j-mracek:
- [x] Remove requirements of dnf-data from multiple components (dnf5, dnf5-daemon, microdnf?)

- [x] Add requirement to libdnf5 to python3-dnf, microdnf, ... because they require /etc/dnf/dnf.conf

- [x] Clean all values in /etc/dnf/dnf.conf and keep only empty main section (distribution overrides has a different location)

evan-goode:
> Remove requirements of dnf-data from multiple components (dnf5, dnf5-daemon, microdnf?)

Removed from `dnf5` and `dnf5daemon-server` in https://github.com/rpm-software-management/dnf5/pull/469. Removed from `microdnf` in https://github.com/rpm-software-management/microdnf/pull/131. 

> Add requirement to libdnf5 to python3-dnf, microdnf, ... because they require /etc/dnf/dnf.conf

Requirement added to `python3-dnf` in https://github.com/rpm-software-management/dnf/pull/1924. Added to `microdnf` in https://github.com/rpm-software-management/microdnf/pull/131.

> Clean all values in /etc/dnf/dnf.conf and keep only empty main section (distribution overrides has a different location)

New /etc/dnf/dnf.conf (now belonging to `libdnf5`) is here: https://github.com/rpm-software-management/dnf5/pull/469

jan-kolarik:
@evan-goode I guess this is done now?

PR #469, opened 2023-04-19
Have DNF update itself to DNF5 

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/416.


dnf5.spec
```diff
@@ -21,6 +21,12 @@ Provides:       microdnf = %{version}-%{release}
 Obsoletes:      microdnf < 4
 %endif
 
+Provides:       dnf = %{version}-%{release}
+Obsoletes:      dnf < 5
+
+Provides:       %{name}-yum = %{version}-%{release}
+Obsoletes:      %{name}-yum < 5
```
j-mracek, line 28: Are you sure about this? I guess that it will obsolete dnf5-yum and this package is not in distribution. I guess that 
correct would be `Obsoletes:      dnf-yum < 5`. Or did I miss something?

dnf5.spec
```diff
@@ -603,6 +612,12 @@ Core DNF5 plugins that enhance dnf5 with builddep and changelog commands.
 %install
 %cmake_install
 
+ln -sr %{buildroot}%{_bindir}/dnf5 %{buildroot}%{_bindir}/dnf
+
+# YUM compatibility
+# ln -sr  %{buildroot}%{confdir}/%{name}.conf %{buildroot}%{_sysconfdir}/yum.conf
```
j-mracek, line 618: Why you keep the comment about yum.conf?

dnf5.spec
```diff
@@ -240,11 +248,15 @@ License:        LGPL-2.1-or-later
 Requires:       libsolv%{?_isa} >= %{libsolv_version}
 Requires:       librepo%{?_isa} >= %{librepo_version}
 Requires:       sqlite-libs%{?_isa} >= %{sqlite_version}
+Conflicts:      dnf-data < 4.16.0
```
evan-goode, line 251: DNF5 should conflict with any version of `dnf-data` before ownership of /etc/dnf/dnf.conf was moved to libdnf5.

evan-goode:
Currently blocked by https://github.com/rpm-software-management/dnf/pull/1924: `dnf5.spec` needs to know the version of dnf-data when ownership of /etc/dnf/dnf.conf was moved to libdnf5.

Conan-Kudo:
Why not just move the `dnf-data` package itself to dnf5 instead of moving its contents to libdnf5?

dnf5/config/etc/dnf/dnf.conf
```diff
@@ -0,0 +1,3 @@
+# see `man dnf.conf` for defaults and possible options
+
+[main]
```
jan-kolarik, line 3: This is different compared to the default `dnf.conf` in the `dnf` component. Doesn't it mean, that when deploying a fresh system having the dnf5 already as the default, then the users would have different configuration than those upgrading from `dnf`?

evan-goode: Yes, that was one of the requirements in https://github.com/rpm-software-management/dnf5/issues/417. My understanding is that we used to have distributions set their default config values via providing their own dnf.conf, but now in DNF 5 there is a different mechanism for distribution overrides so we no longer need to prepopulate dnf.conf.

In general, I think it is best to leave default values out of configuration files, or keep them commented out. That way, if we want to change the defaults, we can do so in e.g. `config_main.cpp`, and a user will only have an option set in dnf.conf if they have set it themselves.

jan-kolarik (approved):
LGTM

PR #476, opened 2023-04-20
transaction: Delete removed groups from system state

m-blaha:
Fixes inconsistency in the system state after group removal.

### DIFF
--- a/libdnf/base/transaction.cpp
+++ b/libdnf/base/transaction.cpp
@@ -724,6 +724,7 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
         std::filesystem::create_directories(comps_xml_dir);
         for (const auto & tsgroup : groups) {
             auto group = tsgroup.get_group();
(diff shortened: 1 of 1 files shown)


libdnf/base/transaction.cpp
```diff
@@ -700,6 +700,9 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
                 system_state.set_group_state(group.get_groupid(), state);
                 // save the current xml group definition
                 group.serialize(comps_xml_dir / (group.get_groupid() + ".xml"));
```
kontura, line 702: I am wondering if we want to keep the serialized group xml.
It would probably be cleaner to also delete it?

m-blaha: Good point.

m-blaha: I've added code for deleting the file.

kontura (approved):
Thanks!

PR #477, opened 2023-04-20
clang-format: use grep -E instead of egrep

praiskup:
The GNU grep installation on Fedora 38 throws this warning: egrep: warning: egrep is obsolescent; using grep -E


jan-kolarik (approved):
Thanks for that!

Issue #138, opened 2022-11-11
Implement group command


mcurlej:
**Command:** ```group```
**Aliases:** ```grp```
**Deprecated aliases:** ```groups```, ```grouplist```, ```groupinstall```, ```groupupdate```, ```groupremove```, ```grouperase```, ```groupinfo```

Groups are virtual collections of packages. DNF keeps track of groups that the user selected (“marked”) installed and can manipulate the comprising packages with simple commands.

- [ ] ```dnf [options] group [summary] <group-spec>```

Display overview of how many groups are installed and available. With a spec, limit the output to the matching groups. summary is the default groups subcommand.

- [x] ```dnf [options] group info <group-spec>```

Display package lists of a group. Shows which packages are installed or available from a repository when ```-v``` is used.

- [x] ```dnf [options] group install [--with-optional] <group-spec>...```

Mark the specified group installed and install packages it contains. Also include optional packages of the group if ```--with-optional``` is specified. All mandatory and Default packages will be installed whenever possible. Conditional packages are installed if they meet their requirement. If the group is already (partially) installed, the command installs the missing packages from the group. Depending on the value of [obsoletes configuration option](https://dnf.readthedocs.io/en/latest/conf_ref.html#obsoletes-conf-option-label) group installation takes obsoletes into account.

- [x] ```dnf [options] group list <group-spec>...```

List all matching groups, either among installed or available groups. If nothing is specified, list all known groups. ```--installed``` and ```--available``` options narrow down the requested list. Records are ordered by the display_order tag defined in comps.xml file. Provides a list of all hidden groups by using option ```--hidden```. Provides group IDs when the ```-v``` or ```--ids``` options are used.

- [x] ```dnf [options] group remove <group-spec>...```

Mark the group removed and remove those packages in the group from the system which do not belong to another installed group and were not installed explicitly by the user.

- [x] ```dnf [options] group upgrade <group-spec>...```

Upgrades the packages from the group and upgrades the group itself. The latter comprises of installing packages that were added to the group by the distribution and removing packages that got removed from the group as far as they were not installed explicitly by the user.

Groups can also be marked installed or removed without physically manipulating any packages:

- [x] ```dnf [options] group mark install <group-spec>...```

Mark the specified group installed. No packages will be installed by this command, but the group is then considered installed.

- [x] ```dnf [options] group mark remove <group-spec>...```

Mark the specified group removed. No packages will be removed by this command.

glum23:
[...]

PR #478, opened 2023-04-20
dnf5: Replacement of old "dnf group mark install/remove" commands

m-blaha:
This is replacement of "dnf group mark install/remove" commands. I certainly can implement those as standalone commands, but it would introduce a lot of code duplicity. `dnf group mark install` is basically the same command (including all command line switches) as `dnf group install`. The only difference is that the former needs to set one GoalJobSettings flag more (`settings.group_no_packages = true`). This will ensure, that only groups are going to be part of the transaction and no packages are changed (and similarly the `dnf group mark remove`).
If needed, I think a compatible alias can be introduced later - `dnf group mark install` -> `dnf group install --no-packages`.

For https://github.com/rpm-software-management/dnf5/issues/138

### DIFF
--- a/include/libdnf/base/goal_elements.hpp
+++ b/include/libdnf/base/goal_elements.hpp
@@ -159,6 +159,10 @@ public:
         return group_package_types ? &group_package_types.value() : nullptr;
     }
 
+    /// If set to true, group operations (install / remove) will only work
+    /// with the group itself, but will not add to the transaction any packages.
+    bool group_no_packages{false};
+
     /// Set whether hints should be reported
     bool report_hint{true};
     /// Set skip_broken, AUTO means that it is handled according to the default behavior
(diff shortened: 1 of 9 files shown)


j-mracek:
I think it is impossible to add compatibility to `dnf group mark install` because the new system state requires binding to particular group.

include/libdnf/base/goal_elements.hpp
```diff
@@ -159,6 +159,11 @@ struct GoalJobSettings : public ResolveSpecSettings {
         return group_package_types ? &group_package_types.value() : nullptr;
     }
 
+    /// If set to true, group operations (install / remove) will only work
+    /// with the group itself, but will not add to the transaction any packages.
+    /// Used by `group mark` command.
```
kontura, line 164: This comment is a bit outdated.

PR #488, opened 2023-04-24
Filter latest per argument for download command

j-mracek:
It fixes the issue when `dnf download pkg-1 pkg-2` where the previous implementation will download only pkg-2.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -109,10 +109,11 @@ void DownloadCommand::run() {
         libdnf::rpm::PackageQuery package_query(full_package_query);
         auto option = dynamic_cast<libdnf::OptionString *>(pattern.get());
         package_query.resolve_pkg_spec(option->get_value(), {}, true);
+        package_query.filter_priority();
+        package_query.filter_latest_evr();
         result_query |= package_query;
     }
-    result_query.filter_priority();
-    result_query.filter_latest_evr();
+
     if (resolve_option->get_value()) {
         auto goal = std::make_unique<libdnf::Goal>(ctx.base);
         libdnf::GoalJobSettings settings;
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -57,6 +57,8 @@ void DownloadCommand::set_argument_parser() {
 
     alldeps_option = dynamic_cast<libdnf::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf::OptionBool>(new libdnf::OptionBool(false))));
+    destdir_option =
+        dynamic_cast<libdnf::OptionString *>(parser.add_init_value(std::make_unique<libdnf::OptionString>(".")));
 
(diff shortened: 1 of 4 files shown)


m-blaha:
Are you also gonna add a test scenario for it?

j-mracek:
Yes, that my plan

j-mracek:
But I realized, that the test will require `--destdir` options

evan-goode (approved):
LGTM, `package_query` could now be reverted to a PackageSet rather than a PackageQuery but it doesn't make a big difference I think.

dnf5/commands/download/download.cpp
```diff
@@ -141,8 +152,9 @@ void DownloadCommand::run() {
         }
     }
 
+    printf("tests %s\n", destdir_option->get_value().c_str());
```
evan-goode, line 155: printf left in?

dnf5/commands/download/download.cpp
```diff
@@ -71,8 +73,17 @@ void DownloadCommand::set_argument_parser() {
     alldeps->set_const_value("true");
     alldeps->link_value(alldeps_option);
 
-    cmd.register_named_arg(resolve);
+    auto destdir = parser.add_new_named_arg("destdir");
+    destdir->set_long_name("destdir");
+    destdir->set_description(
+        "Set directory to copy packages to. Default location is the current directory");
```
evan-goode, line 79: Maybe "set directory to download packages to" rather than "copy packages to"?

dnf5/commands/download/download.cpp
j-mracek: Sorry, I forge to remove it

j-mracek:
A've added a description of the new option and replaced word `current` by `working`. Additionaly I realized that `--destdir` has an alias `--downloaddir`.

doc/commands/download.8.rst
```diff
@@ -44,6 +44,8 @@ Options
 ``--alldeps``
     | To be used together with ``--resolve``, it downloads all dependencies, not skipping the already installed ones.
 
+``--destdir=<path>``
+    Set directory to download packages to. Default location is the working directory
```
jan-kolarik, line 48: Update of the description is still needed, see above: "used for downloading ... to the current working directory.". Also minor: add dot there at the end of the sentence :slightly_smiling_face:

jan-kolarik: Another minor: maybe add a new example to the section below?

j-mracek: Done

jan-kolarik:
Thanks for fixing this use case!

jan-kolarik:
Just for info, I've noticed some other use cases not yet covered in the implementation, so I will create another PR related to the `download` command.

j-mracek:
> Just for info, I've noticed some other use cases not yet covered in the implementation, so I will create another PR related to the `download` command.

Just a note, I've created https://github.com/rpm-software-management/dnf5/issues/494 to track all missing parts in download command. Please don't hesitate to update the issue.

PR #495, opened 2023-04-25
packit: update global version variables

inknos:
With this pr packit should set the version correctly in the specfile.


.packit.yaml
```diff
@@ -1,18 +1,37 @@
...
+
+# add or remove files that should be synced
 files_to_sync:
   - dnf5.spec
+  - .packit.yaml
+
+actions:
+  fix-spec-file:
+    # set the correct version for global project_version_{major,minor,patch} macros needed by %build
+    bash -c "sed -i -r \"s/^%global project_version_major (\s*)\S+/%global project_version_major \1`echo ${PACKIT_PROJECT_VERSION} | cut -d. -f3`/\" dnf5.spec"
```
m-blaha, line 14: I think the requested field index is incorrect here, should be 1:  "... cut -d. **-f1** ..."

.packit.yaml
```diff
@@ -1,18 +1,37 @@
...
+    bash -c "sed -i -r \"s/^%global project_version_minor (\s*)\S+/%global project_version_minor \1`echo ${PACKIT_PROJECT_VERSION} | cut -d. -f3`/\" dnf5.spec"
```
m-blaha, line 15: "... cut -d. **-f2** ..."

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,30 @@
...
+          fetch-depth: 0
+      - name: Prepare release content
+        uses: packit/prepare-release@v1
+        with:
+          version: ${{ inputs.version }}
+          specfiles: dnf5.spec
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          labels: release
```
m-blaha, line 26: I'm not sure how it works, but shouldn't we create the label `release` first?

m-blaha (changes requested):

.packit.yaml
```diff
@@ -1,18 +1,37 @@
...
+downstream_package_name: dnf5
 
 jobs:
   - job: propose_downstream
     trigger: release
     dist_git_branches:
-        - fedora-rawhide
-        - fedora-38
+      - fedora-rawhide
+      - fedora-38
```
m-blaha, line 28: I might be missing something, but branches in fedora distgit are named without the `fedora-` prefix - `rawhide`, `f38`...

inknos: yes, after dnf5 is in fedora 37. for now, we only want f38 and 39 I believe

inknos: sorry I meant after dnf5 is in fedora 38, 39, 40. if we do fedora-all it will target fedora 37 I guess

.github/workflows/prepare-release.yml
inknos: created https://github.com/rpm-software-management/dnf5/labels/release

.packit.yaml
inknos: They are aliases. Maybe packit is smart enough to make a release on all fedoras in which the package is released. I would change it later though, just to be safe.

m-blaha (approved):
Thank you Nicola for your work on this!

Issue #494, opened 2023-04-25
Download command


j-mracek:
- [x] ``dnf download  <pkg-spec>...`` (Core command)

-------
Options
-------

All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.

- [x] ``--resolve``

    Resolves dependencies of specified packages and downloads missing dependencies in the system.

- [x] ``--alldeps``

    When used with ``--resolve``, download all dependencies (do not skip already installed ones).

Remaining options are tracked in separate issues:

- [ ] #945

    Show this help.

  -   Proposing to drop (requires documentation)

- [ ] #946

    Limit the query to packages of given architectures (default is all compatible architectures with
    your system). To download packages with arch incompatible with your system use
    ``--forcearch=<arch>`` option to change basearch.

- [ ] #947

    Download the source rpm. Enables source repositories of all enabled binary repositories.

- [ ] #948

    Download the debuginfo rpm. Enables debuginfo repositories of all enabled binary repositories.

- [ ] #949

    Download directory, default should be loaded from the `destdir` configuration option.

  -   Shall we keep `--downloaddir`, or `--destdir` or both?

- [ ] #497

    Instead of downloading, print list of urls where the rpms can be downloaded.

- [ ] #950

    Limit the protocol of the urls output by the --url option. Options are http, https, rsync, ftp.

j-mracek:
@glum23 Could it be related to the fact that dnf5 does not copy rpms from local repository?

glum23:
@j-mracek I'd say dnf4 does not copy rpms from a local repository too, however, only when it is a file. Here it is a web server on localhost, however, rpm located in other user's username/password protected directory. I'd say it should not be considered as "local so ignored". Anyway, dnf4 in the very same situation downloads the rpm file.

jan-kolarik:
I've noticed following two differences in current `dnf5 download` command behavior compared to the `dnf` when working on the `download` command tests refactoring in `ci-dnf-stack`: https://github.com/rpm-software-management/ci-dnf-stack/pull/1262.

- Command doesn't return an error when non-existing package is passed
- When using the `--resolve` parameter, the matching package is not downloaded if the latest version is already installed on the system

Tests for this functionality already exist in the `ci-dnf-stack` and are disabled for now.

Bavuett:
In the meantime, may I take the implementation of the `--url` option? It looks like a good issue for newbies like me.

j-mracek:
[...]

PR #496, opened 2023-04-25
download: Resolve goal separately for each package

jan-kolarik:
Fix use case when the same package is available in multiple repositories.

We resolve the dependencies separately for each matching package and construct a set of packages instead of a vector to avoid having duplicates in the result.

This is a follow-up to the https://github.com/rpm-software-management/dnf5/pull/488.
Parent issue: https://github.com/rpm-software-management/dnf5/issues/494.
CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1262.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -29,7 +29,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf/rpm/package_set.hpp>
 
 #include <iostream>
-#include <set>
+#include <map>
 
 
 namespace dnf5 {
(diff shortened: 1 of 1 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -112,47 +116,52 @@ void DownloadCommand::configure() {
...
     }
 
     if (!download_pkgs.empty()) {
-        download_packages(download_pkgs, destdir_option->get_value().c_str());
+        libdnf::repo::PackageDownloader downloader(ctx.base);
+
+        // for download command, we don't want to mark the packages for removal
+        downloader.force_caching(true);
+
+        for (auto & pkg : download_pkgs) {
```
j-mracek, line 158: It will sound strange, but we cannot download all packages. We have to download only one package per NEVRA or per file name. The reason is that we download files into one particular directories. It means that RPMs will override each other and because we download them in parallel, the download will fail due to incorrect checksum.

DNF4 implementation:
```
        pkg_dict = {}
        for pkg in pkgs:
            pkg_dict.setdefault(str(pkg), []).append(pkg)

        to_download = []
        cmdline = []
        for pkg_list in pkg_dict.values():
            pkgs_cmdline = [pkg for pkg in pkg_list
                            if pkg.repoid == hawkey.CMDLINE_REPO_NAME]
            if pkgs_cmdline:
                cmdline.append(pkgs_cmdline[0])
                continue
            pkg_list.sort(key=lambda x: (x.repo.priority, x.repo.cost))
            to_download.append(pkg_list[0])
```

dnf5/commands/download/download.cpp
```diff
@@ -112,47 +116,52 @@ void DownloadCommand::configure() {
...
+                    }
+                }
+                for (auto & tspkg : transaction.get_transaction_packages()) {
+                    if (transaction_item_action_is_inbound(tspkg.get_action()) &&
+                        tspkg.get_package().get_repo()->get_type() != libdnf::repo::Repo::Type::COMMANDLINE) {
+                        download_pkgs.insert(tspkg.get_package());
+                    }
+                }
+            } else {
+                download_pkgs.insert(std::move(pkg));
```
j-mracek, line 147: What about to add `pkg` to download set always? It will ensure that package will be downloaded even if it is installed on the system.

jan-kolarik: Yes, this is exactly what I was thinking of when discovering the test case related to `--resolve` functionality. I wasn't really sure if it's really intended to download the target package always, even if it is resolved as already being on the system, but now I am looking to the original `--resolve` option documentation in the dnf and it is talking only about dependencies.

m-blaha: There was different switch for downloading all the dependencies regardless they are installed or not in dnf4 - `--alldeps`.

j-mracek: And `--alldeps` is correctly implemented (load system repository or not)

j-mracek (approved):
LGTM

Issue #397, opened 2023-03-27
DNF repoclosure Plugin


j-mracek:
### Description

repoclosure is a program that reads package metadata from one or more repositories, checks all dependencies, and displays a list of packages with unresolved dependencies.

### Options

- [ ] --arch <arch>

    Query only packages for specified architecture, can be specified multiple times (default is all compatible architectures with your system). To run repoclosure for arch incompatible with your system use --forcearch=<arch> option to change basearch.

- [ ] --best

    Check only the newest packages per arch.

- [ ] --check <repoid>

    Specify repo ids to check, can be specified multiple times (default is all enabled).

- [ ] --newest

    Check only the newest packages in the repos.

- [ ] --pkg <pkg-spec>

    Check closure for this package only.

- [ ] --repo <repoid>

    Specify repo ids to query, can be specified multiple times (default is all enabled).

PR #500, opened 2023-04-28
Repoclosure plugin

m-blaha:
Resolves: https://github.com/rpm-software-management/dnf5/issues/397

It's an equivalent of the dnf4 implementation.

### DIFF
--- a/dnf5-plugins/repoclosure_plugin/repoclosure.cpp
+++ b/dnf5-plugins/repoclosure_plugin/repoclosure.cpp
@@ -1,0 +1,216 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "repoclosure.hpp"
+
+#include "utils/bgettext/bgettext-mark-domain.h"
+
+#include "libdnf-cli/argument_parser.hpp"
+
+#include "libdnf/conf/const.hpp"
+#include "libdnf/rpm/package.hpp"
+#include "libdnf/rpm/package_query.hpp"
+#include "libdnf/utils/format.hpp"
+
+#include <iostream>
+
+
+namespace dnf5 {
+
+void RepoclosureCommand::set_parent_command() {
+    auto * arg_parser_parent_cmd = get_session().get_argument_parser().get_root_command();
+    auto * arg_parser_this_cmd = get_argument_parser_command();
+    arg_parser_parent_cmd->register_command(arg_parser_this_cmd);
+}
+
+void RepoclosureCommand::set_argument_parser() {
+    auto & ctx = get_context();
+    auto & parser = ctx.get_argument_parser();
+
+    auto & cmd = *get_argument_parser_command();
+    cmd.set_description(_("Print list of unresolved dependencies for repositories"));
+
+    auto specs =
+        parser.add_new_positional_arg("specs", libdnf::cli::ArgumentParser::PositionalArg::UNLIMITED, nullptr, nullptr);
+    specs->set_description("List of package specs to check closure for");
+    specs->set_parse_hook_func(
+        [this]([[maybe_unused]] libdnf::cli::ArgumentParser::PositionalArg * arg, int argc, const char * const argv[]) {
+            for (int i = 0; i < argc; ++i) {
+                pkg_specs.emplace_back(argv[i]);
+            }
+            return true;
+        });
+    cmd.register_positional_arg(specs);
+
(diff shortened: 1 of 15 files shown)


doc/dnf5_plugins/repoclosure.8.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    dnf5 repoclosure
+
+Display list of unresolved dependencies for rawhide repository and packages with architecture noarch and x86_64::
+
+    dnf5 repoclosure --repo rawhide --arch noarch --arch x86_64
+
+Display list of unresolved dependencies for zmap package from rawhide repository::
+
+    dnf5 repoclosure --repo rawhide --pkg zmap
```
jan-kolarik, line 74: The `--pkg` option is not mentioned in the `Options` section.

doc/dnf5_plugins/repoclosure.8.rst
```diff
@@ -0,0 +1,78 @@
+..
```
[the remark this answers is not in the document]
m-blaha: Thanks. I've added the changes as a new commit (for easier re-review). If you want me to, I can squash the commit.

dnf5-plugins/repoclosure_plugin/repoclosure.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+
+    // cache query results: reldepid -> whether the reldep is satisfied
+    std::unordered_map<int, bool> resolved;
+    std::vector<std::pair<libdnf::rpm::Package, std::vector<std::string>>> unresolved_packages;
+    for (const auto & pkg : to_check_query) {
+        std::vector<std::string> unsatisfied;
+        for (const auto & reldep : pkg.get_requires()) {
+            int reldep_id = reldep.get_id().id;
```
jan-kolarik, line 167: Do we want to handle there also file requires? I guess the situation for them is more complicated. And the current behavior includes them compared to the `dnf`. When file patterns are dropped like this, the results seem to be the same:
```
if (libdnf::utils::is_file_pattern(reldep.to_string())) {
    continue;
}
```

m-blaha: This PR is 1:1 copy of dnf4 functionality. There are (at least) two fields where it can be enhanced:

- file conflicts (see [Bug 1702652](https://bugzilla.redhat.com/show_bug.cgi?id=1702652) - RFE: extend "dnf repoclosure" with check for file conflicts)
- modularity (see https://github.com/rpm-software-management/dnf5/issues/511)

m-blaha: Oh, sorry - I've mixed file conflicts with file provides...

m-blaha: I think we should check the file requires (and I was convinced that dnf4 does the same - I need to re-check).

jan-kolarik: I've just tested it running the `dnf repoclosure` and `dnf5 repoclosure`. When omitting the file requires, the output is same on my machine.

m-blaha: Hm, this is interesting, because that's why I added explicit `add_item(libdnf::METADATA_TYPE_FILELISTS)` and was testing that the results are the same for dnf4 and dnf5 (except for sorting of unsatisfied package dependencies which dnf4 does not do).
May be there is an issue in libdnf5 regarding loading different sets of repository metadata. Can you remove the repository cache and try again?

jan-kolarik: Yep, you're right. When I remove the cache completely, the repoclosure is working fine with the fresh metadata. Could it be something like that some repositories were marked as expired and refreshed and some were not and therefore some updated requirements are not being provided?

m-blaha: Definitely worth a github issue... I'll try to find a reliable reproducer (as the first step to the solution :)
It looks like in some situations the filelists metadata is not loaded (even though requested).

doc/dnf5_plugins/repoclosure.8.rst
```diff
@@ -0,0 +1,78 @@
...
+``--best``
+    Check only the newest packages per arch.
+
+``--check <repoid>``
+    Specify repo ids to check, can be specified multiple times (default is all enabled).
+
+``--newest``
+    Check only the newest packages in the repos.
+
+``--repo <repoid>``
```
kontura, line 53: I am thinking if this should be mentioned here. 
It is a general dnf option, for example I think `--enable-repo` would also work but it is not mentioned.

Though I see how it is relevant. :thinking:

jan-kolarik: Yes, I guess just command-specific options should be mentioned there. But it's definitely good to have the common use cases in the examples below as it is now.

m-blaha: Well to be honest, I'd like to have different set of options for repoclosure, something like:
--repo <repo to check>
--lookaside <also this repo that can be used to fulfill requirements>
But this design collides with existing `--repo` option which is consistently used in all commands and defines which repositories are available to dnf... And yes, --enable-repo will also work.

So that's why I've decided to include this global option also to repoclosure documentation. But I can try to replace the option in docs by better description of the command behavior.

jan-kolarik: I like the idea that the specific behavior of the option used for this command is described in it's man page. I consider this locality better than describing different behaviors in the general man page, but we could still mention in the general man page there is a difference for `repoclosure`, see its man page (or something like this).

jan-kolarik (approved):
Thanks!

Issue #443, opened 2023-04-11
Implement exclude_from_weak and exclude_from_weak_autodetect configuration


j-mracek:
Both options were introduced after DNF5 fork, therefore there are not included in or code base.

- [x] Implement configuration option
- [x] Add support to the goal for `exclude_from_weak`
- [x] Implement logic for auto-detection

Note - DNF5 does not need any new or modify API, because configuration options are enough for modification of DNF5/libdnf5 behavior

Links:
https://github.com/rpm-software-management/libdnf/pull/1279
https://github.com/rpm-software-management/dnf/pull/1780
https://github.com/rpm-software-management/ci-dnf-stack/pull/1005

PR #504, opened 2023-05-03
Add support for exclude from weaks

j-mracek:
CI tests - https://github.com/rpm-software-management/ci-dnf-stack/pull/1268

Resolves - https://github.com/rpm-software-management/dnf5/issues/443


libdnf/base/goal.cpp
```diff
@@ -1758,6 +1852,14 @@ base::Transaction Goal::resolve() {
         p_impl->rpm_goal.set_installonly_limit(cfg_main.get_installonly_limit_option().get_value());
     }
 
+    // Set exclude weak dependencies from confiduration
```
jan-kolarik, line 1855: Minor: `confiduration` -> `configuration`

libdnf/base/goal.cpp
```diff
@@ -1690,6 +1693,97 @@ void Goal::Impl::add_paths_to_goal() {
...
+        return;
+    }
+    rpm::PackageQuery base_query(base, rpm::PackageQuery::ExcludeFlags::APPLY_EXCLUDES);
+    rpm::ReldepList reldep_list(base);
+
+    std::vector<std::string> installed_names;
+    installed_names.reserve(installed_query.size());
+
+    // Investigate uninstalled recommends of installed packages
+    for (auto pkg : installed_query) {
```
jan-kolarik, line 1720: Better to use `const auto & pkg`?

libdnf/base/goal.cpp
```diff
@@ -1690,6 +1693,97 @@ void Goal::Impl::add_paths_to_goal() {
...
+        }
+    }
+
+    // Investigate supplements of only available packages with a different name to installed packages
+    // We can use base_query, because it is not useful anymore
+    base_query.filter_name(installed_names, sack::QueryCmp::NEQ);
+    // We have to remove all installed packages from testing set
+    base_query -= installed_query;
+    rpm::PackageSet exclude_supplements(base);
+    for (auto pkg : base_query) {
```
jan-kolarik, line 1760: Better to use `const auto & pkg`?

libdnf/base/goal.cpp
```diff
@@ -1690,6 +1693,97 @@ void Goal::Impl::add_paths_to_goal() {
...
+    rpm::PackageQuery base_query(base, rpm::PackageQuery::ExcludeFlags::APPLY_EXCLUDES);
+    rpm::ReldepList reldep_list(base);
+
+    std::vector<std::string> installed_names;
+    installed_names.reserve(installed_query.size());
+
+    // Investigate uninstalled recommends of installed packages
+    for (auto pkg : installed_query) {
+        installed_names.push_back(pkg.get_name());
+        for (auto recommend : pkg.get_recommends()) {
```
jan-kolarik, line 1722: There's probably no reason not having the `ReldepList::add(const Reldep &);`. After making it const we could switch also here and below in supplements to iterate using `const auto &`, but it's not anything major I guess...

libdnf/base/goal.cpp
```diff
@@ -1690,6 +1693,97 @@ void Goal::Impl::add_paths_to_goal() {
...
+            if (libdnf::rpm::Reldep::is_rich_dependency(recommend.to_string())) {
+                // Rich dependencies are skipped because they are too complicated to provide correct result
+                continue;
+            };
+            rpm::PackageQuery query(base_query);
+
+            //  There can be installed provider in different version or upgraded packed can recommend a different
```
jan-kolarik, line 1729: `packed` -> `package`?

libdnf/base/goal.cpp
j-mracek: I was not sure who owns the package object. But I guess that reference should work, because the iterator should be owned by query.

PR #508, opened 2023-05-03
repoquery: queryformat improvements, --depends, --querytags,..

kontura:
Contains:
- Several repoquery improvements to `queryformat`.
- New arguments: `--depends` and `--querytags`.
- Specifying path to `--what*` arguments now pulls in filelists
- Allow using the same `--what*` option multiple times.

The PR got quite big but each commit should be fairly small and simple.

### DIFF
--- a/libdnf-cli/output/repoqueryformat.cpp
+++ b/libdnf-cli/output/repoqueryformat.cpp
@@ -74,7 +74,15 @@ static const std::unordered_map<std::string, Getter> NAME_TO_GETTER = {
     {"repoid", &libdnf::rpm::Package::get_repo_id},
 };
 
-//TODO(amatej): Use keys of NAME_TO_GETTER for --querytags options
+void print_available_pkg_attrs(std::FILE * target) {
+    std::set<std::string> output;
+    for (const auto & pair : NAME_TO_GETTER) {
+        output.insert(pair.first);
+    }
+    for (const auto & line : output) {
+        fmt::print(target, "{}\n", line);
+    }
+}
 
 // Argument format contains partially copied and converted queryformat, for example: "name: %-30{{name".
 // We know the tag "name" is valid so this function check align spec "-30" and if it is valid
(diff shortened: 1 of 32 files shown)


libdnf-cli/output/repoqueryformat.cpp
```diff
@@ -41,11 +42,9 @@ static const std::unordered_map<std::string, Getter> NAME_TO_GETTER = {
     {"release", &libdnf::rpm::Package::get_release},
     {"arch", &libdnf::rpm::Package::get_arch},
     {"evr", &libdnf::rpm::Package::get_evr},
-    {"nevra", &libdnf::rpm::Package::get_nevra},
     {"full_nevra", &libdnf::rpm::Package::get_full_nevra},
-    {"na", &libdnf::rpm::Package::get_na},
     {"group", &libdnf::rpm::Package::get_group},
-    {"packagesize", &libdnf::rpm::Package::get_package_size},
+    {"downloadsize", &libdnf::rpm::Package::get_package_size},
```
m-blaha, line 49: Terminology we use for package sizes is a bit confusing here.
We have `packagesize` used in `Package::get_package_size()`, then there is `downloadsize` used here for query format. And then there is `rpm` and `libsolv` with their own terminology...
Shouldn't we agree on one term and use it consistently everywhere?

kontura:
Rebased.

libdnf-cli/output/repoqueryformat.cpp
kontura: That sounds good.
Possible options could be:
`packagesize` - `libdnf::rpm::Package::get_package_size()`
`downloadsize` - `libdnf::rpm::Package::get_download_size()`

It seems libsolv is using `SOLVABLE_DOWNLOADSIZE` so I would be in favor of renaming the method to `get_download_size()`. Also it seems dnf4 used `Package.downloadsize`. 

Do you have any other ideas? Or which would you prefer?

m-blaha: I'd prefer using `downloadsize`. Packagesize sounds a bit vague to me - it's not completely clear what size we are referring to.

kontura: Ok, I will create a separate PR with that change if that is ok.
It needs to be changed on a bunch of places.

Issue #388, opened 2023-03-22
Add support allow_downgrade configuration from config and commandline


j-mracek:
This is about modifying solver SOLVER_FLAG_ALLOW_DOWNGRADE option. Currently it is implemented in GoalPrivate, but there is no way how to modify it using our API or configuration. The default value is True. This default provide a better UX, but some user prefer to use False.

I am proposing 

- [ ] to add `allow_downgrade` configuration option with default value true
- [ ] add tho command line options `--allow-downgrades`, and  and `--no-downgrades`
NOTE - names of options are not fixed

PR #514, opened 2023-05-05
Implement Allow downgrade

j-mracek:
Resolves: https://github.com/rpm-software-management/dnf5/issues/388

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1274

### DIFF
--- a/dnf5/commands/downgrade/downgrade.cpp
+++ b/dnf5/commands/downgrade/downgrade.cpp
@@ -54,6 +54,7 @@ void DowngradeCommand::set_argument_parser() {
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     auto skip_broken = std::make_unique<SkipBrokenOption>(*this);
+    create_allow_downgrade_options(*this);
 }
 
 void DowngradeCommand::configure() {
--- a/dnf5/include/dnf5/shared_options.hpp
+++ b/dnf5/include/dnf5/shared_options.hpp
@@ -58,6 +58,10 @@ public:
               &command.get_context().base.get_config().get_skip_unavailable_option()) {}
 };
 
+/// Create two options (`--allow-downgrade` and `--no-allow-downgrade`) for a command provided as an argument command.
+/// The values are atored in the `allow_downgrade` configuration option
+void create_allow_downgrade_options(dnf5::Command & command);
+
 
 }  // namespace dnf5
 
--- a/dnf5/shared_options.cpp
+++ b/dnf5/shared_options.cpp
@@ -1,0 +1,52 @@
+/*
+Copyright Contributors to the dnf5 project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Dnf5 is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#include "dnf5/shared_options.hpp"
+
+
+namespace dnf5 {
+
+
+void create_allow_downgrade_options(dnf5::Command & command) {
+    auto * solver_options = command.get_context().get_argument_parser().add_new_group("solver_options");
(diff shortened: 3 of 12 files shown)


dnf5/include/dnf5/shared_options.hpp
```diff
@@ -58,6 +58,29 @@ class SkipUnavailableOption : public libdnf::cli::session::BoolOption {
...
+public:
+    explicit NoDowngradeOption(dnf5::Command & command)
+        : BoolOption(
+              command,
+              "no-downgrade",
+              '\0',
+              _("Disable downgrade of dependencies"),
+              true,
+              &command.get_context().base.get_config().get_allow_downgrade_option()) {}
+};
```
m-blaha, line 83: FWIW, I'd prefer having a pair of options, one of them prefixed by `no-`. E.g. `allow-downgrade / no-allow-downgrade`

m-blaha: And they should be mutually exclusive.

j-mracek: `no-alloew-downgrade` is used by zipper. I also would like to have them exclusive, but this is still under development

dnf5/shared_options.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+#include "dnf5/shared_options.hpp"
+
+
+namespace dnf5 {
+
+
+void create_allow_downgrade_options(dnf5::Command & command) {
+    auto * repoquery_formatting = command.get_context().get_argument_parser().add_new_group("solver_options");
```
m-blaha, line 28: It does not affect functionality at all - but still, can you rename this variable (e.g. `solver_options`)?

dnf5/commands/downgrade/downgrade.cpp
```diff
@@ -60,6 +60,9 @@ void DowngradeCommand::configure() {
     auto & context = get_context();
     context.set_load_system_repo(true);
     context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
+    auto & config = context.base.get_config();
+    /// Allow dowgrade should be the default behavior of the command
+    config.get_allow_downgrade_option().set(libdnf::Option::Priority::COMMANDLINE, true);
```
m-blaha, line 65: Just curious why did you use `COMMANDLINE` priority? Because what I'd do is plain `config.get_allow_downgrade_option().set(true);` (which defaults to RUNTIME).

dnf5/shared_options.cpp
j-mracek: Sure

dnf5/commands/downgrade/downgrade.cpp
j-mracek: I would like to allow `--setopt` or option (e.g. `--allow-downgrade`) overrides. If both will use `runtime` priority then order will be important. I guess that configure section would override the cmd option and that I don't want.

m-blaha: I see.

But I'm quite sure that `--setopt=allow_downgrade=0` will not work with your code, because you both use the same priority (`COMMANDLINE`), but your code is executed later: https://github.com/rpm-software-management/dnf5/blob/4447b5f467baa667399eb837e50bef0ddd3e1cb9/dnf5/main.cpp#L742

That means it will always override values set by commandline args parser which runs here: https://github.com/rpm-software-management/dnf5/blob/4447b5f467baa667399eb837e50bef0ddd3e1cb9/dnf5/main.cpp#L681

Is setting the `allow_downgrade` value for downgrade command even needed? The hard coded default is `true` anyway, and if you remove this line completely, `--setopt` should start working. Of course, the downgrade command will also follow values set in `dnf.conf` file, which is IMHO expected.

j-mracek: This is tricky. What about to use PLUGINDEFAULT?  The only problem what I see is with DROPINCONFIG.

m-blaha: Hm, this feels like misuse of `PLUGINDEFAULT` priority (at least judging by its name, I'm not completely sure what this priority was originally meant for).

What's the use-case you are trying to resolve? Situation when the user puts `allow_downgrade=false` to dnf.conf? 

Maybe what we need in first place is documentation of config options, where should be described what each option does. In this case, that option affects downgrading of **other then directly requested** installed packages during goal resolving. And if someone sets this option in dnf.conf, they should be aware (after reading the doc) that the `downgrade` command is inevitably affected. And for this case we should add the same `--allow-downgrade/--no-allow-downgrade` set of options.

j-mracek: Ok, improvement of documentation is very good suggestion. But I am not sure whether it make sense to include it in this PR. Anyway I have to look to DNF4 code.

m-blaha: Oh, sorry - I didn't mean that you should create a dnf.conf documentation here :) It's definitely a task for separate PR.

PR #521, opened 2023-05-10
Connect destdir cmd option with configuration

j-mracek:
It unify approach how to use destdir variable inside dnf5. Using only configuration value is more transparent.


jan-kolarik (approved):
Great, thanks!

PR #523, opened 2023-05-11
Fix doxygen warnings

m-blaha:
Mostly escaping special characters, fixing parameter names to match function declarations, adding missing parameters.

There is still a lot of warning on undocumented classes and methods, but this is out of scope of the PR.


kontura:
Great, thank you!

PR #530, opened 2023-05-15
Make private one method and remove unused one

j-mracek:
Plugin loading is handled by Base::setup() therefore it is not required to have the method on API.


kontura:
> It seems there are `--allow-downgrade` tests enabled for `dnfdaemon` which doesn't support it.

I made a PR with a fix: https://github.com/rpm-software-management/ci-dnf-stack/pull/1278

jan-kolarik:
LGTM, thanks!

PR #532, opened 2023-05-15
Make strict behavior for several command

j-mracek:
After the change, upgrade, downgrade, distrosync will not ignore broken argument.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -81,6 +81,16 @@ Autoremove command
  * Dropped `<spec>` positional argument since the usecase is sufficiently covered by the `remove` command.
  * Specific `autoremove-n`, `autoremove-na`, and `autoremove-nevra` variants of the command are not supported.
 
+Distro-sync command
+-------------------
+ * When any argument does not match any package or it is not installed, DNF5 fail. The behavior can be modified by
+   the `--skip-unavailable` option.
+
+Downgrade command
+-----------------
+ * When any argument does not match any package or it is not installed, DNF5 fail. The behavior can be modified by
+   the `--skip-unavailable` option.
+
 Group command
 -------------
  * Dropped `group mark install` and `group mark remove` subcommands in favour of the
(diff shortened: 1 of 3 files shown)


j-mracek:
The PR requires to document change in behavior and agreement with comunity.

glum23:
There's a PR with test scenarios for this change: https://github.com/rpm-software-management/ci-dnf-stack/pull/1284 (currently only for upgrade, it is planned to be extended for the other commands).
The test passed for dnf5 build with this PR.

glum23:
While this is a change in comparison with dnf4 behavior, it would be nice to provide a hint like (try to add '--skip-unavailable' or '--skip-broken' to command line to skip...)

j-mracek:
> While this is a change in comparison with dnf4 behavior, it would be nice to provide a hint like (try to add '--skip-unavailable' or '--skip-broken' to command line to skip...)

I've created a new issue for hints - https://github.com/rpm-software-management/dnf5/issues/548

m-blaha:
Since it's a change of behavior compared with dnf4, is it worth a note in `doc/changes.rst`?

doc/changes.rst
```diff
@@ -81,6 +81,16 @@ Autoremove command
  * Dropped `<spec>` positional argument since the usecase is sufficiently covered by the `remove` command.
  * Specific `autoremove-n`, `autoremove-na`, and `autoremove-nevra` variants of the command are not supported.
 
+Distro-sync command
+-------------------
+ * When any argument does not match any package or package is not installed, DNF5 will fail. The behavior can be
+   modify by the `--skip-unavailable` option.
```
m-blaha, line 87: Only minor fixes/enhancements (an article and modify -> modified):

When any argument does not match any package or **a** package is not installed, DNF5 will fail. The behavior can be **modified using** the `--skip-unavailable` option.

glum23:
@j-mracek, the behavior of dnf5 is a bit inconsistent. When there's no downgrade of a package available, it reports on stderr  `Package <whatever> of lowest version already installed, cannot downgrade it.` (BTW, it should be "the lowest version")
When there's no upgrade of a package available, dnf5 does not report anything on stderr.

PR #533, opened 2023-05-16
group info/list: add `--contains-pkgs` option

kontura:

The option is moved from `repoquery` to group info/list commands.
It fits much better into the group sub-commands.

### DIFF
--- a/dnf5/commands/group/arguments.hpp
+++ b/dnf5/commands/group/arguments.hpp
@@ -50,6 +50,18 @@ public:
 };
 
 
+class GroupContainsPkgsOption : public libdnf::cli::session::StringListOption {
+public:
+    explicit GroupContainsPkgsOption(libdnf::cli::session::Command & command)
+        : StringListOption(
+              command,
+              "contains-pkgs",
+              '\0',
+              _("Show only groups containing packages with specified names. List option, supports globs."),
+              "PACKAGE_NAME,...") {}
+};
+
+
 class GroupSpecArguments : public libdnf::cli::session::StringArgumentList {
 public:
     GroupSpecArguments(libdnf::cli::session::Command & command, int nargs)
--- a/include/libdnf/comps/group/query.hpp
+++ b/include/libdnf/comps/group/query.hpp
@@ -54,6 +54,14 @@ public:
         filter(F::name, pattern, cmp);
     }
 
+    /// Filter groups by packages they contain. Keep only groups that contain packages with given names.
+    ///
+    /// @param patterns         A vector of strings (package names) the filter is matched against.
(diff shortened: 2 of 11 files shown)


dnf5/commands/group/arguments.hpp
```diff
@@ -50,6 +50,18 @@ class GroupInstalledOption : public libdnf::cli::session::BoolOption {
...
 
 
+class GroupContainsPkgsOption : public libdnf::cli::session::StringListOption {
+public:
+    explicit GroupContainsPkgsOption(libdnf::cli::session::Command & command)
+        : StringListOption(
+              command,
+              "contains-pkgs",
+              '\0',
+              _("Show only groups containing packges with specified names. List option, supports globs."),
```
jan-kolarik, line 60: Typo `packges` -> `packages`

include/libdnf/comps/group/query.hpp
```diff
@@ -54,6 +54,15 @@ class GroupQuery : public libdnf::sack::Query<Group> {
         filter(F::name, pattern, cmp);
     }
 
+    /// Filter groups by packages they contain. Keep only groups that contain packages with given names.
+    ///
+    /// @param patterns         A vector of strings (package names) the filter is matched against.
+    /// @param cmp_type         A comparison (match) operator, defaults to `QueryCmp::EQ`.
+    ///                         Supported values: `EQ`, `NEQ`, `GLOB`, `NOT_GLOB`.
```
jan-kolarik, line 61: In the usage from `group_info.cpp` we are using also `IGLOB`.

libdnf/comps/group/query.cpp
```diff
@@ -102,5 +103,24 @@ GroupQuery::GroupQuery(const BaseWeakPtr & base, bool empty) : base(base) {
...
+        // Copy group so we can call `get_packages()`, this is needed because `it` is from a std::set and thus const
+        // but `get_packages()` modifies its group (it stores cache of its packages).
+        Group group = *it;
+        bool keep = false;
+        for (const auto & pkg : group.get_packages()) {
+            if (match_string(pkg.get_name(), cmp, patterns)) {
+                keep = true;
+                break;
+            }
+        }
```
jan-kolarik, line 117: Or a simpler C++20 way could be:
```cpp
bool keep = std::ranges::any_of(group.get_packages(), [&](const auto & pkg) {
    return match_string(pkg.get_name(), cmp, patterns);
});
```

include/libdnf/comps/group/query.hpp
kontura: I have removed all of the supported values because it supports many more operators and I don't think we should list them all for all the methods. They are described in the query_cmp.hpp.

jan-kolarik (approved):

PR #539, opened 2023-05-18
Fix nightly builds on different archs than x86_64

jan-kolarik:
Following up to the fix from https://github.com/rpm-software-management/dnf5/pull/95, I've also set up a `x86_64` arch for Python API tests as we use these packages there.

Also some small refactoring and unification in unit tests included.


kontura:
Thanks! Looks good.
I am curios do you know why it started failing only 23 days ago?

jan-kolarik:
That's probably when the Python bindings transaction tests were introduced, also the gpg signature related functionality on the transaction API...

PR #554, opened 2023-05-23
state: Load environmental groups to system repo

m-blaha:
Uses similar approach we already have for groups.
System repo environments are loaded from xml files saved in the system state. In case the xml file with environment definition is missing (e.g. after upgrade from dnf4), a new one is created from available environmental group of the same id (with fallback to create a simple solvable based only on system state data).

### DIFF
--- a/libdnf/repo/repo_sack.cpp
+++ b/libdnf/repo/repo_sack.cpp
@@ -34,6 +34,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf/base/base.hpp"
 #include "libdnf/common/exception.hpp"
+#include "libdnf/comps/environment/query.hpp"
 #include "libdnf/comps/group/query.hpp"
 #include "libdnf/conf/config_parser.hpp"
 #include "libdnf/conf/option_bool.hpp"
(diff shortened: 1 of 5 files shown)


libdnf/repo/repo_sack.cpp
```diff
@@ -610,43 +612,80 @@ void RepoSack::fix_group_missing_xml() {
...
-                // fall-back to creating solvables only from system state
-                solv_repo->create_group_solvable(group_id, system_state.get_group_state(group_id));
+        }
+        if (!environments_missing_xml.empty()) {
+            libdnf::comps::EnvironmentQuery available_environments(base);
+            available_environments.filter_installed(false);
+            for (const auto & environment_id : environments_missing_xml) {
+                bool xml_saved = false;
+                if (directory_exists) {
+                    // try to find the repoid in availables
```
kontura, line 657: Both comments for groups and environments probably shouldn't say `repoid` here?

kontura (approved):
I have only a minor question about a comment, otherwise it looks good.

libdnf/repo/repo_sack.cpp
m-blaha: Yeah, sorry. Fixed.

PR #558, opened 2023-05-24
rpm::Package: rename get_package_size() to get_download_size()

kontura:
Since we have `rpm::Package::get_install_size()` it seems clearer to have `rpm::Package::get_download_size()` to complement it.

Also
- libsolv is using SOLVABLE_DOWNLOADSIZE
- dnf4 used Package.downloadsize
- queryformat uses `downloadsize` tag

Follow up for: https://github.com/rpm-software-management/dnf5/pull/508#discussion_r1203741313

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -582,7 +582,7 @@ static void print_transaction_size_stats(Context & context) {
     for (const auto & trans_pkg : context.get_transaction()->get_transaction_packages()) {
         const auto pkg = trans_pkg.get_package();
         if (transaction_item_action_is_inbound(trans_pkg.get_action())) {
-            const auto pkg_size = pkg.get_package_size();
+            const auto pkg_size = pkg.get_download_size();
             in_pkgs_size += pkg_size;
             if (!pkg.is_available_locally()) {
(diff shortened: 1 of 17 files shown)


jan-kolarik:
Yep, I definitely agree.

kontura:
Test update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1292

Conan-Kudo:
I'm good with this change.

jrohel:
I think either name is fine. It is a package size. And yes, if we want the package to be downloaded, it is equal to the download size (the size of the pure downloaded data, we ignore other layers - the overhead of the download protocol, any additional compression and encryption during the download).

pmatilai:
If it's a package size then better call it just that without assuming API consumer intentions. "Download size" seems to imply there's may be download overhead compared to actual size and whatnot.

jrohel:
To be honest. I think the original name package_size is more correct than download_size. But that's my feeling. And we also deal with consistency. So take that as a weak opinion. That's why I wrote in the previous post that both are fine.

j-mracek (approved):
I like the change. It even unifies the naming with libsolv.

Issue #537, opened 2023-05-16
Please implement `update --downloadonly` 


mark-hyde:
Hi,
I'm testing how dnf 5 works with auter and have found that there is currently no `--downloadonly` option in the dnf5 `update` subcommand.

https://github.com/rackerlabs/auter/issues/245

A typical auter use-case is to have a cronjob run at the start of a month, (which runs `dnf/yum update --downloadonly`) which resolves all dependencies and downloads the packages. 
These can then be applied at a later date (typically staging devices might have packages installed a week or 2 before production), but because the packages were downloaded at the same time, you get pretty much the same packages in both staging and production servers even though they might have been patched some weeks apart.

Could you please implement `dnf5 update --downloadonly` 

As a workaround I'm considering running 'dnf5 check-update', using awk to grab the 1st column, and feed that into dnf download with the --resolve flag, but I'm not entirely convinced yet that this will reliably give the same result.

mark-hyde:
Hi,
I've 
Could you also document this option in `dnf5 --help`? Auter checks this to see if the option is available.
While I could write around that, it would be better if it were documented in the help anyway.

PR #559, opened 2023-05-24
Add support for downloadonly in DNF5

j-mracek:
Resolve: https://github.com/rpm-software-management/dnf5/issues/537

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -769,12 +769,16 @@ int main(int argc, char * argv[]) try {
 
             dnf5::print_transaction_size_stats(context);
 
-            for (const auto & tsflag : base.get_config().get_tsflags_option().get_value()) {
-                if (tsflag == "test") {
-                    std::cout
-                        << "Test mode enabled: Only package downloads, pgp key installations and transaction checks "
-                           "will be performed."
-                        << std::endl;
+            if (base.get_config().get_downloadonly_option().get_value()) {
+                std::cout << "The operation will only download packages for the transaction." << std::endl;
+            } else {
+                for (const auto & tsflag : base.get_config().get_tsflags_option().get_value()) {
+                    if (tsflag == "test") {
+                        std::cout << "Test mode enabled: Only package downloads, pgp key installations and transaction "
+                                     "checks "
+                                     "will be performed."
+                                  << std::endl;
+                    }
                 }
             }
 
--- a/dnf5/commands/downgrade/downgrade.cpp
+++ b/dnf5/commands/downgrade/downgrade.cpp
@@ -52,6 +52,7 @@ void DowngradeCommand::set_argument_parser() {
     cmd.register_positional_arg(keys);
 
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
+    create_downloadonly_option(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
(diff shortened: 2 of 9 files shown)


jan-kolarik:
The PR itself looks great, but I wanted to test the parameter in CLI and it is unknown. I think we are missing the `owner.opt_binds().add("downloadonly", downloadonly);` in the main config...

dnf5/main.cpp
```diff
@@ -769,12 +769,16 @@ int main(int argc, char * argv[]) try {
...
-                           "will be performed."
-                        << std::endl;
+            if (base.get_config().get_downloadonly_option().get_value()) {
+                std::cout << "The operation will only download packages for the transaction." << std::endl;
+            } else {
+                for (const auto & tsflag : base.get_config().get_tsflags_option().get_value()) {
+                    if (tsflag == "test") {
+                        std::cout << "Test mode enabled: Only package downloads, pgp key installations and transaction "
+                                     "checks "
+                                     "will be performed."
```
jrohel, line 779: The message is badly cut.

jrohel:
@jan-kolarik
> I think we are missing the owner.opt_binds().add("downloadonly", downloadonly); in the main config. Can you please fix that?

I don't think so. As you can see in the source code, `downloadonly` is a "runtime only option".
`OptionBool downloadonly{false};  // runtime only option`

And from my point of view, `downloadonly` is a specific option of dnf5 application. It is not a libdnf(5) option. Do we want to add this option to the "dnf.conf " file?

jan-kolarik:
> @jan-kolarik
> 
> > I think we are missing the owner.opt_binds().add("downloadonly", downloadonly); in the main config. Can you please fix that?
> 
> I don't think so. As you can see in the source code, `downloadonly` is a "runtime only option". `OptionBool downloadonly{false}; // runtime only option`
> 
> And from my point of view, `downloadonly` is a specific option of dnf5 application. It is not a libdnf(5) option. Do we want to add this option to the "dnf.conf " file?

I see. I was confused that the `--downloadonly` parameter was not recognized for me when testing the PR and I just noticed the binding is missing and it might be the cause. Anyway, I tried to pull the PR again today and everything seems working correctly, so it was just probably my broken local setup.

jan-kolarik (approved):

j-mracek:
> > @jan-kolarik
> > > I think we are missing the owner.opt_binds().add("downloadonly", downloadonly); in the main config. Can you please fix that?
> > 
> > 
> > I don't think so. As you can see in the source code, `downloadonly` is a "runtime only option". `OptionBool downloadonly{false}; // runtime only option`
> > And from my point of view, `downloadonly` is a specific option of dnf5 application. It is not a libdnf(5) option. Do we want to add this option to the "dnf.conf " file?
> 
> I see. I was confused that the `--downloadonly` parameter was not recognized for me when testing the PR and I just noticed the binding is missing and it might be the cause. Anyway, I tried to pull the PR again today and everything seems working correctly, so it was just probably my broken local setup.

In the first version I enabled binding to allow `--setopt=downloadonly=true` but I was not sure whether we want to support that. Be honest I did not found any reason to not add it also.

dnf5/main.cpp
jan-kolarik: It looks weird, but static tests are not complaining and also the actual output looks OK, so I don't consider this as a problem.

PR #563, opened 2023-05-26
downloaders: Pass fail_fast and resume by setters

### LINKED ISSUE #528: Improve the download methods API
Current `download` methods for the `FileDownloader` and `PackageDownloader` are passing two `bool` parameters:
- `fail_fast`: whether to fail the whole download on a first error or keep downloading.
- `resume`: whether to try to resume the download if a destination package already exists.

It would be probably better to use the `download` method without arguments and introduce new setters on the `*Downloader` classes to configure these values to see the workflow of the individual operations more clearly.

jan-kolarik:
To make the API and code usages more explicit.

Closes #528.
CI tests fix: https://github.com/rpm-software-management/ci-dnf-stack/pull/1297.

### DIFF
--- a/include/libdnf/repo/package_downloader.hpp
+++ b/include/libdnf/repo/package_downloader.hpp
@@ -72,12 +72,11 @@ public:
     /// @param value If true, packages will be kept on the disk after downloading
     /// regardless the `keepcache` option value, if false, it enforces packages removal after
     /// the next successful transaction.
-    void force_keep_packages(bool value) { keep_packages = value; }
+    void force_keep_packages(bool value);
 
 private:
     class Impl;
     std::unique_ptr<Impl> p_impl;
-    std::optional<bool> keep_packages;
 };
 
 /// Wraps librepo PackageTarget
--- a/include/libdnf/repo/package_downloader.hpp
+++ b/include/libdnf/repo/package_downloader.hpp
@@ -53,9 +53,17 @@ public:
     void add(const libdnf::rpm::Package & package, const std::string & destination, void * user_data = nullptr);
 
     /// Download the previously added packages.
-    /// @param fail_fast Whether to fail the whole download on a first error or keep downloading.
-    /// @param resume Whether to try to resume the download if a destination package already exists.
-    void download(bool fail_fast, bool resume);
+    void download();
+
+    /// Configure whether to fail the whole download on a first error or keep downloading.
+    /// @param value If true, download will fail on the first error, otherwise it continues.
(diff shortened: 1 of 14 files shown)


include/libdnf/repo/package_downloader.hpp
```diff
@@ -53,9 +53,17 @@ class PackageDownloader {
...
+
+    /// Configure whether to fail the whole download on a first error or keep downloading.
+    /// @param value If true, download will fail on the first error, otherwise it continues.
+    /// This is set to true by default.
+    void set_fail_fast(bool value);
+
+    /// Configure whether to try resuming the download if a destination package already exists.
+    /// @param value If true, download is tried to be resumed, otherwise it starts over.
+    /// This is set to true by default.
+    void set_resume(bool value);
```
j-mracek, line 66: In this approach the behavior will be not modifiable from DNF5 manager. Only API user can make a decision what approach should be used. Could you imagine a situation when resume would create problem?

include/libdnf/repo/package_downloader.hpp
```diff
@@ -53,9 +53,17 @@ class PackageDownloader {
...
     /// Download the previously added packages.
-    /// @param fail_fast Whether to fail the whole download on a first error or keep downloading.
-    /// @param resume Whether to try to resume the download if a destination package already exists.
-    void download(bool fail_fast, bool resume);
+    void download();
+
+    /// Configure whether to fail the whole download on a first error or keep downloading.
+    /// @param value If true, download will fail on the first error, otherwise it continues.
+    /// This is set to true by default.
+    void set_fail_fast(bool value);
```
j-mracek, line 61: I think that for fail_fast it is the best approach. :)

j-mracek (commented):
The PR looks good, but I am not sure only about `void set_resume(bool value);`

include/libdnf/repo/package_downloader.hpp
jan-kolarik: If we would provide configuration options for `resume` or `fail_fast` then I understand it like these values would have global meaning. But we are already using specific values for different scenarios inside the libdnf. So the question is, how it will work then? Should the user configuration override the values we currently pass to the downloader in the library?

j-mracek: make sense, I overlooked the internal modification for download pgpkeys. it makes no sense to resume interrupted downloads for very small files.

j-mracek:
I've tested the PR locally and everything is ok.

PR #565, opened 2023-05-26
DNF5 should not deprecate DNF on Fedora38

inknos:


dnf5.spec
```diff
@@ -247,15 +251,19 @@ License:        LGPL-2.1-or-later
...
 Requires:       sqlite-libs%{?_isa} >= %{sqlite_version}
+%if 0%{?fedora} > 38
 Conflicts:      dnf-data < 4.16.0
+%endif
 
 %description -n libdnf5
 Package management library.
 
 %files -n libdnf5
+%if 0%{?fedora} > 38
```
jan-kolarik, line 262: Here we would need to do the changes also in `dnf` spec I guess as this is not already being provided there...

dnf5.spec
```diff
@@ -521,6 +529,9 @@ Requires:       libdnf5%{?_isa} = %{version}-%{release}
 Requires:       libdnf5-cli%{?_isa} = %{version}-%{release}
 Requires:       dbus
 Requires:       polkit
+%if 0%{?fedora} <= 38
+Requires:       dnf-data
```
jan-kolarik, line 533: We probably also want this for dnf5 package back, see [here](https://github.com/rpm-software-management/dnf5/pull/469/files#diff-3764aa091c8d6fcbde25e8587620de2d599df9a8a2897c372975ecfa4d6b4980L15).

inknos: fixed

jan-kolarik (approved):
LGTM

jan-kolarik:
This PR is addressing the https://github.com/rpm-software-management/dnf5/issues/567 issue.

evan-goode (approved):

PR #568, opened 2023-05-26
repoquery: add rq command alias

tmzullinger:
Running `repoquery` is common enough that many folks have become accustomed to the `rq` alias in dnf.


j-mracek:
I will test it locally, because our CI has not a good day. Thank you for the contribution.

dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
```diff
@@ -145,6 +145,13 @@ descr = "Alias for 'remove'"
 group_id = 'commands-compatibility-aliases'
 complete = false
 
+['rq']
+type = 'command'
+attached_command = 'repoquery'
+descr = "Alias for 'repoquery'"
+group_id = 'commands-compatibility-aliases'
+complete = true
```
kontura, line 153: We generally don't have command auto-completion turned on for these shortcuts, can you set
```
complete = false
```

tmzullinger: Sure, done.  Would a follow-up and/or separate patch to change the `ls` alias be welcome as well?  Then there would be no 2-character aliases which are completed.

dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
```diff
@@ -1,4 +1,4 @@
-version = '1.0'
+qersion = '1.0'
```
m-blaha, line 1: Oh, this is a typo which needs to be fixed, I guess.

m-blaha (changes requested):

dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
tmzullinger: Yuck.  I was clearly typing too early yesterday.  Sorry about that.

PR #571, opened 2023-05-29
Catch up cmake version with spec version

kontura:
This is required.
It is checked during build.
Follow up to https://github.com/rpm-software-management/dnf5/pull/565


kontura:
@inknos @jan-kolarik The CI builds are failing with a couple of problems. Should I look into those or is it expected due to the dnf4/dnf5 switch?

jan-kolarik:
@kontura Yes, I think this is expected and it's a consequence of the changes delivered during Thursday. We are planning to work on that today.

PR #573, opened 2023-05-29
Add a bunch of repoquery options

kontura:
Add several repoquery options.

### DIFF
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -19,6 +19,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "repoquery.hpp"
 
+#include "libdnf-cli/output/changelogs.hpp"
 #include "libdnf-cli/output/repoquery.hpp"
 
 #include "libdnf/utils/patterns.hpp"
@@ -265,6 +266,9 @@ void RepoqueryCommand::set_argument_parser() {
     advisory_bz = std::make_unique<BzOption>(*this);
     advisory_cve = std::make_unique<CveOption>(*this);
 
+    changelogs = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this, "changelogs", '\0', "Display package changelogs.", false);
+
     std::vector<std::string> pkg_attrs_options{
         "conflicts",
         "depends",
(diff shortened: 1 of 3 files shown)


dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -256,6 +256,16 @@ void RepoqueryCommand::set_argument_parser() {
         "Limit to unneeded installed packages (i.e. packages that were installed as "
         "dependencies but are no longer needed).",
         false);
+    extras = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this, "extras", '\0', "Limit to packages that are not present in any of of available repositories.", false);
```
m-blaha, line 264: Maybe we should be more explicit here and state that the packages are **installed** - `Limit to installed packages...`. Does it make the description too long?
And there is a duplicated **of** - `...in any of of available...`

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -256,6 +256,16 @@ void RepoqueryCommand::set_argument_parser() {
...
+    upgrades = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this,
+        "upgrades",
+        '\0',
+        "Limit to packages that provide an upgrade for some already installed package.",
+        false);
+    recent = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this, "recent", '\0', "Limit to only recently edited packages.", false);
```
m-blaha, line 272: I know it's also in dnf4 docs, but I consider word `edited` a bit odd in this context. Wouldn't be `changed` or `added` a better fit?

dnf5/commands/repoquery/repoquery.cpp
kontura: I like it, I have also rephrased the ending a bit.

dnf5/commands/repoquery/repoquery.cpp
kontura: `added` seems a bit misleading to me but `changed` is better, thanks. :+1:

kontura:
> Please, add the new options also to `doc/commands/repoquery.8.rst` documentation.

Right, all of the new options are missing the docs I was thinking I will add it all at once in a separate PR if that is ok??
(There is also a task for that in https://github.com/rpm-software-management/dnf5/issues/122#issue-1437992549)

PR #583, opened 2023-05-31
aliases: don't complete ls alias

tmzullinger:
No other 2-character aliases are completed.  To be consistent, don't complete the `ls` alias either.

In #568, it was noted that completion is not generally enabled for these shortcut aliases.  That seems like a good reason to disable completion of the `ls` alias as well.


kontura:
Thanks!

PR #595, opened 2023-06-07
repoquery: add --srpm and --sourcerpm

kontura:

### DIFF
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -272,6 +272,12 @@ void RepoqueryCommand::set_argument_parser() {
         *this, "recent", '\0', "Limit to only recently changed packages.", false);
     installonly = std::make_unique<libdnf5::cli::session::BoolOption>(
         *this, "installonly", '\0', "Limit to installed installonly packages.", false);
+    srpm = std::make_unique<libdnf5::cli::session::BoolOption>(
+        *this,
+        "srpm",
+        '\0',
+        "After filtering is finished use packages' corresponding source RPMs for output (enables source repositories).",
+        false);
 
     advisory_name = std::make_unique<AdvisoryOption>(*this);
     advisory_security = std::make_unique<SecurityOption>(*this);
@@ -360,6 +366,10 @@ void RepoqueryCommand::configure() {
                 extras->get_value() || upgrades->get_value() || recent->get_value()
             ? Context::LoadAvailableRepos::ENABLED
             : Context::LoadAvailableRepos::NONE);
+
+    if (srpm->get_value()) {
+        context.base.get_repo_sack()->enable_source_repos();
+    }
 
     if (changelogs->get_value()) {
         context.base.get_config().get_optional_metadata_types_option().add_item(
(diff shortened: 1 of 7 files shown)


kontura:
Rebased.

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -611,37 +629,20 @@ void RepoqueryCommand::run() {
...
-            result_pset |= package_query;
-        }
-    }
 
     if (querytags_option->get_value()) {
         libdnf::cli::output::print_available_pkg_attrs(stdout);
     } else if (changelogs->get_value()) {
-        libdnf::rpm::PackageQuery wrapped(result_pset);
-        libdnf::cli::output::print_changelogs(wrapped, full_package_query, false, 0, 0);
+        libdnf::cli::output::print_changelogs(full_package_query, full_package_query, false, 0, 0);
```
pkratoch, line 662: I was confused about this, because the second argument for `print_changelogs` changed during the move, but it doesn't have any effect... However, the query is then unnecessarily  copied and filtered, so wouldn't it be better to pass an empty query or change it in the `print_changelogs`? Neither is necessary in this PR, though (it's actually smaller now than it was before).

kontura: Yes, I agree. I tried to improve the `print_changelogs` in a new commit. Please take a look.

libdnf-cli/output/changelogs.cpp
```diff
@@ -71,9 +65,12 @@ void print_changelogs(
             });
 
         // filter changelog
-        if (upgrades) {
+        if (filter.first == ChangelogFilterType::UPGRADES) {
             // Find the newest changelog on the installed version of the
             // package, and filter out any changelogs newer than that
+            auto & installed = std::get<libdnf::rpm::PackageQuery>(filter.second);
+            installed.filter_installed();
```
pkratoch, line 72: The query is now created for each srpm, although it's the same for all of them. Can it be moved out of the loop?

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -272,6 +272,12 @@ void RepoqueryCommand::set_argument_parser() {
         *this, "recent", '\0', "Limit to only recently changed packages.", false);
     installonly = std::make_unique<libdnf::cli::session::BoolOption>(
         *this, "installonly", '\0', "Limit to installed installonly packages.", false);
+    srpm = std::make_unique<libdnf::cli::session::BoolOption>(
+        *this,
+        "srpm",
+        '\0',
+        "For selected packages format corresponding source RPMs (enables source repositories).",
```
pkratoch, line 279: Hm, I don't understand the word "format" in this context. Previously, there was "operate on", which seems more intuitive, but maybe I am missing something?

dnf5/commands/repoquery/repoquery.cpp
pkratoch: Thanks. FYI, I encountered an unrelated issue with the changelog command (https://github.com/rpm-software-management/dnf5/issues/627) when testing it. Just mentioning if you run into the same.

libdnf-cli/output/changelogs.cpp
kontura: Oh, I will move the filtering out of the loop. Thanks!

dnf5/commands/repoquery/repoquery.cpp
pkratoch: Well, other options are applied to the `srpms` - the query options. In dnf5, I don't see it in the documentation (at least not on the first glance), but there is `info` or the `pkg_attrs_options` (`requires`, `provides`...).

But I see what you mean and the "operate on" is not super clear either. The proposed sentence looks easier to understand.

kontura: I have updated it.

PR #596, opened 2023-06-07
Add dnf5-unstable copr repo to dnf5 CI copr builds

kontura:
This is needed because since rpm-software-management/dnf#1937 new dnf-data (in dnf-nightly) requires `/etc/dnf/dnf.conf` but in rawhide and dnf-nightly nothing provides that. It is provided by new libdnf5 currently only present in dnf5-unstable.

It should make the CI Copr builds pass on rawhide.


kontura:
The builds are still not passing because the change needs to be merged to take effect. I believe the relevant code is not taken from the PR.

Issue #112, opened 2022-10-29
--forcearch is missing


berolinux:
dnf 4 has the --forcearch option, useful for installing foreign architecture chroots e.g. for use with qemu.

dnf5 doesn't have this option; and if an equivalent exists, it's hidden so well that it triggers a PBKAC with me ;)

j-mracek:
`--forcearch` modify `arch` substitution or configuration and `ignorearch`.

```
        if self.conf.ignorearch:
            self._rpm_probfilter.add(rpm.RPMPROB_FILTER_IGNOREARCH)
```

We can implement the same option or we can use a setter for arch substitution, but it requires also handle rpm flag to ignore architecture.

The option is required for dependency check for repositories with a different architecture.

praiskup:
This is used/needed by Mock too.

PR #599, opened 2023-06-07
Support --forcearch for many commands

evan-goode:
Added --forcearch to:

- `dnf5 distro-sync`
- `dnf5 download`
- `dnf5 group`
- `dnf5 install`
- `dnf5 makecache`
- `dnf5 repo`
- `dnf5 repoquery`
- `dnf5 search`
- `dnf5 swap`

The --forcearch=<arch> argument overrides the "arch" and "basearch" substitution vars and tells librpm to allow installations of packages not matching the architecture of the host system. It enables several use cases, including:

- Install packages to be used with something like qemu-user-static
- Search repos for packages built for a different architecture
- Download packages to be installed on another system

Resolves https://github.com/rpm-software-management/dnf5/issues/112.

A good Fedora package to test this with is `Random123-doc` since it has a version for `x86_64` and `aarch64` (though it probably should be `noarch` since it's documentation?) but has no dependencies.

The `forcearch` argument/conf option should override the `arch` and `basearch` vars as well as the `ignorearch` conf option, but there didn't seem to be a standard place we do post-processing of config like this. I put the logic  [here](https://github.com/evan-goode/dnf5/blob/ac70b2c810d73eecb3724deefb32630af779e44a/libdnf/base/base.cpp#L193) in `Base::setup()` since it needs to happen in after config is loaded and before the RPM pool is set up with the architecture. But maybe there should be a `Base::apply_vars_overrides()` or something so `Base::setup()` doesn't get so cluttered.

If this looks good, I can add the docs changes to this PR. Tests are also still on the way, there don't seem to be any tests in dnf-ci-stack already for `--forcearch`.

### DIFF
--- a/dnf5/commands/distro-sync/distro-sync.cpp
+++ b/dnf5/commands/distro-sync/distro-sync.cpp
@@ -52,6 +52,7 @@ void DistroSyncCommand::set_argument_parser() {
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     auto skip_broken = std::make_unique<SkipBrokenOption>(*this);
+    create_forcearch_option(*this);
 }
 
 void DistroSyncCommand::configure() {
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -79,6 +79,7 @@ void DownloadCommand::set_argument_parser() {
 
     cmd.register_named_arg(alldeps);
     create_destdir_option(*this);
+    create_forcearch_option(*this);
     cmd.register_named_arg(resolve);
     cmd.register_positional_arg(keys);
 }
--- a/dnf5/commands/group/group.cpp
+++ b/dnf5/commands/group/group.cpp
@@ -25,6 +25,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "group_remove.hpp"
 #include "group_upgrade.hpp"
 
+#include <dnf5/shared_options.hpp>
+
 namespace dnf5 {
 
 void GroupCommand::set_parent_command() {
@@ -36,6 +38,7 @@ void GroupCommand::set_parent_command() {
 
 void GroupCommand::set_argument_parser() {
     get_argument_parser_command()->set_description("Manage comps groups");
+    create_forcearch_option(*this);
 }
 
 void GroupCommand::register_subcommands() {
--- a/dnf5/commands/install/install.cpp
+++ b/dnf5/commands/install/install.cpp
@@ -54,6 +54,7 @@ void InstallCommand::set_argument_parser() {
 
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
     create_downloadonly_option(*this);
+    create_forcearch_option(*this);
 
     advisory_name = std::make_unique<AdvisoryOption>(*this);
     advisory_security = std::make_unique<SecurityOption>(*this);
--- a/dnf5/commands/list/list.cpp
+++ b/dnf5/commands/list/list.cpp
@@ -21,6 +21,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "utils/bgettext/bgettext-mark-domain.h"
 
+#include <dnf5/shared_options.hpp>
 #include <libdnf5-cli/output/package_list_sections.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 
@@ -56,6 +57,8 @@ void ListCommand::set_argument_parser() {
         });
     specs->set_complete_hook_func([&ctx](const char * arg) { return match_specs(ctx, arg, true, true, false, false); });
     cmd.register_positional_arg(specs);
+
+    create_forcearch_option(*this);
 
     show_duplicates = std::make_unique<libdnf5::cli::session::BoolOption>(
         *this, "showduplicates", '\0', "Show all versions of the packages, not only the latest ones.", false);
--- a/dnf5/commands/makecache/makecache.cpp
+++ b/dnf5/commands/makecache/makecache.cpp
@@ -19,6 +19,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "makecache.hpp"
 
+#include <dnf5/shared_options.hpp>
 #include <fmt/format.h>
 
 #include <iostream>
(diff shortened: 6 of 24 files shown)


libdnf/rpm/transaction.hpp
```diff
@@ -334,6 +334,7 @@ class Transaction {
 
     std::map<unsigned int, rpmte> implicit_ts_elements;  // elements added to the librpm transaction by librpm itself
     bool downgrade_requested{false};
+    bool ignorearch{false};
```
jrohel, line 337: Do we need a private copy of `ignorearch` stored in the class?

j-mracek: I think that the value could be used directly from `base.conf`. Or is there any reason why would you like to store it here?

libdnf/base/base.cpp
```diff
@@ -190,6 +190,12 @@ void Base::setup() {
 
     get_vars()->load(vars_installroot, config.get_varsdir_option().get_value());
 
+    if (!config.get_forcearch_option().get_value().empty()) {
+        config.get_ignorearch_option().set(Option::Priority::RUNTIME, true);
+        get_vars()->set("arch", config.get_forcearch_option().get_value(), libdnf::Vars::Priority::CONFIG);
+        get_vars()->set("basearch", config.get_forcearch_option().get_value(), libdnf::Vars::Priority::CONFIG);
+    }
```
j-mracek, line 197: I am  not in favor of a redundancy because of this part of the code. It is possible to modify vars from multiple side including commandline. In the code above, you try to handle priority and this could be a problem. Lets use as a test command `dnf install acpi  --setvar=arch=x86_64 --forcearch=s390`. I believe that the final value would be `x86_64`, because it has command-line priority and cannot be overridden by the Config priority.

libdnf/base/base.cpp
```diff
@@ -190,6 +190,12 @@ void Base::setup() {
 
     get_vars()->load(vars_installroot, config.get_varsdir_option().get_value());
 
+    if (!config.get_forcearch_option().get_value().empty()) {
+        config.get_ignorearch_option().set(Option::Priority::RUNTIME, true);
+        get_vars()->set("arch", config.get_forcearch_option().get_value(), libdnf::Vars::Priority::CONFIG);
+        get_vars()->set("basearch", config.get_forcearch_option().get_value(), libdnf::Vars::Priority::CONFIG);
```
j-mracek, line 196: I think we should add another logic here  or define what is allowed for forcearch.
My system can detect that I have `amd64`, then arch == `amd64`. And base arch is generated from that value  using internal get_base_arch(). Base arch is important for repository path substitution therefore using any value will lead into repository paths that are not available.

Dnf4 validates value and convert it from arch to basearch:
```
 sudo dnf --forcearch amd64 install acpi
Fedora 38 - x86_64 - Updates                                                                                                                                                                            24 kB/s |  19 kB     00:00    
Fedora 38 - x86_64 - Updates                                                                                                                                                                           2.5 MB/s | 2.5 MB     00:01    
Fedora Modular 38 - x86_64 - Updates                                                                                                                                                                    15 kB/s |  18 kB     00:01    
Dependencies resolved.
=======================================================================================================================================================================================================================================
 Package                                              Architecture                                           Version                                                      Repository                                              Size
=======================================================================================================================================================================================================================================
Installing:
 acpi                                                 x86_64                                                 1.7-20.fc38                                                  fedora                                                  28 k

Transaction Summary
=======================================================================================================================================================================================================================================
Install  1 Package

Total download size: 28 k
Installed size: 52 k
Is this ok [y/N]: 
```
```
sudo dnf --forcearch and64 install acpi
usage: dnf [-c [config file]] [-q] [-v] [--version] [--installroot [path]] [--nodocs] [--noplugins] [--enableplugin [plugin]] [--disableplugin [plugin]] [--releasever RELEASEVER] [--setopt SETOPTS] [--skip-broken] [-h]
           [--allowerasing] [-b | --nobest] [-C] [-R [minutes]] [-d [debug level]] [--debugsolver] [--showduplicates] [-e ERRORLEVEL] [--obsoletes] [--rpmverbosity [debug level name]] [-y] [--assumeno] [--enablerepo [repo]]
           [--disablerepo [repo] | --repo [repo]] [--enable | --disable] [-x [package]] [--disableexcludes [repo]] [--repofrompath [repo,path]] [--noautoremove] [--nogpgcheck] [--color COLOR] [--refresh] [-4] [-6]
           [--destdir DESTDIR] [--downloadonly] [--comment COMMENT] [--bugfix] [--enhancement] [--newpackage] [--security] [--advisory ADVISORY] [--bz BUGZILLA] [--cve CVES] [--sec-severity {Critical,Important,Moderate,Low}]
           [--forcearch ARCH]
Command line error: argument --forcearch: invalid choice: 'and64' (choose from 'aarch64', 'alpha', 'alphaev4', 'alphaev45', 'alphaev5', 'alphaev56', 'alphaev6', 'alphaev67', 'alphaev68', 'alphaev7', 'alphapca56', 'amd64', 'armv5tejl', 'armv5tel', 'armv5tl', 'armv6hl', 'armv6l', 'armv7hl', 'armv7hnl', 'armv7l', 'armv8hl', 'armv8l', 'athlon', 'geode', 'i386', 'i486', 'i586', 'i686', 'ia32e', 'ia64', 'loongarch64', 'mips', 'mips64', 'mips64el', 'mipsel', 'noarch', 'ppc', 'ppc64', 'ppc64iseries', 'ppc64le', 'ppc64p7', 'ppc64pseries', 'riscv128', 'riscv32', 'riscv64', 's390', 's390x', 'sh3', 'sh4', 'sh4a', 'sparc', 'sparc64', 'sparc64v', 'sparcv8', 'sparcv9', 'sparcv9v', 'x86_64')

```

libdnf/rpm/transaction.hpp
evan-goode: No reason, I've changed it to use the config option directly.

evan-goode:
> It seems for me as too many different changes in a single commit, like:
> 
>     * introducing a new shared option `--forcearch`
> 
>     * usage of the new option in various commands
> 
>     * new `Priority::CONFIG` added
> 
>     * use of `ignorearch` in `libdnf::rpm::transaction`
> 
> 
> Do you think it could be separated into multiple commits?

Yes, thanks for the suggestion!

evan-goode:
@j-mracek thank you for your review, I understand what you're suggesting, and I agree, it's simpler this way.

I changed `--forcearch` to not have its own config option and to just override the existing `arch` var and the `ignorearch` config option. Now the base arch detection works as expected; `get_base_arch` is used as usual.

This approach has the drawback that `forcearch` is not configurable in `dnf.conf`, but DNF 4 didn't have that anyway.

j-mracek:
LGTM, may I ask you for a rebase?

evan-goode:
Rebased, added support for a couple more commands, and added a page in the documentation for `--forcearch`.

j-mracek:
LGTM

Issue #126, opened 2022-11-09
Adding additional command for Modularity


mcurlej:
- [x] #648
- [x] #649
- [x] #650

pkratoch:
Related (already merged) PRs:
https://github.com/rpm-software-management/dnf5/pull/387 - modules: Change State to set and get the whole ModuleState
https://github.com/rpm-software-management/dnf5/pull/401 - modules: Add ModuleDB to store the runtime states of all modules
These are needed to manage module states to make modules enabled, disabled or neither.

pkratoch:
Module enable PR: https://github.com/rpm-software-management/dnf5/pull/606

j-mracek:
According to description the scope of the issue is resolved.

PR #606, opened 2023-06-09
modules: Implement module enable

pkratoch:
Related issue: https://github.com/rpm-software-management/dnf5/issues/126

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -203,6 +203,9 @@ public:
             case libdnf::transaction::TransactionItemAction::REPLACED:
                 break;
             case libdnf::transaction::TransactionItemAction::REASON_CHANGE:
+            case libdnf::transaction::TransactionItemAction::ENABLE:
+            case libdnf::transaction::TransactionItemAction::DISABLE:
+            case libdnf::transaction::TransactionItemAction::RESET:
                 throw std::logic_error(fmt::format(
                     "Unexpected action in TransactionPackage: {}",
                     static_cast<std::underlying_type_t<libdnf::base::Transaction::TransactionRunResult>>(
--- a/include/libdnf-cli/output/transaction_table.hpp
+++ b/include/libdnf-cli/output/transaction_table.hpp
@@ -248,6 +248,53 @@ private:
 };
 
 
+class ActionHeaderPrinterModule {
+public:
+    ActionHeaderPrinterModule(struct libscols_table * table) : table(table) {}
+
+    template <class T>
+    struct libscols_line * print(const T & tsmodule) {
+        if (!current_action || *current_action != tsmodule.get_action() || !current_reason ||
+            *current_reason != tsmodule.get_reason()) {
+            auto reason = tsmodule.get_reason();
+            auto action = tsmodule.get_action();
+            current_header_line = scols_table_new_line(table, NULL);
+            std::string text;
+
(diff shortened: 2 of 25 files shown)


libdnf/base/goal.cpp
```diff
@@ -2060,6 +2064,23 @@ bool Goal::get_allow_erasing() const {
 }
 
 base::Transaction Goal::resolve() {
+    if (!p_impl->base->is_initialized()) {
+        throw RuntimeError(M_("Base not initialized"));
+    }
```
pkratoch, line 2069: Not sure what to throw here.

j-mracek: What about:
```
libdnf_user_assert(p_impl->base->is_initialized(), "Base instance was not fully initialized by Base::setup()");
```

dnf5/context.cpp
```diff
@@ -202,7 +202,7 @@ class RpmTransCB : public libdnf::rpm::TransactionCallbacks {
             case libdnf::transaction::TransactionItemAction::REMOVE:
             case libdnf::transaction::TransactionItemAction::REPLACED:
                 break;
-            case libdnf::transaction::TransactionItemAction::REASON_CHANGE:
+            default:
```
j-mracek, line 205: In DNF5 we try to avoid default cases, because it hides missing handling in switches during compilation and it is discovered during runtime when value is used. If there is no default case and we add a new value to enum, the compilation will tell us all places where enum is used in a switch - safer approach. Is there any reason why you would prefer to use default case here?

include/libdnf-cli/output/transaction_table.hpp
```diff
@@ -114,7 +117,7 @@ class ActionHeaderPrinter {
                 case libdnf::transaction::TransactionItemAction::REASON_CHANGE:
                     text = "Changing reason";
                     break;
-                case libdnf::transaction::TransactionItemAction::REPLACED:
```
j-mracek, line 117: I have the same note like above.

include/libdnf-cli/output/transaction_table.hpp
```diff
@@ -242,6 +245,53 @@ class ActionHeaderPrinterGroup {
...
+                case libdnf::transaction::TransactionItemAction::ENABLE:
+                    text = "Enabling module streams";
+                    break;
+                case libdnf::transaction::TransactionItemAction::DISABLE:
+                    text = "Disabling module streams";
+                    break;
+                case libdnf::transaction::TransactionItemAction::RESET:
+                    text = "Resetting module streams";
+                    break;
+                default:
```
j-mracek, line 271: I think the default case is used correctly here.

include/libdnf-cli/output/transaction_table.hpp
```diff
@@ -267,6 +317,9 @@ class TransactionSummary {
             case libdnf::transaction::TransactionItemAction::REASON_CHANGE:
                 reason_changes++;
                 break;
+            default:
```
j-mracek, line 320: I think it would be better to explicitly list all actions here even if we are going to skip them with Todo.

dnf5/context.cpp
pkratoch: I see, that makes sense. I listed all the options, then.

j-mracek:
LGTM

Issue #574, opened 2023-05-29
dnf5daemon-server install packages without checking gpg key


timlau:
While implementing dnf5daemon support for  https://github.com/timlau/yumex-ng i found out the the **repo_key_import_request** signal is not sent when a key need to be imported, it just install the package with out any issues.

### Enable a new copr repo

sudo dnf copr enable timlau/rasmil -y

### Test with dnf5 cli

**This work as expected, dnf5 ask for key import**

```
sudo dnf5 install python3-rasmil
Updating and loading repositories:
Repositories loaded.
Package                       Arch   Version                       Repository                   Size
Installing:                                                                                         
 python3-rasmil               noarch 1.0.2-2.fc38                  copr:copr.fedorainfracl  76.5 KiB
Installing dependencies:                                                                            
 libxfce4util                 x86_64 4.18.1-2.fc38                 fedora                  889.9 KiB
 xfconf                       x86_64 4.18.1-1.fc38                 updates                 832.6 KiB

Transaction Summary:
 Installing:        3 packages

Total size of inbound packages is 440 KiB. Need to download 0 B.
After this operation 2 MiB will be used (install 2 MiB, remove 0 B).
Is this ok [y/N]: y
[1/3] python3-rasmil-0:1.0.2-2.fc38.noarch                  100% |   0.0   B/s |   0.0   B |  00m00s
>>> Already downloaded
[2/3] xfconf-0:4.18.1-1.fc38.x86_64                         100% |   0.0   B/s |   0.0   B |  00m00s
>>> Already downloaded
[3/3] libxfce4util-0:4.18.1-2.fc38.x86_64                   100% |   0.0   B/s |   0.0   B |  00m00s
>>> Already downloaded
----------------------------------------------------------------------------------------------------
[3/3] Total                                                 100% |   0.0   B/s |   0.0   B |  00m00s
[1/4] https://download.copr.fedorainfracloud.org/results/ti 100% |   1.7 KiB/s | 994.0   B |  00m01s
----------------------------------------------------------------------------------------------------
[4/4] Total                                                 100% |   0.0   B/s |   0.0   B |  00m00sImporting PGP key 0xD3337195
 UserId     : "timlau_rasmil (None) <timlau#rasmil@copr.fedorahosted.org>"
 Fingerprint: B2C5988089B26F72266DF052B12760C9D3337195
 From       : https://download.copr.fedorainfracloud.org/results/timlau/rasmil/pubkey.gpg
[...]

Issue #557, opened 2023-05-24
Unify the code related to keys importing


jan-kolarik:
`libdnf::repo::*Callbacks` are using the virtual method approach to allow client overriding or configuring custom behavior for individual situations. We should use this also in `libdnf::base::Transaction::check_gpg_signatures` where a custom function for confirming keys import is passed from CLI use case now. The current approach there is unwanted as we expose `std::function` on the public API which is not supported by SWIG. Also this `user_confirm_key` function defined in `context.cpp` is basically a copy of the `KeyImportRepoCB::user_confirm_key` method, so this should be also unified.

This is a follow-up to https://github.com/rpm-software-management/dnf5/pull/404.

Issue #556, opened 2023-05-24
Transaction::run() does not verify signatures


j-mracek:
Transaction::run() does not verify signature even when it is supposed to do according to configuration.

There is a method to do the verification (`check_gpg_signatures`) but it is not clear whether it is API, how to set callbacks, and it is not described in tutorials.

Shall we make Transaction::run()  sensitive for signature verification by default?

Related: https://github.com/rpm-software-management/dnf5/issues/555

PR #607, opened 2023-06-09
Verify transaction PGP signatures automatically

m-blaha:
Currently signatures verification requires to manually call `transaction.check_gpg_signatures()` API.
This PR removes the API and changes the workflow so that the signature verification is done automatically during the transaction run.

Resolves: https://github.com/rpm-software-management/dnf5/issues/556
Resolves: https://github.com/rpm-software-management/dnf5/issues/574
Resolves: https://github.com/rpm-software-management/dnf5/issues/557

Tests adujstments: https://github.com/rpm-software-management/ci-dnf-stack/pull/1314

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -376,38 +376,11 @@ std::chrono::time_point<std::chrono::steady_clock> RpmTransCB::prev_print_time =
 }  // namespace
 
 
-bool Context::check_gpg_signatures(libdnf5::base::Transaction & transaction) {
-    bool any_import_confirmed = false;
-    std::function<bool(const libdnf5::rpm::KeyInfo &)> bound_user_confirm =
-        [&](const libdnf5::rpm::KeyInfo & key_info) {
-            auto import_confirmed = user_confirm_key(base.get_config(), key_info);
-            if (import_confirmed) {
-                any_import_confirmed = true;
-            }
-            return import_confirmed;
-        };
-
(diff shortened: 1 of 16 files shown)


libdnf/rpm/rpm_signature.cpp
```diff
@@ -88,12 +88,14 @@ KeyInfo::KeyInfo(
     const std::string & key_id,
     const std::vector<std::string> & user_ids,
     const std::string & fingerprint,
+    const long int timestamp,
```
jrohel, line 91: `timestamp` is passed by value. The `const` keyword is not appropriate here.

m-blaha: Yes, thanks!

include/libdnf/repo/repo.hpp
```diff
@@ -100,6 +100,8 @@ class Repo {
     // @replaces libdnf:repo/Repo.hpp:method:Repo.setCallbacks(std::unique_ptr<RepoCallbacks> && callbacks)
     void set_callbacks(std::unique_ptr<libdnf::repo::RepoCallbacks> && callbacks);
 
+    std::unique_ptr<libdnf::repo::RepoCallbacks> & get_callbacks();
```
jan-kolarik, line 103: Better to have a docstring there as it is public API.

m-blaha: Oh! Thanks for the remainder.

jan-kolarik (approved):

Issue #572, opened 2023-05-29
Set temporar cachedir for binding UNIT tests


j-mracek:
To ensure stability DNF5 must always use unique temporary directory for unit tests and clean it everytime. I believe that C++ test already adopted that approach  but tests for bindings not yet.

Why this is important?

When you use `make test` then it will keep cached repositories on the system and the cache could interact with the next run and create a fail for no obvious reason.

PR #616, opened 2023-06-14
test: Add cachedir to python3 bindings tutorial

pkratoch:
Without setting cachedir to a temporary directory, when testing the python3 bindings tutorial, the user cache directory is used and the test repository rpm-repo1 is cached there. When dnf5 is then built second time, the newly created repository metadata differ from the cached ones and the tests fail with:

terminate called after throwing an instance of 'Swig::DirectorMethodException'
  what():  SWIG director method error. Error detected when calling 'DownloadCallbacks.mirror_failure'

Resolves: https://github.com/rpm-software-management/dnf5/issues/572

### DIFF
--- a/test/python3/libdnf5/tutorial/session/create_base.py
+++ b/test/python3/libdnf5/tutorial/session/create_base.py
@@ -28,6 +28,12 @@ base.load_config_from_file()
 base_config = base.get_config()
 base_config.installroot = installroot
 
+# Optionally set cachedir.
+#
+# By default, system cachedir or user cachedir is used, but we can set it to a
+# different location. This is useful for tests.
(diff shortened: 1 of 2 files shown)


pkratoch:
The problem I see with this fix is that the cachedir is not needed for the tutorial, but only for testing the tutorial. But if it's overridden outside the tutorial, it might hide future problem in the tutorial itself. For that reason, I put it inside the tutorial, but I'll be happy for any suggestion how to do it better.

j-mracek:
LGTM.

j-mracek:
1 test failed but the change is in Python tutorial (not related) therefore it can be merged.

PR #619, opened 2023-06-15
Document change in disable_excludes option

jan-kolarik:
The `*` glob character is used now to disable all configuration file excludes instead of the `all` value as before in DNF.


kontura:
This matches https://github.com/rpm-software-management/dnf5/commit/e9dc7314c7cc050c3eba95e68d92104bbe4805f0

Thanks

Issue #627, opened 2023-06-20
Changelog --since not working


pkratoch:
There are two problems:

- [x] Unexpected value results in "Aborted (core dumped)"
- [x] Date in the YYYY-MM-DD format (as documented in help) gives random results

## Unexpected value results in "Aborted (core dumped)"

**Actual output:**

```
# dnf5 changelog dnf --since --help
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid date: --help
Aborted (core dumped)

# dnf5 changelog dnf --since yesterday
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid date: yesterday
Aborted (core dumped)
```

**Expected output:**

A helpful message/error explaining how to use it.

## Date in the YYYY-MM-DD format (as documented in help) gives random results

**Actual output:**
```
# dnf5 changelog dnf --since 2023-01-01
Updating and loading repositories:
Repositories loaded.
Listing changelogs since Tue Jul  5 21:27:04 2072
Changelogs for dnf-4.15.1-1.fc37.noarch

# dnf5 changelog dnf --since 2023-01-01
Updating and loading repositories:
Repositories loaded.
Listing changelogs since Tue Jan 23 09:52:48 2046
Changelogs for dnf-4.15.1-1.fc37.noarch
```

**Expected output:**

```
# dnf5 changelog dnf --since 2023-01-01
Updating and loading repositories:
Repositories loaded.
Listing changelogs since Sun Jan 01 00:00:00 2023
Changelogs for dnf-4.15.1-1.fc37.noarch
(...)
```

PR #629, opened 2023-06-20
changelog: Fix handling of the --since option

jan-kolarik:
Use the `libdnf::cli::ArgumentParserError` exception instead of the generic runtime error for invalid date format with `--since` option. This way it is processed in `main` instead of aborting and dumping the core.

Also initialize the `time_m` struct with default values, otherwise it results in returning random values.

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1322.
Closes: #627.


pkratoch:
Thanks, LGTM. But I will wait with the merge, partly not to disturb the huge libdnf to libdnf5 rename, partly to give others chance to look at if they want (because this is open only for a very short time).

pkratoch (approved):

PR #630, opened 2023-06-20
Do not download updateinfo and comps metadata by default

kontura:
Updateinfo and comps metadata were always downloaded and loaded by default. This is not necessary because each command configures the `optional_metadata_types` to download/load only what it requires.

Also there was a bug in detection whether updateinfo metadata are needed.

For: https://bugzilla.redhat.com/show_bug.cgi?id=2214520

In my testing when loading updateinfo metadata memory requirements spike at ~623 MiB but without them its only ~132 MiB (on F38 with clean cache). I will further investigate why the updateinfo metadata use so much during libsolv loading, it seems unexpected to me given their size.

### DIFF
--- a/dnf5/commands/check-upgrade/check-upgrade.cpp
+++ b/dnf5/commands/check-upgrade/check-upgrade.cpp
@@ -86,6 +86,15 @@ void CheckUpgradeCommand::configure() {
         context.base.get_config().get_optional_metadata_types_option().add(
             libdnf5::Option::Priority::RUNTIME, libdnf5::OPTIONAL_METADATA_TYPES);
     }
+    context.update_repo_metadata_from_advisory_options(
+        advisory_name->get_value(),
+        advisory_security->get_value(),
+        advisory_bugfix->get_value(),
+        advisory_enhancement->get_value(),
+        advisory_newpackage->get_value(),
+        advisory_severity->get_value(),
+        advisory_bz->get_value(),
+        advisory_cve->get_value());
 }
 
 std::unique_ptr<libdnf5::cli::output::PackageListSections> CheckUpgradeCommand::create_output() {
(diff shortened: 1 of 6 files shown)


libdnf/conf/config_main.cpp
```diff
@@ -187,8 +187,7 @@ class ConfigMain::Impl {
     OptionBool debug_solver{false};
     OptionStringList installonlypkgs{INSTALLONLYPKGS};
     OptionStringList group_package_types{GROUP_PACKAGE_TYPES};
-    OptionStringSet optional_metadata_types{
-        OptionStringSet::ValueType{libdnf::METADATA_TYPE_COMPS, libdnf::METADATA_TYPE_UPDATEINFO}};
+    OptionStringSet optional_metadata_types{OptionStringSet::ValueType{}};
```
j-mracek, line 190: We can modify in dnf5 default but we should not modify library defaults. This default is insecure. I would rather investigate the cause then disabling loading.

j-mracek:
This patch is not fixing the issue but hides it. It looks like that the issue is in processing updateinfo metadata, because I was unable to reproduce the issue with `--setopt=optional_metadata_types=,comps`. I do not recommend to merge the PR and I would rather wait for the proper solution.

libdnf/conf/config_main.cpp
j-mracek: I added the blocker only because of the change of default type for download.

j-mracek:
LGTM

j-mracek:
@kontura May I ask you for rebase?

kontura:
> @kontura May I ask you for rebase?

Sure, rebased.

j-mracek (approved):

PR #636, opened 2023-06-21
swig: Return None for unset options in Python

### LINKED ISSUE #594: Accessing unset metalink or mirrorlist results in RuntimeError
While testing the new environment fixes from #585 I discovered that you no longer need to use .set() and .get_value() which is great.
But if I try to access these two attributes from a repository's config it raises a RuntimeError of `Option value is not set`

The traceback looks like:

```
Traceback (most recent call last):
  File "/root/./dnf-json", line 397, in solve
    result = solver.depsolve(transactions)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/./dnf-json", line 345, in depsolve
    "remote_location": remote_location(package),
                       ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/./dnf-json", line 66, in remote_location
    print(rc.mirrorlist)
          ^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/libdnf5/conf.py", line 2490, in <lambda>
    lambda self, attr=attr: _config_option_getter(self, attr),
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/libdnf5/conf.py", line 2479, in _config_option_getter
    return getattr(config_object, option_name)().get_value()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/libdnf5/conf.py", line 1008, in get_value
    return _conf.OptionString_get_value(self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Option value is not set

```

I would expect to get an empty string if it hasn't been set.

jan-kolarik:
When using Python bindings, return a `None` value when accessing values of unset configuration options.

The solution is simplified to do so if any runtime error occurred within the `get_value()` option call.

API unit tests included.

Closes #594.

### DIFF
--- a/bindings/libdnf5/conf.i
+++ b/bindings/libdnf5/conf.i
@@ -128,7 +128,10 @@ wrap_unique_ptr(StringUniquePtr, std::string);
 import re
 
 def _config_option_getter(config_object, option_name):
-    return getattr(config_object, option_name)().get_value()
+    try:
+        return getattr(config_object, option_name)().get_value()
+    except RuntimeError:
+        return None
 
 def _config_option_setter(config_object, option_name, value):
     getattr(config_object, option_name)().set(value)
--- a/test/python3/libdnf5/conf/test_option.py
+++ b/test/python3/libdnf5/conf/test_option.py
@@ -70,3 +70,12 @@ class TestConfigurationOptions(base_test_case.BaseTestCase):
         option.lock('')
 
         self.assertRaises(RuntimeError, option.set, False)
+
+    def test_get_unset_option_by_attribute(self):
+        config = self.base.get_config()
+        destdir = config.destdir
+        self.assertEqual(destdir, None)
+
+    def test_get_unknown_option_by_attribute(self):
+        config = self.base.get_config()
+        self.assertRaises(AttributeError, lambda: config.xyz)


bindings/libdnf5/conf.i
```diff
@@ -128,7 +128,10 @@ wrap_unique_ptr(StringUniquePtr, std::string);
 import re
 
 def _config_option_getter(config_object, option_name):
-    return getattr(config_object, option_name)().get_value()
+    try:
+        return getattr(config_object, option_name)().get_value()
+    except RuntimeError:
```
j-mracek, line 133: I think it is not ideal. The except part quite wide, but this is a consequence of unfinished error handling in swig part. What about to add a todo to use more targeted error handling when it will be available?

jan-kolarik: I know it's broad, but do we actually want to throw any exception when accessing an attribute like this in Python? Is there some case we want to throw an exception?

j-mracek: I am not sure whether it is valid, but I see ways how to generate an exception for different case that option is not set. I don't know what happen when attribute does not exists. If it generates runtime error then it is expected that it will fail.

j-mracek: Thank you for clarification.

PR #637, opened 2023-06-21
dnf5: --allowerasing for reinstall command

m-blaha:
It was clearly forgotten to implement for reinstall command.


pmatilai:
But... how does reinstall legitimately erase anything?

j-mracek:
> But... how does reinstall legitimately erase anything?

This situation is really rare. It could be useful only when you have broken systems or the package with conflict was installed by rpm without dependency check.

j-mracek:
LGTM

PR #643, opened 2023-06-22
Typo: pacakge -> package

m-blaha:
There are only two occurences actually visible to the user, but still...


jan-kolarik:
It seems this needs also fix in CI ...

m-blaha:
wow, we are testing presence of the typo in the output :)
OK, I'll fix also the tests.

PR #652, opened 2023-06-22
Add placeholder test to test/python3/libdnf5_cli

evan-goode:
The tests started failing after updating to Python 3.12 because of this change in unittest:
https://discuss.python.org/t/unittest-fail-if-zero-tests-were-discovered/21498 If no tests are discovered, unittest now fails with a nonzero exit code.

This adds a "placeholder" test that does nothing, just to get unittest to succeed.


kontura:
Thanks!

Issue #180, opened 2022-11-30
Decide how to ship bgettext (and others from common/utils/ directory) headers


m-blaha:
There is `#include "utils/bgettext/bgettext-mark-common.h"` in `include/libdnf/common/exception.hpp` which makes `dnf install dnf5-devel` insufficient for building a third-party plugin. The bgettext/* headers are not part of any dnf5 devel packages.

jan-kolarik:
After the team discussion, we decided to resolve this by exposing the needed methods to the public API.

evan-goode:
Whoops, I meant to assign myself on a different issue. Didn't mean to steal this from ya @jan-kolarik

jan-kolarik:
> Whoops, I meant to assign myself on a different issue. Didn't mean to steal this from ya @jan-kolarik

Haha, no worry! Let's leave this unassigned for now, so anyone can take it if interested. I still have several things in my queue :slightly_smiling_face:

jan-kolarik:
This could be now closed as the PR is merged?

PR #655, opened 2023-06-23
bgettext: Include header files in libdnf5 devel package

jrohel:
The header file from bgettext is used in the libdnf5 public header file. So the bgettext header files must also be public - part of the libdnf5 devel package.

Issue: https://github.com/rpm-software-management/dnf5/issues/180


j-mracek (approved):
LGTM

Issue #589, opened 2023-06-02
dnf5 repoquery --location option


praiskup:
There used to be `--qf %location` possibility, but seems removed now even in the recent DNF4.
So we migrate prunerepo project to `--location`.

Neither of those options are in DNF5, and we'll need it.

kontura:
DNF5 we definitely support both `--location` and queryformat `location` tag.
But yea it is still in the works, also tracked here: https://github.com/rpm-software-management/dnf5/issues/122

praiskup:
```
08:02:47 ~$ rpm -q dnf5 dnf5-plugins
dnf5-5.0.13-20230605010115.16.g632dd459.fc38.x86_64
dnf5-plugins-5.0.13-20230605010115.16.g632dd459.fc38.x86_64
08:02:52 ~$ dnf5 repoquery --location
Unknown argument "--location" for command "repoquery". Add "--help" for more information about the arguments.
```
Did you want to write `s/we definitely support/we definitely plan to support/`? :-) I'm just confused.

kontura:
oh sorry, yes. It was supposed to be: `s/we/will/`.

Issue #497, opened 2023-04-27
dnf download `--url`


j-mracek:
Additional resources:

DNF4 implementation - https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/download.py (between lines 112 and 125)

Requirements:
Each package will be represented only by one url

Problems or recommendations:
RPM location could contain the whole url (then don't merge it with url of mirror)
Skip mirrors with unsupported protocols by Curl (rsync).
Metalinks and mirrorlists should be taken as primary source for urls - DNF5 use that approach also.

Bavuett:
Hi, @j-mracek. I've worked on the issue but I'm encountering some difficulties. After reading the code in the [`download.cpp`](https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/download/download.cpp) file I figured out that I needed to print a combination of the values returned by the `get_baseurl()` and the `get_location()` functions. However, I've noticed that with the packages I tried (`code` and `python`) the base location is not included in their `repomd.xml` file which, as I understood from the code base, the function uses to get the URL.

As a draft, I'm currently printing the repo's mirrors along with the location that `get_location()` returns. This does the job with packages such as `python` which specify different mirrors for their repos and the URLs are correct too, but this doesn't work with other packages such as `code`.

Am I missing something? What should I change in [my code](https://github.com/Bavuett/dnf5/commit/0fe50a8c576f6ca796626ed0766cd938b4e26090)?

j-mracek:
I updated the issue description with some additional information. Please ping me, if I forget to describe something.

I checked your code and I think it will generate all possible urls for each package. Usually users use the feature to download packages using a different downloaded, therefore they need only one url per package. If there will be more urls per package they will have a difficulty to known what is an alternative and what is a different package.

kontura:
@Bavuett are you still working on this?
The same functionality (`pkg->get_remote_location()`) is also needed for `repoquery --location` (https://github.com/rpm-software-management/dnf5/issues/589) so I was wondering what is the status.

kontura:
PR: https://github.com/rpm-software-management/dnf5/pull/664 add `get_remote_locations(...)` to `libdnf5::rpm::package`, this should be useful for this.

derickdiaz:
In dnf4, I'm I correct to believe it is only returning the first url that matches the schema supplied? I've attached the link where the code lives. 

[...]

Issue #366, opened 2023-03-16
Package object should implement remote_location()


bcl:
In libdnf4 the package object method `.remote_location()` returned the remote url where the package could be downloaded from. This is a core part of how osbuild works, so it would be very useful if libdnf5 could provide the same information.

I have worked around it temporarily in this code:
https://github.com/osbuild/osbuild-composer/pull/3332/commits/d6f4f66a5214249a8e342fc04d1cf26664dd1580

but note that this doesn't provide the final resolved url, it provides the intermediate url pointing to the mirror url used to setup the repo.

gotmax23:
This is also a blocker for [fedrq#20](https://todo.sr.ht/~gotmax23/fedrq/20).

For the record, here is dnf4's implemention:

https://github.com/rpm-software-management/dnf/blob/110c40194f23d9587b02dd221e0162332f66793c/dnf/package.py#L292-L303

https://github.com/rpm-software-management/dnf/blob/110c40194f23d9587b02dd221e0162332f66793c/dnf/repo.py#L636-L659

kontura:
PR: https://github.com/rpm-software-management/dnf5/pull/664

kontura:
We have added `get_remote_locations()` it returns a vector of urls.

PR #664, opened 2023-06-26
repoquery: add `--disable-modular-filtering` and `--location`

kontura:
For: https://github.com/rpm-software-management/dnf5/issues/589
For: https://github.com/rpm-software-management/dnf5/issues/366
It should also be used for: https://github.com/rpm-software-management/dnf5/issues/497

Unlike the dnf4 version the package now returns a `std::vector` of all its remote locations.
It also takes into account pkg baseurl (if there is one) which should ensure that the returned location is always the same one which dnf (librepo) would attempt to download first.

Tests update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1328

### DIFF
--- a/libdnf5/rpm/package.cpp
+++ b/libdnf5/rpm/package.cpp
@@ -33,6 +33,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include <fcntl.h>
 #include <librepo/checksum.h>
+#include <librepo/util.h>
 #include <unistd.h>
 
 #include <filesystem>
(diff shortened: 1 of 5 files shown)


libdnf5/rpm/package.cpp
```diff
@@ -300,6 +301,58 @@ std::string Package::get_location() const {
...
+    }
+
+    auto location = get_location();
+    if (location.find("://") != std::string::npos) {
+        return {location};
+    }
+
+    std::vector<std::string> remote_locations;
+
+    auto pkg_baseurl = get_baseurl();
```
j-mracek, line 320: I am not sure why we test exclusively `get_baseurl();` and then `auto repo_baseurls = get_repo()->get_baseurls();`.

May I ask you for explanation?

Idditionally I think that the order is not perfect one. I would recommend to add first mirrors (from metalinks, mirrorlist) and the last add baseurls.

There is a problem in librepo with order of alternatives. We are unable to handle together baseurls and mirrorlists/metalinks. When baseurl is set it use only baseurl. I think `baseurl` should be used as a last option not the first one to split downloads to multiple providers.

kontura: There are actually two baseurls:
- repo baseurl, configured in repo file. This is the base url for the whole repo (alternative repo source). You are correct about librepo I will switch the order of the loops.
- second is package baseurl, this can theoretically be different for each package. It looks like:
```xml
<package type="rpm">
  <name>CQRlib</name>
  <arch>src</arch>
  <version epoch="0" ver="1.1.1" rel="4.fc29"/>
  <checksum type="sha256" pkgid="YES">36eb2f6e63d3df64e61b14e570d35997a0f9357414ac1f6a637c48b971f968e5</checksum>
  <summary>ANSI C API for quaternion arithmetic and rotation</summary>
  <description>CQRlib is an ANSI C implementation of a utility library for quaternion
arithmetic and quaternion rotation math.</description>
  <url>http://cqrlib.sourceforge.net/</url>
  <size package="7780" installed="1628" archive="1888"/>
  <location xml:base="http://localhost:43223" href="src/CQRlib-1.1.1-4.fc29.src.rpm"/>    <-----------
  <format>
    <rpm:license>LGPLv2+</rpm:license>
    <rpm:vendor>Fedora Project</rpm:vendor>
    <rpm:group>System Environment/Libraries</rpm:group>
    <rpm:buildhost>77a43894e301</rpm:buildhost>
    <rpm:sourcerpm></rpm:sourcerpm>
    <rpm:header-range start="4504" end="6850"/>
    <rpm:provides>
      <rpm:entry name="CQRlib" flags="EQ" epoch="0" ver="1.1.1" rel="4.fc29"/>
    </rpm:provides>
  </format>
</package>
```

I was thinking that if a repo says the package is elsewhere it probably doesn't make sense to try to find it in the repo location. Therefore there is a separate check for package baseurl and if present it returns only the one location.

kontura:
Rebased to resolve a conflict.

include/libdnf5/repo/repo.hpp
```diff
@@ -327,6 +327,10 @@ class Repo {
     // @replaces libdnf:repo/Repo.hpp:method:Repo.getMirrors()
     std::vector<std::string> get_mirrors() const;
 
+    /// Returns baseurls of repository
+    /// @return A vector of baseurls
+    std::vector<std::string> get_baseurls() const { return config.get_baseurl_option().get_value(); };
```
j-mracek, line 332: I am curios whether we need that. If I am not mistaken we already has several option like that, but do we want to make the same pattern for all configuration?

On the other hand it is not wrong and it might  be useful if the option is called often.

kontura: It is not needed, I have dropped it.
We can always add it later if we want.

j-mracek:
LGTM

PR #676, opened 2023-06-28
nevra: Fix evrcpm to work with empty epoch

m-blaha:
The evrcmp template function can work also with Nevra objects which can
have empty epoch. Comparing such object to another object with epoch=0
produces incorrect results.


jan-kolarik (approved):
LGTM

PR #679, opened 2023-06-28
repoquery: update and unify repoquery manpage

kontura:

It also fixes a typo in userinstalled option description.

### DIFF
--- a/doc/commands/repoquery.8.rst
+++ b/doc/commands/repoquery.8.rst
@@ -40,69 +40,231 @@ Options
 =======
 
 ``--advisories=ADVISORY_NAME,...``
-    | Consider only content contained in advisories with specified name.
+    | Limit to packages in advisories with specified name.
     | This is a list option.
     | Expected values are advisory IDs, e.g. `FEDORA-2201-123`.
 
 ``--advisory-severities=ADVISORY_SEVERITY,...``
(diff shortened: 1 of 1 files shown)


dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -271,7 +271,7 @@ void RepoqueryCommand::set_argument_parser() {
     recent = std::make_unique<libdnf5::cli::session::BoolOption>(
         *this, "recent", '\0', "Limit to only recently changed packages.", false);
     installonly = std::make_unique<libdnf5::cli::session::BoolOption>(
-        *this, "installonly", '\0', "Limit to installed installonly packages.", false);
```
m-blaha, line 274: I think that's not a typo, but on purpose to emphasize that only `installed` packages are considered.

kontura: I see, you are correct. I will drop the commit.
Do you have any idea why is that? The filter would make sense to me also on available packages. Plus it can be combined with `--installed`.

m-blaha: I think the filter was implemented according to dnf4 version which behaves the same.
But you are right I guess.

j-mracek:
LGTM

PR #680, opened 2023-06-29
Prevent repositories from being loading multiple times

### LINKED ISSUE #439: Disallow loading repositories multiple times
This is a follow-up to #412 where the issue regarding [`libdnf::repo::RepoSack::update_and_load_enabled_repos()`](https://github.com/rpm-software-management/dnf5/blob/ce88399e3da1fcc6d555fefadc9037628674a325/include/libdnf/repo/repo_sack.hpp#L110) was discovered. When calling this function multiple times then the same packages with different solver ids are loaded into sack, which is undesirable.

softwaresale:
- Make some methods private (as specified in the issue)
- Add static flag to `RepoSack::update_and_load_enabled_repos` to ensure that it cannot be called more than once.

Fixes #439


libdnf5/repo/repo_sack.cpp
```diff
@@ -468,6 +468,12 @@ void RepoSack::update_and_load_repos(libdnf5::repo::RepoQuery & repos, bool impo
 
 
 void RepoSack::update_and_load_enabled_repos(bool load_system) {
+    static bool called = false;
```
softwaresale, line 471: This might be a coarse way of solving this. I also considered adding a flag on the class, but I wasn't sure if repositories can only be loaded once per program invocation, or only once for this given `RepoSack` instance. This can definitely be changed

jan-kolarik:
@softwaresale Thank you for your PR. However, the situation regarding the user workflows appears to be unclear and requires further discussion. Since it's currently vacation time, I guess that we will have more information in about 3 weeks.

jan-kolarik:
I am bringing some new info regarding this PR.

We want to keep only one way for common API users to load repositories, otherwise it will cause problems related to solver, modular filtering etc.

Therefore the only official way should be the `update_and_load_enabled_repos` method. It should be guarded to be run at most once in its instance. Please change the current guard to class member and also use the `libdnf_user_assert` macro which allows the bindings users to handle this exception when using the API.

The other `update_and_load_repos` method should be marked as experimental, f.e. with @warning tag, because we cannot remove / deprecate it yet as we still have a few tests / 3rd party use cases that need it.

Also we want to add a per-repo loading guard to avoid loading same repo instances multiple times when f.e. called using the experimental API. This could be done by adding the guard into the `Repo::load()` method, so it will check if it was already called for a given repo instance and eventually immediately return.

If you want any further info or you don't have capacity to work on this anymore, please let me know. You can add the mentioned additional functionalities as new commits to have the individual changes separated and documented in the git.

softwaresale:
I believe I made the appropriate changes, but please correct me if I'm wrong. I read [in a comment](https://github.com/rpm-software-management/dnf5/issues/439#issuecomment-1547665152) that I should make `Repo::load` private. However, that is causing a compilation error. Guidance on this?

Also, I have not added any unit tests. Would you like me to write something?

include/libdnf5/repo/repo.hpp
```diff
@@ -385,6 +372,21 @@ class Repo {
     std::unique_ptr<SolvRepo> solv_repo;
 
     WeakPtrGuard<Repo, false> data_guard;
+
+    bool loaded{false};
```
jan-kolarik, line 376: Please move this member together with other privates, f.e. right below `bool expired{false};`

include/libdnf5/repo/repo.hpp
```diff
@@ -385,6 +372,21 @@ class Repo {
...
+    void download_metadata(const std::string & destdir);
+
+    /// Loads the repository objects into sacks.
+    ///
+    /// Also writes the libsolv's solv/solvx cache files.
+    void load();
+
+    /// Append a rpm database into the system repository. The type of the repo must be Type::SYSTEM.
+    // TODO(jrohel) this will add packages with conflicting rpmdb ids, which will break some operations
+    void load_extra_system_repo(const std::string & rootdir);
```
jan-kolarik, line 389: Please rebase your branch against current `main` state. These three methods: `download_metadata()`, `load()` and `load_extra_system_repo()` are already marked as deprecated or not API, so we don't have to do anything there. As we have already moved to stable API we cannot easily just drop public methods.

include/libdnf5/repo/repo_sack.hpp
```diff
@@ -165,6 +165,7 @@ class RepoSack : public sack::Sack<Repo> {
 
     repo::Repo * system_repo{nullptr};
     repo::Repo * cmdline_repo{nullptr};
+    bool repos_updated_and_loaded{false};
```
jan-kolarik, line 168: OK, just in this file also please mark the `update_and_load_repos` method as experimental, using f.e. the @warning tag as described in my https://github.com/rpm-software-management/dnf5/pull/680#issuecomment-1665187488.

jan-kolarik:
> Also, I have not added any unit tests. Would you like me to write something?

It would be nice if we add at least testing of the `update_and_load_enabled_repos`. Check that calling it for second time causes the exception. The proper place should be `test/libdnf5/repo/test_repo.cpp`. If you don't feel adding the test there, I can do it afterwards, no problem...

softwaresale:
I rebased on my _own_ `main` and not `upstream` main. Whoops. The project _should_ build now...

Also, I misread your comment. I'll fix that test commit.

jan-kolarik (approved):
LGTM, thank you very much for your contribution!

test/libdnf5/repo/test_repo.cpp
```diff
@@ -138,3 +138,54 @@ void RepoTest::test_load_repo_nonexistent() {
     repos.filter_id(repoid);
     CPPUNIT_ASSERT_THROW(repo_sack->update_and_load_repos(repos), libdnf5::repo::RepoDownloadError);
 }
+
+void RepoTest::test_load_repo_twice() {
```
jan-kolarik, line 142: Just last request to change from me. We can drop this test. Unfortunately, AFAIK we cannot easily test through API that the given repo is already loaded. The current test is checking some counters for loading metadata, but these are skipped anyway even when calling repo->load() multiple times. So better just delete this test and keep only the `test_update_and_load_enabled_repos_twice_fails` here.

PR #683, opened 2023-06-29
changelogs: Fix `count` variable type

pkratoch:
The variable `count` in `print_changelogs` used to be `int32_t`, but was changed to `int64_t`. Because of that, the build on i686 arch fails with:

```
error: conversion from 'int64_t' {aka 'long long int'} to
'__gnu_cxx::__normal_iterator<libdnf5::rpm::Changelog*,
std::vector<libdnf5::rpm::Changelog> >::difference_type' {aka 'int'} may
change value [-Werror=conversion]
```

---

The issue blocks build on Fedora rawhide: https://koji.fedoraproject.org/koji/taskinfo?taskID=102735120

### DIFF
--- a/dnf5-plugins/changelog_plugin/changelog.cpp
+++ b/dnf5-plugins/changelog_plugin/changelog.cpp
@@ -117,8 +117,8 @@ void ChangelogCommand::configure() {
 void ChangelogCommand::run() {
     auto & ctx = get_context();
 
-    std::pair<libdnf5::cli::output::ChangelogFilterType, std::variant<libdnf5::rpm::PackageQuery, int64_t>> filter = {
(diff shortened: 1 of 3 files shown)


jan-kolarik:
It looks it should work, also on `i686`. We are currently running CI only against `x86_64` unfortunately. This could be probably improved.

Also one other thing. The `libdnf5-cli` are public headers. I am not sure if the `changelogs.hpp` header will be in the future also wrapped using SWIG. In that case we would need to get rid of the `std::variant` from the interface and I think it would be even better for readability, because it seems the logic there could be easily refactored to several overloads...

jan-kolarik (approved):
LGTM, created follow-up ticket for mentioned SWIG issue.

PR #686, opened 2023-06-29
repoquery: Do not treat packages with GROUP reason as userinstalled

ignatenkobrain:


Conan-Kudo (approved):

j-mracek (changes requested):
I do not agree with the change because it will break autoremove functionality. The filter is used in unneeded filter. The change does not provide any description why the change is required and what king of issue it resolves.

Conan-Kudo:
Hmm, that reasoning makes sense. To be clear, I thought `GROUP` basically was the equivalent of dep-installed for packages that users installed by requesting a group. If it's not that, then this change makes no sense.

ignatenkobrain:
`dnf5 repoquery --userinstalled` shows packages from GROUPs.

j-mracek:
The command that you are looking for is:
`sudo dnf5 repoquery --installed --qf '%{name}.%{arch} & %{reason}\n' | grep '& User'`
I am really sorry but the PR cannot be merged because it will break DNF5 behavior.

Issue #617, opened 2023-06-14
Using a different security policy from DNF4?


praiskup:
When using `--installroot` with `dnf-3`, the transaction fails:
```
...
Importing GPG key 0xC105B9DE:
 Userid     : "CentOS-6 Key (CentOS 6 Official Signing Key) <centos-6-key@centos.org>"
 Fingerprint: C1DA C52D 1664 E8A4 386D BA43 0946 FCA2 C105 B9DE
 From       : /usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-6
error: Certificate 0946FCA2C105B9DE:
  The certificate is expired: The primary key is not live
Key import failed (code 2). Failing package is: less-436-13.el6.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-6
...
```

But with dnf5 it succeeds with just:

```
...
[93/93] Total                                                                                                                                                                                                                                                                                                                                                                                                                                       100% |   8.6 MiB/s |  54.8 MiB |  00m06s
Verifying PGP signatures
...
```

Related [fedora devel discussion](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/KOR3HE2VHHIPDBLDLXTYRMON6JQXCHMW/).

dnf.conf file:

```
[main]
keepcache=1
debuglevel=2
reposdir=/dev/null
logfile=/var/log/yum.log
retries=20
obsoletes=1
gpgcheck=0
assumeyes=1
syslog_ident=mock
syslog_device=
metadata_expire=0
mdpolicy=group:primary
best=1
protected_packages=
user_agent=Mock (epel-6-x86_64-bootstrap; x86_64)

# repos
[base]
name=CentOS-$releasever - Base
enabled=1
baseurl=https://vault.centos.org/6.10/os/$basearch/
failovermethod=priority
gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-6
gpgcheck=1
skip_if_unavailable=False

[updates]
name=CentOS-$releasever - Updates
enabled=0
baseurl=https://vault.centos.org/6.10/updates/$basearch/
failovermethod=priority
gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-6
gpgcheck=1
skip_if_unavailable=False

[extras]
name=CentOS-$releasever - Extras
[...]

Issue #457, opened 2023-04-14
Flaky gpg CI test


m-blaha:
`dnf/gpg.feature:47  Install subkey signed package with masterkey signed dependency`

Sometimes it fails to import the subkey. I didn't investigate whether the issue is in the test suite or in dnf5.

```
2023-04-12T17:17:20.6554596Z       Verifying PGP signatures
2023-04-12T17:17:20.6554905Z       Importing PGP key 0x1F8F17F7
2023-04-12T17:17:20.6555244Z        UserId     : "dnf-ci-gpg"
2023-04-12T17:17:20.6555599Z        Fingerprint: 201362D5BD4B1EFA0B669CBFC21DAB591F8F17F7
2023-04-12T17:17:20.6556146Z        From       : file:///opt/ci/dnf-behave-tests/fixtures/gpgkeys/keys/dnf-ci-gpg/dnf-ci-gpg-public
2023-04-12T17:17:20.6556547Z       Key imported successfully.
2023-04-12T17:17:20.6556842Z       Importing PGP key 0x75195502
2023-04-12T17:17:20.6557210Z        UserId     : "dnf-ci-gpg-subkey"
2023-04-12T17:17:20.6557580Z        Fingerprint: B99BAD0A41D440A26C51931DF895092F75195502
2023-04-12T17:17:20.6558137Z        From       : file:///opt/ci/dnf-behave-tests/fixtures/gpgkeys/keys/dnf-ci-gpg-subkey/dnf-ci-gpg-subkey-public
2023-04-12T17:17:20.6558560Z       [36m[1m
2023-04-12T17:17:20.6558879Z       Last Command stderr:[90m
2023-04-12T17:17:20.6559506Z       Failed to import public key "file:///opt/ci/dnf-behave-tests/fixtures/gpgkeys/keys/dnf-ci-gpg-subkey/dnf-ci-gpg-subkey-public" to rpmdb.
2023-04-12T17:17:20.6560472Z       GPG check for package "filesystem-3.9-2.fc29.x86_64" (/tmp/dnf_ci_installroot_lq6jgqzq/var/cache/dnf/dnf-ci-gpg-cdd86aa09da6d53e/packages/filesystem-3.9-2.fc29.x86_64.rpm) from repo "dnf-ci-gpg" has failed: import of the key didn't help, wrong key?
2023-04-12T17:17:20.6561046Z       Signature verification failed
```

m-blaha:
[...]

PR #697, opened 2023-07-05
gpgcheck security and usability fixes

evan-goode:
As enumerated by Panu here: https://github.com/rpm-software-management/dnf5/issues/617#issuecomment-1592633793:

- [x] The gpgcheck option in the main config should not override the per-repo setting.
- [x] DNF 5 should not print the misleading message "Verifying PGP signatures" to the console when in fact GPG checking is off.
  - This message is not present as of bfedca82d84d5588f4f57142fe7953b520eedefb, but we may want to print a warning to the console anyway, ideally after the transaction is run so it's visible to the user
- [x] When key import fails, a more specific error message should be displayed. For the test case in #617, DNF 4 prints "The certificate is expired: The primary key is not live" whereas DNF 5 just shows "Failed to import public key". The underlying issue is that DNF 5 swallows output from rpm-sequoia, mentioned here: https://github.com/rpm-software-management/dnf5/issues/457.
- [x] Transaction should stop immediately after key import fails

Resolves https://github.com/rpm-software-management/dnf5/issues/617.

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -979,7 +979,7 @@ bool Transaction::Impl::check_gpg_signatures() {
                         signature_problems.push_back(
                             err_msg + import_repo_keys_result_to_string(ImportRepoKeysResult::ALREADY_PRESENT));
                         result = false;
-                        continue;
+                        break;
                     }
                     processed_repos.emplace(repo_id);
 
@@ -989,14 +989,17 @@ bool Transaction::Impl::check_gpg_signatures() {
                         if (check_again != libdnf5::rpm::RpmSignature::CheckResult::OK) {
                             signature_problems.push_back(err_msg + _("Import of the key didn't help, wrong key?"));
                             result = false;
+                            break;
                         }
                     } else {
                         signature_problems.push_back(err_msg + import_repo_keys_result_to_string(import_result));
                         result = false;
+                        break;
                     }
                 } else {
                     signature_problems.push_back(err_msg + rpm_signature.check_result_to_string(check_result));
                     result = false;
+                    break;
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


j-mracek:
It looks like that many tests fails on output check (stderr) where new warning is present (`Warning: skipped PGP checks for 1 package(s).`).

evan-goode:
> It looks like that many tests fails on output check (stderr) where new warning is present (`Warning: skipped PGP checks for 1 package(s).`).

Yeah, I want to get some feedback to make sure these changes are reasonable before rewriting all those tests to include the warning message.

The most important changes in this PR are 5ab4eece9a670c22907a0b0c0c63a3fdc3a8c421 (Don't allow main gpgcheck=0 to override repo config) and 01c98675afdb147fda8e4beedeba2be5e0be51cd (Include RPM logs in KeyImportError).

We may not want the warning when skipping GPG checks, which would certainly require a lot of changes to the tests.

libdnf5/base/transaction.cpp
```diff
@@ -981,7 +979,7 @@ bool Transaction::Impl::check_gpg_signatures() {
                         signature_problems.push_back(
                             err_msg + import_repo_keys_result_to_string(ImportRepoKeysResult::ALREADY_PRESENT));
                         result = false;
-                        continue;
+                        break;
```
m-blaha, line 982: I'm not sure about all these `breaks`. Doesn't it mean that we will return to the user only the first signature problem instead of accumulating all of them?

m-blaha: Oh, right. For larger transaction the new behaviour is definitely better because the root cause of the issue is more clearly visible.

m-blaha (approved):
Please, prepare a PR with tests adjustments before merging this, so it could be properly tested.

j-mracek:
I like the change, but I didn't spend enough time to provide an approval.

j-mracek:
When a patch for CI-dnf tests will be available I think we can merge it. Or from the second direction - the absence of patch of CI blocks a merge of this PR.

j-mracek (approved):

evan-goode:
> When a patch for CI-dnf tests will be available I think we can merge it. Or from the second direction - the absence of patch of CI blocks a merge of this PR.

Thanks. CI PR here: https://github.com/rpm-software-management/ci-dnf-stack/pull/1343

Issue #704, opened 2023-07-07
dnf history info


11raob:
This command provides me no output.  Is it implemented?

jn64:
Seems like in dnf5 you must specify a transaction ID (or range), unlike dnf4 which defaults to showing the last transaction without arg.

PR #713, opened 2023-07-12
history: info: when no tran id specified use last transaction

kontura:
Otherwise the output was empty which is not very user friendly.

Closes: https://github.com/rpm-software-management/dnf5/issues/704

### DIFF
--- a/dnf5/commands/history/history_info.cpp
+++ b/dnf5/commands/history/history_info.cpp
@@ -37,8 +37,15 @@ void HistoryInfoCommand::set_argument_parser() {
 }
 
 void HistoryInfoCommand::run() {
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

jan-kolarik (approved):

jan-kolarik:
Maybe we could also add the CI test case for this. I will try to add one.

kontura:
> Maybe we could also add the CI test case for this. I will try to add one.

I think there already is one in https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/history-info.feature I also wanted to look into but I forgot about it.

Issue #547, opened 2023-05-21
`BaseGoalTest::test_install_multilib_all` test fails on openSUSE Tumbleweed x86_64


Conan-Kudo:
While packaging dnf5-5.0.9 for openSUSE, I discovered that one of the tests fail: `BaseGoalTest::test_install_multilib_all`

Test output:

```
[  678s] BaseGoalTest::test_install_multilib_all
[  678s] Dnf log:
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] created directory(s) /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm mode 0755
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] created directory(s) /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm mode 0755
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] created directory(s) /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm mode 0755
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] created directory(s) /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm mode 0755
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] opening  db index       /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm/Packages.db mode=0x0
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] disabling fsync on database
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] opening  db index       /tmp/libdnf5_unittest.Wavua8/installroot/usr/lib/sysimage/rpm/Index.db mode=0x0
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] opening  db index       Name tag=1000
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] index Name needs creating
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] opening  db index       Basenames tag=1117
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] index Basenames needs creating
[  678s] 2023-05-21T22:07:48+0000 [6973] TRACE [rpm] opening  db index       Group tag=1016
[...]

PR #715, opened 2023-07-12
Set the implicitobsoleteusescolors flag for the solver pool

Conan-Kudo:
This ensures that RPM multi-arch package handling (commonly known as "multilib") consistently works.

This also resolves the issue with the multilib test in the test suite failing because libsolv was somehow not compiled with the flag enabled by default.

Fixes: #547


m-blaha (approved):
Thank you!

Issue #711, opened 2023-07-12
dnf5 make cache --refresh


11raob:
Run (as superuser or root)

# dnf makecache --refresh

Receive the following output---
root@machinename:~# dnf makecache  --refresh 
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot create directory: Not a directory [/var/cache/libdnf5/fastestmirror.cache/attrs]
Aborted (core dumped)

jan-kolarik:
Bugzilla ticket: https://bugzilla.redhat.com/show_bug.cgi?id=2222143.

PR #717, opened 2023-07-12
dnf5: Fix core dump on --refresh switch usage

m-blaha:
The current implementation blindly believes that each entry under the
cache dir is a directory representing a repository cache. Unfortunately
that's not always true, libdnf5 stores also files here (e.g.
fastestmirror.cache or temporary_files.toml).
In such case it leads to a coredump on attempt to create a
subdirectory `attrs` of this file.

Fixes: https://github.com/rpm-software-management/dnf5/issues/711

### DIFF
--- a/dnf5/commands/clean/clean.cpp
+++ b/dnf5/commands/clean/clean.cpp
@@ -22,6 +22,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include <libdnf5-cli/argument_parser.hpp>
 #include <libdnf5/repo/repo_cache.hpp>
+#include <libdnf5/utils/bgettext/bgettext-lib.h>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 
 #include <filesystem>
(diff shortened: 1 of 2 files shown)


dnf5/commands/clean/clean.cpp
```diff
@@ -131,6 +131,9 @@ void CleanCommand::run() {
     std::error_code ec;
     libdnf5::repo::RepoCache::RemoveStatistics statistics{};
     for (const auto & dir_entry : std::filesystem::directory_iterator(cachedir, ec)) {
+        if (!dir_entry.is_directory()) {
+            continue;
+        }
         libdnf5::repo::RepoCache cache(ctx.base.get_weak_ptr(), dir_entry.path());
 
         if (required_actions & CLEAN_ALL) {
```
evan-goode, line 139: Should exceptions from these cache operations also be caught?

evan-goode (approved):

dnf5/commands/clean/clean.cpp
m-blaha: The exception is caught, so dnf5 will not end up with core dump (all command.run() execution is inside a try/catch block). But you are right, it should be caught also here, otherwise the `clean` command execution will fail on the first inaccessible directory.

jan-kolarik:
Thanks!

Issue #566, opened 2023-05-26
Missing virtual provides for command


j-mracek:
DNF5 as well as dnf has modular command therefore therefore it is quite useful to mark packages that provides particular command by virtual provide.

In DNF4 we use `dnf-command(<command_name>)` and when user requests unknown command it suggest to install such a command using the virtual provide.

What about to use the sama schema for `dnf5`?

Because dnf5 provides DNF and man pages refers to dnf including examples we can use `dnf-command(<command_name>)`.

Or we can use `dnf5-command(<command_name>)`.

jkonecny12:
We are facing this issue on https://github.com/rhinstaller/anaconda/actions/runs/5372767043/jobs/9746479214#step:3:290 -- I'll fix it in our tests to use the package directly but would be great to have resolved soon to avoid another issues like this.

Also couldn't this be resolved based on the DNF version. For so it would always install the correct version? My point is that it will break stuff if you add a new provides. However, I'm not sure how this is implemented.

evan-goode:
I think the only option here is to use `dnf5-command(<command_name>)` if we must support having a DNF 4 and DNF 5 stack installed simultaneously.

If DNF 5 plugins provided `dnf-command(<command_name>)`, then imagine the following situation:
- User has `dnf5`, `python3-dnf` and `dnf-plugins-core` installed, but they're missing `dnf5-plugins`.
- User wants to use `dnf5 builddep`, but `dnf5 install dnf-command(builddep)` will do nothing since `dnf-plugins-core` already provides `dnf-command(builddep)`.

A possibility is to have DNF 4 plugins instead provide `dnf4-command(<command_name>)` in Fedora 39+, but that seems like it could be messy.

PR #718, opened 2023-07-12
Add provides for dnf5-command(<command-name>)

evan-goode:
This adds `Provide`s for `dnf5-command(<command-name>)` for all the commands in `dnf5`, and all the commands provided by `dnf5-plugins`.

Also adds a hint like we had for DNF 4, so when trying to run a missing command:

```
$ dnf foo
Unknown argument "foo" for command "dnf5". Add "--help" for more information about the arguments.
It could be a command provided by a plugin, try: dnf install dnf5-command(foo)
```

Resolves https://github.com/rpm-software-management/dnf5/issues/566

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -738,6 +738,14 @@ int main(int argc, char * argv[]) try {
                 }
             }
             std::cerr << ex.what() << _(". Add \"--help\" for more information about the arguments.") << std::endl;
+            if (auto * unknown_arg_ex = dynamic_cast<libdnf5::cli::ArgumentParserUnknownArgumentError *>(&ex)) {
+                if (unknown_arg_ex->get_command() == "dnf5") {
+                    std::cout << fmt::format(
+                                     "It could be a command provided by a plugin, try: dnf install dnf5-command({})",
+                                     unknown_arg_ex->get_argument())
+                              << std::endl;
+                }
+            }
             return static_cast<int>(libdnf5::cli::ExitCode::ARGPARSER_ERROR);
         }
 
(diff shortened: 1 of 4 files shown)


dnf5/main.cpp
```diff
@@ -738,6 +738,14 @@ int main(int argc, char * argv[]) try {
                 }
             }
             std::cerr << ex.what() << _(". Add \"--help\" for more information about the arguments.") << std::endl;
+            if (auto * unknown_arg_ex = dynamic_cast<libdnf5::cli::ArgumentParserUnknownArgumentError *>(&ex)) {
+                if (unknown_arg_ex->get_command() == "dnf5") {
+                    std::cout << fmt::format(
+                                     "It could be a command provided by a plugin, try: dnf install dnf5-command({})",
```
jan-kolarik, line 744: The warning is currently shown even when invalid parameter is passed to `dnf5`:
```
[root@c3c9bc2f0fde dnf5]# dnf5 --param
Unknown argument "--param" for command "dnf5". Add "--help" for more information about the arguments.
It could be a command provided by a plugin, try: dnf install dnf5-command(--param)
```

evan-goode: Ah thanks. Now I'm also checking that the argument doesn't begin with "-".

evan-goode:
Maybe we should wait for Jarda/others to weigh in on `dnf-command(<command-name>)` vs `dnf5-command(<command-name>)` before merging.

j-mracek:
The `dnf-command()` require is used by following packages therefore if we will use 'dnf-command()` there might be a benefit with following (not significant).
dnf5 repoquery --whatrequires='dnf-command(*)'
fedora-review
mkosi-initrd
plasma-drkonqi

negative part for `dnf-command()`
In next few Fedoras we can get installed DNF4 stack when command of dnf5 will be unknown and we will suggest to run dnf5 install 'dnf-command(<command-name>)' .

I think `dnf5-command(<command-name>) ` is a better choice.

j-mracek (approved):

Issue #720, opened 2023-07-13
dnf5 will not install an older kernel


AdamWill:
kernels being parallel-installable in Fedora, it should be fine to install one that is *older* than the latest one currently installed. Previous versions of dnf and yum let you do this, no problem.

Attempting to do it with dnf5, though, results in an error. Here, I have kernel-6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39 installed, and I want to install kernel-6.5.0-0.rc0.20230706gitc17414a273b8.12.fc39 alongside it:

    Transaction failed: Rpm transaction failed.
      - package kernel-modules-core-6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39.x86_64 (which is newer than kernel-modules-core-6.5.0-0.rc0.20230706gitc17414a273b8.12.fc39.x86_64) is already installed
      - package kernel-core-6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39.x86_64 (which is newer than kernel-core-6.5.0-0.rc0.20230706gitc17414a273b8.12.fc39.x86_64) is already installed
      - package kernel-modules-6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39.x86_64 (which is newer than kernel-modules-6.5.0-0.rc0.20230706gitc17414a273b8.12.fc39.x86_64) is already installed
      - package kernel-6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39.x86_64 (which is newer than kernel-6.5.0-0.rc0.20230706gitc17414a273b8.12.fc39.x86_64) is already installed

This happens whether you use `dnf install` or `dnf downgrade`.

AdamWill:
Running `dnf-3 install` instead of `dnf install`, the operation works fine.

jan-kolarik:
There is also related Bugzilla ticket: https://bugzilla.redhat.com/show_bug.cgi?id=2221308.

jan-kolarik:
Could you please provide a reproducer and dnf5 version? I tried to install the `6.5.0-0.rc1.20230711git3f01e9fed845.12.fc39.x86_64` kernel in Rawhide container and then downgrade to `6.3.12-200.fc38` with dnf5 `5.0.15` and everything seems OK.

jan-kolarik:
Maybe it is also connected with the currently running kernel version and new dnf5 having the `allow_vendor_change` option set to `False` by default.

m-blaha:
[...]

PR #723, opened 2023-07-14
rpm::Transaction: Enable installonly pkg downgrade

m-blaha:
To correctly filter out the
"package kernel-x (which is newer than kernel-y) is already installed"
RPM transaction problem, we need to determine whether a downgrade
operation is present in the transaction.

The current implementation relies on the presence of a "downgrade"
operation in the solver results. However, for install-only packages,
there is no "downgrade" operation; only an "install" operation.

This patch introduces detection to identify if an install-only package
of an older version is present in the transaction. It then sets ignored
problems in the RPM transaction accordingly.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2221308
Resolves: https://github.com/rpm-software-management/dnf5/issues/720

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1337

### DIFF
--- a/libdnf5/rpm/transaction.cpp
+++ b/libdnf5/rpm/transaction.cpp
@@ -24,6 +24,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5/base/transaction.hpp"
 #include "libdnf5/common/exception.hpp"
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):
Seems reasonable to me.

kontura (approved):

PR #725, opened 2023-07-14
Implement module disable and reset

pkratoch:
Resolves: https://github.com/rpm-software-management/dnf5/issues/649
Resolves: https://github.com/rpm-software-management/dnf5/issues/650

### DIFF
--- a/include/libdnf5/base/goal_elements.hpp
+++ b/include/libdnf5/base/goal_elements.hpp
@@ -102,7 +102,10 @@ enum class GoalAction {
     DOWNGRADE,
     REMOVE,
     RESOLVE,
-    REASON_CHANGE
+    REASON_CHANGE,
+    ENABLE,
+    DISABLE,
+    RESET
 };
 
 /// Convert GoalAction enum to user-readable string
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -218,6 +218,14 @@ void Goal::add_module_enable(const std::string & spec) {
     p_impl->module_specs.push_back(std::make_tuple(GoalAction::ENABLE, spec));
 }
 
+void Goal::add_module_disable(const std::string & spec) {
+    p_impl->module_specs.push_back(std::make_tuple(GoalAction::DISABLE, spec));
+}
+
+void Goal::add_module_reset(const std::string & spec) {
+    p_impl->module_specs.push_back(std::make_tuple(GoalAction::RESET, spec));
+}
+
 void Goal::add_install(const std::string & spec, const libdnf5::GoalJobSettings & settings) {
     p_impl->add_spec(GoalAction::INSTALL, spec, settings);
 }
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -534,6 +534,15 @@ GoalProblem Goal::Impl::add_specs_to_goal(base::Transaction & transaction) {
             case GoalAction::REASON_CHANGE: {
                 libdnf_throw_assertion("Unsupported action \"REASON_CHANGE\"");
             }
+            case GoalAction::ENABLE: {
+                libdnf_throw_assertion("Unsupported action \"ENABLE\"");
+            }
+            case GoalAction::DISABLE: {
+                libdnf_throw_assertion("Unsupported action \"DISABLE\"");
+            }
+            case GoalAction::RESET: {
+                libdnf_throw_assertion("Unsupported action \"RESET\"");
+            }
         }
     }
     return ret;
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -103,7 +103,7 @@ public:
     GoalProblem resolve_group_specs(std::vector<GroupSpec> & specs, base::Transaction & transaction);
     void add_resolved_group_specs_to_goal(base::Transaction & transaction);
     void add_resolved_environment_specs_to_goal(base::Transaction & transaction);
-    GoalProblem add_module_specs_to_goal(base::Transaction & transaction);
+    void add_module_specs_to_goal();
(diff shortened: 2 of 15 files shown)


pkratoch:
Note: I'm not sure how module disable should handle dependencies:

- If module A requires module B and both are enabled, and I attempt to disable module A, only module A gets disabled.
- If module A requires module B and both are enabled, and I attempt to disable module B, dnf reports an error with resolving modules.

It's the same for dnf4. If it's not an expected behavior, I propose to file an issue.

test/libdnf5/module/test_module.cpp
```diff
@@ -653,3 +655,112 @@ void ModuleTest::test_module_enable() {
...
+}
+
+
+void ModuleTest::test_module_disable_enabled() {
+    add_repo_repomd("repomd-modules");
+
+    // Set state of module berries to ENABLED
+    (base.*get(priv_impl()))
+        ->get_system_state()
+        .set_module_state("berries", libdnf5::system::ModuleState({"main", ModuleStatus::ENABLED, {}}));
```
pkratoch, line 702: I am accessing the system state here and changing the status of one module. You may notice the system state is accessed later as well. But for some reason, if I use the same instance for the system state, the state change doesn't propagate correctly. I will investigate this further, but could it be related to the private accessor hack?

kontura:
> Note: I'm not sure how module disable should handle dependencies:
> 
>     * If module A requires module B and both are enabled, and I attempt to disable module A, only module A gets disabled.
> 
>     * If module A requires module B and both are enabled, and I attempt to disable module B, dnf reports an error with resolving modules.
> 
> 
> It's the same for dnf4. If it's not an expected behavior, I propose to file an issue.

I think this recent https://bugzilla.redhat.com/show_bug.cgi?id=2216407 is related.

libdnf5/module/module_sack.cpp
```diff
@@ -739,6 +739,24 @@ ModuleSack::resolve_active_module_items() {
...
 
 
+ModuleQuery module_spec_to_query(BaseWeakPtr & base, const std::string & module_spec) {
+    auto nsvcaps = Nsvcap::parse(module_spec);
+
+    ModuleQuery query(base, true);
+    for (auto & nsvcap : nsvcaps) {
+        ModuleQuery nsvcap_query(base, false);
+        nsvcap_query.filter_nsvca(nsvcap);
+        query |= nsvcap_query;
```
j-mracek, line 749: May I ask you why you would like to use all possibilities as a candidate? For RPM we use the first valid candidate form that match. 
Practically I think there will be not real difference in result, because module spec parsing is more exact.

libdnf5/base/goal.cpp
```diff
@@ -534,11 +542,70 @@ GoalProblem Goal::Impl::add_specs_to_goal(base::Transaction & transaction) {
...
+            case GoalAction::ENABLE:
+                module_sack.p_impl->enable(spec);
+                break;
+            case GoalAction::DISABLE:
+                module_sack.p_impl->disable(spec);
+                break;
+            case GoalAction::RESET:
+                module_sack.p_impl->reset(spec);
+                break;
+            case GoalAction::INSTALL:
```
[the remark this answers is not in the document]
pkratoch: Yeah, it won't have the same format as the messages thrown in `add_specs_to_goal`, but I think it would be better. Changed it.

libdnf5/module/module_sack.cpp
pkratoch: Oh, right, I didn't realize it's sufficient to take the first candidate only. I changed it.

libdnf5/module/module_sack.cpp
```diff
@@ -739,6 +739,24 @@ ModuleSack::resolve_active_module_items() {
...
+    }
+
+    if (query.empty()) {
+        throw NoModuleError(M_("No such module: {}"), module_spec);
```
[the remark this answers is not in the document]
pkratoch: To fail when no module is found for given argument. Do you have another approach in mind?

include/libdnf5/base/goal_elements.hpp
```diff
@@ -102,7 +102,10 @@ enum class GoalAction {
     DOWNGRADE,
     REMOVE,
     RESOLVE,
-    REASON_CHANGE
+    REASON_CHANGE,
+    ENABLE,
```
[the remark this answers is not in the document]
pkratoch: Hm, I think `ENABLE` would be worse if we had another thing in the future that would get enabled/disabled but for some reason, it couldn't use the same action. But I can't think of any such scenario.

test/libdnf5/module/test_module.cpp
j-mracek: I am sorry, I have no idea

j-mracek: I can only guess that if the modification are not propagated, it sounds like that changes are performed on copy instance.

j-mracek:
LGTM

j-mracek:
I am sorry, am still looking at the PR.

j-mracek:
There is a problem with handling of multiple incorrect arguments where only the firs one is reported:
```
dnf5 module reset nodej adasd
Repositories loaded.
No such module: nodej
```
The issue is caused by raising an error: `throw NoModuleError(M_("No such module: {}"), module_spec)`

libdnf5/module/module_sack.cpp
```diff
@@ -739,6 +739,18 @@ ModuleSack::resolve_active_module_items() {
 }
 
 
+ModuleQuery module_spec_to_query(BaseWeakPtr & base, const std::string & module_spec) {
```
j-mracek, line 742: I guess that the function should be static or in anonymous name space. Or did I overlooked something?

pkratoch: Fixed.

j-mracek:
LGTM

PR #732, opened 2023-07-17
Remove unneeded unused configuration priority

j-mracek:


Conan-Kudo:
Aren't we finally going to add support for drop-in config files?

j-mracek:
Yes we are adding it, but the implementation that we planned does not need that priority. The drop in support is supposed be available in a week or weeks.

Conan-Kudo (approved):
👍🏾

PR #733, opened 2023-07-17
Allow obsoletion of protected packages

evan-goode:
libdnf5 companion to https://github.com/rpm-software-management/libdnf/pull/1610

### DIFF
--- a/libdnf5/rpm/solv/goal_private.cpp
+++ b/libdnf5/rpm/solv/goal_private.cpp
@@ -580,12 +580,15 @@ libdnf5::GoalProblem GoalPrivate::protected_in_removals() {
     auto & pool = get_rpm_pool();
 
     libdnf5::solv::SolvMap pkg_remove_list(pool->nsolvables);
-    for (auto index = 0; index < removes.size(); ++index) {
-        pkg_remove_list.add_unsafe(removes[index]);
+    for (const auto & remove : removes) {
+        pkg_remove_list.add_unsafe(remove);
     }
-    for (auto index = 0; index < obsoleted.size(); ++index) {
-        pkg_remove_list.add_unsafe(obsoleted[index]);
-    }
+    // We want to allow obsoletion of protected packages, so we do not consider
+    // obsoletes here, only removes. Previously, obsoletion of protected
+    // packages was disallowed, but there needed to be some mechanism for
+    // obsoleting/swapping a protected package, such as to obsolete `dnf` in
+    // favor of `dnf5`. Obsoleting a package is much harder to do accidentally
+    // than removing it.
 
     libdnf5::solv::SolvMap protected_pkgs(pool->nsolvables);
     if (protected_packages) {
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Could you please put the details in the commit message instead of referring to another PR?

evan-goode:
> Could you please put the details in the commit message instead of referring to another PR?

Thanks, I added some details to the commit message.

j-mracek:
This change might be problematic, because you can remove protected package in two steps `protected` -> `obsoleted` -> `removed`. But I agree to try it. It might be interesting to search in dnf/yum commit history when and why dnf/yum started with protection from obsoleting packages.

m-blaha:
I didn't find any particular reason - the obsoleted packages are just there from the beginning.

Here is the forefather of the current implementation - https://github.com/rpm-software-management/yum/commit/ab34fabf5ce21ea63aa989a65aad06a71fd604dc .

And here is how the protected packages yum plugin looked like when it was moved from plugins to core: https://github.com/rpm-software-management/yum-utils/commit/610ef70f88ea9bf72eaa129c8da83e9cc421d23b

evan-goode:
Ah, this change would unfortunately break the way we protect the running kernel from being removed. The failing test is:
```
  @bz1698145
  Scenario: Running kernel is protected against obsoleting                                     # dnf/protect-running-kernel.feature:44
```
I think it would be best to find another way to allow upgrading DNF 4 to DNF 5 while keeping DNF 4 "protected".

evan-goode:
Per @j-mracek's recommendation, I am reopening this and treating the running kernel as a special case.

So now with this patch, obsoleting protected packages is allowed, except for the running kernel, which cannot be removed or obsoleted.

I think this approach is reasonable; there should be _some_ way for the distribution to "unprotect" packages.

One use case that might be broken by this change is a user protecting a package to prevent it from being obsoleted by something else in the distribution. But that's a problem better suited for `excludepkgs`.

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -580,11 +580,8 @@ libdnf5::GoalProblem GoalPrivate::protected_in_removals() {
     auto & pool = get_rpm_pool();
 
     libdnf5::solv::SolvMap pkg_remove_list(pool->nsolvables);
-    for (auto index = 0; index < removes.size(); ++index) {
-        pkg_remove_list.add_unsafe(removes[index]);
-    }
-    for (auto index = 0; index < obsoleted.size(); ++index) {
-        pkg_remove_list.add_unsafe(obsoleted[index]);
+    for (const auto & remove : removes) {
```
j-mracek, line 583: May I ask you to add a comment that explains why ignoring obsoleted packages is intended.

j-mracek: The rest looks good to me.

evan-goode: Good call, I added a comment.

PR #737, opened 2023-07-19
history info: add groups and environments to ouput

kontura:
It also updates storing groups/environments to db with id and repoid.

Tests update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1338

### DIFF
--- a/include/libdnf5-cli/output/transactioninfo.hpp
+++ b/include/libdnf5-cli/output/transactioninfo.hpp
@@ -22,6 +22,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #define LIBDNF5_CLI_OUTPUT_TRANSACTIONINFO_HPP
 
 #include "libdnf5-cli/output/key_value_table.hpp"
+#include "libdnf5-cli/tty.hpp"
 
 #include <libdnf5/transaction/transaction.hpp>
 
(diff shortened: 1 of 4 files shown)


libdnf5/transaction/transaction.cpp
```diff
@@ -170,13 +170,17 @@ void Transaction::fill_transaction_environments(
...
         auto environment = tsenv.get_environment();
+        new_env.set_environment_id(environment.get_environmentid());
         new_env.set_name(environment.get_name());
         new_env.set_translated_name(environment.get_translated_name());
         libdnf5::comps::PackageType package_types{libdnf5::comps::PackageType::MANDATORY};
         if (tsenv.get_with_optional()) {
             package_types |= libdnf5::comps::PackageType::OPTIONAL;
         }
         new_env.set_package_types(package_types);
+        if (tsenv.get_environment().get_repos().begin() != tsenv.get_environment().get_repos().end()) {
```
evan-goode, line 181: Looks good. I would use `tsenv.get_environment().get_repos().empty()` here and below.

evan-goode (approved):

libdnf5/transaction/transaction.cpp
kontura: I like it, thanks!
Updated.

evan-goode (approved):

PR #739, opened 2023-07-19
dnf5: Fix printing advisories for the running kernel

m-blaha:
The main problem was that --contains-pkgs values for default
(--available) behavior need to be applied after running kernel packages are
injected to packages query.

Unfortunately the problem was not caught, because the test for printing
advisories for the running kernel was also incorrect.
Fix for the test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1339


kontura:
I see.
Thank you for properly fixing this!

The failing test is unrelated. 
(I am looking into it)

PR #741, opened 2023-07-20
config: Add support for included configuration files

jrohel:
The format of the "dnf.conf" configuration file for libdnf5 is compatible with dnf4.
It is based on the INI format. It is a simple text file format that contains "key=value" pairs organized into sections, blank lines, and comments.

This extension adds support for included configuration files. It introduces a special comment format for this - `#!include_config <absolute_path>`.
From the point of view of the INI parser and the original dnf4/yum, this is a comment.

Also adds `#!include "/etc/dnf/libdnf5.conf.d/*.conf"` to the distributed "dnf.conf".

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -293,6 +293,7 @@ Package management library.
 %files -n libdnf5
 %if 0%{?fedora} > 38 || 0%{?rhel} > 9
 %config(noreplace) %{_sysconfdir}/dnf/dnf.conf
+%dir %{_sysconfdir}/dnf/libdnf5.conf.d
 %dir %{_sysconfdir}/dnf/vars
 %dir %{_sysconfdir}/dnf/protected.d
 %else
--- a/dnf5/config/etc/dnf/dnf.conf
+++ b/dnf5/config/etc/dnf/dnf.conf
@@ -1,3 +1,0 @@
-# see `man dnf.conf` for defaults and possible options
-
-[main]
--- a/include/libdnf5/conf/config_parser.hpp
+++ b/include/libdnf5/conf/config_parser.hpp
@@ -34,35 +34,35 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 namespace libdnf5 {
 
-/// Error accessing config file other than ENOENT; e.g. we don't have read permission
-class InaccessibleConfigError : public Error {
-public:
-    using Error::Error;
-    const char * get_domain_name() const noexcept override { return "libdnf5"; }
-    const char * get_name() const noexcept override { return "InaccessibleConfigError"; }
-};
-
-/// Configuration file is missing
-class MissingConfigError : public Error {
-public:
-    using Error::Error;
-    const char * get_domain_name() const noexcept override { return "libdnf5"; }
-    const char * get_name() const noexcept override { return "MissingConfigError"; }
-};
-
-/// Configuration file is invalid
-class InvalidConfigError : public Error {
-public:
-    using Error::Error;
-    const char * get_domain_name() const noexcept override { return "libdnf5"; }
-    const char * get_name() const noexcept override { return "InvalidConfigError"; }
-};
-
 class ConfigParserError : public Error {
 public:
     using Error::Error;
     const char * get_domain_name() const noexcept override { return "libdnf5"; }
     const char * get_name() const noexcept override { return "ConfigParserError"; }
+};
+
+/// Error accessing config file other than ENOENT; e.g. we don't have read permission
+class InaccessibleConfigError : public ConfigParserError {
+public:
+    using ConfigParserError::ConfigParserError;
+    const char * get_domain_name() const noexcept override { return "libdnf5"; }
+    const char * get_name() const noexcept override { return "InaccessibleConfigError"; }
+};
+
+/// Configuration file is missing
+class MissingConfigError : public ConfigParserError {
+public:
+    using ConfigParserError::ConfigParserError;
+    const char * get_domain_name() const noexcept override { return "libdnf5"; }
+    const char * get_name() const noexcept override { return "MissingConfigError"; }
+};
+
+/// Configuration file is invalid
+class InvalidConfigError : public ConfigParserError {
+public:
+    using ConfigParserError::ConfigParserError;
+    const char * get_domain_name() const noexcept override { return "libdnf5"; }
+    const char * get_name() const noexcept override { return "InvalidConfigError"; }
 };
 
 class ConfigParserSectionNotFoundError : public ConfigParserError {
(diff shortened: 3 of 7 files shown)


m-blaha:
Nice! We should provide some tests for this feature though.

j-mracek:
May I ask you to reference original bugzilla?

j-mracek:
@jrohel One test is failing because the test expects empty stderr, but it contains `repo_comps: Document is empty`. May I ask you to investigate the test?

jrohel:
@j-mracek I think this is a problem with the recent `createrepo_c`. It is not the problem of this PR.

Conan-Kudo:
Can we default to `/usr/share/dnf/conf.d` being searched too?

jrohel:
@Conan-Kudo
libdnf5 reads distribution-specific configuration from `/usr/share/dnf5/libdnf.conf` (Another question if this is a good place). So if the distribution puts a file in this place containing the line:
`#!include_config /usr/share/dnf/conf.d/*.conf` then achieves what you want.

I don't want it in this PR. But anyone can prepare a PR with the appropriate configuration.

jrohel:
Rebased to resolve conflict.

jrohel:
@keszybz 
I'm back from vacation and reading your post. So:

> Please just define a list of directories to include directly in the code. When this is made part of the configuration, things become much more complicated:

I was also thinking about simpler options. Hard-coded list of directories. In that case, it is necessary to specify a list of directories and their priorities. One would then expect the main configuration file "dnf.cong" to be loaded first, and then the configuration from the included directories. The last loaded value wins. But the next step now is to program the "config-manager". Now what? Which configuration file will "config-manager" edit. The main one (/etc/dnf/dnf.conf) ? And will it have less weight than the configuration in the included directories? Or the included directories are loaded first and then the main configuration file. What will the priorities be? There are many scenarios and solutions. In the current solution, the user can see in the main configuration file "dnf.conf" which directories are inluded, in which order and when. Which options are before the include and which are after. So "config-manager" can add the settings to the end of the "/etc/dnf/dnf.conf" file (after `#!include_config <absolute_path>`) and win. I'm not saying it's an ideal solution. I am open to change. But there is not much time and other solutions have to count on cooperation with "config-manager".

Personally, as a server administrator, I was used to the fact that if additional configuration was loaded from drop-in directories, I could clearly see it in the main configuration file as an include. Examples from Fedora workstation (and there are more on my PC and servers):
ipsec.conf: `include /etc/ipsec.d/*.conf`
krb5.conf: `includedir /etc/krb5.conf.d/`
ld.so.conf: `include ld.so.conf.d/*.conf`
logrotate.conf: `include /etc/logrotate.d`
my.cnf: `!includedir /etc/my.cnf.d`
dnsmasq.conf: `conf-file=/etc/dnsmasq.more.conf`
              `conf-dir=/etc/dnsmasq.d`
rsyslog.conf: `include(file="/etc/rsyslog.d/*.conf" mode="optional")`
ssh_config: `Include /etc/ssh/ssh_config.d/*.conf`
sshd_config: `Include /etc/ssh/sshd_config.d/*.conf`
sudoers: `#includedir /etc/sudoers.d`

> magic syntax with the include-in-a-comment must be added. This suddenly means that comments are not comments anymore, they become meaningful.

Yes it is true. I do not like it. It's a "hack" how to add it to the INI file without breaking the ini parses.
I'm not the first to do something like this. E.g. this in "/etc/sudoers":
```
## Read drop-in files from /etc/sudoers.d (the # here does not mean a comment)
#includedir /etc/sudoers.d
```

And the standard first line in scripts `#!/bin/sh` or another interpreter.

>the base file is needed for the drop-ins to work. A long-standing goal is for things to work without any fixed file in /etc. Also, the user cannot just do echo blah >/etc/dnf/dnf.conf, they need to remember to include the magic include syntax.

The user will write his changes to his own file in the included directory. That's why they were introduced. So the user does not need to edit the main configuration file. And it will have its changes in its own separate file. If the user prefers to use "echo blah >/etc/dnf/dnf.conf" -> it completely overwrites the existing "dnf.conf" with its content. In other words, it wants to have its completely new settings in this file instead of using the included directories. Does he want to use included directories?

> If I'm reading the patch correctly, this does a glob but doesn't do sorting. The ordering of files must be defined and must be not be locale-dependent.

The patch uses the `glob` function. According to the documentation, it sorts the file names alphabetically before returning them. Of course, I have no problem with adding sorting if the `glob` function sorts wrongly.

jrohel:
@keszybz 
The libeconf solution has its pros and cons. It would mean more intervention in the loading logic of the dnf5 configuration. That could probably still be done. However, it is necessary that the resulting solution is compatible with "config-manager". https://dnf-plugins-core.readthedocs.io/en/latest/config_manager.html

In the old DNF it was simple. No drop-in directories, one configuration file. But the user wants a drop-in directory in dnf5 and at the same time the "config-manager" functionality is preserved.

If I make it compatible with libeconf, I can think of one solution. The configuration manager writes its file to the drop-in directory and ensures that it is loaded last. Will the file be named "zzz-config-manager.conf" so that it is last in alphabetical order?

keszybz:
Re sorting: you are right. `glob()` sorts by default. The only caveat is that glob sorting is locale dependent, which is not good. (In particular, `a < A` in en_US.UTF-8 and `A < a` in C.UTF-8, so this is not just a theoretical matter. The issue can occur even with "normal" ASCII names.)

> So the user does not need to edit the main configuration file.

The goal is for things to work if the main configuration file is there and when it is removed (empty `/etc`, factory resets, etc.). By defining the include path in the configuration file, the possibility of having just having some drop-ins with no main configuration file is closed.

> `#includedir /etc/sudoers.d`

An important thing is that two+ levels of configuration drop-in locations should be supported. Both systemd and libeconf do this. For example, `{/usr/lib,/usr/local/lib,/etc,/run}/something.d/*.conf`. This means that the distro can provide configuration overrides in packages (using `/usr/lib`), and non-distro packages can use `/usr/local/lib`, and the local maintainer can use `/etc` but also `/run` for temporary configuration. The drop-ins in all locations should be globbed and sorted together and loaded in order.

This cannot really be expressed using the glob-in-a-comment approach.

> The configuration manager writes its file to the drop-in directory and ensures that it is loaded last. Will the file be named "zzz-config-manager.conf" so that it is last in alphabetical order?

Yeah, something like that. Right now there are no configuration drop-ins, so you can define a convention. Generally, numerical prefixes are nice because they provide an obvious ordering. For example, you can say that config-manager will use drop-in `99-config-manager.conf`, so all user configuration should use names `nn-*.conf` that sort earlier than that.

jrohel:
@keszybz 
OK. As I said before. The libeconf solution has its pros and cons. However, even this PR is not ideal.
I will prepare a second solution. New PR with hardcoded list of drop-in directories and behaviour similar to libeconf. Then we can choose.

keszybz:
> @keszybz OK. As I said before. The libeconf solution has its pros and cons. However, even this PR is not ideal. I will prepare a second solution. New PR with hardcoded list of drop-in directories and behaviour similar to libeconf. Then we can choose.

That'd be great. FWIW, I didn't want to suggest that using libeconf is preferred. It's a fairly heavy-weight solution, and open-coding this can be done in <100 lines.

jrohel:
A competing solution using a fixed drop-in directory list and file loading order inspired by the "libeconf" PR https://github.com/rpm-software-management/dnf5/pull/813 .

Issue #738, opened 2023-07-19
Non working distro-sync


praiskup:
I have `dnf5` installed from `dnf copr enable copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf5-unstable`:

    dnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64
    libdnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64

I disabled the Copr repository, and I fail to "downgrade" the package using the
distro-sync command

    $ dnf copr disable copr.fedorainfracloud.org/rpmsoftwaremanagement/dnf5-unstable
    $ dnf copr distro-sync dnf5

I can not even downgrade for that matter...

    $ dnf downgrade dnf5
    Problem: conflicting requests
      - cannot install both dnf5-5.0.13-2.fc38.x86_64 and dnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64
      - cannot install both dnf5-5.0.6-2.fc38.x86_64 and dnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64
      - problem with installed package 

(I have alias dnf=dnf5)

evan-goode:
Likely another issue caused by us changing the default value of `allow_vendor_change` to `false` (https://github.com/rpm-software-management/dnf5/issues/722, https://github.com/rpm-software-management/dnf5/issues/701, https://github.com/rpm-software-management/dnf5/pull/623).

Does it work with `--setopt=allow_vendor_change=true`?

j-mracek:
The question is whether we should revert `allow_vendor_change`  to `true`.
Basically there are several problems
- command line packages are refused to get installed when they have a different vendor
  - This could be fixed
- disabling copr repository results in refusing installation of packages from original repositories - skipping all upgrades
  - I thing it is not fixable, because it is the principle of the feature
- cryptic solver message
  - it is impossible to guess that the real issue is related to a different vendor of both packages
  - `- cannot install both dnf5-5.0.13-2.fc38.x86_64 and dnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64`

@Conan-Kudo @dcantrell @kontura @evan-goode @m-blaha @jrohel @pkratoch @inknos  I think it would be better to revert the default of `allow_vendor_change` for now to decrease pressure on DNF5 Fedora 39 release. What do you think?

m-blaha:
[...]

Issue #722, opened 2023-07-14
dnf update local file doesn't work as expected


rwmjones:
I have this installed now:
```
libisofs-1.5.6-1.fc38.x86_64
dnf5-5.0.15-3.fc39.x86_64
```
and I built a local update for `libisofs`, but it cannot be updated and the messages are weird:
```
$ sudo dnf update $PWD/x86_64/libisofs-1.5.6-2.fc39.x86_64.rpm 
Updating and loading repositories:
Repositories loaded.
Nothing to do.

$ sudo dnf update $PWD/x86_64/libisofs-1.5.6-2.fc39.x86_64.rpm --best
Updating and loading repositories:
Repositories loaded.
Nothing to do.

$ sudo dnf install $PWD/x86_64/libisofs-1.5.6-2.fc39.x86_64.rpm --best
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: problem with installed package 
  - cannot install both libisofs-1.5.6-1.fc38.x86_64 and libisofs-1.5.6-2.fc39.x86_64
  - cannot install both libisofs-1.5.4-5.fc38.x86_64 and libisofs-1.5.6-2.fc39.x86_64
  - conflicting requests
```

I suspect this is caused because the new package has impossible to satisfy dependencies, but it should say that (if that's the case).

rwmjones:
Actually my last paragraph is not correct.  The new package was installable using `rpm` directly:

```
$ sudo rpm -Uvh $PWD/x86_64/libisofs-1.5.6-2.fc39.x86_64.rpm 
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
Updating / installing...
   1:libisofs-1.5.6-2.fc39            ################################# [ 50%]
Cleaning up / removing...
   2:libisofs-1.5.6-1.fc38            ################################# [100%]
```

So I think dnf is wrong here.

rwmjones:
Also downgrading doesn't work, with equally weird message:
```
$ sudo dnf downgrade libisofs --releasever=39 
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: cannot install both libisofs-1.5.6-1.fc39.x86_64 and libisofs-1.5.6-2.fc39.x86_64
  - cannot install the best candidate for the job
  - problem with installed package 
```

(even `--nobest` doesn't work)

jan-kolarik:
[...]

Issue #701, opened 2023-07-07
dnf install *.rpm failing with cannot install both versions


juhp:
I built a newer version of a package locally (pkgconf) and tried to install it (ie update the version via dnf):

```shellsession
⬢fedora39$ sudo "/usr/bin/dnf" install "/home/petersen/pkgconf/RPMS/x86_64/pkgconf-1.9.5-1.fc39.x86_64.rpm" "/home/petersen/pkgconf/RPMS/x86_64/libpkgconf-1.9.5-1.fc39.x86_64.rpm" "/home/petersen/pkgconf/RPMS/noarch/pkgconf-m4-1.9.5-1.fc39.noarch.rpm" "/home/petersen/pkgconf/RPMS/x86_64/pkgconf-pkg-config-1.9.5-1.fc39.x86_64.rpm"
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem 1: problem with installed package 
  - cannot install both pkgconf-1.9.4-2.fc39.x86_64 and pkgconf-1.9.5-1.fc39.x86_64
  - conflicting requests
 Problem 2: problem with installed package 
  - cannot install both libpkgconf-1.9.4-2.fc39.x86_64 and libpkgconf-1.9.5-1.fc39.x86_64
  - conflicting requests
 Problem 3: problem with installed package 
  - cannot install both pkgconf-m4-1.9.4-2.fc39.noarch and pkgconf-m4-1.9.5-1.fc39.noarch
  - conflicting requests
 Problem 4: problem with installed package 
  - cannot install both pkgconf-pkg-config-1.9.4-2.fc39.x86_64 and pkgconf-pkg-config-1.9.5-1.fc39.x86_64
  - conflicting requests
```
This is regression compared to dnf-3: though it works with `--allowerasing`.

The #142 gives the impression that this is supposed to work?

kontura:
I think this might be because of the switch of `allow_vendor_change` default configuration to `false`.
https://dnf5.readthedocs.io/en/latest/changes.html#default-of-allow-vendor-change-configuration-option-changed-to-false

Can you try with `--setopt=allow_vendor_change=true`?

juhp:
I suspect you are right - need to retest...

PR #621, opened 2023-06-18
Modify default value for best configuration

j-mracek:
best=true is more secure and will allow us to discover and fix broken packages in distribution rather than skipping them in transaction. The value is easily overridden from commandline and configuration, therefore the change is not show stopper.

Discussion:
https://github.com/rpm-software-management/dnf5/discussions/564

PR #745, opened 2023-07-20
Revert default of allow_vendor_change to true

j-mracek:
We discovered several issues related to the feature.

Command line packages are refused to get installed when they have a different vendor.

Disabling copr repository results in refusing installation of packages and upgrades from original repositories - skipping all upgrades.

Cryptic solver message
 - it is impossible to guess that the real issue is related to a different vendor of both packages
 - cannot install both dnf5-5.0.13-2.fc38.x86_64 and dnf5-5.1.0-20230719020203.0.gc17c4f61.fc38.x86_64 repoquery info does not show vendor

Closes: #738
Closes: #701  
Closes: #722  

Requires revert of https://github.com/rpm-software-management/dnf5/pull/621

### DIFF
--- a/libdnf5/conf/config_main.cpp
+++ b/libdnf5/conf/config_main.cpp
@@ -210,7 +210,7 @@ class ConfigMain::Impl {
     OptionBool gpgkey_dns_verification{false};
     OptionBool obsoletes{true};
     OptionBool exit_on_lock{false};
-    OptionBool allow_vendor_change{false};
+    OptionBool allow_vendor_change{true};
     OptionSeconds metadata_timer_sync{60 * 60 * 3};  // 3 hours
     OptionStringList disable_excludes{std::vector<std::string>{}};
     OptionEnum<std::string> multilib_policy{"best", {"best", "all"}};  // :api


Conan-Kudo:
We definitely need to add some output for sticky vendor cases so that the behavior is obvious and people know what to do. Do we want to have a session of some kind to figure it out? I'd really like for us to have this in as default at some point soon, especially because vendor switching should be purposeful rather than implicit.

inknos:
> We definitely need to add some output for sticky vendor cases so that the behavior is obvious and people know what to do. Do we want to have a session of some kind to figure it out? I'd really like for us to have this in as default at some point soon, especially because vendor switching should be purposeful rather than implicit.

Hey @Conan-Kudo, I opened an issue to track that https://github.com/rpm-software-management/dnf5/issues/750

Meanwhile, I would go on with this PR to fix fedora 39 urgencies.

Conan-Kudo:
I'm fine with this, we should track for enabling by Fedora 40, though.

j-mracek:
> I'm fine with this, we should track for enabling by Fedora 40, though.

I think Fedora 40 is not realistic.

Issue #721, opened 2023-07-14
Cannot configure logdir outside of installroot


DaanDeMeyer:
I would like to use installroot and configure the logdir to be outside of it. Currently the logdir is unconditionally put inside of the installroot. Please make this configurable.

evan-goode:
Thanks, yes this would be nice to have. Currently `reposdir` and `varsdir` are treated as relative to the host system, not the installroot, when they are specified on the command line via `--setopt=reposdir=/path/to/reposdir`. A good solution might be to have the same behavior for the `logdir` option.

I opened https://github.com/rpm-software-management/dnf5/pull/754 which would allow you to set a logdir outside the installroot with `--setopt=logdir=/path/to/logdir`.

PR #754, opened 2023-07-20
Treat logdir relative to host when specified on command-line

evan-goode:
There should be a way to set a logdir outside the installroot [1].

Currently, when `--setopt=varsdir` and `--setopt=reposdir` are specified on the command line rather than in a config file, they are treated as relative to the host, not the installroot. This patch changes `logdir` to behave the same way.

[1] https://github.com/rpm-software-management/dnf5/issues/721

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -794,13 +794,13 @@ int main(int argc, char * argv[]) try {
             close(fd);
         }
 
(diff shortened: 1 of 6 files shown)


DaanDeMeyer:
This would work for me

kontura (approved):

PR #755, opened 2023-07-21
Add --no-autoremove option for remove command plus alias

j-mracek:
It is an option to prevent removal of unused dependencies.

The option was requested here: https://pagure.io/fesco/issue/3039

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1342


m-blaha (approved):
Thank you!

Issue #141, opened 2022-11-11
Implement info command


mcurlej:
- [x] **Command:** ```info```

- [x] **Aliases:** ```if```
  - drop the alias - people must be not lazy to write additional two letters

```dnf [options] info [<package-file-spec>...]```

Lists description and summary information about installed and available packages.

The info command limits the displayed packages the same way as the [list command](https://dnf.readthedocs.io/en/latest/command_ref.html#list-command-label).

This command by default does not force a sync of expired metadata. See also [Metadata Synchronization](https://dnf.readthedocs.io/en/latest/command_ref.html#metadata-synchronization-label).

m-blaha:
Implemented by https://github.com/rpm-software-management/dnf5/pull/298

PR #756, opened 2023-07-21
Document dropped "if" alias of info command

m-blaha:
It's confusing and does not save that many keystrokes.

With this the https://github.com/rpm-software-management/dnf5/issues/141 is resolved.


j-mracek:
LGTM

Issue #766, opened 2023-07-23
dnf install --advisory crash


mattiaverga:
Using dnf5.x86_64 5.1.0-1.fc39

```
$ sudo dnf install --refresh --advisory=FEDORA-2023-840c288dcc \*
[sudo] password di rpmbuild: 
Updating and loading repositories:
 Fedora - Rawhide - Developmental packages for th 100% |  20.7 KiB/s |  10.5 KiB |  00m01s
 Fedora rawhide openh264 (From Cisco) - x86_64    100% |   2.5 KiB/s | 989.0   B |  00m00s
Repositories loaded.
/usr/include/c++/13/bits/stl_vector.h:1125: constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = s_Solvable*; _Alloc = std::allocator<s_Solvable*>; reference = s_Solvable*&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
```

PR #778, opened 2023-07-26
Fix out-of-bounds access in Goal::Impl::add_install_to_goal

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/766

The issue was assuming that the `std::vector` `tmp_solvables` has at least one element. Sometimes it is empty, such as when installing with both (1) a glob pattern and (2) an advisory filter that filters out all packages matched by the glob, e.g. this command:

```dnf5 install --advisories=nonexistent \*```


evan-goode:
CI PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1348

kontura (approved):
Looks good.
Thanks!

PR #779, opened 2023-07-27
Use focusbest: prefer latest deps versions over smaller transactions

kontura:
I chose the name `--prefer-latest` but I would be happy to change it if there is a better name.
I think `--prefer-latest` is better than `focus-best` (used by libsolv) because it is more descriptive and doesn't confuse the behavior with the `--best` option.

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1350

### DIFF
--- a/libdnf5/rpm/solv/goal_private.cpp
+++ b/libdnf5/rpm/solv/goal_private.cpp
@@ -376,6 +376,9 @@ libdnf5::GoalProblem GoalPrivate::resolve() {
     libsolv_solver.set_flag(SOLVER_FLAG_ALLOW_VENDORCHANGE, vendor_change);
     libsolv_solver.set_flag(SOLVER_FLAG_DUP_ALLOW_VENDORCHANGE, vendor_change);
 
+    // Ensure the solver tries to use the latest versions of dependencies, even if it results in a bigger transaction
+    libsolv_solver.set_flag(SOLVER_FLAG_FOCUS_BEST, 1);
+
     if (libsolv_solver.solve(job)) {
         return libdnf5::GoalProblem::SOLVER_ERROR;
     }
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -103,7 +103,7 @@ Provides:       dnf5-command(makecache)
 
 %global libmodulemd_version 2.5.0
 %global librepo_version 1.15.0
-%global libsolv_version 0.7.21
+%global libsolv_version 0.7.25
 %global sqlite_version 3.35.0
 %global swig_version 4
 %global zchunk_version 0.9.11
--- a/libdnf5/CMakeLists.txt
+++ b/libdnf5/CMakeLists.txt
@@ -51,7 +51,7 @@ pkg_check_modules(LIBMODULEMD REQUIRED modulemd-2.0>=2.11.2)
 list(APPEND LIBDNF5_PC_REQUIRES "${LIBMODULEMD_MODULE_NAME}")
 target_link_libraries(libdnf5 PRIVATE ${LIBMODULEMD_LIBRARIES})
 
-pkg_check_modules(LIBSOLV REQUIRED libsolv>=0.7.21)
+pkg_check_modules(LIBSOLV REQUIRED libsolv>=0.7.25)
 list(APPEND LIBDNF5_PC_REQUIRES "${LIBSOLV_MODULE_NAME}")
 target_link_libraries(libdnf5 PRIVATE ${LIBSOLV_LIBRARIES})
 


j-mracek:
What about to create a configuration option for that?

About naming - `best` is more accurate from technical point of view. But users thing about `best` option like about candidates with the higher version but in some cases best candidate is a different candidate. I think that `latest` is not the right term, because latest build might be not the build with the highest version.

kontura:
> What about to create a configuration option for that?

Sure we can add an option, I just wasn't sure since the behavior is quite specific to just the install command for now.

> 
> About naming - `best` is more accurate from technical point of view. But users thing about `best` option like about candidates with the higher version but in some cases best candidate is a different candidate. I think that `latest` is not the right term, because latest build might be not the build with the highest version.

With `best` there is also a difference in behavior. `focusbest` will not fail if it cannot install the highest version.
But still possibly `--prefer-best`? or `--prefer-highest`?

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -376,6 +376,10 @@ libdnf5::GoalProblem GoalPrivate::resolve() {
     libsolv_solver.set_flag(SOLVER_FLAG_ALLOW_VENDORCHANGE, vendor_change);
     libsolv_solver.set_flag(SOLVER_FLAG_DUP_ALLOW_VENDORCHANGE, vendor_change);
 
+    // Ensure solver tries to use the latests version of dependecies even if it results in a bigger transaction
```
jan-kolarik, line 379: Minor typo -> `Ensure the solver tries to use the latest version of dependencies, even if it results in a bigger transaction`

kontura:
As agreed on our meeting I have set the flag to be used always.

If it turns out to be problematic we can revert it and possibly add it as an option.

m-blaha (approved):
Looking forward to test this on users :).
I'd like to see it also in dnf4.

j-mracek:
LGTM

j-mracek:
There are some failing tests. But I quite confused, tests complains about something (missing upgrade, weak-dep) but output from failed commands shows correct result. @kontura May I ask you for investagation?

kontura:
> There are some failing tests. But I quite confused, tests complains about something (missing upgrade, weak-dep) but output from failed commands shows correct result. @kontura May I ask you for investigation?

Yes, I am already looking into it.
The weak-dep is changing reason. It is no longer installed as a weak dependency but as a normal dependency.
The spec contains:
```spec
Requires:       (milk or soymilk)
Recommends:     milk
```
and even with weak deps turned off (`--setopt=install_weak_deps=False`) it installs `milk` despite that `soymilk` is already installed.

The other problem is that during an upgrade of a specific package it also upgrades dependencies of the package.

j-mracek:
> > There are some failing tests. But I quite confused, tests complains about something (missing upgrade, weak-dep) but output from failed commands shows correct result. @kontura May I ask you for investigation?
> 
> Yes, I am already looking into it. The weak-dep is changing reason. It is no longer installed as a weak dependency but as a normal dependency. The spec contains:
> 
> ```python
> Requires:       (milk or soymilk)
> Recommends:     milk
> ```

I guess that would be a minor issue or no issue at all. The issue would be if the package would not have `Recommends: milk` but it stills pull milk into transaction.

> 
> and even with weak deps turned off (`--setopt=install_weak_deps=False`) it installs `milk` despite that `soymilk` is already installed.
> 

Thit is an issue and we should report that issue to `libsolv` with a test case.

> The other problem is that during an upgrade of a specific package it also upgrades dependencies of the package.
> 
> The question is if these are expected?

I think this is expected. I believe that this might cause a negative feedback. The flag limits option for dependencies, therefore dependencies are supposed to bi in the highest version.

kontura:
> > and even with weak deps turned off (`--setopt=install_weak_deps=False`) it installs `milk` despite that `soymilk` is already installed.
> 
> Thit is an issue and we should report that issue to `libsolv` with a test case.

I have created: https://github.com/openSUSE/libsolv/issues/539

j-mracek:
It looks like that the blocking issue is resolved in libsolv, therefore we need to wait until 0.7.25 is released.

j-mracek:
@kontura We know that libsolv-0.7.25 resolves the issue with the flag. What about to increase requirement of libsolv to 0.7.25? The feature is already controversial enough and it would be good to avoid additional pain. What do you think?

kontura:
> @kontura We know that libsolv-0.7.25 resolves the issue with the flag. What about to increase requirement of libsolv to 0.7.25? The feature is already controversial enough and it would be good to avoid additional pain. What do you think?

I have increased the required `libsolv` version.

ppisar:
Please note that libsolv-0.7.25 is not yet available in Fedora <https://bugzilla.redhat.com/show_bug.cgi?id=2239215>.

j-mracek:
> Please note that libsolv-0.7.25 is not yet available in Fedora https://bugzilla.redhat.com/show_bug.cgi?id=2239215.

I know, therefore I would like to ensure the availability by require before release to downstreams.

j-mracek:
LGTM, CI-dnf-stack PR fixes failed tests

PR #780, opened 2023-07-27
Modify dnf and microdnf obsoletes for RHEL

j-mracek:
The patch prevents replacement of microdnf and dnf by dnf5 in RHEL10 and previous versions.

PR #783, opened 2023-07-28
Postpone dnf5-as-dnf until Fedora 41

yselkowitz:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/EYE2JY537OM7GFW46EK7YIBLHJ52USAZ/


j-mracek:
There is already PR that handles this. The previous PR is blocked, because it also requires a synchronization at least with DNF.

yselkowitz:
Covered by #780

PR #786, opened 2023-08-01
cachedir: Use system dir when cacheonly enabled

jan-kolarik:
If `cacheonly` is enabled for all data and `cachedir` is not explicitly configured, use the system cachedir location, so unprivileged users can read system cached data.

`build_cache` is turned off when loading metadata only from cache as there is no need to rebuild solv cache files. And also based on the root mask we might not be able to create those files when running with system cache as unprivileged user.

### DIFF
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -183,6 +183,11 @@ void Base::setup() {
         config.get_logdir_option().set(Option::Priority::INSTALLROOT, full_path);
     }
 
+    // If cachedir is not explicitly specified and cacheonly is activated, use the system cachedir location.
+    if (config.get_cacheonly_option().get_value() == "all") {
+        config.get_cachedir_option().set(Option::Priority::DEFAULT, config.get_system_cachedir_option().get_value());
+    }
+
     load_plugins();
     p_impl->plugins.init();
 
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -188,6 +188,11 @@ void Base::setup() {
         config.get_cachedir_option().set(Option::Priority::DEFAULT, config.get_system_cachedir_option().get_value());
     }
 
+    // If only cached metadata are used, disable rebuilding solv cache files.
+    if (config.get_cacheonly_option().get_value() != "none") {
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Proposed changes seem not enough as running dnf with cacheonly as unprivileged user is giving following error:
```
filesystem error: cannot create temporary file: Permission denied [/var/cache/libdnf5/fedora-376ef8e983c65ce0/solv/fedora.solv.XXXXXX]
```
Switching to draft as more changes will follow ...

kontura:
I have put together a simple test for this: https://github.com/rpm-software-management/ci-dnf-stack/pull/1361

kontura:
Shouldn't we first check if the user cache is empty/expired? Before using the system one.
If an unprivileged user creates his own cache and then runs in `cacheonly` mode (and there is no system cache) it will fail even though he has the cache available.

I am also thinking about one related feature. Not sure if possible or in scope of this PR but what if unprivileged user (even without `cacheonly`) always first checks his own cache and if unavailable/expired he checks the system cache and only if that is also unavailable/expired he downloads new metadata?

jan-kolarik:
> Shouldn't we first check if the user cache is empty/expired? Before using the system one. If an unprivileged user creates his own cache and then runs in `cacheonly` mode (and there is no system cache) it will fail even though he has the cache available.

It sounds like a good idea. But if we want to make this change, we need to change documentation as now it matches the dnf behavior and it starts with "Run entirely from system cache ...". I am also not sure if there are some expected use cases connected with this.

> I am also thinking about one related feature. Not sure if possible or in scope of this PR but what if unprivileged user (even without `cacheonly`) always first checks his own cache and if unavailable/expired he checks the system cache and only if that is also unavailable/expired he downloads new metadata?

I also like that, but it's connected with what I mentioned above. Lets put it on the next discussion meeting.

j-mracek:
There is additional issue in DNF with cache - when user sets cachedir and performs operation as root and then as normal user then we have also problems with intermissions.

j-mracek:
I opened a discussion for ideas - https://github.com/rpm-software-management/dnf5/discussions/845.

Be honest I would like to block this PR before we agree about a future path.

Issue #792, opened 2023-08-02
Regression in Vars handling


j-mracek:
There are 3 issues covered by CI scenarios (`vars.feature`) that fails with DNF5

- [x] Scenario: Variables arch supports basearch `loongarch64` {}
Substitution of multiple vars in section `[dnf-ci-test-$arch-$basearch]` fails

- [x] Scenario: Using dnf with non-files in /etc/dnf/vars
- bz2091636

- [x] Scenario: Ignoring variable files with invalid encoding
- bz2141215

m-blaha:
The issue with invalid encoding doesn't seem to affect dnf5. The problems with loongarch64 and non-files in the 'vars' directory have been resolved by https://github.com/rpm-software-management/dnf5/pull/797 and https://github.com/rpm-software-management/dnf5/pull/798, respectively.

PR #798, opened 2023-08-03
vars: Ignore unreadable vars files.

m-blaha:
If there is an item in the vars directory that is either inaccessible
(e.g. permission denied) or cannot be read from (e.g. a directory),
skip this item and write a warning to the log.

For: https://github.com/rpm-software-management/dnf5/issues/792


jan-kolarik (approved):
LGTM

Issue #710, opened 2023-07-11
Add support for `$releasever_major`, `$releasever_minor`, and limited shell expansion


Conan-Kudo:
### Description of request
Zypper defines in its man page that two additional variables are auto-defined based on `$releasever`'s definition: `$releasever_major` and `$releasever_minor`.

Per `zypper(8)`:

> In addition `$releasever_major` will be set to the leading portion up to
> (but not including) the 1st dot; `$releasever_minor` to the trailing portion
> after the 1st dot. If there’s no dot in `$releasever`, `$releasever_major` is
> the same as `$releasever` and `$releasever_minor` is empty.

This can be verified by reading the man page of zypper on any Fedora system.

In addition, it seems that SUSE has [implemented very limited shell expansion capabilities](https://doc.opensuse.org/projects/libzypp/HEAD/structzypp_1_1repo_1_1RepoVarExpand.html) for supporting SUSE Linux Enterprise upgrades due to how their OBS names the repos.

For example, this is a common thing to see in repo files for SLE when trying to use repos from the openSUSE Build Service (or other OBS instances making SLE packages): `SLE_${releasever_major}${releasever_minor:+_SP$releasever_minor}`.

Having support for this would make it tremendously easier to slot in DNF in place of Zypper in SUSE distributions.

The implementation in libzypp is here:

* header: https://github.com/openSUSE/libzypp/blob/d03a02d986f43266af81c73cced1b310e8615971/zypp/repo/RepoVariables.h
* source: https://github.com/openSUSE/libzypp/blob/d03a02d986f43266af81c73cced1b310e8615971/zypp/repo/RepoVariables.cc

### Benefit to RHEL/CentOS

Starting with EPEL 10, [EPEL will be versioning its repositories with minor versions (`epel10.Y`) targeting RHEL and the major version-only branch (`epel10`) will target CentOS Stream 10](https://discussion.fedoraproject.org/t/epel-10-proposal/44304).

EPEL would really like this feature so that we can have a singular `epel-release` package that seamlessly works on CentOS Stream and derivatives (such as RHEL, Alma, etc.).

Conan-Kudo:
cc: @davide125, @carlwgeorge, @tdawson

gotmax23:
+1. It'd be great to have this in time for EL 10.

j-mracek:
Zypper uses following code for parsing `releasever`:
```

        /** Split \c releasever at \c '.' and store major/minor parts as requested. */
[...]

PR #800, opened 2023-08-03
Add support for $releasever_major, $releasever_minor, and limited shell expansion

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/710.

Add `$releasever_major` and `$releasever_minor`: whenever the `releasever` var is set, automatically derive and set the `releasever_major` and `releasever_minor` vars by splitting `releasever` on the first dot. In the future (or now, possibly), we may also want to add `releasever_patch`.

Compatible with this implementation in libzypp:
https://github.com/openSUSE/libzypp/blob/d03a02d986f43266af81c73cced1b310e8615971/zypp/repo/RepoVariables.cc#L478-L495

Also adds support for `${variable:-word}` and `${variable:+word}` expansions of vars, similar to the functionality in POSIX shell.

`${variable:-word}` means if `variable` is unset or empty, the expansion of `word` is substituted. Otherwise, the value of `variable` is substituted.

`${variable:+word}` means if `variable` is unset or empty, nothing is substituted. Otherwise, the expansion of `word` is substituted.

Both support nested variable expressions; you could do `${var0:-${var1:+${var2:+$var3}}}`

Zypper also supports these expansions, see here: https://doc.opensuse.org/projects/libzypp/HEAD/structzypp_1_1repo_1_1RepoVarExpand.html

### DIFF
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -137,6 +137,13 @@ private:
         const std::function<const std::unique_ptr<const std::string>()> & get_value,
         Priority prio);
 
+    /// @brief Expand variables in a subexpression
+    ///
+    /// @param text String with variable expressions
+    /// @param depth The recursive depth
+    /// @return Pair of the resulting string and the number of characters scanned in `text`
+    std::pair<std::string, size_t> substitute_expression(std::string_view text, unsigned int depth) const;
+
     BaseWeakPtr base;
     std::map<std::string, Variable> variables;
 };
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -88,6 +88,12 @@ public:
 
     static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
 
+    /// @brief Split releasever on the first "." into its "major" and "minor" components
+    ///
+    /// @param releasever A releasever string, possibly containing a "."
+    /// @return releasever_major, releasever_minor
+    static std::tuple<std::string, std::string> split_releasever(const std::string & releasever);
+
 private:
     friend class Base;
 
@@ -120,6 +126,16 @@ private:
     /// Reads environment variables that match "DNF[0-9]" and
     /// "DNF_VAR_[A-Za-z0-9_]+" patterns. The "DNF_VAR_" prefix is cut off.
     void load_from_env();
+
+    /// @brief Set a variable to a value, only obtaining the value if needed using `get_value`
+    ///
+    /// @param name Name of the variable
+    /// @param get_value Function that returns the (optional) value for the variable
+    /// @param prio Source/Priority of the value
+    void set_lazy(
+        const std::string & name,
+        const std::function<const std::unique_ptr<const std::string>()> & get_value,
+        Priority prio);
 
     BaseWeakPtr base;
     std::map<std::string, Variable> variables;
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -94,12 +94,6 @@ public:
 
     static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
 
-    /// @brief Split releasever on the first "." into its "major" and "minor" components
-    ///
-    /// @param releasever A releasever string, possibly containing a "."
-    /// @return releasever_major, releasever_minor
-    static std::tuple<std::string, std::string> split_releasever(const std::string & releasever);
-
 private:
     friend class Base;
 
@@ -149,6 +143,12 @@ private:
     /// @param depth The recursive depth
     /// @return Pair of the resulting string and the number of characters scanned in `text`
     std::pair<std::string, size_t> substitute_expression(std::string_view text, unsigned int depth) const;
+
+    /// @brief Split releasever on the first "." into its "major" and "minor" components
+    ///
+    /// @param releasever A releasever string, possibly containing a "."
+    /// @return releasever_major, releasever_minor
+    static std::tuple<std::string, std::string> split_releasever(const std::string & releasever);
 
     BaseWeakPtr base;
     std::map<std::string, Variable> variables;
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -23,9 +23,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/base/base_weak.hpp"
 
 #include <map>
+#include <set>
 #include <string>
 #include <vector>
-
 
 namespace libdnf5 {
 
@@ -69,6 +69,12 @@ public:
     /// @param value Value to be stored in variable
     /// @param prio Source/Priority of the value
     void set(const std::string & name, const std::string & value, Priority prio = Priority::RUNTIME);
+
+    /// @brief Checks whether a variable is read-only
+    ///
+    /// @param name Name of the variable
+    /// @return true if the variable is read-only, false if it is writable
+    bool is_read_only(const std::string & name) const;
 
     /// @brief Checks if there is an variable with name equivalent to name in the container.
     ///
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -29,6 +29,14 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 namespace libdnf5 {
 
+// Thrown when attempting to set a read-only variable
+class ReadOnlyVariableError : public Error {
+    using Error::Error;
+
+    const char * get_domain_name() const noexcept override { return "libdnf5"; }
+    const char * get_name() const noexcept override { return "ReadOnlyVariableError"; }
+};
+
 /// @class Vars
 ///
 /// @brief Class for reading and substituting DNF vars (arch, releasever, etc.).
@@ -68,6 +76,7 @@ public:
     /// @param name Name of the variable
     /// @param value Value to be stored in variable
     /// @param prio Source/Priority of the value
+    /// @throw ReadOnlyVariableError if the variable is read-only
     void set(const std::string & name, const std::string & value, Priority prio = Priority::RUNTIME);
 
     /// @brief Checks whether a variable is read-only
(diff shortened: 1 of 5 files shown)


include/libdnf5/conf/vars.hpp
```diff
@@ -86,7 +95,13 @@ struct Vars {
     /// @param name Name of the variable
     const Variable & get(const std::string & name) const { return variables.at(name); }
 
-    static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
+    static std::optional<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
```
j-mracek, line 98: I am sorry, but we cannot use std::optional because it is not supported by SWIG.

include/libdnf5/conf/vars.hpp
```diff
@@ -70,6 +71,14 @@ struct Vars {
     /// @param prio Source/Priority of the value
     void set(const std::string & name, const std::string & value, Priority prio = Priority::RUNTIME);
 
+    /// @brief Set a variable to a value, only obtaining the value if needed using `get_value`
+    ///
+    /// @param name Name of the variable
+    /// @param get_value Function that returns the (optional) value for the variable
+    /// @param prio Source/Priority of the value
+    void set_lazy(
```
j-mracek, line 79: I am not sure whether it is a good idea to detect vars not at the same time point. Right not it happen at Base::setup() which is a mandatory step. Is there a particular reason to do it this way? Originally we were thinking about locking vars like with configuration, but the feature has low priority therefore it is not yet implemented.

include/libdnf5/conf/vars.hpp
evan-goode: Ah, forgot about that, that's a shame. I'll use `unique_ptr` instead.

include/libdnf5/conf/vars.hpp
evan-goode: `set_lazy` doesn't actually defer setting the var, it's the same as `set` but only evaluates the value to be set if it's necessary to do so, i.e. when the incoming priority is greater than the existing priority.

It's a somewhat common pattern, often you will see a `log.Println(() => string)` function that only evaluates the lambda if necessary.

I added it to deduplicate some of the code in `detect_vars`.

libdnf5/conf/vars.cpp
```diff
@@ -223,17 +223,52 @@ std::string Vars::substitute(const std::string & text) const {
...
+    if (it != variables.end() && prio < it->second.priority) {
+        return;
+    }
+
+    // Whenever releasever is set, split it into major and minor parts
+    if (name == "releasever") {
+        const auto [releasever_major, releasever_minor] = split_releasever(value);
+        set("releasever_major", releasever_major, prio);
+        set("releasever_minor", releasever_minor, prio);
+    }
```
j-mracek, line 345: I am not sure whether automatic detection during the setting is the right approach. Anyway it is not wrong.

I have some concerns to priority and setting empty values.
The current code using the same priority for derived variables. Why cannot we use a lover priority? Example user define from commandline `releasever_major` and then `releasever` but it overrides `releasever_major` because autodetected value has the same priority.

With empty value I guess that we should not substitute `$releasever_minor` when `releasever` = f38. DNF does that in other places (when no releasever (in installroot), it does not substitute it) and I somehow feel it  would be good to keep the same pattern.  This behavior is different from shell, but it helps with debugging.

include/libdnf5/conf/vars.hpp
j-mracek: Thank you very much for explanation. Do you think we need it as API?

libdnf5/conf/vars.cpp
m-blaha: Hmm, playing with priorities and setting `releasever_major` with artificially lowered priority doesn't seem right to me. I believe the root issue here lies in the design itself - `releasever_major` and `releasever_minor` are dependent on the value of the `releasever` variable and should be treated as kind of  "readonly" (I'd even consider disallowing independent setting, allowing changes only through `releasever`).

j-mracek: That would be possible, but it would require to implement locking mechanism for Vars, like we have for configuration and it will create an exception in the logic.

evan-goode: > Hmm, playing with priorities and setting `releasever_major` with artificially lowered priority doesn't seem right to me. I believe the root issue here lies in the design itself - `releasever_major` and `releasever_minor` are dependent on the value of the `releasever` variable and should be treated as kind of "readonly" (I'd even consider disallowing independent setting, allowing changes only through `releasever`).

I agree, I think this is the better approach, I made `releasever_major` and `releasever_minor` read-only. Let me know what you think.

libdnf5/conf/vars.cpp
```diff
@@ -35,7 +35,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <cstring>
 #include <filesystem>
 #include <memory>
-#include <optional>
```
Conan-Kudo, line 38: Why is `optional` getting removed here? It seems independent from the rest of the changes? Can we split this to a separate commit at least?

Conan-Kudo (changes requested):

libdnf5/conf/vars.cpp
evan-goode: Whoops, thanks. I've put it back in.

Conan-Kudo:
@evan-goode Can you rebase this to get rid of the conflicts?

evan-goode:
> @evan-goode Can you rebase this to get rid of the conflicts?

Yep, rebased.

include/libdnf5/conf/vars.hpp
Conan-Kudo: You could just move it to the `private` section of the class if it's only used in the Vars class.

evan-goode: OK, done.

libdnf5/conf/vars.cpp
Conan-Kudo: I think this sufficiently addressed this feedback.

Conan-Kudo: I think this sufficiently addresses the feedback. And this is the only feedback blocking me from merging this.

@j-mracek are we good with this now?

include/libdnf5/conf/vars.hpp
```diff
@@ -88,6 +88,12 @@ struct Vars {
 
     static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
 
+    /// @brief Split releasever on the first "." into its "major" and "minor" components
+    ///
+    /// @param releasever A releasever string, possibly containing a "."
+    /// @return releasever_major, releasever_minor
+    static std::tuple<std::string, std::string> split_releasever(const std::string & releasever);
```
j-mracek, line 101: Be honest I would not prefer to add the method to API and make it private. If we will add support of releasever_patch value, the method will be obsoleted.

libdnf5/conf/vars.cpp
```diff
@@ -180,67 +182,243 @@ std::unique_ptr<std::string> Vars::detect_release(const BaseWeakPtr & base, cons
...
+    return std::make_tuple(releasever_major, releasever_minor);
+}
+
+bool Vars::is_read_only(const std::string & name) {
+    return READ_ONLY_VARIABLES.contains(name);
 }
 
 void Vars::set(const std::string & name, const std::string & value, Priority prio) {
+    if (is_read_only(name)) {
+        throw RuntimeError(M_("Variable \"{}\" is read-only"), name);
```
j-mracek, line 377: Please document that it can throw an error. I have to test, but I think that it will create a problem with setvars option or other channels that can modify variables.

j-mracek: I've tested it and it works like expected. I am still not sure whether RuntimeError is the best exception.

j-mracek: I know that it is used in the rest of the module but I feel that libdnf5 exception that inherits from runtime would be more appropriate. What do you think?

include/libdnf5/conf/vars.hpp
evan-goode: Done, makes sense. An API user could always read the value of `releasever_major`, `releasever_minor` after setting `releasever`.

libdnf5/conf/vars.cpp
evan-goode: I agree, I added a `ReadOnlyVariableError` class, and I put a `/// @throw ReadOnlyVariableError` in the docstring. Let me know if I should document it anywhere else.

I added a test here https://github.com/rpm-software-management/ci-dnf-stack/pull/1359 to ensure read-only variables cannot be set.

j-mracek: Additionally I asked @jrohel about opinion for locking mechanism, because he is an author of config locking mechanism.

jrohel: This is not about locking. This code implements a hard-coded list of read-only variable names. The value of these variables is not locked, but is calculated based on the value of another variable. In other words, the value of these variables is not locked, they can be changed at any time, indirectly, by changing the value of another variable.

libdnf5/conf/vars.cpp
```diff
@@ -179,54 +180,170 @@ std::unique_ptr<std::string> Vars::detect_release(const BaseWeakPtr & base, cons
...
-
 std::string Vars::substitute(const std::string & text) const {
-    std::string res = text;
+    return substitute_expression(text, 0).first;
+}
 
-    if (variables.empty()) {
-        return res;
+const unsigned int MAXIMUM_EXPRESSION_DEPTH = 32;
+std::pair<std::string, size_t> Vars::substitute_expression(const std::string_view & text, unsigned int depth) const {
```
jrohel, line 192: Please always pass `std::string_view` by value. 
`const std::string_view & text` -> `std::string_view text `

evan-goode: Done

include/libdnf5/conf/vars.hpp
```diff
@@ -68,8 +76,15 @@ struct Vars {
...
     /// @param value Value to be stored in variable
     /// @param prio Source/Priority of the value
+    /// @throw ReadOnlyVariableError if the variable is read-only
     void set(const std::string & name, const std::string & value, Priority prio = Priority::RUNTIME);
 
+    /// @brief Checks whether a variable is read-only
+    ///
+    /// @param name Name of the variable
+    /// @return true if the variable is read-only, false if it is writable
+    bool is_read_only(const std::string & name);
```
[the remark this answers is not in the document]
evan-goode: Done

PR #803, opened 2023-08-04
Configure copr repo dnf5-testing

inknos:
This pr enable dnf5 builds that obsolete dnf5
it is used by https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-testing/
related to: https://github.com/rpm-software-management/dnf/pull/1976

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -2,6 +2,8 @@
 %global project_version_minor 1
 %global project_version_patch 1
 
+%bcond obsolete_dnf %[0%{?fedora} > 40 || 0%{?rhel} > 10]
+
 Name:           dnf5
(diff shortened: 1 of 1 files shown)


dnf5.spec
```diff
@@ -2,6 +2,8 @@
 %global project_version_minor 1
 %global project_version_patch 1
 
+%bcond_with obsolete_dnf
```
gotmax23, line 5: This seems unnecessarily convoluted. How about changing this to

```suggestion
%bcond obsolete_dnf %[0%{?fedora} >= 41 || 0%{?rhel} >= 11]
```

and using simple `%if %{with obsolete_dnf}` and `%if %{without obsolete_dnf}` conditionals everywhere?

j-mracek: I like it.

Issue #795, opened 2023-08-02
Running DNF5 with i386 basearch 


nuqleo:
Hi,

I know i386 architecture support is being removed, but tried to test dnf5 in 32bit container and VM using build repo https://kojipkgs.fedoraproject.org/repos/rawhide/latest/i386/

When dnf5 running in 32bit container on x86_64 host it works fine

```
# dnf5 repolist 
repo id                                                                      repo name
rawhide                                                                      Fedora - Rawhide - x86_64
```
```
# dnf5 install less
Updating and loading repositories:
Repositories loaded.
Package                  Arch    Version                  Repository        Size
Installing:                                                                     
 less                    i686    633-2.fc39               rawhide      384.5 KiB

Transaction Summary:
 Installing:        1 packages

Total size of inbound packages is 181 KiB. Need to download 181 KiB.
After this operation 385 KiB will be used (install 385 KiB, remove 0 B).
Is this ok [y/N]: y
[1/1] less-0:633-2.fc39.i686            100% | 154.1 KiB/s | 181.3 KiB |  00m01s
--------------------------------------------------------------------------------
[1/1] Total                             100% | 153.9 KiB/s | 181.3 KiB |  00m01s
Running transaction
[1/3] Verify package files              100% |   0.0   B/s |   1.0   B |  00m00s
[2/3] Prepare transaction               100% |  58.0   B/s |   1.0   B |  00m00s
[3/3] Installing less-0:633-2.fc39.i686 100% |  21.0 MiB/s | 387.9 KiB |  00m00s
>>> Running trigger-install scriptlet: glibc-common-0:2.37.9000-18.fc39.i686
>>> Stop trigger-install scriptlet: glibc-common-0:2.37.9000-18.fc39.i686
```

But inside 32-bit kernel VM with setup identical to container dnf5 fails to install packages
```
# dnf5 repolist 
repo id                             repo name                                   
rawhide                             Fedora - Rawhide - i386 
```
```
# dnf5 install less
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - package less-633-2.fc39.i686 does not have a compatible architecture
```
Should dnf5 work with i386 basearch?

m-blaha:
Can you share `uname -a` output from the VM?

nuqleo:
```
# uname -a
[...]

PR #805, opened 2023-08-07
vars: Change detecting arch mechanism

m-blaha:
Currently used `rpmGetArchInfo()` API corresponds to basearch and not the
actual architecture, causing problems with incompatible architectures
during transaction resolving.
To really get the actual architecture, we need to use
`rpmExpand("%_host_cpu", NULL)`, which is also what RPM uses to detect
the architecture.

Resolves: https://github.com/rpm-software-management/dnf5/issues/795
(also see the discussion there for details)

### DIFF
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -29,6 +29,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <dirent.h>
 #include <rpm/rpmdb.h>
 #include <rpm/rpmlib.h>
+#include <rpm/rpmmacro.h>
 #include <rpm/rpmts.h>
 #include <sys/types.h>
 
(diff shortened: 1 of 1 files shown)


pmatilai:
I seem to have missed the boat on this one, but I'll point it out anyway: the commit message is quite wrong, obviously rpm cannot use that macro to *detect* the architecture. The macro is set by the architecture detection code.

This also causes a memory leak, as rpmExpand() always returns malloced memory. You'd want to call it once on early init and remember thereafter (to protect against somebody changing the macro too).

m-blaha:
Oh, thanks for noticing this. I'll fix the memory leak in follow-up pull request.
And sorry for the clumsy commit message. I wanted to express that rpm and dnf5 use the same value for architecture detected by the same code, but I obviously messed it up...

PR #806, opened 2023-08-08
Repoquery reorganize and small improvements

kontura:

There is quite a bit of code reorganization, I am trying to make the slur of repoquery options more understandable but if you don't consider these changes worth it (they do complicate the history and could be confusing) I can drop them.

There is also a couple of small improvements like: option conflicts, docs improvements, fix for `--recent`, one better exception and faster `--querytags`.

### DIFF
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -471,7 +471,7 @@ void RepoqueryCommand::run() {
     libdnf5::rpm::PackageQuery base_query(ctx.base, flags, false);
     libdnf5::rpm::PackageQuery full_package_query(ctx.base, flags, true);
 
-    // First filter by pkg_specs - it can narrow the query the most
+    // First filter by pkg_specs - it should be in SIMPLE FILTERS but it can narrow the query the most
     if (pkg_specs.empty()) {
         full_package_query |= base_query;
     } else {
@@ -490,6 +490,8 @@ void RepoqueryCommand::run() {
         }
     }
 
+    // APPLY FILTERS ONLY FOR INSTALLED PACKAGES
+
     if (leaves_option->get_value()) {
         full_package_query.filter_leaves();
     }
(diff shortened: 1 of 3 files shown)


j-mracek:
I have a question related to Daemon - is there a way how to share repoquery command code with Daemon. Repoquery is not only about rpm queries, but it also contains some logic (--resolve, --recursive, output formatting, ...). What do you think?

kontura:
> I have a question related to Daemon - is there a way how to share repoquery command code with Daemon. Repoquery is not only about rpm queries, but it also contains some logic (--resolve, --recursive, output formatting, ...). What do you think?

For output formatting it should work fine, it is separated in `libdnf5-cli` in `repoqueryformat.cpp` so `dnf5daemon-client` can use that.
There is some logic related to `--exactdeps` which I think could be moved to `PackageQuery` because it is basically `PackgeQuery::resolve_pkg_spec` that takes a vector of specs.

With `--resolve` it is a bit more complicated. Mainly I would like to change the option to something like:
 `--providers-of=<requires,recommends,provides...>`.
Because the `dnf4` implementation overloads the formatting options `--requires`, `--recommends`, `--provides`,... and also is quite non-descriptive. (I have this only locally so far, waiting on other PRs.)
However together with the `--recursive` these are basically transforms of a PackageSet into a different PackageSet. Perhaps we could add them to API to `PackageSet` or `PackageQuery` namespace?

kontura:
I have thought about it some more and the output formatting might be problematic after all. It [relies quite a bit](https://github.com/rpm-software-management/dnf5/blob/main/libdnf5-cli/output/repoqueryformat.cpp#L41-L81) on pointers to `libdnf5::rpm::Package` methods. 
This will not easily work with the daemon's `DbusPackageWrapper`.

kontura:
Rebased to resolve conflicts.

jan-kolarik (approved):
LGTM, nice commits organization!

jan-kolarik:
Maybe the CI tests for conflicting options could be also added (later).

Issue #150, opened 2022-11-11
Implement repoinfo command


mcurlej:
- [x] **Command:** ```repoinfo```
  - command is implemented but it provides only limited information in comparison to DNF4
  - Priority medium

- [x] An alias for the [repolist](https://dnf.readthedocs.io/en/latest/command_ref.html#repolist-command-label) command that provides more detailed information like ```dnf repolist -v```.
  - Proposing to drop
  - Priority medium

Tracked in another issues:

- [ ] Unify the output with original dnf and create a possibility to add custom fields to the output ?
  - https://github.com/rpm-software-management/dnf5/issues/842

glum23:
[RHEL related] When some repos are signed, dnf4 on RHEL is expected to import all required keys in one transaction and to provide info about it (dnf5 currently does not provide anything like that). In the example below, 2 repos (repo1 and repo2) are signed. See also RHEL bug 1781601.
~~~
# dnf -y repoinfo repo1
My Local Repo repo1                            26 kB/s | 659  B     00:00    
My Local Repo repo1                           2.3 MB/s | 2.3 kB     00:00    
Importing GPG key 0x9440AFDC:
 Userid     : "key1"
 Fingerprint: 5CE4 74FC 4EB6 8F1B D1D5 797E F5A2 5AB3 9440 AFDC
 From       : /tmp/tmp.dtiqa65REx/key1/key1-public
My Local Repo repo1                            57 kB/s | 1.0 kB     00:00    
My Local Repo repo2                            79 kB/s | 659  B     00:00    
My Local Repo repo2                           2.3 MB/s | 2.3 kB     00:00    
Importing GPG key 0xA701D86D:
 Userid     : "key2"
 Fingerprint: FC6E C162 F264 58B8 3726 FBC0 1C67 3F29 A701 D86D
 From       : /tmp/tmp.dtiqa65REx/key2/key2-public
My Local Repo repo2                            65 kB/s | 1.0 kB     00:00    
Repo-id            : repo1
Repo-name          : My Local Repo repo1
Repo-status        : enabled
Repo-revision      : 1682068598
Repo-updated       : Fri 21 Apr 2023 05:16:37 AM EDT
Repo-pkgs          : 2
Repo-available-pkgs: 2
Repo-size          : 13 k
Repo-baseurl       : file:///tmp/tmp.dtiqa65REx/repo1/
[...]

PR #807, opened 2023-08-08
Implementation of repo info command

jan-kolarik:
Provide `repo info` command with attributes included in original dnf command and several more:
- status: `enabled` / `disabled`
- priority
- cost
- type: `available` / `commandline`
- skip if unavailable: `true` / `false`
- PGP info
  - keys (from the `gpgkey` option)
  - verify repodata: `true` / `false` (from the `repo_gpgcheck` option)
  - verify packages: `true` / `false` (from the `gpgcheck` option) 

Related issue: https://github.com/rpm-software-management/dnf5/issues/150.
CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1363.

### DIFF
--- a/dnf5/commands/repo/repo_info.cpp
+++ b/dnf5/commands/repo/repo_info.cpp
@@ -76,6 +76,7 @@ public:
     uint64_t get_size() const { return size; }
     uint64_t get_pkgs() const { return pkgs; }
     uint64_t get_available_pkgs() const { return available_pkgs; }
+    std::vector<std::string> get_mirrors() const { return repo.get_mirrors(); }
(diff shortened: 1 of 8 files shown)


jan-kolarik:
Number of total packages at the end of the output is not provided for now. Could be easily implemented in addition if requested.

jan-kolarik:
Also, the time of last cache update is currently displayed only within `Metadata expire` column together with the expire seconds value, as before with dnf command. It might make sense to also put it as a separate key.

m-blaha (approved):
Thank you!

PR #813, opened 2023-08-10
Add support for drop-in configuration directories 

jrohel:
Introduces a new method `Base::load_config()` to replace `Base::load_config_from_file()`.

In the original implementation, the `Base` class constructor loads the default configuration from the distribution configuration file ("/usr/share/dnf5/libdnf.conf"). And later, the user configuration (by default "/etc/dnf/dnf.conf") was loaded by calling the `Base::load_config_from_file()` method.

**New implementation:**
1. The `Base` class constructor does not load any configuration file.
2. The application calls `Base::load_config()` method instead of `Base::load_config_from_file()`.

**The order of loading the configuration files by the method `Base::load_config()`:**
1. Creates an alphabetically sorted list of names of all configuration files in the distribution configuration drop-in directory ("/usr/share/dnf5/libdnf.conf.d") and the user configuration drop-in directory ("/etc/dnf/libdnf5.conf.d"). If the file with the same name is present in both drop-in directories, only the file from the user configuration drop-in directory is added to the sorted list. So, the distribution configuration drop-in file can be simply masked by creating a file with the same name in the user configuration drop-in directory.
2. Retrieves the configuration from the files in the sorted list. Follows the order. The configuration from the next file overrides the previous one - the last option value wins.
3. Loads the user configuration file (by default "/etc/dnf/dnf.conf"), if exists. If file does not exist and was explicitly requested by the user, the method throws an exeption.

---------------------------------------------
### Example configuration:
**User configuration files:**
/etc/dnf/dnf.conf
/etc/dnf/libdnf5.conf.d/20-user-settings.conf
/etc/dnf/libdnf5.conf.d/60-something.conf
/etc/dnf/libdnf5.conf.d/80-user-settings.conf

**Distribution configuration files:**
/usr/share/dnf5/libdnf.conf.d/50-something.conf
/usr/share/dnf5/libdnf.conf.d/60-something.conf
/usr/share/dnf5/libdnf.conf.d/90-something.conf

**Resulting file loading order by defaul**t
("/usr/share/dnf5/libdnf.conf.d/60-something.conf" is skipped, masked by the user file "/etc/dnf/libdnf5.conf.d/60-something.conf"):

1. /etc/dnf/libdnf5.conf.d/20-user-settings.conf
2. /usr/share/dnf5/libdnf.conf.d/50-something.conf
3. /etc/dnf/libdnf5.conf.d/60-something.conf
4. /etc/dnf/libdnf5.conf.d/80-user-settings.conf
5. /usr/share/dnf5/libdnf.conf.d/90-something.conf
6. /etc/dnf/dnf.conf

### DIFF
--- a/include/libdnf5/base/base.hpp
+++ b/include/libdnf5/base/base.hpp
@@ -79,10 +79,16 @@ public:
     /// Returns a pointer to a locked "Base" instance or "nullptr" if no instance is locked.
     static Base * get_locked_base() noexcept;
 
+    /// Loads main configuration.
+    /// The file defined in the current configuration and files in the drop-in directories are used.
+    void load_config();
+
+    /// @deprecated Don't use it! It will be removed in Fedora 40. It was intended for internal use only.
     /// Call a function that loads the config file, catching errors appropriately
     void with_config_file_path(std::function<void(const std::string &)> func);
 
-    /// Loads main configuration from file defined by the current configuration.
+    /// @deprecated It is redundant. It calls `load_config()`.
+    /// Loads main configuration.
     void load_config_from_file();
 
     /// @return a reference to configuration
--- a/include/libdnf5/base/base.hpp
+++ b/include/libdnf5/base/base.hpp
@@ -132,23 +132,6 @@ private:
     friend class libdnf5::repo::RepoSack;
     friend class libdnf5::repo::SolvRepo;
 
-    /// Loads the default configuration. To load distribution-specific configuration.
-    void load_defaults();
-
-    //TODO(jrohel): Make public?
-    /// Loads main configuration from file defined by path.
-    void load_config_from_file(const std::string & path);
-
-    //TODO(jrohel): Make public? Will we support drop-in configuration directories?
-    /// Loads main configuration from files with ".conf" extension from directory defined by dir_path.
-    /// The files in the directory are read in alphabetical order.
-    void load_config_from_dir(const std::string & dir_path);
-
-    //TODO(jrohel): Make public? Will we support drop-in configuration directories?
-    /// Loads main configuration from files with ".conf" extension from directory defined by the current configuration.
-    /// The files in the directory are read in alphabetical order.
-    void load_config_from_dir();
-
     /// Load plugins according to configuration
     void load_plugins();
 
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -111,18 +111,6 @@ Base * Base::get_locked_base() noexcept {
     return locked_base;
 }
 
-void Base::load_defaults() {
-    const std::string file_path = LIBDNF5_DISTRIBUTION_CONFIG_FILE;
-    try {
-        log_router.debug("Loading default configuration from \"{}\"", file_path);
-        ConfigParser parser;
-        parser.read(file_path);
-        config.load_from_parser(parser, "main", vars, *get_logger(), Option::Priority::DEFAULT);
-    } catch (const libdnf5::MissingConfigError & ex) {
-        log_router.debug("Configuration file \"{}\" not found", file_path);
-    }
-}
-
 void Base::load_config() {
     fs::path conf_file_path{config.get_config_file_path_option().get_value()};
     fs::path conf_dir_path{CONF_DIRECTORY};
@@ -157,12 +145,6 @@ void Base::load_config() {
     }
 }
 
-void Base::load_config_from_file(const std::string & path) {
-    ConfigParser parser;
-    parser.read(path);
-    config.load_from_parser(parser, "main", vars, *get_logger());
-}
-
 void Base::with_config_file_path(std::function<void(const std::string &)> func) {
     std::filesystem::path conf_path{config.get_config_file_path_option().get_value()};
     const auto & conf_path_priority = config.get_config_file_path_option().get_priority();
@@ -187,34 +169,6 @@ void Base::with_config_file_path(std::function<void(const std::string &)> func) 
 
 void Base::load_config_from_file() {
     load_config();
-}
-
-void Base::load_config_from_dir(const std::string & dir_path) {
-    std::vector<std::filesystem::path> paths;
-    std::error_code ec;
-    for (auto & dentry : std::filesystem::directory_iterator(dir_path, ec)) {
-        auto & path = dentry.path();
-        if (path.extension() == ".conf") {
-            paths.push_back(path);
-        }
-    }
-    if (ec) {
-        log_router.warning("Cannot read configuration from directory \"{}\": {}", dir_path, ec.message());
-        return;
-    }
-    std::sort(paths.begin(), paths.end());
-    for (auto & path : paths) {
-        load_config_from_file(path);
-    }
-}
-
-void Base::load_config_from_dir() {
-    std::filesystem::path conf_dir{libdnf5::CONF_DIRECTORY};
-    if (!config.get_use_host_config_option().get_value()) {
-        conf_dir = config.get_installroot_option().get_value() / conf_dir.relative_path();
-    }
-
-    load_config_from_dir(conf_dir);
 }
 
 void Base::load_plugins() {
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -33,7 +33,49 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <cstdlib>
 #include <filesystem>
 #include <mutex>
+#include <string_view>
 #include <vector>
+
+namespace fs = std::filesystem;
+
+namespace {
+
+// Creates an alphabetically sorted list of all files with `file_extension` from `directories`.
+// If a file with the same name is in multiple directories, only the first file found is added to the list.
+// Directories are traversed in the same order as they are in the input vector.
+[[nodiscard]] std::vector<fs::path> create_sorted_file_list(
+    const std::vector<fs::path> & directories, std::string_view file_extension) {
+    std::vector<fs::path> paths;
+
+    for (const auto & dir : directories) {
+        std::error_code ec;
+        for (const auto & dentry : fs::directory_iterator(dir, ec)) {
+            const auto & path = dentry.path();
+            if (dentry.is_regular_file() && path.extension() == file_extension) {
+                const auto & path_fname = path.filename();
+                bool found{false};
+                for (const auto & path_in_list : paths) {
+                    if (path_fname == path_in_list.filename()) {
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found) {
+                    paths.push_back(path);
+                }
+            }
+        }
+    }
+
+    // sort all drop-in configuration files alphabetically by their names
+    std::sort(paths.begin(), paths.end(), [](const fs::path & p1, const fs::path & p2) {
+        return p1.filename() < p2.filename();
+    });
+
+    return paths;
+}
+
+}  // namespace
 
 namespace libdnf5 {
 
@@ -46,9 +88,7 @@ Base::Base(std::vector<std::unique_ptr<Logger>> && loggers)
       repo_sack(get_weak_ptr()),
       rpm_package_sack(get_weak_ptr()),
       transaction_history(get_weak_ptr()),
-      vars(get_weak_ptr()) {
-    load_defaults();
-}
+      vars(get_weak_ptr()) {}
 
 Base::~Base() = default;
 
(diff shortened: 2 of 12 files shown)


jrohel:
This solution uses a fixed list of drop-in directories. The file loading order algorithm is inspired by `libeconf` https://github.com/openSUSE/libeconf.

So it is a competing implementation to PR https://github.com/rpm-software-management/dnf5/pull/741. Competing PR introduces support for including configuration files using a special comment directive in the configuration file.

Note to reviewer:
There is no need to waste time reviewing the implementation. **What is important is whether we agree on this support for drop-in configuration directories from the user's point of view. Implementation is now a detail. Subsequently, I will rework it while adding support for "config-manager".**

m-blaha:
>     1. The `Base` class constructor does not load any configuration file.
> 
>     2. The application calls `Base::load_config()` method instead of `Base::load_config_from_file()`.

Just a question - does it still make sense to have separate `base.load_config()` and `base.setup()` methods? Or can we squash the functionalities into only one call?
Or is the intention here to enable additional runtime configuration adjustments for API users?

Conan-Kudo:
> /etc/dnf/dnf.conf
> /etc/dnf/libdnf5.conf.d/20-user-settings.conf
> /usr/share/dnf5/libdnf.conf.d/50-something2.conf
> /etc/dnf/libdnf5.conf.d/60-something2.conf
> /etc/dnf/libdnf5.conf.d/80-user-settings.conf
> /usr/share/dnf5/libdnf.conf.d/90-something2.conf
> /etc/dnf/libdnf5.conf.d/99-config-manager.conf

I'm trying to understand this order, does this mean that DNF basically stops reading config if `/etc/dnf/dnf.conf` exists? My general understanding of how this should work is that `/usr` configs are read first, then `/etc` configs afterward will overwrite any set values from `/usr`. I'm slightly confused if this actually does that.

jrohel:
@Conan-Kudo 
The distribution config file `/usr/share/dnf5/libdnf.conf` is readed only if the user config file `/etc/dnf/dnf.conf` does not exist.

"libeconf" https://github.com/openSUSE/libeconf does a similar thing. 
`/usr/vendor/example.suffix` is read only if `/etc/example.suffix` does not exist.

ppisar:
Which configuration among dnf.conf and libdnf5.conf.d take precedence?

jrohel:
@Conan-Kudo 
All configuration files from drop-in directories are placed in one alphabetically sorted list by name. If a file with the same name is in both drop-in directories (distribution and user), only the one from the user ("/etc/dnf/libdnf5.conf.d") is used.

jrohel:
@ppisar 
As I described. The last value loaded wins (has the highest priority). The order can be seen in the example in the description. "/etc/dnf/dnf.conf" is loaded first and then the files from the "drop-in" directories. So the configuration from the "drop-in" directories has a higher priority than from "/etc/dnf/dnf.conf".

Conan-Kudo:
That makes sense.

So how will this extend to repo configuration? Because that's the big one that both Fedora and openSUSE would like to solve too.

jrohel:
@Conan-Kudo 
> So how will this extend to repo configuration? Because that's the big one that both Fedora and openSUSE would like to solve too.

This PR only addresses loading the libdnf configuration. The "[main]" section in the configuration files.

### What about repositories configuration? So far I have this idea:
"libdnf5" now loads repositories configuration from `etc/yum.repos.d`, `/etc/distro.repos.d` directories.

**We can:**
a) add another drop-in directory (for example `/etc/dnf/repos-override.d`) to override repositories configuration.
b) or load repositories configuration overrides from drop-in configuration files read in this PR (we will also support sections other than "[main]").
c) or load the repositories configuration from the drop-in configuration directories mentioned in this PR, but the files will have a different extension than ".conf" (eg ".repo").

In any case, this would be to support overwriting the configuration of existing repositories. Creating new repositories using these drop-in directories will not be supported.

I am considering glob support for bulk modification of repositories parameters.
E.g. override proxy settings for all repositories whose "id" begins with "fedora":
```
[fedora*]
proxy=http://xyz.cc
```
But those are just my ideas. I don't know yet what I will implement or how we will agree in the team.

Conan-Kudo:
So, in openSUSE, we have the following repo dirs: `/etc/dnf/repos.d`, `/etc/yum.repos.d`, and `/etc/distro.repos.d`.

Packaged configs go in the latter two directories, whereas the first path is where user-installed ones go.

I would _like_ to change this to have packaged repository configs move to `/usr` and leave `/etc` entirely for user/admin stuff.

This would also include moving the packaged GPG keys to `/usr`, but that's not going to be hard anyway.

So the idea would be to have `/usr/share/dnf5/yum.repos.d` for packaged stuff, then `/usr/share/dnf5/yum.repos.override.d` for packaged overrides (e.g. `epel-release` turning on CRB in RHEL/CentOS, for example).

For user-controlled things, in addition to the existing directory paths in `/etc`, having `.override.d` variants could make sense.

jrohel:
@Conan-Kudo 
Changing the repo dirs list is easy. This is given by the configuration value `reposdir`. And changing the default value in the code is also technically easy - changing the value of the constant. I write "technically" on purpose. I can't change randomly for compatibility. Here it will be more about agreeing what the right directories (paths) are.

In my description above it was about the introduction of "overrides". That is, configuration files that will not be able to define new repositories, but only modify existing ones.

But we are already running outside the scope of this PR. It handles the main configuration of libdnf (the "[main]" section). No repositories configuration.

jrohel:
@m-blaha 
> Just a question - does it still make sense to have separate base.load_config() and base.setup() methods? Or can we squash the functionalities into only one call?

I'm not sure. I'd leave it as is for now.

include/libdnf5/base/base.hpp
```diff
@@ -79,11 +79,8 @@ class Base {
     /// Returns a pointer to a locked "Base" instance or "nullptr" if no instance is locked.
     static Base * get_locked_base() noexcept;
 
-    /// Call a function that loads the config file, catching errors appropriately
-    void with_config_file_path(std::function<void(const std::string &)> func);
-
-    /// Loads main configuration from file defined by the current configuration.
-    void load_config_from_file();
```
j-mracek, line 86: We cannot remove those methods. We can mar them deprecated with a reason and then we can remove them. Please also add suggested removal window - like Fedora xx

libdnf5/base/base.cpp
```diff
@@ -71,21 +71,77 @@ Base * Base::get_locked_base() noexcept {
...
+    // it will be loaded instead of the distribution configuration file.
+    if (user_defined_config_file_name || fs::exists(conf_file_path)) {
+        parser.read(conf_file_path);
+    } else if (fs::exists(distribution_conf_file_path)) {
+        parser.read(distribution_conf_file_path);
+    }
+
+    // If the name of the configuration file was not specified, the files from the drop-in directories are also loaded.
+    // An alphabetically sorted list of all configuration files from the drop-in directories is created. If there is
+    // a file of the same name in both the user and distribution drop-in directories, only the user file is used.
```
j-mracek, line 104: I am not sure about this logic. If user will specify -C /<path>, then all distribution and /etc overrides will be ignored. May I ask you for more details or some use case when this behavior is beneficial?

include/libdnf5/base/base.hpp
jrohel: Modified. I recommend removing it as soon as possible. So I wrote in Fedora 40. Or can we do it in Fedora 39?

j-mracek:
Drop in configuration must be also documented in `doc` part. It doesn't have to be included in this PR, but at least it must be tracked in a separate issue.

libdnf5/base/base.cpp
jrohel: @j-mracek 
`-C, --cacheonly`
You probably meant `-c [config file], --config [config file]`. Not yet implemented in dnf5. So `-c user_config_file`.
What should I do? The user has explicitly specified that they want configuration from this file. Overwriting it with configuration from drop-in directories is probably not what he wants. Therefore, in this case, the configuration from the drop-in directories is not loaded.
A compromise can still be made in this case. Change loading order. Load configuration files from drop-in directories and then from an explicitly user-defined file. But it will be so weird. The user changes the path to the main configuration file, but not the path to the drop-in directories.

j-mracek:
Notes from discussion:

### Requirements

When user specify configuration files, it should be not overridden by drop in configuration

### Solutions

1. Do not load drop in overrides when user specify path to main configuration

- allow to be more compatible with e-conf library when main configuration file is loaded first and then overrides
- When user will override the default path with the same path, DNF might behave differently, because overrides will be not applied (`best`, `skip_if_unavailable`)
- It will require to discourage user to use `/etc/dnf/dnf.conf` because the value migh get overridden by drop in configuration files

2. Load main configuration file as last

- Main config will be the main config containing overrides with higher priority
- Overrides will be used only for overrides hard-coded internals, but not for main config
- We do not plan to ship any value in `/etc/dnf/dnf.conf` therefore there will be no difference with a default deployment
- Cannot disable drop-in overrides without additional configuration option

3. Load main configuration files with higher priority

- I think this solution (or solution 1) will be required when drop-in paths will be configurable.
- Can be used as an upgrade from `solution 2` when disablement of drop in overrides will be required

j-mracek:
I would prefer to use Solution 2 - `Load main configuration file as last`, but there might be some draw back that I overlooked.

j-mracek:
@keszybz May I ask you for your opinion?

ppisar:
I agree with the requirement that a user's configuration has a higher priority / is loaded later than a distribution-wide configuration.
One can have --config-file option to replace /etc/dnf/dnf.conf with a user supplied file and --config-directory option to replace /etc/dnf/libdnf5.conf.d directory with a user supplied directory.
Should we stop loading distribution-wide directory if user passes --config-file? I don't think so. The reason is that "dnf" and "dnf --config-file /etc/dnf/dnf.conf", albeit reading the same files, would behave differently. Then how could a user inhibit loading distribution-wide configuration? I would add another option --no-implicit-config. That would simply disable loading any configuration, both from directories and files, distribution-wide and localhost-specific which were not explicitly pointed with --config-file and --config-directory.

As a side note, I think that introducing configuration locations of various priorities is not much helpful: Instead of fighting over a line number in a single configuration file, users and packagers will fight over the highest file name. Therefore I'd like to decrease the amount of various locations to a bare minimum. I understand that configuration directories are easier for populating by third-parties than editing a single file. Therefore I recommend removing a distribution-wide file /usr/share/dnf5/libdnf.conf completely. Regarding /etc/dnf/dnf.conf file, I would discourage user's from using it. The only remaining function of it would be a backward compatibility and a last resort place with guaranteed hightest priority.

ppisar:
I think we will need to implement inhibition of system-wide configuration for DNF5 tests.

include/libdnf5/base/base.hpp
```diff
@@ -79,10 +79,16 @@ class Base {
...
+    /// Loads main configuration.
+    /// The file defined in the current configuration and files in the drop-in directories are used.
+    void load_config();
+
+    /// @deprecated Don't use it! It will be removed in Fedora 40. It was intended for internal use only.
     /// Call a function that loads the config file, catching errors appropriately
     void with_config_file_path(std::function<void(const std::string &)> func);
 
-    /// Loads main configuration from file defined by the current configuration.
+    /// @deprecated It will be removed in Fedora 40. It is redundant. It calls `load_config()`.
```
j-mracek, line 90: Proposing to remove `It will be removed in Fedora 40`, because the removal will have negative consequences on multiple projects.

jrohel: @j-mracek OK, removed.

j-mracek (approved):
LGTM

birdie-github:
> Based on discussions in the team, I changed the loading order of the configuration files. I also edited the description accordingly.

Is this now the status quo?

```
Resulting file loading order by default
("/usr/share/dnf5/libdnf.conf.d/60-something.conf" is skipped,
masked by the user file "/etc/dnf/libdnf5.conf.d/60-something.conf"):

    /etc/dnf/libdnf5.conf.d/20-user-settings.conf
    /usr/share/dnf5/libdnf.conf.d/50-something.conf
    /etc/dnf/libdnf5.conf.d/60-something.conf
    /etc/dnf/libdnf5.conf.d/80-user-settings.conf
    /usr/share/dnf5/libdnf.conf.d/90-something.conf
    /etc/dnf/dnf.conf
```

If files are loaded in this exact order, then `dnf.conf` settings will override whatever the user has specified, IOW `dnf.conf` settings cannot be overridden.

j-mracek:
> > Based on discussions in the team, I changed the loading order of the configuration files. I also edited the description accordingly.
> 
> Is this now the status quo?
> 
> ```
> Resulting file loading order by default
> ("/usr/share/dnf5/libdnf.conf.d/60-something.conf" is skipped,
> masked by the user file "/etc/dnf/libdnf5.conf.d/60-something.conf"):
> 
>     /etc/dnf/libdnf5.conf.d/20-user-settings.conf
>     /usr/share/dnf5/libdnf.conf.d/50-something.conf
>     /etc/dnf/libdnf5.conf.d/60-something.conf
>     /etc/dnf/libdnf5.conf.d/80-user-settings.conf
>     /usr/share/dnf5/libdnf.conf.d/90-something.conf
>     /etc/dnf/dnf.conf
> ```
> 
> If files are loaded in this exact order, then `dnf.conf` settings will override whatever the user has specified, IOW `dnf.conf` settings cannot be overridden.

DNF5 component will ship only empty `/etc/dnf/dnf.conf` therefore there is no requirement to override a content of the configuration file. The overrides will be used to overrides hard-coded configuration of the library.

Issue #399, opened 2023-03-27
Handle duplicate REPOIDs somehow


praiskup:
DNF4 warns like this:

> Repository REPOID is listed more than once in the configuration

And uses just one of them?  Or merges the config?

With DNF5 it seems the `REPOID` is loaded twice as `libdnf::Repo`, per:

```
$ dnf5 repo list
REPOID   Some repo name
REPOID   Some repo name - the duplicate
```

j-mracek:
This issue might create a problem that will be very difficult to discover. The report is related to repofrompath issue - https://github.com/rpm-software-management/dnf5/issues/794

praiskup:
The current solution though isn't ideal - this is how every single DNF5 command ends now for me:

> Failed to create repo "copr:copr.fedorainfracloud.org:project:foo": Id is present more than once in the configuration

The Copr inter-project dependency was designed so that some of those repositories _may_ be defined twice on the system, and that DNF is able to handle that.  Consider that user enables projects `foo` and `baz`, both `foo` and `baz` though depend on `bar`.  This basically leads to duplicated `bar` on the user's system.

m-blaha:
Oh, sorry for breaking it...

m-blaha:
@praiskup I've created PR https://github.com/rpm-software-management/dnf5/pull/814 which will hopfully fix the issue.

praiskup:
Hats off, it's a pleasure to report here. Thank you for such a quick reaction and fix!

PR #814, opened 2023-08-11
repo: Don't error out on duplicate repo id

m-blaha:
It turned out that with copr inter-project dependency it's possible to
have a repo id present in configuration files multiple times.
With this patch dnf5 will not error out, but only log an error and
ignore duplicate repo.
For repositories defined by --repofrompath option the previous behaviour
is preserved, and dnf5 still errors out.
Now the behaviour is consistent with dnf4.

Fixes: https://github.com/rpm-software-management/dnf5/issues/399


kontura:
LGTM

Issue #804, opened 2023-08-07
`dnf.conf` max parallel downloads error message


supakeen:
Hi, when setting `/etc/dnf/dnf.conf`'s `max_parallel_downloads` to a sufficiently large value you used to get an error in `dnf4`:

```
user@muja dnf € sudo dnf up
Error: Bad value of LRO_MAXPARALLELDOWNLOADS.
Error: Bad value of LRO_MAXPARALLELDOWNLOADS.
Error: Bad value of LRO_MAXPARALLELDOWNLOADS.
Error: Bad value of LRO_MAXPARALLELDOWNLOADS.
```

`dnf5` instead tells you it can't reach mirrors:

```
user@muja dnf € sudo dnf5 up
Updating and loading repositories:
Failed to download metadata (metalink: "https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-38&arch=x86_64") for repository "fedora-modular"
```

Which is a lot less descriptive.

evan-goode:
Thanks, I can reproduce this. It should definitely give a proper error message.

PR #819, opened 2023-08-14
Print error messages in nested errors

evan-goode:
Currently, for `libdnf5::Errors`, only the outermost error message is printed, even if the error contains nested errors, which can cause some crucial details to be left out.

Resolves https://github.com/rpm-software-management/dnf5/issues/804.


jan-kolarik:
I've added a simple CI test for this: https://github.com/rpm-software-management/ci-dnf-stack/pull/1364.

jan-kolarik (approved):
LGTM

PR #820, opened 2023-08-15
Add support for repos configuration overrides

jrohel:
After the repositories configuration is loaded, new code is run that loads the override files from the `/usr/share/dnf5/repos.override.d` and `/etc/dnf/repos.overide.d` directories.
  
The format of override files is the same as regular ".repo" files. However, override files can modify the settings of already existing repositories. They cannot create new repositories.
Overide files support glob in the repository ID (in the configuration file section name) to support bulk modification of repositories parameters.

**The order of processing the files with overrides:**
1. Creates an alphabetically sorted list of all filenames with overrides in the distribution override directory ("/usr/share/dnf5/repos.override.d") and the user override directory ("/etc/dnf/repos.overide.d"). If a file with the same name is present in both override directories, only the file from the user override directory is added to the sorted list. So, the distribution override file can be simply masked by creating a file with the same name in the user override directory.
2. Retrieves the overrides from the files in the sorted list. Follows the order. The override from the next file overrides the previous one - the last override value wins.

The PR also includes a commit that adds support for the ']' character in the section header name in `IniParser`. Allows to use more complex glob patterns for repository IDs.

-------------------------------------------------
### Example of the repos configuration override file:
```
# Enable `skip_if_unavailable` for all repositories
[*]
skip_if_unavailable = true

# And then disable `skip_if_unavailable` for repositories with id prefix "fedora"
[fedora*]
skip_if_unavailable = false
```

-------------------------------------------------------------------
### An example showing the order in which override files are processed:
**Files with user repos overrides:**
/etc/dnf/repos.overide.d/20-user-overrides.repo
/etc/dnf/repos.overide.d/60-something2.repo
/etc/dnf/repos.overide.d/80-user-overrides.repo
/etc/dnf/repos.overide.d/99-config-manager.repo

**Files with distribution repos overrides:**
/usr/share/dnf5/repos.overide.d/50-something2.repo
/usr/share/dnf5/repos.overide.d/60-something2.repo
/usr/share/dnf5/repos.overide.d/90-something2.repo

**Resulting file processing order:**
1. /etc/dnf/repos.overide.d/20-user-overrides.repo
2. /usr/share/dnf5/repos.overide.d/50-something2.repo
3. /etc/dnf/repos.overide.d/60-something2.repo
4. /etc/dnf/repos.overide.d/80-user-overrides.repo
5. /usr/share/dnf5/repos.overide.d/90-something2.repo
6. /etc/dnf/repos.overide.d/99-config-manager.repo

### DIFF
--- a/include/libdnf5/repo/repo_sack.hpp
+++ b/include/libdnf5/repo/repo_sack.hpp
@@ -74,9 +74,10 @@ public:
 
     /// Creates new repositories from the Base's configuration file (the /
     /// "config_file_path" configuration option) and from directories defined by
-    /// the "reposdir" configuration option.
+    /// the "reposdir" configuration option. Repository overrides are then applied.
     ///
-    /// Calls `create_repos_from_config_file()` and `create_repos_from_reposdir()`.
+    /// Calls `create_repos_from_config_file()`, `create_repos_from_reposdir()`,
+    /// and loads repository configuration overrides.
     void create_repos_from_system_configuration();
 
     /// Creates a new repository from a libsolv testcase file.
@@ -154,6 +155,10 @@ private:
 
     explicit RepoSack(const libdnf5::BaseWeakPtr & base) : base(base) {}
     explicit RepoSack(libdnf5::Base & base);
+
+    /// Loads repositories configuration overrides from drop-in directories. No new repositories are created.
+    /// Only the configuration of the coresponding existing repositories is modified.
+    void load_repos_configuration_overrides();
 
     WeakPtrGuard<RepoSack, false> sack_guard;
 
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -23,6 +23,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "conf/config.h"
 #include "solv/pool.hpp"
 #include "utils/dnf4convert/dnf4convert.hpp"
+#include "utils/fs/utils.hpp"
 
 #include "libdnf5/conf/config_parser.hpp"
 #include "libdnf5/conf/const.hpp"
@@ -37,45 +38,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <vector>
 
 namespace fs = std::filesystem;
-
-namespace {
-
-// Creates an alphabetically sorted list of all files with `file_extension` from `directories`.
-// If a file with the same name is in multiple directories, only the first file found is added to the list.
-// Directories are traversed in the same order as they are in the input vector.
-[[nodiscard]] std::vector<fs::path> create_sorted_file_list(
-    const std::vector<fs::path> & directories, std::string_view file_extension) {
-    std::vector<fs::path> paths;
-
-    for (const auto & dir : directories) {
-        std::error_code ec;
-        for (const auto & dentry : fs::directory_iterator(dir, ec)) {
-            const auto & path = dentry.path();
-            if (dentry.is_regular_file() && path.extension() == file_extension) {
-                const auto & path_fname = path.filename();
-                bool found{false};
-                for (const auto & path_in_list : paths) {
-                    if (path_fname == path_in_list.filename()) {
-                        found = true;
-                        break;
-                    }
-                }
-                if (!found) {
-                    paths.push_back(path);
-                }
-            }
-        }
-    }
-
-    // sort all drop-in configuration files alphabetically by their names
-    std::sort(paths.begin(), paths.end(), [](const fs::path & p1, const fs::path & p2) {
-        return p1.filename() < p2.filename();
-    });
-
-    return paths;
-}
-
-}  // namespace
 
 namespace libdnf5 {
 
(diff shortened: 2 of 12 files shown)


libdnf5/utils/iniparser.cpp
```diff
@@ -25,6 +25,49 @@ namespace libdnf5 {
...
+        } else {
+            if (ch == '\0' && ch == '\n' && ch == '\r') {
+                return std::string::npos;
+            }
+            if (ch == ']') {
+                bracket = true;
+            }
+        }
+    }
+}
```
ppisar, line 49: The find_end_of_section_name() will exhibit an undefined behavior if "str" does not contain an unescaped ']', '\0', '\n', or '\r'. In that case "idx" will grow indefinitely and the function will overread a heap buffer.
I recommend either limit the for() cycle with str.end() and return std::string::npos when reaching it, or document that the function assumes a '\0'-terminated string.

jrohel: @ppisar 
https://cplusplus.com/reference/string/string/operator[]/
`If pos is equal to the [string length](https://cplusplus.com/string::length), the function returns a reference to the null character that follows the last character in the string (which should not be modified).`

So if the string/memory is not corrupted, `std::string` will always have a null character at the end. It is defined in the standard.

The problem occurs when someone calls the function with the parameter `pos > str.size()` -> skips the terminating null character. But this is not possible from the outside, the function is in an anonymous namespace -> only available locally. But I will modify the code, what if someone makes a mistake in the future.

However, I found another bug in the code. How could I write that. In this line:
`if (ch == '\0' && ch == '\n' && ch == '\r') {`
there should be `if (ch == '\0' || ch == '\n' || ch == '\r') {`

Thanks for the review.

jrohel:
I fixed the `find_end_of_section_name` internal function in `IniParser` and added unit tests for ini files with incorrect syntax.

libdnf5/utils/iniparser.cpp
jrohel: I fixed the find_end_of_section_name internal function in IniParser and added unit tests for ini files with incorrect syntax.

jrohel:
I added CI tests https://github.com/rpm-software-management/ci-dnf-stack/pull/1368

Conan-Kudo:
We need to be able to move packaged repos out of `/etc` too, but this is a good start.

jrohel:
I rebased the code to resolve conflicts and deduplicate `create_sorted_file_list` function.

libdnf5/utils/iniparser.cpp
```diff
@@ -25,6 +25,53 @@ namespace libdnf5 {
...
+
+// Returns the input string without the undoubled characters ']'.
```
ppisar, line 55: s/without/with/

ppisar:
Otherwise it looks good for my eye untrained for C++.

libdnf5/utils/iniparser.cpp
jrohel: @ppisar OK. Comment corrected.

ppisar (approved):
I don't see any mistake. Yet I did not test the code.

include/libdnf5/repo/repo_sack.hpp
```diff
@@ -74,11 +74,16 @@ class RepoSack : public sack::Sack<Repo> {
...
+    /// the "reposdir" configuration option. Repository overrides are then applied.
     ///
-    /// Calls `create_repos_from_config_file()` and `create_repos_from_reposdir()`.
+    /// Calls `create_repos_from_config_file()`, `create_repos_from_reposdir()`,
+    /// and `load_repos_configuration_overrides()`.
     void create_repos_from_system_configuration();
 
+    /// Loads repositories configuration overrides from drop-in directories. No new repositories are created.
+    /// Only the configuration of the coresponding existing repositories is modified.
+    void load_repos_configuration_overrides();
```
j-mracek, line 85: What about to make it private?

libdnf5/base/base.cpp
```diff
@@ -38,45 +39,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 namespace fs = std::filesystem;
 
-namespace {
```
j-mracek, line 41: I am not 100% sure but the commit might be merged with previous commit or is there any reason why you would prefer the split?

libdnf5/utils/iniparser.cpp
```diff
@@ -25,6 +25,53 @@ namespace libdnf5 {
...
+            if (ch == ']') {
+                bracket = true;
+            }
+        }
+    }
+}
+
+// Returns the input string with the undoubled characters ']'.
+// A sequence of two chars ']' is returned as a single char ']'.
+std::string section_name_decode_bracket(const std::string & section_name) {
```
j-mracek, line 57: I guess it is required for glob support in section header like [fedora[2-8]]. @jrohel Is it correct?

j-mracek: What about to use the sema logic like we have in NEVRA parser - when the second `[` then the parser skips everything until the next `]`. What do you think?

include/libdnf5/repo/repo_sack.hpp
jrohel: OK. Good idea.

libdnf5/utils/iniparser.cpp
jrohel: @j-mracek OK. I changed the logic to be compatible with the NEVRA parser.

j-mracek:
There is one problem - PR does not contain documentation, but it can be delivered later, but it requires to create an issue.

Conan-Kudo:
> There is one problem - PR does not contain documentation, but it can be delivered later, but it requires to create an issue.

I've created an issue to track this as #883

Conan-Kudo (approved):

j-mracek:
LGTM

PR #830, opened 2023-08-22
dnf5: Configure secondary logger for dnf5

m-blaha:
The logger writes messages with level ERROR and higher directly to the 
stderr so that user can see them.
Currently some libdnf5 errors are buried in the log file and user is not aware
of the problem until they read the log. This change should increase the
visibility of such problems.

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -60,6 +60,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/logger/factory.hpp>
 #include <libdnf5/logger/global_logger.hpp>
 #include <libdnf5/logger/memory_buffer_logger.hpp>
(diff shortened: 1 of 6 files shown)


m-blaha:
The solving errors are now both logged and separately printed by dnf5 to stderr. Which results in duplicated messages in stderr, for example:

This part comes from the new stderr logger:
```
ERROR Problem: package filesystem-3.9-2.fc29.x86_64 requires setup, but none of the providers can be installed
  - conflicting requests
  - package setup-2.12.1-1.fc29.noarch is filtered out by exclude filtering
```
and this is current transaction errors output:
```
Problem: package filesystem-3.9-2.fc29.x86_64 requires setup, but none of the providers can be installed
  - conflicting requests
  - package setup-2.12.1-1.fc29.noarch is filtered out by exclude filtering
```
So the PR still needs a bit of work... But the priority is not high.

Issue #834, opened 2023-08-23
Enhance libdnf5::create_file_logger() to allow user-defined file names


m-blaha:
Currently only hard-coded "dnf5.log" file name is used.
But different tools (like dnf5daemon, dnf5-automatic) will need to log into different locations.

blocks: https://github.com/rpm-software-management/dnf5/issues/57

PR #835, opened 2023-08-24
logger: Possibility to set log file name

m-blaha:
I think responsibility for setting the log file name should be on the API
users, and not the library.
This patch enhances `create_file_logger()` method to accept log file name
parameter and also deprecates the variant without the name.

Resolves: https://github.com/rpm-software-management/dnf5/issues/834

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -71,6 +71,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <filesystem>
 #include <iostream>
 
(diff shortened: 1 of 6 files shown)


jan-kolarik:
Looks good. What about CLI users, do we want to allow changing log file name also for them? In that case, configuration option might probably be better?

m-blaha:
Thinking about it more, I would prefer to postpone introducing a new config option until someone has a real use-case for it.
(it's in-line with our "be reluctant to introduce new config options since it's hard to get rid of them" policy ;) )

jan-kolarik (approved):

PR #836, opened 2023-08-25
changes.rst: --noautoremove was added back

evan-goode:
The `--noautoremove` option was added back to the `remove` command, so we don't need to mention it in `changes.rst`.

See https://github.com/rpm-software-management/dnf5/pull/755


kontura (approved):
Thanks!

PR #850, opened 2023-08-30
Avoid reinstalling installonly packages marked for ERASE

kontura:
https://github.com/openSUSE/libsolv/issues/540

https://bugzilla.redhat.com/show_bug.cgi?id=2163474
https://issues.redhat.com/browse/RHEL-1253

### DIFF
--- a/libdnf5/rpm/solv/goal_private.cpp
+++ b/libdnf5/rpm/solv/goal_private.cpp
@@ -19,6 +19,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "goal_private.hpp"
 
+#include "rpm/package_sack_impl.hpp"
 #include "solv/pool.hpp"
 
 #include "libdnf5/common/exception.hpp"
@@ -223,6 +224,9 @@ int obsq_cmp(const Id * ap, const Id * bp, const ObsoleteCmpData * s_cb) {
     return *ap - *bp;
 }
 
+inline bool name_solvable_cmp_key(const Solvable * first, const Solvable * second) {
+    return first->name < second->name;
+}
 
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -283,6 +288,16 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
                 Id action = SOLVER_ERASE;
                 if (j < static_cast<int>(installonly_limit)) {
                     action = SOLVER_INSTALL;
+                } else {
+                    // We want to avoid reinstalling packages marked for ERASE, therefore
+                    // if identical unused provider is available we need to lock it.
+                    Solvable * s = spool.id2solvable(id);
+                    for (Id aup_id : available_unused_providers) {
+                        Solvable * aup_s = spool.id2solvable(aup_id);
+                        if (solvable_identical(s, aup_s) != 0) {
```
j-mracek, line 297: I am not sure whether the test is required. I understand we want to avoid reinstall but what about the install of another version to prevent  removal of dependent package.

j-mracek: But it is not wrong - the patch will prevents only reinstall and other cases are less likely.

kontura: I have updated the PR to match the old libdnf PR: https://github.com/rpm-software-management/libdnf/pull/1625

kontura:
Back-ported tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1392

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -270,6 +284,7 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
 
         const InstallonlyCmpData installonly_cmp_data{spool, running_kernel};
         q.sort(&installonly_cmp, &installonly_cmp_data);
+        std::sort(available_unused_providers.begin(), available_unused_providers.end(), nevra_solvable_cmp_key);
```
j-mracek, line 287: If I am not mistaken, the original patch for libdnf uses name cmp key and not nevra for both - sorting and lower_bound. May I ask you why is there a different implementation?

kontura: Oh right, I forgot about that.
I think I added that just because `nevra_solvable_cmp_key` is already implemented.
I can add a new function to compare just the names if you prefer that?
The `arch` and `evr` sorting is not needed here so it would be more efficient.

j-mracek: Yes sorting using only name will be more efficient,  but sorting by NEVRA is not wrong. Please feel free to implement it according to your preferences.

kontura: Ok, I have added the new comparator for names only.

I have also rebased the PR.

j-mracek (approved):
LGTM

Issue #614, opened 2023-06-14
Add ranges support to glob matching packages


kontura:
E.g. `*top[-a-f]*-1.[!1-4]*.x86_64` should match: `desktop-utils-1:1.0-1.x86_64` and `toped-1:1.0-1.x86_64`.

This is currently failing due to nevra parsing in: 
`libdnf::rpm::Nevra::parse(const std::string & nevra_str, const std::vector<Form> & forms)`

Specifically the `-` in ranges are mistaken for nevra separators.

There are [ci-dnf-stack tests](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/repoquery/globs.feature) that fail on this.

kontura:
There is still one test case that is not working: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/repoquery/globs.feature#L187-L197 so lets keep this open for now.

PR #852, opened 2023-08-30
Fix algorithm in package NEVRA filter

j-mracek:
DNF5 detects first, whether provided argument contains epoch (find ':') then it construct package strings IDs accordingly.

CI-tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1372

Closes: https://github.com/rpm-software-management/dnf5/issues/614

### DIFF
--- a/libdnf5/rpm/package_query.cpp
+++ b/libdnf5/rpm/package_query.cpp
@@ -1942,14 +1942,26 @@ void PackageQuery::PQImpl::filter_nevra(
         case libdnf5::sack::QueryCmp::LTE:
             filter_nevra_internal_str<cmp_lte>(pool, c_pattern, sorted_solvables, filter_result);
             break;
-        case libdnf5::sack::QueryCmp::GLOB:
-            filter_glob_internal<&libdnf5::solv::RpmPool::get_nevra>(
-                pool, c_pattern, *pkg_set.p_impl, filter_result, 0);
-            break;
-        case libdnf5::sack::QueryCmp::IGLOB:
-            filter_glob_internal<&libdnf5::solv::RpmPool::get_nevra>(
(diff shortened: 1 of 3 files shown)


libdnf5/solv/pool.cpp
```diff
@@ -79,6 +79,71 @@ Pool::~Pool() {
 }
 
 
+/// Returns a temporrary object allocated by pool_alloctmpspace
```
evan-goode, line 82: Spelling: temporrary -> temporary, also in `pool.hpp`

libdnf5/rpm/package_query.cpp
```diff
@@ -1943,12 +1945,24 @@ void PackageQuery::PQImpl::filter_nevra(
             filter_nevra_internal_str<cmp_lte>(pool, c_pattern, sorted_solvables, filter_result);
             break;
         case libdnf5::sack::QueryCmp::GLOB:
-            filter_glob_internal<&libdnf5::solv::RpmPool::get_nevra>(
-                pool, c_pattern, *pkg_set.p_impl, filter_result, 0);
+            found = pattern.find(':');
```
evan-goode, line 1948: Nit: I think it would be more readable to declare `found` down here instead of above the switch

evan-goode (changes requested):

libdnf5/solv/pool.cpp
j-mracek: Thank you

libdnf5/rpm/package_query.cpp
j-mracek: Fixed

evan-goode (approved):

Issue #855, opened 2023-08-31
forcearch expect to have rpm database initialised


tpgxyz:
Hi,

I’m trying to use dnf5 with mock and found this issue:
`[tpg@omv-rockpro64 dnf]$ sudo dnf5 --installroot /var/lib/mock//cooker-aarch64/root/ --releasever 0 install basesystem-build dwz dnf5 magic-devel --refresh --setopt=deltarpm=False --setopt=install_weak_deps=False --no-gpgchecks --no-docs --forcearch=aarch64
Updating and loading repositories:
warning: using dummy database, installs not possible
Failed to load system repo from root "/": no error `

chroot is empty(except dnf.conf for repository metadata) so as rpm db inside: 
`[tpg@omv-rockpro64 dnf]$ ls /var/lib/mock/cooker-aarch64/root/var/lib/rpm/ | wc -l
0
`

`[tpg@omv-rockpro64 dnf]$ cat /var/lib/mock/cooker-aarch64/root/var/log/dnf5.log 
2023-08-31T15:52:12+0000 [58910] INFO DNF5 start
2023-08-31T15:52:12+0000 [58910] DEBUG Loading default configuration from "/usr/share/dnf5/libdnf.conf"
2023-08-31T15:52:12+0000 [58910] DEBUG Configuration file "/usr/share/dnf5/libdnf.conf" not found
2023-08-31T15:52:12+0000 [58910] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/builddep_cmd_plugin.so"
2023-08-31T15:52:12+0000 [58910] INFO Loaded dnf5 plugin "builddep" ("/usr/lib64/dnf5/plugins/builddep_cmd_plugin.so") version "1.0.0"
2023-08-31T15:52:12+0000 [58910] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/changelog_cmd_plugin.so"
2023-08-31T15:52:12+0000 [58910] INFO Loaded dnf5 plugin "changelog" ("/usr/lib64/dnf5/plugins/changelog_cmd_plugin.so") version "1.0.0"
2023-08-31T15:52:12+0000 [58910] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/copr_cmd_plugin.so"
2023-08-31T15:52:12+0000 [58910] INFO Loaded dnf5 plugin "copr" ("/usr/lib64/dnf5/plugins/copr_cmd_plugin.so") version "0.1.0"
[...]

Issue #600, opened 2023-06-07
Is get_base_arch available in python?


bcl:
I need to be able to map the arch back to basearch. Currently I'm using code I borrowed from dnf.rpm, but I see that `libdnf/conf/vars.cpp` has a `get_base_arch` function. But I don't see it available anywhere in the Python API. If it isn't available, could it be added?

j-mracek:
Thank you very much for your report. @bcl May I ask you for additional information for which component and for which purpose you need the method. I am asking because there is a project that can provide such a mapping and we are looking for customers (DNF5 would be a customer).

bcl:
In https://github.com/osbuild/osbuild-composer/pull/3332/files#diff-cc316e51a644ae96748aa77302da8fa1d0e24bd9afa43163b3cd53b43eaaa7beR28 I am using it to setup the basearch substitution for `dnf5-json` in osbuild-composer.

And in lorax (this is an older PR, needs updating for latest dnf5 changes) - https://github.com/weldr/lorax/pull/1319/files#diff-6252bbe9a917d5070d076793ce4bf11888c3a7720c4ceaca07b0a9079fa00a90R71 I am using it to setup basearch for the package install.

I'm fine with using a library as long as it's well maintained. But I'd prefer to just depend on libdnf5 if it is possible.

PR #856, opened 2023-09-01
base: Call rpmReadConfigFiles() earlier

m-blaha:
Currently, libdnf5 initializes librpm only when detecting `arch` or
`releasever` variables. In case both these values are provided using
command line options (--releasever, --forcearch), `rpmReadConfigFiles()`
is not called at all, resulting in uninitialized librpm.

This patch moves `init_lib_rpm()` call to the `Base` class constructor,
ensuring that `rpmReadConfigFiles()` is always called.

Resolves: https://github.com/rpm-software-management/dnf5/issues/855

### DIFF
--- a/include/libdnf5/rpm/arch.hpp
+++ b/include/libdnf5/rpm/arch.hpp
@@ -1,0 +1,34 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef LIBDNF5_RPM_ARCH_HPP
+#define LIBDNF5_RPM_ARCH_HPP
+
+#include <string>
+#include <vector>
+
+
+namespace libdnf5::rpm {
+
+std::vector<std::string> get_supported_arches();
+
+}  // namespace libdnf5::rpm
+
+#endif  // LIBDNF5_RPM_ARCH_HPP
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -182,8 +182,9 @@ void Base::setup() {
     auto & config = get_config();
     auto & installroot = config.get_installroot_option();
     installroot.lock("Locked by Base::setup()");
+    auto vars = get_vars();
 
-    get_vars()->load(vars_installroot, config.get_varsdir_option().get_value());
+    vars->load(vars_installroot, config.get_varsdir_option().get_value());
 
     // TODO(mblaha) - move system state load closer to the system repo loading
     std::filesystem::path system_state_dir{config.get_system_state_dir_option().get_value()};
@@ -220,7 +221,7 @@ void Base::setup() {
     pool_setdisttype(**pool, DISTTYPE_RPM);
     // TODO(jmracek) - architecture variable is changable therefore architecture in vars must be synchronized with RpmPool
     // (and force to recompute provides) or locked
-    pool_setarch(**pool, get_vars()->get_value("arch").c_str());
+    pool_setarch(**pool, vars->get_value("arch").c_str());
     pool_set_rootdir(**pool, installroot.get_value().c_str());
 
     p_impl->plugins.post_base_setup();
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -109,10 +109,10 @@ static const char * get_base_arch(const char * arch) {
     return nullptr;
 }
 
-static void init_lib_rpm() {
+static void init_lib_rpm(const char * arch) {
     static bool lib_rpm_initiated{false};
     if (!lib_rpm_initiated) {
-        if (rpmReadConfigFiles(nullptr, nullptr) != 0) {
+        if (rpmReadConfigFiles(nullptr, arch) != 0) {
             throw RuntimeError(M_("failed to read rpm config files"));
         }
         lib_rpm_initiated = true;
@@ -128,7 +128,6 @@ static constexpr const char * DISTROVERPKGS[] = {
     "suse-release"};
 
 static std::string detect_arch() {
-    init_lib_rpm();
     std::string value{};
     char * tmp = rpmExpand("%{_host_cpu}", NULL);
     if (tmp != nullptr) {
@@ -143,7 +142,6 @@ static std::string detect_arch() {
 
 
 std::unique_ptr<std::string> Vars::detect_release(const BaseWeakPtr & base, const std::string & install_root_path) {
-    init_lib_rpm();
     std::unique_ptr<std::string> release_ver;
 
     libdnf5::rpm::RpmLogGuard rpm_log_guard(base);
@@ -433,6 +431,12 @@ void Vars::load(const std::string & installroot, const std::vector<std::string> 
 }
 
 void Vars::detect_vars(const std::string & installroot) {
+    const char * arch = nullptr;
+    if (contains("arch")) {
+        arch = get_value("arch").c_str();
+    }
+    init_lib_rpm(arch);
+
     set_lazy(
         "arch", []() -> auto { return std::make_unique<std::string>(detect_arch()); }, Priority::AUTO);
 
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -19,6 +19,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5/conf/vars.hpp"
 
+#include "rpm/arch_private.hpp"
 #include "rpm/rpm_log_guard.hpp"
 #include "utils/fs/file.hpp"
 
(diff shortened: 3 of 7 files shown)


j-mracek:
LGTM. @m-blaha Do you think whether is it possible to create a test for out CI?

m-blaha:
I can create a test for this particular occurrence of working with initialized rpm - make sure that `dnf5 install --releasever=xx --forcearch=yy` works correctly.

m-blaha:
PR with test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1376

libdnf5/base/base.cpp
```diff
@@ -82,13 +84,25 @@ namespace libdnf5 {
 static std::atomic<Base *> locked_base{nullptr};
 static std::mutex locked_base_mutex;
 
+static void init_lib_rpm() {
+    static bool lib_rpm_initiated{false};
+    if (!lib_rpm_initiated) {
+        if (rpmReadConfigFiles(nullptr, nullptr) != 0) {
```
pmatilai, line 90: I don't know what exactly --forcearch does inside libdnf, but since it's obviously about overriding the detected arch: it might make sense to pass it down to rpm too, ie as the second argument of rpmReadConfigFiles().

m-blaha: Hm, that might make things more complicated than I assumed :)
Is it possible to call `rpmReadConfigFiles()` multiple times? Currently the function is called very early, even before the `--forcearch` switch is processed (for example because detecting the releasever needs to query rpm database). So if it was possible, we could use the second call of `rpmReadConfigFiles` to "adjust" the architecture during processing of `--forcearch` switch. If this is not possible (which I assume), we need to handle arch and releasever variables differently and initialize rpmdb later.
Then we might run into kind of chicken and egg problem - reading the configuration needs variables (thus to have rpm initialized) and initialization of rpm needs `arch` value which is obtained from configuration (or --forcearch parameter).

m-blaha:
@j-mracek can you please review again?

- librpm is now initialized a bit later - in base.setup(). We need to do it after commandline argument parsing (to be able to extract --forcearch value), but before loading variables.
- added passing --forcearch value also to `rpmReadConfigFiles()` (according to Panu's comment)
- added handling of unknown arch used in --forcearch with explanatory message to the user

libdnf5/base/base.cpp
pmatilai: Calling rpmReadConfigFiles() multiple times is perfectly okay (rpmbuild itself does this), just remember to call rpmFreeRpmrc() for each round.

Conan-Kudo (approved):

include/libdnf5/rpm/arch.hpp
```diff
@@ -0,0 +1,37 @@
...
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef LIBDNF5_RPM_ARCH_HPP
+#define LIBDNF5_RPM_ARCH_HPP
+
+#include "libdnf5/common/exception.hpp"
+
+#include <sstream>
```
j-mracek, line 26: May be I overlooked something, but it looks like that 
```
#include "libdnf5/common/exception.hpp"

#include <sstream>
```
are not required.

m-blaha: Oh, this copy and paste (mainly because of this ugly license text). Looking forward to have short, neat SPDX identifiers instead...

m-blaha: Fixed.

include/libdnf5/rpm/arch.hpp
```diff
@@ -0,0 +1,34 @@
...
+#ifndef LIBDNF5_RPM_ARCH_HPP
+#define LIBDNF5_RPM_ARCH_HPP
+
+#include <string>
+#include <vector>
+
+
+namespace libdnf5::rpm {
+
+std::vector<std::string> get_supported_arches();
```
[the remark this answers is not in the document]
j-mracek: May be we need it, but I am confused another thing - structure from which the vector is generated.

It is c structure and it is not nice for our usage. The construct is <basearch>, <arches> therefore you have to iterate over the whole structure. If we would like to improve the search then we can create std::unordered_map<string, string> where key will be <arch> and the value <basearch>. The problem of this approach is that map is bigger then the original structure, and <Basearch> strings will be present multiple times, but it can be improved by another approach
```
std::unordered_map<string, int>  // <arch>, <index>
std::array<std::string>, where index could be used to retrieve the Basearch.
```
@m-blaha What do you think? @jrohel May I ask you for your opinion?

libdnf5/base/base.cpp
```diff
@@ -221,7 +233,13 @@ void Base::setup() {
     auto & installroot = config.get_installroot_option();
     installroot.lock("Locked by Base::setup()");
 
-    get_vars()->load(vars_installroot, config.get_varsdir_option().get_value());
+    auto vars = get_vars();
+    const char * arch = nullptr;
+    if (vars->contains("arch")) {
+        arch = vars->get_value("arch").c_str();
+    }
+    init_lib_rpm(arch);
```
j-mracek, line 241: May I ask you to add a note that when locking of vars will be available we should lock all vars or arch variable.

Are you sure that the order of following lines is correct? I somehow have feeling that we should first load and then init rpm with architecture.
```
init_lib_rpm(arch);
vars->load(vars_installroot, config.get_varsdir_option().get_value());
```

include/libdnf5/rpm/arch.hpp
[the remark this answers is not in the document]
m-blaha: It's used by dnf5 to generate an error message if unsupported arch is used.

libdnf5/base/base.cpp
m-blaha: The thing is, that during `vars->(load)` the architecture and releasever are also detected (if missing). And this detection needs rpm correctly initialized.

include/libdnf5/rpm/arch.hpp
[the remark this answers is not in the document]
m-blaha: Regarding the `ARCH_MAP` data structure, you are right. I guess it was copied from libdnf where probably the C structure was needed. In dnf5 we can replace it by C++ structure (like std::unordered_map).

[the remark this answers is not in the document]
j-mracek: may be `std::map` could be better if you want to have get_supported_arches ordered without ordering.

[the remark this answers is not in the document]
jrohel: - I would prefer `std::map` over `std::unordered_map` for these tiny maps.
- And inside `std::string_view`. No need to copy those strings to the heap.

But it will still take up more memory than the original code.

libdnf5/base/base.cpp
m-blaha: I've added a comment on it.

include/libdnf5/rpm/arch.hpp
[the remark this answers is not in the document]
m-blaha: @j-mracek do you want me to change also the underlying `ARCH_MAP` structure, or is TODO sufficient for this PR?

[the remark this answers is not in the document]
jrohel: @m-blaha Why do you actually want to replace the C structure? How many times do we call `get_base_arch`? Is it worth it to construct a map?
The cost to create a map will be several times higher than calling `get_base_arch`.

Is the requirement or do we need to intensively call `get_base_arch`? (Note: If so, then code can be written to create a sorted reverse mapping (array of native -> base structs) with little cost directly at compile time.)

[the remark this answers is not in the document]
j-mracek: I would like to clarify internal stricture, because there is a method (public) which depends on. The method returns a vector with strings.

libdnf5/base/base.cpp
```diff
@@ -183,7 +195,17 @@ void Base::setup() {
...
-    get_vars()->load(vars_installroot, config.get_varsdir_option().get_value());
+    // Initialize rpm before calling vars->load(), because it also detects
+    // default arch and releasever values, and the detection relies on rpm
+    // being initialized.
+    auto vars = get_vars();
+    const char * arch = nullptr;
+    if (vars->contains("arch")) {
+        arch = vars->get_value("arch").c_str();
+    }
+    init_lib_rpm(arch);
```
j-mracek, line 206: I have additional  comment. If I am not mistaken, we pass rpm an architecture, but in this time we can still use a different architecture, because architecture might be modify or set by `vars->load(vars_installroot, config.get_varsdir_option().get_value());`. It means that we might use for POOL a different architecture that we provide to RPM. @pmatilai If we will use the same architecture in libsolv pool and we will initialize rpm config `rpmReadConfigFiles(nullptr, arch)` with the same architecture, could we stop use `RPMPROB_FILTER_IGNOREARCH` for transaction where architecture used for transaction calculation differs from the system architecture.

m-blaha: Gosh, that's right! Thanks for catching it. So I'll move the code back to vars.cpp and just make sure that the code is called unconditionally.

m-blaha: @j-mracek the issue with architecture provided by vars file should be resolved. Now the rpm is initialized during Vars::detect_vars() call.

include/libdnf5/rpm/arch.hpp
j-mracek, line 30: There is somehow related issue - https://github.com/rpm-software-management/dnf5/issues/600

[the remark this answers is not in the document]
j-mracek: I am fine even with TODO.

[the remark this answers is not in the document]
j-mracek: @jrohel - I was thinking that the map is created during compilation. Or did I overlooked something?

libdnf5/conf/vars.cpp
```diff
@@ -433,12 +377,21 @@ void Vars::load(const std::string & installroot, const std::vector<std::string>
 }
 
 void Vars::detect_vars(const std::string & installroot) {
+    const char * arch = nullptr;
+    if (contains("arch")) {
+        arch = get_value("arch").c_str();
+    }
+    init_lib_rpm(arch);
+
```
j-mracek, line 385: This is very minor comment - you already have an information for test whether arch is set or not therefore it is not required to ask set_lazy to check it again. 
```
    set_lazy(
        "arch", []() -> auto { return std::make_unique<std::string>(detect_arch()); }, Priority::AUTO);
```

libdnf5/conf/vars.cpp
```diff
@@ -433,12 +377,21 @@ void Vars::load(const std::string & installroot, const std::vector<std::string>
 }
 
 void Vars::detect_vars(const std::string & installroot) {
+    const char * arch = nullptr;
+    if (contains("arch")) {
+        arch = get_value("arch").c_str();
+    }
+    init_lib_rpm(arch);
```
j-mracek, line 384: I think this is the perfect place. Thank you.

j-mracek:
LGTM

Issue #858, opened 2023-09-01
repo info: Sort repositories in output alphabetically


m-blaha:
Currently repos are printed in random order.
Normally the priority of the issue would be low, but since it makes `repo info` CI tests flaky (see https://github.com/rpm-software-management/dnf5/actions/runs/6047597173/job/16412339868), I would argue for a higher priority.

ppisar:
That makes sense.

PR #860, opened 2023-09-01
dnf5: Sort repos in 'repo info' command output

m-blaha:
With this should `repo info` always print repositories alphabetically sorted.

Fixes: https://github.com/rpm-software-management/dnf5/issues/858


jan-kolarik (approved):
LGTM, thanks!

Issue #863, opened 2023-09-04
Missing libdnf5-plugins in license enumaration in COPYING.md


ppisar:
"libdnf5-plugins" files are LGPL-2.1-or-later, but "COPYING.md" file does not mention that directory in "LGPLv2.1+" section":

~~~~
LGPLv2.1+
---------
Alternatively, all of the source code in 'libdnf5', 'libdnf5-cli', 'include'
and 'bindings' subdirectories of the libdnf source code distribution
~~~~

The directory should be named there. Otherwise, it makes in impression that the code is gpl-2.0-or-later:

~~~~
GPLv2+
------
The entire code base may be distributed under the terms of the
GNU General Public License version 2 and any later version (GPLv2+).
~~~~

This was spotted in <https://github.com/rpm-software-management/dnf5/pull/857#discussion_r1314297085>.

ppisar:
Addressed in PR#864.

PR #864, opened 2023-09-04
doc: Add libdnf5-plugins to lgpl-2.1-or-later list in COPYING.md

ppisar:
https://github.com/rpm-software-management/dnf5/issues/863


jrohel:
Thanks

Issue #172, opened 2022-11-24
Apply --debugsolver option for modules and groups


j-mracek:
- [x] [modules](https://github.com/rpm-software-management/dnf5/pull/870)
- [x] [groups](https://github.com/rpm-software-management/dnf5/pull/874)

Follow-up issue:
- #963

jan-kolarik:
@j-mracek Can we consider this as completed? I am not really sure what is really meant by these optional features.

j-mracek:
> @j-mracek Can we consider this as completed? I am not really sure what is really meant by these optional features.

In DNF4 `--debugsolver` only provides an information what solver sees, but it does not dump information about dnf configuration, repo configuration or variables. The optional items were only a idea wheres the option may dump additional information related to configuration, repo configuration or variables including releasever.

@jan-kolarik What do you think - shall we add it or keep it like it is?

jan-kolarik:
> > @j-mracek Can we consider this as completed? I am not really sure what is really meant by these optional features.
> 
> In DNF4 `--debugsolver` only provides an information what solver sees, but it does not dump information about dnf configuration, repo configuration or variables. The optional items were only a idea wheres the option may dump additional information related to configuration, repo configuration or variables including releasever.
> 
> @jan-kolarik What do you think - shall we add it or keep it like it is?

I guess that the logic behind this is already implemented [here](https://github.com/rpm-software-management/dnf5/pull/877) and [here](https://github.com/rpm-software-management/dnf5/pull/875)? So the only thing would be to possibly redirect it to the file instead to the standard output?

jan-kolarik:
I am closing this as completed. Optional functionality is now tracked in a separate issue.

PR #870, opened 2023-09-05
Add debug solver data for modules

jan-kolarik:
Save debug solver data into the `debugdir` / "modules" subdirectory after module goal is resolved.
Also, use canonical format when displaying debug data file path to the user and show just the path to the parent directory as there could be more debug data subdirectories (e.g. like "modules") inside.

Related issue: https://github.com/rpm-software-management/dnf5/issues/172.


ppisar:
Looks good to me.

Conan-Kudo (approved):

PR #871, opened 2023-09-05
Remove accidentally pushed files

jan-kolarik:
Delete files that were pushed by mistake during refactoring.


evan-goode (approved):

evan-goode:
Probably should keep these editor-specific files out of the project .gitignore, but I find `~/.config/git/ignore` useful to automatically ignore this kind of thing.

Issue #405, opened 2023-03-28
DNF config-manager Plugin


j-mracek:
Documentation if dnf4 command - https://dnf-plugins-core.readthedocs.io/en/latest/config_manager.html

I am proposing to not keep DNF4 compatibility, because the syntax was not the best one and we also have new requirements due to support of drop-in directories for configuration.

- [x] Investigate options how to implement configuration overrides for main config
    - implemented in https://github.com/rpm-software-management/dnf5/pull/813
- [x] Investigate options how to implement configuration overrides for repositories
    - implemented in https://github.com/rpm-software-management/dnf5/pull/820
 - Implement mechanism to synchronize dnf and subscription-manager

- [ ] implement configuration management and repositories
    - Will be implemented in https://github.com/rpm-software-management/dnf5/pull/907
 - Configuration of repositories could be handled by a different command that for main repositories (under `repo` command)
 - missing feature in DNF4 - unable do dump repository configuration
     - Implementation of master configuration dump and repository configuration dump https://github.com/rpm-software-management/dnf5/pull/877. Unlike dnf4, it can be combined with any command - they are global dnf5 options, not part of config-manager. 
     - In addition, dump of variables are implemented. https://github.com/rpm-software-management/dnf5/pull/875. (missing in DNF4)
 - `--save` option was somehow strange in DNF4

Additional inspiration for handling of drop-in configuration:
https://github.com/rpm-software-management/libdnf/issues/1387
https://github.com/rpm-software-management/dnf/pull/1414

ptoscano:
>     * [ ]  Investigate options how to implement configuration overrides for repositories
> 
>     * Implement mechanism to synchronize dnf and subscription-manager

I started a separate discussion about this, i.e. #422.

inknos:
Issues reported by the lack of config-manager plugin here https://bodhi.fedoraproject.org/updates/FEDORA-2023-69dfac15db#comment-3063374

msrb:
Fedora CI relies on the config-manager plugin as well:

```
Unknown argument "config-manager" for command "dnf5". Add "--help" for more information about the arguments.
```

juhp:
Currently I use `dnf-3 config-manager` as a workaround in the meantime.

msrb:
Thanks!

Is `dnf-3 config-manager` something we can expect to exist in Fedora 40+ (and later CentOS Stream, etc)? Or is it just a temporary workaround that might go away in the not so distant future?

[...]

PR #875, opened 2023-09-06
[dnf5] Implement new argument "--dump-variables"

jrohel:
If dnf5 is run with this argument, variable values are added to standard output.

The old DNF4 supports this argument only with the `config-manager` plugin. If the user wants to know the values of the variables with which a certain command will run, he must first use `config-manager` and then run the desired command with the same settings (same environment and configuration files) and hope that nothing changed in between (e.g. thanks to a plugin).

In other words, this PR implements part of the functionality of the dnf4 `config-manager` plugin, but for general use.

Related to: https://github.com/rpm-software-management/dnf5/issues/405

### DIFF
--- a/dnf5/include/dnf5/context.hpp
+++ b/dnf5/include/dnf5/context.hpp
@@ -97,6 +97,11 @@ public:
 
     bool get_quiet() const { return quiet; }
 
+    /// Set to true to print information about variables
+    void set_dump_variables(bool enable) { this->dump_variables = enable; }
+
+    bool get_dump_variables() const { return dump_variables; }
+
     Plugins & get_plugins() { return *plugins; }
 
     libdnf5::Goal * get_goal(bool new_if_not_exist = true);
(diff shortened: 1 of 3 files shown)


m-blaha:
Just an idea - would it make sense to implement this as a new command (e.g. `dnf5 debug`)? It could present to the user more information - variables, all applied config options (since configuration is spread among multiple files, this might be interesting), versions of libraries in use (libdnf5, libsolv, librpm, librepo...) and so on.

Copr plugin has the same subcommand used for similar purpose.

jrohel:
Rebased to resolve conflicts.

ppisar:
There is missing a documentation in doc/dnf5.8.rst.

ppisar:
I somehow find it quite obtrusive that "./dnf5/dnf5 --dump-variables" does not work also alone:

~~~~
$ ./dnf5/dnf5 --dump-variables
Missing command. Add "--help" for more information about the arguments.
~~~~

But maybe that's on purpose.

jrohel:
@ppisar
I intended that `--dump-variables` would print out what variables the given command was running with.
I am considering whether to modify the behavior of dnf5. Now a command must always be entered, otherwise dnf5 will print "Missing command..."

ppisar (approved):
Documentation and standalone invocation works now.

PR #878, opened 2023-09-07
goal: Do not add lower versions to upgrade jobs

m-blaha:
For UPGRADE* actions, do not add packages that are older than the
currently installed versions to the libsolv job. This decreases the
number of potential solutions and helps resolve some edge cases.

See https://issues.redhat.com/browse/RHEL-1448 for one of those edge
cases.

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1380

### DIFF
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -1478,15 +1478,16 @@ GoalProblem Goal::Impl::add_up_down_distrosync_to_goal(
         filter_candidates_for_advisory_upgrade(base, query, settings.advisory_filter.value(), obsoletes);
     }
 
-    if (minimal) {
-        query.filter_earliest_evr();
-    }
-
     switch (action) {
         case GoalAction::UPGRADE_MINIMAL:
         case GoalAction::UPGRADE:
             query.filter_available();
             remove_older_versions(query);
+
+            if (minimal) {
+                query.filter_earliest_evr();
+            }
+
             // Given that we use libsolv's targeted transactions, we need to ensure that the transaction contains both
             // the new targeted version and also the current installed version (for the upgraded package). This is
             // because if it only contained the new version, libsolv would decide to reinstall the package even if it
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -71,6 +71,30 @@ void add_obsoletes_to_data(const libdnf5::rpm::PackageQuery & base_query, libdnf
     data |= obsoletes_query;
 }
 
+/// Removes from the query available packages with versions lower than or equal
+/// to the installed ones. Similar to filter_upgrades(), but retains also
+/// installed packages and packages whose names are not installed.
+/// The method is called during processing UPGRADE* actions to prune versions
+/// available to the solver only to real upgrades. There are edge cases
+/// where adding old versions to solver job cause problems.
+/// See https://issues.redhat.com/browse/RHEL-1448 for details about such case.
+static void remove_older_versions(libdnf5::rpm::PackageQuery & query) {
+    libdnf5::rpm::PackageQuery installed(query.get_base(), libdnf5::rpm::PackageQuery::ExcludeFlags::IGNORE_EXCLUDES);
+    installed.filter_installed();
+    installed.filter_latest_evr();
+
+    // Keep only packages from priority repositories. Otherwise libsolv may
+    // upgrade to higher version of package from non-priority repo.
+    // Previously we passed to libsolv all versions of package and it was libsolv's
+    // responsibility to select correct upgrade.
+    query.filter_priority();
+
+    libdnf5::rpm::PackageQuery to_remove(query);
+    to_remove.filter_available();
+    to_remove.filter_nevra(installed, libdnf5::sack::QueryCmp::LTE);
+
+    query.difference(to_remove);
+}
 
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


libdnf5/base/goal.cpp
```diff
@@ -509,6 +509,13 @@ GoalProblem Goal::Impl::add_specs_to_goal(base::Transaction & transaction) {
                         base, query, settings.advisory_filter.value(), cfg_main.get_obsoletes_option().get_value());
                 }
 
+                // Remove packages of versions lower than the installed ones.
+                // Substract downgrades instead of filter_upgrades() to keep
+                // installed packages in the query.
+                rpm::PackageQuery tmp_downgrades(query);
```
j-mracek, line 515: I am quite curious, why you want to keep packages with the same version as installed in the query?

m-blaha: Well, I don't :) I've just overlooked it. Thanks for noticing.

m-blaha: Fixed.

m-blaha: > I recall that "dnf upgrade" reinstalls packages which were originally installed unsigned from a local build and then they appeared properly signed in a repository. That could warrant leaving same versions in the query.

I need to check this. We do not have a test for it, I suppose?

m-blaha: @ppisar it seems that you are refering to behaviour described in this libsolv issue: https://github.com/openSUSE/libsolv/issues/287
The trigger for re-installation was not whether the package was or was not signed, but more generally that the package was changed (for example has different build time).
This behavior was changed in dnf4 here: https://github.com/rpm-software-management/libdnf/pull/786 (and the same behavior we have in dnf5).
So dnf4 does not reinstall package with the same nevra but different buildtime for quite a long time. But distrosync command should trigger the reinstallation.

libdnf5/base/goal.cpp
```diff
@@ -71,6 +71,19 @@ void add_obsoletes_to_data(const libdnf5::rpm::PackageQuery & base_query, libdnf
     data |= obsoletes_query;
 }
 
+/// Removes from the query available packages with versions lower than or equal
+/// to the installed ones. Similar to filter_upgrades(), but retains also
+/// installed packages and packages whose names are not installed.
+static void remove_older_versions(libdnf5::rpm::PackageQuery & query) {
+    libdnf5::rpm::PackageQuery installed(query.get_base());
```
j-mracek, line 78: I would suggest to initialize query without excluded packages

`libdnf5::rpm::PackageQuery installed(query.get_base(), rpm::PackageQuery::ExcludeFlags::IGNORE_EXCLUDES)`

libdnf5/base/goal.cpp
```diff
@@ -71,6 +71,19 @@ void add_obsoletes_to_data(const libdnf5::rpm::PackageQuery & base_query, libdnf
...
+    installed.filter_installed();
```
j-mracek, line 79: This is very weak suggestion because it does not change the result.
What about to add additional filter  - `latest()`? Installonly packages will be handled in the next query only once.

libdnf5/base/goal.cpp
```diff
@@ -509,6 +522,8 @@ GoalProblem Goal::Impl::add_specs_to_goal(base::Transaction & transaction) {
                         base, query, settings.advisory_filter.value(), cfg_main.get_obsoletes_option().get_value());
                 }
 
+                remove_older_versions(query);
```
j-mracek, line 525: May I ask you to reference the bug or write a note why it is important?

libdnf5/base/goal.cpp
m-blaha: Makes sense, fixed.

libdnf5/base/goal.cpp
m-blaha: Added. (although I did not measure impact on performance here).

libdnf5/base/goal.cpp
m-blaha: Sure. Added to `remove_older_versions()` function description.

libdnf5/base/goal.cpp
ppisar: Yes, that libsolv ticket describes my experience. If DNF stopped doing it, the then j-mracek's proposal is fine.

j-mracek (approved):

m-blaha:
Looks like some tests with repo priorities started failing. I'll investigate.

j-mracek:
We have a problem, if we do this we will have a problem with repo priority test cases. Basically we have two options, we can keep all packages from proiority repo and then apply removal of old packages (logic from check update)

m-blaha:
> We have a problem, if we do this we will have a problem with repo priority test cases. Basically we have two options, we can keep all packages from proiority repo and then apply removal of old packages (logic from check update)

I've added code to pre-select upgrade candidates considering also repository priorities. However, I'm starting to feel that we might be exceeding our responsibilities by intervening in matters that should be handled by libsolv. To be honest, I'm leaning towards closing this PR and simply documenting in the upgrade code why including all versions of the package in the libsolv job is important.

m-blaha:
@ppisar unfortunately, the `--best` does not help here and the result remains the same.
Unversioned obsoletes are obstacle, that's hard to get around. I don't see other solution than rename the package and start again from scratch.

PR #879, opened 2023-09-07
Add ccache Github workflow 

inknos:


.github/workflows/ccache-build.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      # this step needs to be incorporated in our container that we use for testing
+      - name: Install dependencies
+        run: |
+          dnf install -y ccache clang && dnf builddep -y ./dnf5.spec
```
inknos, line 26: @m-blaha about this, would you rather update our current container that we use for CI or should we create a separate image for building?

inknos:
looks like the names the workflows are not updated. I'll retry to run them tomorrow

.github/workflows/ccache-build.yaml
m-blaha: I'd prefer to keep one container.

m-blaha (approved):
LGTM, thanks!

.github/workflows/ccache-build.yaml
inknos: also, the COPR secret needs to be fixed. I'll do those two fixes before merging

PR #885, opened 2023-09-15
Don't drop cached packages when only removing packages (RhBug:2237883)

jan-kolarik:
In order to have a easier workflow for making package caching exceptions for some commands and also to match the `keepcache` behavior with the original dnf in this matter, the cached files are now removed only when `keepcache` option is turned off at the moment.

We don't want to remove the stored packages cache after running the `remove` command transaction, therefore we override the `keepcache` option to true. This is useful for use cases when the previous transaction downloaded a large amount of data, but didn't finished correctly. Then we'd like to fix the system state by removing some problematic package. We don't want to lose our cached data in this case.

CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1381
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2237883

### DIFF
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -124,6 +124,10 @@ public:
     /// To watch progress or trigger actions during specific transactions events,
     /// setup the `callbacks` object.
     ///
+    /// After a successful transaction, any temporarily downloaded packages are removed
+    /// if the 'keepcache' option is set to 'false' and the transaction involved an inbound action.
+    /// Otherwise, the packages are preserved on the disk.
+    ///
     /// @return An enum describing the result of running the transaction.
     TransactionRunResult run();
 
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -869,18 +869,21 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
 
     if (ret == 0) {
         // removes any temporarily stored packages from the system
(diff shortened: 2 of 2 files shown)


jan-kolarik:
The current PR is creating an exception for the `remove` command to not removed cached packages from the disk disregarding the currently setup `keepcache` option. What about other related commands, like `autoremove` or `group remove`? Or do we want to configure the keeping cache behavior in general for all transactions with only package removals? In that case, we'd need to document that properly.

dnf5/commands/remove/remove.cpp
```diff
@@ -61,6 +61,7 @@ void RemoveCommand::configure() {
     auto & context = get_context();
     context.set_load_system_repo(true);
     context.set_load_available_repos(Context::LoadAvailableRepos::NONE);
+    context.base.get_config().get_keepcache_option().set(libdnf5::Option::Priority::RUNTIME, true);
```
j-mracek, line 64: I have a controversial idea. Please take a seat before continue the reading. What about to move the logic to core behavior. It means in `transaction.cpp` extend the condition `if (!config.get_keepcache_option().get_value()) {` by `OR` transaction does not have any inbound package. I believe that kind of information is available to that part of the code. This behavior will unify behavior with daemon and other API users. Please take it as subject for a discussion.

j-mracek:
LGTM

j-mracek (approved):

j-mracek:
@jan-kolarik Thank you very much for your contribution.

Issue #587, opened 2023-06-01
Implement logic to check whether package requires system reboot


evan-goode:
DNF 4 has `base.reboot_needed(): bool` which checks whether a system reboot is recommended following the transaction. In DNF 4, the logic is very crude, this function simply checks the set of packages involved in the transaction against a hardcoded set of "important" packages which includes `systemd`, `kernel`, and `glibc`.

In DNF 5, we should investigate better ways to do this check. jmracek proposed using the `<reboot_suggested>True</reboot_suggested>` advisory, which I think is the best approach, but has the disadvantage that advisories are mostly only supported by RHEL and Fedora repos. I'm not sure how someone would flag a package in their COPR repository as "needs reboot".

Another idea is to examine the files changed in the transaction and use some heuristic to make a decision. Or some combination of these three methods (hardcoded list, advisories, checking files).

This logic will be needed for the DNF 5 needs-restarting plugin (https://github.com/rpm-software-management/dnf5/issues/389) and by the `reboot` [option](https://dnf.readthedocs.io/en/latest/automatic.html?highlight=reboot#commands-section), if we choose to keep it, in whatever replaces DNF automatic (https://github.com/rpm-software-management/dnf5/issues/57).

API users may also want to use this functionality, see https://bugzilla.redhat.com/show_bug.cgi?id=2164835.

- [x] Open a discussion with fedora-devel

Issue #743, opened 2023-07-20
DNF needs-restarting Plugin


j-mracek:

evan-goode:
See also https://github.com/rpm-software-management/dnf5/issues/587

j-mracek:
My apology for the movement.

m-blaha:
FYI - this dnf4 issue reveals a problem with restart needed detection in case the RTC is set in the local time zone: https://bugzilla.redhat.com/show_bug.cgi?id=2237223

evan-goode:
> FYI - this dnf4 issue reveals a problem with restart needed detection in case the RTC is set in the local time zone: https://bugzilla.redhat.com/show_bug.cgi?id=2237223

Thanks, I think asking systemd for the boot time will solve that issue for the DNF 5 implementation: https://github.com/rpm-software-management/dnf5/pull/887/commits/cd366af6578584368ac0a1d37ce77c18aa1ba2d6

If desired I can backport that to DNF 4 needs-restarting. As I understand it this issue would mostly affect users who (1) are in a time zone later than UTC and (2) are dual-booting Windows.

PR #887, opened 2023-09-15
Implement dnf5 needs-restarting plugin

evan-goode:
I'd like the DNF 5 needs-restarting command to behave a bit different from `dnf4 needs-restarting`. Here's what I propose:

- `dnf5 needs-restarting` (no flag): I'd like this to simply report whether the system needs to be restarted. If there are "core packages" (such as `kernel` or `glibc`) that have been updated since boot, or if any advisory with the `reboot_suggested` flag applies to any of the packages that have been updated/installed since boot, `dnf5 needs-restarting` will list these packages and exit with code 1 to indicate a reboot is needed. Otherwise, it will exit with code 0. This proposed behavior is the same as the current behavior of `dnf4 needs-restarting -r`, but it would properly honor the `reboot_suggested` advisory rather than simply checking the hardcoded list of "core packages".

    I think this should be the default behavior of `needs-restarting` (rather than being behind a flag) because it's probably the most common use case. Here's an example of a user running `needs-restarting` when they needed `needs-restarting -r`: https://bugzilla.redhat.com/show_bug.cgi?id=2237223.

- Regarding the behavior of `dnf4 needs-restarting` (no flag): currently, this command lists processes with open outdated files and recommends the user restart them. I think we should leave this behavior out of `dnf5 needs-restarting` for two reasons. First, [tracer](https://github.com/FrostyX/tracer) and `dnf-plugins-extras-tracer` implement this approach well, and we don't need to duplicate their functionality. Second, this whole strategy of looking for open files is a bit of a hack and IMO shouldn't be included in the core DNF plugins. It might work most of the time, but there are nuances and edge cases here, e.g. when a process maps a library without opening it: https://bugzilla.redhat.com/show_bug.cgi?id=1131307, or if, hypothetically, a process reads a file into memory, closes it, then encrypts the copy in memory. There's just no way to tell that the process is "using" that outdated file.

- `dnf5 needs-restarting -s`: I'd like this to list systemd services that need to be restarted, but rather than check whether the service has any open outdated files, check the dependencies of the package that provides the service. I haven't implemented this yet in this PR, but here's an outline:

```
for each running systemd $service
    find the $package that provides the unit file for $service, if any
        recursively get all $dependencies of $package
        if $package or any $dependency has been updated since $service started
            recommend restart of $service
```

For https://github.com/rpm-software-management/dnf5/issues/743

Resolves https://github.com/rpm-software-management/dnf5/issues/587


ppisar:
> dnf5 needs-restarting will list these packages and exit with code 1 to indicate a reboot is needed.

Would it be possible the reverse the exit codes? Now the codes match to "does-not-need-restart" question.

evan-goode:
> > dnf5 needs-restarting will list these packages and exit with code 1 to indicate a reboot is needed.
> 
> Would it be possible the reverse the exit codes? Now the codes match to "does-not-need-restart" question.

Hmm, yes using code 0 (truthy) to mean "no restart" does make scripts read a bit awkward, I see your point:

```
if dnf5 needs-restarting; then
  echo The system DOES NOT need a reboot
fi
```

Similar tools like tracer, Debian's checkrestart, and `dnf4 needs-restarting` use these exit codes though, so I think still it would be best to keep consistent with them. checkrestart [mentions](https://manpages.debian.org/bullseye/debian-goodies/checkrestart.8.en.html) the exit code may be consumed by monitoring tools like Nagios, where the convention is "exit zero is good".

I think of "no reboot needed" as the default, unsurprising, comfortable result and "reboot needed" as the less common result that needs to be handled. The exit codes make more sense from that perspective.

evan-goode:
Alright, I think this is ready for review now.

dnf5-plugins/needs_restarting_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+add_library(needs_restarting_cmd_plugin MODULE needs_restarting.cpp needs_restarting_cmd_plugin.cpp)
+
+# disable the 'lib' prefix in order to create needs_restarting_cmd_plugin.so
+set_target_properties(needs_restarting_cmd_plugin PROPERTIES PREFIX "")
+
+pkg_check_modules(SDBUS_CPP REQUIRED sdbus-c++)
+
+target_link_libraries(needs_restarting_cmd_plugin PRIVATE libdnf5 libdnf5-cli ${SDBUS_CPP_LIBRARIES})
+target_link_libraries(needs_restarting_cmd_plugin PRIVATE dnf5)
```
jan-kolarik, line 12: Minor: just, is there any reason why we don't have all the links inside the single `target_link_libraries` call?

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+    return dependencies;
+}
+
+void NeedsRestartingCommand::system_needs_restarting(Context & ctx) {
+    const auto boot_time = get_boot_time();
+
+    libdnf5::rpm::PackageQuery base_query{ctx.base};
+
+    libdnf5::rpm::PackageQuery installed{base_query};
```
jan-kolarik, line 169: I don't see `base_query` used anywhere else, could you drop it and make this: `libdnf5::rpm::PackageQuery installed{ctx.base};`?

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+    installed.filter_installed();
+
+    libdnf5::rpm::PackageQuery reboot_suggested{installed};
```
jan-kolarik, line 172: Well, or maybe create just a single `reboot_suggested` query at all when I am reading this now :)

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+    std::vector<libdnf5::rpm::Package> need_reboot = {};
+    for (const auto & pkg : reboot_suggested) {
+        if (pkg.get_install_time() > static_cast<unsigned long long>(boot_time)) {
+            need_reboot.push_back(pkg);
+        }
+    }
+
+    if (need_reboot.empty()) {
+        std::cout << "No core libraries or services have been updated since boot-up.\n"
```
jan-kolarik, line 183: Minor: would be better to use `std::endl` instead of `\n` on the output (more occurrences in this file).

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
+    std::unique_ptr<sdbus::IConnection> connection;
+    try {
+        connection = sdbus::createSystemBusConnection();
+    } catch (const sdbus::Error & ex) {
+        throw libdnf5::cli::CommandExitError(1, M_("Couldn't connect to D-Bus: {}"), ex.what());
+    }
+
+    auto systemd_proxy = sdbus::createProxy("org.freedesktop.systemd1", "/org/freedesktop/systemd1");
```
jan-kolarik, line 204: I would make the constants for these D-Bus paths, they are already used on multiple places here.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+        unit_file_to_service.insert(std::make_pair(fragment_path, Service{unit_name, start_timestamp_us}));
+    }
+
+    // Iterate over each file from each installed package and check whether it
+    // is a unit file for a running service. This is much faster than running
+    // filter_file on each unit file.
+    libdnf5::rpm::PackageQuery base_query{ctx.base};
+
+    libdnf5::rpm::PackageQuery installed{base_query};
```
jan-kolarik, line 259: Probably again, we could squash the queries into a single one?

libdnf5/rpm/package_query.cpp
```diff
@@ -2828,4 +2828,52 @@ void PackageQuery::filter_installonly() {
...
+    for (const auto & pkg : *this) {
+        if (CORE_PACKAGE_NAMES.contains(pkg.get_name())) {
+            filter_result.add_unsafe(pkg.get_id().id);
+        }
+    }
+
+    libdnf5::advisory::AdvisoryQuery advisories{p_impl->base};
+    auto reboot_advisories =
+        advisories.get_advisory_packages_sorted_by_name_arch_evr() |
+        std::views::filter([](const auto & advisory_pkg) { return advisory_pkg.get_reboot_suggested(); });
```
jan-kolarik, line 2866: So I was testing the new `needs-restarting` plugin in CLI inside a Fedora container and I experienced a VERY slow execution time when running the `dnf5 needs-restarting`, like above 10 seconds compared to <1s using the `dnf4 needs-restarting -r`. I figured out it is because of using this range view. I didn't analyse what is going on in the background, but just changing this line to `std::erase_if(reboot_advisories, [](const auto & pkg) { return !pkg.get_reboot_suggested(); });` results again in similar execution time as with dnf4.

doc/dnf5_plugins/index.rst
```diff
@@ -8,6 +8,7 @@ DNF5 Plugins
 
     builddep.8
     copr.8
+    needs_restarting.8
```
jan-kolarik, line 11: We still need to deploy the new man page, it is missing in the `doc/conf.py.in`, `doc/CMakeLists.txt` and `dnf5.spec` files.

jan-kolarik: And please also add a reference to new plugin command into `dnf5.8.rst`.

dnf5-plugins/needs_restarting_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+install(TARGETS needs_restarting_cmd_plugin LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/dnf5/plugins/)
```
jan-kolarik, line 14: Please also add the new provide `Provides: dnf5-command(needs-restarting)` into the spec file.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+    //      the start of the epoch, and (2) machines with the RTC in local time
+    //      (see `timedatectl status`).
+    // - /proc/uptime
+    //      Seconds field of /proc/uptime subtracted from the current time.
+    //      Works for machines without RTC iff the current time is reasonably
+    //      correct. Does not work on containers which share their kernel with
+    //      the host---there, the host kernel uptime is returned
+
+    // First, ask systemd for the boot time. If systemd is available, this is
+    // the best option.
```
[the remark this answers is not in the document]
jan-kolarik: Thanks for the clarification. Currently, I don't have any ideas for improving this specific containerized use case. Our solution for calculating boot time, including your changes there, is already quite complex. I believe it's acceptable as it is now, and we can wait for potential reports.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,306 @@
...
+        const uint64_t systemd_boot_time_us =
+            proxy->getProperty("UserspaceTimestamp").onInterface(SYSTEMD_MANAGER_INTERFACE);
+
+        const time_t systemd_boot_time = static_cast<long>(systemd_boot_time_us) / (1000L * 1000L);
+
+        if (systemd_boot_time != 0) {
+            return systemd_boot_time;
+        }
+    } catch (const sdbus::Error & ex) {
+        // Some D-Bus error, maybe we're inside a container.
```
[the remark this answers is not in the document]
jan-kolarik: Oh, you've done it pretty thoroughly, nice 🙂

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,316 @@
...
+        throw libdnf5::cli::SilentCommandExitError(1);
+    }
+}
+
+void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
+    std::unique_ptr<sdbus::IConnection> connection;
+    try {
+        connection = sdbus::createSystemBusConnection();
+    } catch (const sdbus::Error & ex) {
+        throw libdnf5::cli::CommandExitError(1, M_("Couldn't connect to D-Bus: {}"), ex.what());
```
jan-kolarik, line 213: Copr build is failing for now. There seems to be some issue when passing the given arguments here to the underlying Error classes, although at first glance, I don't see why...

include/libdnf5/rpm/package_query.hpp
```diff
@@ -31,6 +31,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/common/sack/exclude_flags.hpp"
 #include "libdnf5/common/sack/query_cmp.hpp"
 
+#include <ranges>
```
jan-kolarik, line 34: We can drop this now.

evan-goode: Done

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
evan-goode: OK, the Copr build is succeeding now after creating a `std::string` from `ex.what()`. Not sure why it only failed in Copr when passing the `char *`.

libdnf5/rpm/package_query.cpp
```diff
@@ -2828,4 +2828,51 @@ void PackageQuery::filter_installonly() {
...
+
+    const auto & cmp_naevr = libdnf5::rpm::cmp_naevr<const advisory::AdvisoryPackage &, const Package &>;
+    for (const auto & pkg : *this) {
+        auto lower = std::lower_bound(reboot_advisories.begin(), reboot_advisories.end(), pkg, cmp_naevr);
+        if (lower != reboot_advisories.end() && lower->get_nevra() == pkg.get_nevra()) {
+            filter_result.add_unsafe(pkg.get_id().id);
+        }
+    }
+
+    *p_impl &= filter_result;
```
kontura, line 2875: Oh, this is my bad, I chose a wrong name for the `get_advisory_packages_sorted_by_name_arch_evr()` function. I really need to at least document it.

The problem is that it returns the advisory packages sorted by libsolv `id`s of name, arch and evr, not the actual strings they represent. It is a different sorting than `cmp_naevr` and therefore the `std::lower_bound(...)` doesn't work correctly.

Perhaps the best solution would be to replace this block with:
```cpp
PQImpl::filter_sorted_advisory_pkgs(*this, reboot_advisories, libdnf5::sack::QueryCmp::EQ);
```
(It should also be faster because it compares the `id`s directly and doesn't resolve them into strings.)

kontura (changes requested):

libdnf5/rpm/package_query.cpp
kontura: I have created: https://github.com/rpm-software-management/dnf5/pull/998

evan-goode: Ah thanks, I thought the results from `filter_reboot_suggested` looked correct, but you're right, it wasn't working. I think it's correct now, . Clearly some tests are needed still.

jan-kolarik (approved):

jan-kolarik:
Oh, I just realized that the existing tests aren't being executed for dnf5 at the moment. We should attempt to reuse at least some of them.

evan-goode:
CI PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1403

evan-goode:
> Just a couple minor typos, otherwise LGTM.
> 
> Also I believe the changes in `rpm/nevra.hpp` are technically not longer needed but we can keep them if you think they might be useful in the future.

True, I think we might as well keep them.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -0,0 +1,325 @@
...
+    //      Reflects the time the first process was run after booting. This
+    //      works for all known cases except (1) machines without a RTC---they
+    //      awake at the start of the epoch, and (2) machines with the RTC in
+    //      local time (see `timedatectl status`). This method will be correct
+    //      in container environments, even those running on hosts without an
+    //      RTC, since the correct time should be known by the time the
+    //      container is started.
+    // - /proc/uptime
+    //      Seconds field of /proc/uptime subtracted from the current time.
+    //      Works for machines without RTC iff the current time is reasonably
```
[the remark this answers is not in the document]
evan-goode: That's intentional, iff is shorthand for "if and only if".

doc/changes.rst
```diff
@@ -129,6 +129,14 @@ Module command
 --------------
  * Dropped ``--all`` option since this behavior is the default one.
 
+Needs-restarting command
+------------------------
+ * ``needs-restarting`` no longer scans for open files to determine whether any outdated files still in use. The default behavior is now the ``--reboothint`` behavior of DNF 4 needs-restarting, which reports whether a system reboot is recommended depending on which packages have been updated since the most recent boot.
```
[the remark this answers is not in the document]
evan-goode: Thanks, done!

jan-kolarik:
I just noticed the `tmp` commit present in the PR, probably an oversight... 🙂

evan-goode:
> I just noticed the `tmp` commit present in the PR, probably an oversight... 🙂

Sigh... thank you. One more force push...

jan-kolarik:
> > I just noticed the `tmp` commit present in the PR, probably an oversight... 🙂
> 
> Sigh... thank you. One more force push...

Great and thank you for your patience with such an exhaustive review 🙂 I believe we can proceed with the merge now, as Ales mentioned only minor typos.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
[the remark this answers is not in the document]
kontura: I see, okok. :slightly_smiling_face:

Issue #889, opened 2023-09-18
I have a suggestion for an improvement (that can be implemented right now, or separately): What about printing "{} s since Unix epoch" instead of "unrepresentable"?


j-mracek:
I have a suggestion for an improvement (that can be implemented right now, or separately): What about printing "{} s since Unix epoch" instead of "unrepresentable"?

_Originally posted by @ppisar in https://github.com/rpm-software-management/dnf5/pull/844#discussion_r1311528643_

j-mracek:
I created an issue to unblock merge of PR

PR #891, opened 2023-09-18
Print more useful form for large epochs that don't fit into `time_t`

kontura:
Follow up to: https://github.com/rpm-software-management/dnf5/pull/844
Closes: https://github.com/rpm-software-management/dnf5/issues/889


ppisar:
The code is good. The CI failure are due to an RPM signing failure in COPR.

ppisar (approved):

PR #894, opened 2023-09-19
Make get_base_arch() public

m-blaha:
Let's also standardize the rpm.arch interface by consistently using the std::string type for both arguments and return values. While this might not be the most optimal choice (as the underlying structure is char *), I prioritize consistency over performance in this case. I did consider using std::string_view, but it's not supported by SWIG.

Unit tests included.

Resolves: https://github.com/rpm-software-management/dnf5/issues/600

### DIFF
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -30,6 +30,7 @@
 }
 
 %{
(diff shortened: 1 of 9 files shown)


j-mracek (approved):
LGTM

Issue #842, opened 2023-08-28
repo info: Provide parameter for adding custom fields


jan-kolarik:
Currently, `repo info` contains a lot of values in the output, more than the original dnf version (see merged [PR](https://github.com/rpm-software-management/dnf5/pull/807)).

We could change the behavior to match the output with original dnf command and add an option like `--add-values` to allow users adding more values to the output, like:

```
dnf5 repo info --add-values=priority,expired,...
```

Related issue: #150.

jan-kolarik:
As discussed in the closed [PR](https://github.com/rpm-software-management/dnf5/pull/896), this feature is currently not planned to be implemented.

PR #896, opened 2023-09-19
repoinfo: Implement add-values option

jan-kolarik:
Create `add-values` option for the `repo info` command to allow adding additional repo values to the output. 

Match the default output with the original dnf behavior and allow user to provide more repo values to the output using the `add-values` option.

The related logic was also added to the daemon side.

Added option is now documented in the `repo` command man page.

CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1383.
Closes: https://github.com/rpm-software-management/dnf5/issues/842.

### DIFF
--- a/dnf5/commands/repo/arguments.hpp
+++ b/dnf5/commands/repo/arguments.hpp
@@ -57,6 +57,18 @@ public:
 };
 
 
+class RepoAddValuesOption : public libdnf5::cli::session::AppendStringListOption {
+public:
+    explicit RepoAddValuesOption(libdnf5::cli::session::Command & command)
+        : AppendStringListOption(
+              command,
+              "add-values",
+              '\0',
+              _("Add specified repo values to the output. List option."),
+              _("VALUE_NAME,...")) {}
+};
+
+
 }  // namespace dnf5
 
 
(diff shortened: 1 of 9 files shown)


kontura:
We may have already discussed this and agreed on it, if that is the case please disregard my comment.

However I am feeling somewhat conflicted about this option. Between the `--dump-repo-config=..` and json outputs (which I think will contain all values?) this feels unnecessary. Is there an actual user requirement for this?

If I am not mistaken it is also another new output configuration approach. Something like `repoquery --queryformat` but different. 

Personally I would keep the current dnf5 behavior - always printing them.

jan-kolarik:
> We may have already discussed this and agreed on it, if that is the case please disregard my comment.
> 
> However I am feeling somewhat conflicted about this option. Between the `--dump-repo-config=..` and json outputs (which I think will contain all values?) this feels unnecessary. Is there an actual user requirement for this?
> 
> If I am not mistaken it is also another new output configuration approach. Something like `repoquery --queryformat` but different.
> 
> Personally I would keep the current dnf5 behavior - always printing them.

I think the parent issue wasn't discussed publicly so far, just internally with Jarda. 

You're right that we are introducing here some behavior which is not consistent with other commands. The main intent was to match the default output with the original dnf behavior. I guess that when we're gonna implement the formatted output (which is planned for this command), we can then introduce something like verbose mode for all the respective commands and put more things (everything which makes sense) to the formatted output.

jan-kolarik:
After the team meeting discussion, we decided not to introduce an additional parameter like this. We will keep the new values in the dnf5 repo info output.

PR #906, opened 2023-09-20
Hardcode protection of dnf5 package

evan-goode:
DNF 5, like DNF 4, should be marked as protected to prevent accidental removal.

Doing it this way, rather than using a file in /etc/dnf/protected.d, means that DNF 4 will still be able to remove DNF 5. I think this is the ideal functionality: DNF 4 will be able to remove DNF 5 and vice-versa, but neither package manager will accidentally remove itself.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2221907 "dnf5 is not marked as protected, removal leaves system more or less bricked ".

### DIFF
--- a/libdnf5/conf/config_main.cpp
+++ b/libdnf5/conf/config_main.cpp
@@ -293,7 +293,7 @@ class ConfigMain::Impl {
     OptionString proxy_username{nullptr};
     OptionString proxy_password{nullptr};
     OptionStringSet proxy_auth_method{"any", "any|none|basic|digest|negotiate|ntlm|digest_ie|ntlm_wb", false};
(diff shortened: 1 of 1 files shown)


evan-goode:
Related: https://github.com/rpm-software-management/libdnf/pull/1621
Related: https://github.com/rpm-software-management/dnf/pull/1991

inknos:
the failing test is due to an unresolved issue with gh secrets. I tried to build it locally and installing it in a f38 container and it was fine, so I would approve this PR to

inknos (approved):

Conan-Kudo (approved):

PR #907, opened 2023-09-21
[dnf5] config-manager plugin

jrohel:
CI tests https://github.com/rpm-software-management/ci-dnf-stack/pull/1402

Manage libdnf5 configuration.

Subcommands:
- addrepo - Add repositories from the specified configuration file or define a new repository using user options
- setopt - Set configuration and repositories options
- unsetopt - Uset/remove configuration and repositories options
- setvar - Set variables
- unsetvar - Unset/remove variables

Note:
Main configuration:
libdnf5 reads the distribution configuration from the drop-in directory
("/usr/share/dnf5/libdnf.conf.d") and system configuration from drop-in
directory ("/etc/dnf/libdnf5.conf.d"). Last, it loads the system configuration
file (by default "/etc/dnf/dnf.conf"). The latter has the highest priority
and is modified by config-manager.

Repository configuration:
Libdnf5 loads the repositories configuration and then loads the configuration
overrides. Configuration overrides are stored in files in
the "/usr/share/dnf5/repos.override.d" and "/etc/dnf/repos.override.d"
directories. The files are sorted alphabetically. The override from the next
file overrides the previous one - the last override value wins.
The config-manager writes the repositories configuration changes to the file
"/etc/dnf/repos.override.d/99-config-manager.repo".


jrohel:
I put back the work in progress. I am thinking about solving a corner case.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+#include <regex>
+#include <string_view>
+
+namespace dnf5 {
+
+using namespace libdnf5::cli;
+
+namespace {
+
+const std::filesystem::path CFG_MANAGER_REPOS_OVERRIDE_FILENAME = "99-config_manager.repo";
```
j-mracek, line 41: The same path is already defined in `shared.hpp`. Do we need to define the file name also here?

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+const std::filesystem::path CFG_MANAGER_REPOS_OVERRIDE_FILEPATH =
```
j-mracek, line 43: What about to define the file path in `shared.hpp`?

j-mracek:
@jrohel May I ask you to enable CI tests or to create new test cases for config-manager?

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+    libdnf5::REPOS_OVERRIDE_DIR / CFG_MANAGER_REPOS_OVERRIDE_FILENAME;
+
+
+// Extracts a specific part of the URL from a URL string.
```
j-mracek, line 47: May I ask you to describe situation when methods returns an empty string? There is a test in code whether the function returns an empty string and the value is not used for anything else.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+    auto & ctx = get_context();
+    auto & base = ctx.base;
+    auto logger = base.get_logger();
+
+    const auto & repo_dirs = base.get_config().get_reposdir_option().get_value();
+    if (repo_dirs.empty()) {
+        throw ConfigManagerError(M_("Missing path to repository configuration directory"));
+    }
+
+    std::filesystem::path repo_dir = repo_dirs.front();
```
j-mracek, line 244: Note - this is simple and consistent way to choose a path from list.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+    if (!repofile_url.empty()) {
+        url = repofile_url;
+    } else if (auto it = repo_opts.find("baseurl"); it != repo_opts.end() && !it->second.empty()) {
+        const auto urls = libdnf5::OptionStringList(std::vector<std::string>{}).from_string(it->second);
+        if (urls.empty()) {
+            throw ConfigManagerError(M_("Bad baseurl: {}={}"), it->first, it->second);
+        }
+        url = urls[0];
+    }
+    if (url.empty()) {
```
j-mracek, line 256: May I ask you why you did not continue with `else if` => skip `if (url.empty()) {`

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+    const auto & repo_dirs = base.get_config().get_reposdir_option().get_value();
+    if (repo_dirs.empty()) {
+        throw ConfigManagerError(M_("Missing path to repository configuration directory"));
+    }
+
+    std::filesystem::path repo_dir = repo_dirs.front();
+
+    std::string url;
+    if (!repofile_url.empty()) {
+        url = repofile_url;
```
j-mracek, line 248: A suggestion - What about to make the code more condense?
```
    if (!repofile_url.empty()) {
        add_repo_from_repofile(repofile_url, repo_dir);
        return;
```

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+namespace {
+
+const std::filesystem::path CFG_MANAGER_REPOS_OVERRIDE_FILENAME = "99-config_manager.repo";
+
+const std::filesystem::path CFG_MANAGER_REPOS_OVERRIDE_FILEPATH =
+    libdnf5::REPOS_OVERRIDE_DIR / CFG_MANAGER_REPOS_OVERRIDE_FILENAME;
+
+
+// Extracts a specific part of the URL from a URL string.
+std::string get_url_part(const std::string & url, CURLUPart what_part) {
```
j-mracek, line 48: May I ask you to describe that it returns an empty string if part does not exist?

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+    curl_url_cleanup(c_url);
+    return ret;
+}
+
+
+// Computes CRC32 checksum of input string.
+// Slow bitwise implementation. Used to calculate the checksum of the omitted part of long URLs.
+uint32_t crc32(std::string_view input) {
```
j-mracek, line 68: Just a question - would be possible to replace that code by something from a standard library?

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+        save_filename = std::filesystem::path(get_url_part(full_url, CURLUPART_PATH)).filename();
+    }
+    if (!save_filename.ends_with(".repo")) {
+        save_filename += ".repo";
+    }
+    auto dest_path = repo_dir / save_filename;
+
+    test_filepath(dest_path);
+
+    auto tmpfilepath = dest_path.string() + ".XXXXXX";
```
j-mracek, line 301: May I ask you to add a comment what the  the code does between line 301 and 307. It looks like that it only check whether it is possible to create a temp file.

dnf5-plugins/config-manager_plugin/addrepo.hpp
```diff
@@ -0,0 +1,57 @@
...
+class ConfigManagerAddRepoCommand : public Command {
+public:
+    explicit ConfigManagerAddRepoCommand(Context & context) : Command(context, "addrepo") {}
+    void set_argument_parser() override;
+    void configure() override;
+
+private:
+    void add_repo_from_repofile(const std::string & url, const std::filesystem::path & repo_dir);
+    void create_repo(const std::string & url, const std::filesystem::path & repo_dir);
+    void test_filepath(const std::filesystem::path & dest_path) const;
```
j-mracek, line 42: May I ask you for a description what the method does - if I am not mistaken it check whether file described in path exists.

dnf5-plugins/config-manager_plugin/addrepo.hpp
```diff
@@ -0,0 +1,57 @@
...
+    void test_if_ids_not_already_exist(
```
j-mracek, line 43: May I ask you for a description of the method. The name is quite self descriptive but additional information how arguments are used might be helpful.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+    auto fd = mkstemp(tmpfilepath.data());
```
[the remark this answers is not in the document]
j-mracek: Anyway we have to describe the behavior in man pages.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,462 @@
...
+        libdnf5::ConfigParser parser;
+        parser.read(tmpfilepath);
+        std::vector<std::string> repo_ids;
+        repo_ids.reserve(parser.get_data().size());
+        for (const auto & [repo_id, opts] : parser.get_data()) {
+            repo_ids.emplace_back(repo_id);
+        }
+        test_if_ids_not_already_exist(repo_ids, dest_path);
+
+        // Test if the repository options can be set.
```
j-mracek, line 328: I like the test. There might be some legacy configuration option that might cause the problem. As I said, this is a good approach.

j-mracek:
I would like to open a discussion related to plural form of sub-commands:
`setvars`
`setopts`

j-mracek:
Problems related to absence of directory

Creating a repository from file in emptuy installroot ends in
```
filesystem error: cannot create temporary file: No such file or directory [/tmp/testconf/etc/yum.repos.d/rpmsoftwaremanagement-dnf5-deps-fedora-rawhide.repo.DKkNpb]
```
Setvars:
```
dnf5 config-manager setvars  acpi=ss --installroot=/tmp/testconf
Cannot write variable to file "/tmp/testconf/etc/dnf/vars/acpi": basic_ios::clear: iostream error
```

```
dnf5 config-manager setopts strict=false --installroot=/tmp/testconf2
filesystem error: cannot open file: No such file or directory [/tmp/testconf2/etc/dnf/dnf.conf]
```
Errors are misleading, because the presence of directory will resolve the issue.

j-mracek:
This is not a problem, but interesting discovery - Command `dnf5 config-manager setopts strict=false --installroot=/tmp/testconf2` creates RPMDB. It is unneeded, but I think there is a low chance that it will create any problem.

dnf5-plugins/config-manager_plugin/setvars.cpp
```diff
@@ -0,0 +1,105 @@
...
+            }
+            return true;
+        });
+    cmd.register_positional_arg(vars_vals);
+}
+
+
+void ConfigManagerCommandSetVars::configure() {
+    auto & ctx = get_context();
+
```
j-mracek, line 83: What about to add to all configure steps -> `ctx.set_load_system_repo(false);`. I think that loading RPMDB is not required by the plugin. What do you think?

jrohel: @j-mracek `load_system_repo` is `false` by default. So adding `ctx.set_load_system_repo(false);` does not change anything.
The plugin does not need RPMDB. If the RPMDB is created, it will be caused by something other than loading. Maybe librpm will do this during architecture detection? Maybe ... I don't know, I need to investigate.

jrohel:
> If I am not mistaken DNF5 uses singular forms like `install`, `remove` even if it is possible to perform multiple operations.

Comparing `setopts` and `install`? From my point of view, this is an inappropriate comparison. `install` is like `set`. `setvars` and `setopts` are like `installpackages` (which is not used in dnf).

j-mracek:
> > If I am not mistaken DNF5 uses singular forms like `install`, `remove` even if it is possible to perform multiple operations.
> 
> Comparing `setopts` and `install`? From my point of view, this is an inappropriate comparison. `install` is like `set`. `setvars` and `setopts` are like `installpackages` (which is not used in dnf).

I would like to double-check that proposed names are the best option. After release it is difficult to modify.

pkratoch:
I am for singular form (i.e. `setopt` etc.).
1. There is also subcommand `addrepo` (and not `addrepos`).
2. It will be less confusing with the options `--setopt` and `--setvar` that already exist.
3. When we discussed the naming previously, the singular form didn't seem unnatural. It's only now that it's put in comparison.

jan-kolarik:
I am also for singular forms. I like keeping the consistency with already existing names in DNF4/DNF5 for the same or very similar behavior. I believe it will also improve the transition for users to DNF5.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: No. Removed.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: Moved to `shared.hpp`.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: I extended the description.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: I reworked the code. And moved it to the `create_repo` method.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: Reworked.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: Described.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jrohel: No. The code will produce an open temporary file. It then closes it but does not remove it. In the following code, this temporary file is used to store the copied/downloaded configuration.

dnf5-plugins/config-manager_plugin/addrepo.hpp
jrohel: I added docs strings.

dnf5-plugins/config-manager_plugin/addrepo.hpp
jrohel: I added doc string

dnf5-plugins/config-manager_plugin/addrepo.cpp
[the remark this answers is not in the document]
jrohel: I added checks if the directory exists. If not, according to the new option `--create_missing_dirs`, the directories (missing paths elements) are created or `ConfigManagerError` exception is thrown.

jrohel:
- I rebased the PR.
- I have made changes to reflect the comments.
- I also added support for the new `--create-missing-dirs` option. Defines how the plugin should behave if the target directory does not exist. It either prints an error message about it or creates the missing directories.

It remains to solve the plural/singular form of sub-commands.

dnf5-plugins/config-manager_plugin/addrepo.cpp
jan-kolarik: If I understand correctly, this is used in a specific use case only for generating shorter URLs. Why not utilize a hash function (like SHA1, etc.) from the existing dependency `libcrypto` and extract a substring of the appropriate length from the result?

jrohel: @jan-kolarik I don't currently know why, but there is indeed a dependency on `libcrypto` in "dnf5.spec". There is a line: `BuildRequires: pkgconfig(libcrypto)`.
Are we really using `libcrypto` in libdnf5? Yes `libcurl` which we depend on is compiled against `libcrypto` in Fedora. But libcurl also supports other crypto libraries.

jan-kolarik: I was curious, so I've queued [a testing scratch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=108418756) without `BuildRequires: pkgconfig(libcrypto)` dependency and it really seems unused.

jrohel:
Plural/singular - current state:
Runnings variables: `dnf5 --setvar=var1=value --setvar=var2=value --setvar=var3=value ...`
Configure variables: `dnf5 config-manager setvars var1=value var2=value var3=value ...`
`dnf5 config-manager unsetvars var1 var2 var3 ...`

Runnings options: `dnf5 --setopt=[repo.]opt1=value --setopt=[repo.]opt2=value --setopt=[repo.]opt3=value ...`
Configure options: `dnf5 config-manager setopts [repo.]opt1=value [repo.]opt2=value [repo.]opt3=value ...`
`dnf5 config-manager ussetopts [repo.]opt1 [repo.]opt2 [repo.]opt3 ...`

Enabled running repositories: `dnf5 --repo=repo1,repo2,...`
1. Add one new repositories from user options: `dnf5 config-manager addrepo [id=repoid] --set=repo_opt1=value --set=repo_opt2=value ...`
2. Add one repository config file (can contain more repositories): `dnf5 config-manager addrepo --from-repofile=<URL|filepath`
Maybe different commands can be implemented for these two cases.

ppisar:
I'm almost indifferent. I had to choose I would use a singular form. Actually if this pull request had a manual page, it would be easier to consider this question from a usability point of view.

dnf5-plugins/config-manager_plugin/addrepo.cpp
j-mracek: Please don't use SHA1 for any purpose. It is not available in FIPS mode in RHEL9.

j-mracek: I am not sure whether it is a good idea but what about to use SHA2 but from that long hash use only few first characters. I believe that it is available from libsolv API.

ppisar: Well, SHA-256 is also not recommended by NIST now. In general, a choice of hashing algorithm depends on what you want to use the hash for.

If it is for cryptographic purposes, then it you should go with something implemented by a certified library and be prepared to change the algorithm during a life cycle of the code (i.e. always handle hash value and hash algorithm as a pair in your data and code).

If it is for mapping data onto memory buckets, e.g. for implementing associative arrays, then choose an algorithm witch is designed for it (i.e. fair and fast). E.g. MurmurHash. MD5 and SHA-1 are needlessly slow.

If it is for error detection, then CRC can be pretty fine.

For the last to uses, it's unfortunate that they are usually implemented as a copylib library. Not as a standalone, dynamic library. It's similar to Base64.

jrohel:
The singular form was chosen. I updated the PR.

dnf5-plugins/config-manager_plugin/addrepo.cpp
j-mracek: I believe that this part can be used like it is, it could be modify in future because the hash is not something that cannot be changed. I do not consider it as a blocking problem.

dnf5-plugins/config-manager_plugin/shared.hpp
```diff
@@ -0,0 +1,108 @@
...
+struct ConfigManagerError : public libdnf5::Error {
+    using Error::Error;
+    const char * get_domain_name() const noexcept override { return "dnf5"; }
+    const char * get_name() const noexcept override { return "ConfigManagerError"; }
+};
+
+// Checks if the `path` directory exists. If not, according to the `create_missing_dirs` argument,
+// the directories (missing paths elements) are created or `ConfigManagerError` exception is thrown.
+inline void resolve_missing_dir(const std::filesystem::path & path, bool create_missing_dirs) {
+    if (!std::filesystem::exists(path)) {
```
j-mracek, line 46: This is real nitpicking - if the path is a file, the test pass because the path exists.
```
$ mkdir /tmp/config-manager/etc/ -p
$ touch /tmp/config-manager/etc/dnf
$ dnf5 --installroot=/tmp/config-manager config-manager setopt gpgcheck=false
```
There is `std::filesystem::is_directory` method. But if the path is a `symlink` to an existing directory then it will not work according to expectation - pass.

In other cases it works perfectly.

jrohel:
- rebased the PR
- fixed saving the repository override configuration in the installation root different from "/".
- `setopt` does not use RepoQuery to be consistent with the implementation of other subcommands
- if the path exists, the `resolve missing dir` function tests whether it is a directory or a symlink to an existing directory, if not it throws an exception

And I created a lot of CI tests https://github.com/rpm-software-management/ci-dnf-stack/pull/1402.

dnf5-plugins/config-manager_plugin/addrepo.cpp
ppisar: Yesterday I learnt that zlib has crc32() and crc32_z() in its public API. From zlib.h:

~~~~
     uLong crc = crc32(0L, Z_NULL, 0);

     while (read_buffer(buffer, length) != EOF) {
       crc = crc32(crc, buffer, length);
     }
     if (crc != original_crc) error();
~~~~

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -0,0 +1,519 @@
...
+        if (parser.has_section(repo_id)) {
+            // If the repository with the id already exists, it will be removed.
+            parser.remove_section(repo_id);
+        }
+    }
+
+    parser.add_section(repo_id);
+
+    // Sets the default repository name. May be overwritten with "--set=name=<name>".
+    parser.set_value(repo_id, "name", "created by dnf5 config-manager");
```
j-mracek, line 425: What about to include repoid in the name? Proposing something like `<RepoID> created by dnf5 config-manager`. Keeping like it is is also OK, therefore the change is optional.

j-mracek:
I would also suggest to add a warning when user want to unset a value but value is not in configuration file. Unset of option for repository should be explicitly documented in man pages, because unset only removes overrides but does not touch the original repo file.

j-mracek (approved):
Remaining suggestions or comments are not blocking the merge and could be addressed in the next PR if they are considered as something to worth to implement The PR was tested with provided test cases.

Issue #919, opened 2023-09-26
updateinfo issues


glum23:
According to https://github.com/rpm-software-management/dnf5/issues/155, work on `updateinfo` command has been finished. However:
- Documentation is insufficient. 
In https://dnf5.readthedocs.io/en/latest/changes.html, `updateinfo` is mentioned only once and the mention of it is incorrect (see the next item). It definitely should be listed among changed commands as it default name has been changed to `advisory` while many aliases (including `upif`) are not available now.
`dnf5 updateinfo --help` does not provide any options while several options are listed in the man page (vice versa problem is already tracked for all commands in https://github.com/rpm-software-management/dnf5/issues/644 )
- In the section "Changes on the command line" of https://dnf5.readthedocs.io/en/latest/changes.html, there's explicitly mentioned `dnf updateinfo info -> dnf updateinfo –info` as an example of the new approach in comparison with dnf4, however, the command is now implemented in the old way in dnf5.

kontura:
Thank you for the report!
I have created a PR https://github.com/rpm-software-management/dnf5/pull/920 which documents the changes.

Regarding the `dnf5 updateinfo --help` not showing any options. I think that is expected, the options are for subcommands: `dnf5 updateinfo info --help`.

glum23:
> Thank you for the report! I have created a PR #920 which documents the changes.
Thanks!

> Regarding the `dnf5 updateinfo --help` not showing any options. I think that is expected, the options are for subcommands: `dnf5 updateinfo info --help`.
I'd expect the output of  `dnf5 updateinfo --help` similar to the related man page (especially because all subcommands' options are the same), sth like:
~~~
Usage:
  dnf5 advisory <COMMAND> [GLOBAL OPTIONS] [OPTIONS] [ARGUMENTS]
Command:                 
  list                          List advisories
  info                          Print details about advisories
  summary                       Print summary of advisories
Options:                         
 <list of options>
Arguments:                                     
  advisory-spec                                List of patterns matched against advisory names.
~~~
From my POV, seeking the relevant options in two steps is unnecessarily protracted. However, the current solution is consistent.

kontura:
They are the same for `updateinfo` but it doesn't have to be like that in general.
For example `dnf5 group list --help` vs `dnf5 group install --help`.

But I see your point about the man page. 
I agree it is something to consider.

glum23:
new issue: https://github.com/rpm-software-management/dnf5/issues/921

PR #920, opened 2023-09-27
advisory: document advisory command changes and few clean ups

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/919

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -52,14 +52,14 @@ This allows using a list of authentication methods in configuration files and th
 Changes on the command line:
 ============================
 
-Commands cannot have optional subcommands and optional arguments. Is some cases subcommand can have the same string as
-an argument. It means there is a difficulty to find advisory for package with the name `list`, `info`, or `summary`.
-`dnf history info <transaction ID>` -> `dnf history --info <transaction ID>`
-`dnf updateinfo info` -> `dnf updateinfo --info`
+Commands cannot have optional subcommands and optional arguments. Optional subcommands were ambiguous, it wasn't clear
+whether the input is a command argument or a subcommand. If present, subcommands are now mandatory.
+`dnf history <transaction ID>` -> `dnf history info <transaction ID>`
+`dnf updateinfo` -> `dnf updateinfo summary`
 
-Options that cannot be applied to all command or can be applied but without any effect should be removed from general
-options and implemented only for related commands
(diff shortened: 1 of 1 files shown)


pkratoch:
Hi, the PR doesn't address this part of the issue: "`dnf5 updateinfo --help` does not provide any options while several options are listed in the man page". Do you plan to add it? Otherwise, the issue shouldn't be marked as closed by this.

pkratoch:
Also, I noticed that in the `changes.rst` document, single backticks are used everywhere to indicate command substitution, while for example in `dnf5.8.rst`, double backticks are used. When I build the documentation, the single backtics produce italics instead of literal, so if doesn't show e.g. options correctly. Is it the same for you or is there a special setting I am missing?

kontura:
> Hi, the PR doesn't address this part of the issue: "`dnf5 updateinfo --help` does not provide any options while several options are listed in the man page". Do you plan to add it? Otherwise, the issue shouldn't be marked as closed by this.

That part was extracted into a separate issue: https://github.com/rpm-software-management/dnf5/issues/921 which was closed in the mean time.

 

> Also, I noticed that in the `changes.rst` document, single backticks are used everywhere to indicate command substitution, while for example in `dnf5.8.rst`, double backticks are used. When I build the documentation, the single backtics produce italics instead of literal, so if doesn't show e.g. options correctly. Is it the same for you or is there a special setting I am missing?

Yea, I am seeing the same. It is also here: https://dnf5.readthedocs.io/en/latest/changes.html. I added a commit to switch that to double backticks. What do you think?

pkratoch:
> > Hi, the PR doesn't address this part of the issue: "`dnf5 updateinfo --help` does not provide any options while several options are listed in the man page". Do you plan to add it? Otherwise, the issue shouldn't be marked as closed by this.
> 
> That part was extracted into a separate issue: #921 which was closed in the mean time.

I see, ok.

> > Also, I noticed that in the `changes.rst` document, single backticks are used everywhere to indicate command substitution, while for example in `dnf5.8.rst`, double backticks are used. When I build the documentation, the single backtics produce italics instead of literal, so if doesn't show e.g. options correctly. Is it the same for you or is there a special setting I am missing?
> 
> Yea, I am seeing the same. It is also here: https://dnf5.readthedocs.io/en/latest/changes.html. I added a commit to switch that to double backticks. What do you think?

Looks good, thanks.

Issue #893, opened 2023-09-18
rpm-tmp.XXXXXX files in /var/tmp/


nuqleo:
dnf5 leaves a lot files rpm-tmp.XXXXXX files in /var/tmp/ after transactions

```
[root@rawhide ~]# ls -la /var/tmp/rpm-tmp.*
-rw------- 1 root root  150 сен  1 12:33 /var/tmp/rpm-tmp.02NJZ0
-rw------- 1 root root  254 сен 12 17:23 /var/tmp/rpm-tmp.04urC0
-rw------- 1 root root   38 авг 27 12:39 /var/tmp/rpm-tmp.06y9DG
-rw------- 1 root root   48 сен 18 01:18 /var/tmp/rpm-tmp.0JdhCb
-rw------- 1 root root   49 авг 16 13:11 /var/tmp/rpm-tmp.0NmhSA
-rw------- 1 root root  208 авг 18 20:41 /var/tmp/rpm-tmp.0PqgyJ
-rw------- 1 root root   43 авг 28 12:34 /var/tmp/rpm-tmp.0uI3rY
-rw------- 1 root root   59 авг 17 16:45 /var/tmp/rpm-tmp.0xhzH0
-rw------- 1 root root   48 авг 25 16:11 /var/tmp/rpm-tmp.10zQbM
-rw------- 1 root root   78 сен  1 12:33 /var/tmp/rpm-tmp.19e0D7
...
```
```
[root@rawhide ~]# rpm -q dnf5
dnf5-5.1.2-1.fc40.x86_64
```

jrohel:
These helper files are created by librpm. And librpm does not delete them if the librpm log mask contains RPMLOG_DEBUG.

nuqleo:
Fixed in dnf5-5.1.5

PR #929, opened 2023-10-02
Change librpm log mask value from 0xFF to everything up to RPMLOG_INFO

jrohel:
RPMLOG_DEBUG is fairly low-level and causes librpm to not remove auxiliary temporary files. Therefore, it is not a good idea to use RPMLOG_DEBUG as the default. Everything except RPMLOG_DEBUG is now logged.

However, we should consider introducing an option to enable RPMLOG_DEBUG for debugging purposes.

It solves the issue https://github.com/rpm-software-management/dnf5/issues/893


j-mracek:
LGTM

ppisar (approved):
That's reasonable.

ppisar:
A test failure (COPR_USER secret) is independent.

PR #933, opened 2023-10-03
doc: "group install" and "group remove" require an argument

ppisar:
dnf5-group(8) synopsis gave an impression that a group specifier is optional:

    SYNOPSIS
           dnf5 group <subcommand> [options] [<group-spec>...]

That's not true for "install" and "remove" commands. This patch fixes the manual.

### DIFF
--- a/doc/commands/group.8.rst
+++ b/doc/commands/group.8.rst
@@ -25,7 +25,9 @@
 Synopsis
 ========
 
-``dnf5 group <subcommand> [options] [<group-spec>...]``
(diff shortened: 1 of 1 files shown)


kontura (commented):
Looks good.

Can I just ask if the indentation is on purpose?

It seems inconsistent to me with the rest of the commands.

ppisar:
A change in the indentation was unintentional. (I tried a literal block before to prevent from the blank line separating the the invocations. Though Spinx does not support.) I removed the indentation. Thanks for catching it.

kontura (approved):

PR #939, opened 2023-10-04
doc: Add "group upgrade"

ppisar:
This patch documents "dnf5 group upgrade" subcommand.

Completes: #491


jan-kolarik (approved):
LGTM, thank you!

Issue #872, opened 2023-09-05
Make `--forcearch` option a global one


katexochen:
In dnf4, `--forcearch` was handled as a global option (between "dnf" and "\<command\>", like `dnf --forcearch=aarch64 install nano`.
With dnf5, this usage isn't currently possible. Some tools, including [mkosi](https://github.com/systemd/mkosi) use global options to construct a base command and append arguments dynamically.

ppisar:
I confirm this issue with dnf5-5.1.2-20230906004518.61.g6d7cb379.fc40.x86_64:

~~~~
root@fedora-40:~ # dnf5 --forcearch=x86_64 install dnf5
Unknown argument "--forcearch=x86_64" for command "dnf5". Add "--help" for more information about the arguments.
~~~~

~~~~
root@fedora-40:~ # dnf5 install --forcearch=x86_64 dnf5
Updating and loading repositories:
Repositories loaded.
Package "dnf5-5.1.2-20230906004518.61.g6d7cb379.fc40.x86_64" is already installed.

Nothing to do.
~~~~

DNF5 indeed implements --forcearch option as an command-specific option at 10 subcommands now. I agree that with this amount of subcommands it makes sense to turn it into a global option.

ppisar:
@evan-goode, you can take it. I don't have time for this now.

PR #941, opened 2023-10-04
Make `--forcearch` a global argument

evan-goode:
Many commands respect `--forcearch` (distro-sync, download, install, list, makecache, repoquery, search, etc.) so it makes sense to make `--forcearch` a global argument that can be used with any command.

Resolves https://github.com/rpm-software-management/dnf5/issues/872

### DIFF
--- a/dnf5/include/dnf5/shared_options.hpp
+++ b/dnf5/include/dnf5/shared_options.hpp
@@ -65,14 +65,13 @@ void create_allow_downgrade_options(dnf5::Command & command);
 /// The values are stored in the `destdir` configuration option
 void create_destdir_option(dnf5::Command & command);
 
-/// Create the `--forcearch` option for a command provided as an argument.
-/// The values are stored in the `forcearch` configuration option
-void create_forcearch_option(dnf5::Command & command);
-
 /// Create the `--downloadonly` option for a command provided as an argument.
 /// The values are stored in the `downloadonly` configuration option
 void create_downloadonly_option(dnf5::Command & command);
 
+/// Create the `--forcearch` option for a command provided as an argument.
+/// The values are stored in the `forcearch` configuration option
+[[deprecated("--forcearch is now a global argument")]] void create_forcearch_option(dnf5::Command & command);
 
 }  // namespace dnf5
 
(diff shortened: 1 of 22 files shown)


dnf5/include/dnf5/shared_options.hpp
```diff
@@ -65,10 +65,6 @@ void create_allow_downgrade_options(dnf5::Command & command);
 /// The values are stored in the `destdir` configuration option
 void create_destdir_option(dnf5::Command & command);
 
-/// Create the `--forcearch` option for a command provided as an argument.
-/// The values are stored in the `forcearch` configuration option
-void create_forcearch_option(dnf5::Command & command);
```
j-mracek, line 70: It may sound bad, but this is a change of API.  What about to consider less dramatic deprecation path?

evan-goode: Ah right, the command-line interface should be the same so I assume you're strictly referring to `create_forcearch_option`. Do you think it would be more appropriate to (1) make `create_forcearch_option` a no-op with a deprecation message printed as a warning or (2) throw an error here?

evan-goode: I left `create_forcearch_option` as a no-op with comments, but I did not add a runtime warning.

dnf5/shared_options.cpp
```diff
@@ -106,5 +73,11 @@ void create_downloadonly_option(dnf5::Command & command) {
     command.get_argument_parser_command()->register_named_arg(downloadonly);
 }
 
+void create_forcearch_option(dnf5::Command & command) {
+/// create_forcearch_option is deprecated. --forcearch is now a global
+/// argument.
+#pragma unused(command)
```
j-mracek, line 79: May I ask you what `pragma unused(command)` does?

j-mracek:
If I am not mistaken, we agreed with the team to use deprivation warning raised during compilation, and to keep the same functionality for deprecated method even if it is unused API and there is low chance that is in use from other project.

evan-goode:
OK, I've kept the original `create_forcearch_option` there with the `[deprecated]` attribute. I've tested that it does not compile if you try to reference `create_forcearch_option`.

j-mracek:
I was just on way to merge it and I discovered that there is a merge conflict. @evan-goode May I ask you for a rebase?

j-mracek (approved):

PR #918, opened 2023-09-26
Add module list command

pkratoch:
Closes: https://github.com/rpm-software-management/dnf5/issues/936
Closes: https://github.com/rpm-software-management/dnf5/issues/937

PR #964, opened 2023-10-09
modules: Show default profiles in module list

pkratoch:
Follow-up to https://github.com/rpm-software-management/dnf5/pull/918
Closes: https://github.com/rpm-software-management/dnf5/issues/938

### DIFF
--- a/include/libdnf5-cli/output/modulelist.hpp
+++ b/include/libdnf5-cli/output/modulelist.hpp
@@ -93,10 +93,14 @@ void print_modulelist_table(Query & module_list) {
 
             // Get profile strings (append [d] or [i] if needed)
             std::vector<std::string> profile_strings;
-            for (auto profile : module_item.get_profiles()) {
+            const auto & default_profiles = module_item.get_default_profiles();
+            for (const auto & profile : module_item.get_profiles()) {
                 // TODO(pkratoch): Also show "[i]" for installed profiles
-                // TODO(pkratoch): Make sure defaults are loaded
-                profile_strings.push_back(profile.is_default() ? profile.get_name() + " [d]" : profile.get_name());
+                const auto & name = profile.get_name();
+                profile_strings.push_back(
+                    std::find(default_profiles.begin(), default_profiles.end(), name) != default_profiles.end()
+                        ? name + " [d]"
+                        : name);
             }
 
             add_line_into_modulelist_table(
--- a/include/libdnf5/module/module_item.hpp
+++ b/include/libdnf5/module/module_item.hpp
@@ -132,11 +132,10 @@ public:
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getProfiles()
     std::vector<ModuleProfile> get_profiles() const;
 
-    // TODO(pkratoch): Similar implementation as in ModulePackageContainer::getDefaultProfiles().
     /// @return The default profiles.
     //
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getDefaultProfile()
-    std::vector<ModuleProfile> get_default_profiles() const;
+    std::vector<std::string> get_default_profiles() const;
 
     /// Get the `static_context` value (needed for proper behaviour of modular solver).
     ///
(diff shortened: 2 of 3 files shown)


include/libdnf5/module/module_item.hpp
```diff
@@ -132,11 +132,10 @@ class ModuleItem {
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getProfiles()
     std::vector<ModuleProfile> get_profiles() const;
 
-    // TODO(pkratoch): Similar implementation as in ModulePackageContainer::getDefaultProfiles().
     /// @return The default profiles.
     //
     // @replaces libdnf:module/ModuleItem.hpp:method:ModuleItem.getDefaultProfile()
-    std::vector<ModuleProfile> get_default_profiles() const;
+    std::vector<std::string> get_default_profiles() const;
```
jan-kolarik, line 138: Hmm, this is a change to public API. But so far the method didn't have an implementation, so it couldn't get referenced anywhere, I guess it's OK, right?

include/libdnf5-cli/output/modulelist.hpp
```diff
@@ -93,10 +93,14 @@ void print_modulelist_table(Query & module_list) {
...
+            const auto & default_profiles = module_item.get_default_profiles();
+            for (const auto & profile : module_item.get_profiles()) {
                 // TODO(pkratoch): Also show "[i]" for installed profiles
-                // TODO(pkratoch): Make sure defaults are loaded
-                profile_strings.push_back(profile.is_default() ? profile.get_name() + " [d]" : profile.get_name());
+                const auto & name = profile.get_name();
+                profile_strings.push_back(
+                    std::find(default_profiles.begin(), default_profiles.end(), name) != default_profiles.end()
+                        ? name + " [d]"
+                        : name);
```
jan-kolarik, line 103: I was testing it and yes, it seems working correctly. But what is the purpose of the not used `profile.is_default()` method then?

include/libdnf5/module/module_item.hpp
pkratoch: Yes, that was exactly my thinking. I am not sure if there could be any problem with it in this case...

jan-kolarik: I believe there should not be any problem. AFAIK if someone would reference this method, the linking would fail, so there cannot be any valid consumer so far.

include/libdnf5-cli/output/modulelist.hpp
pkratoch: I am still not sure what to do with it. The problem is already in libmodulemd, where `modulemd_profile_is_default` means something else than `modulemd_defaults_v1_get_default_profiles_for_stream_as_strv`. If I understood correctly, is connected to there being two possible module yamls, one that actually defines the module streams and one that is used by a build system to create modules. So far, I don't know what exactly the other one is used for and whether it's useful for dnf, so I plan to find out and either change it or document it, but the `modulemd_defaults_v1_get_default_profiles_for_stream_as_strv` is definitely correct to use in this case.

But now that I think about it, we can probably wait with this PR a little bit in case I find any changes or documentation that would be suitable to be added here as well.

ppisar: modulemd_profile_is_default() operates on ModulemdProfile object. That object represents a set of profiles defined in a modulemd YAML document (a module build) or modulemd-packager YAML document (a module to be built). A modulemd document never defines a default profile. This property only exists in modulemd-packager YAML documents. I.e. when building a module. Hence it never works for modules after building you can see in YUM repositories.

In YUM repositories a default profile is defined in modulemd-defaults YAML documents which are represented with ModulemdDefaultsV1 objects and that information is accessible with modulemd_defaults_v1_get_default_profiles_for_stream_as_strv() method. The information about a default profile is separated from module builds in order to be able to change that information without rebuilding a module.

ppisar: In other words, Pavla's code is correct.

jan-kolarik (approved):

PR #978, opened 2023-10-17
Add TransactionReplay and use it history replay command

kontura:
There are two approaches (APIs):
 - load the stored transaction and use it to fill `Goal` that is then resolved which creates the transaction
 - load the stored transaction and create `base::Transaction` directly, no resolving is done

If we want to use `TransactionReplay` for history undo operation we will need to be able to resolve the loaded transaction.
On the other hand `SystemUpgrade` plugin doesn't require the second resolving.

To make both available I have added a `--resolve` switch to `history replay` command.

For: https://github.com/rpm-software-management/dnf5/issues/999

### DIFF
--- a/include/libdnf5/base/goal_elements.hpp
+++ b/include/libdnf5/base/goal_elements.hpp
@@ -169,6 +169,9 @@ public:
     /// If set to true, group operations (install / remove) will only work
     /// with the group itself, but will not add to the transaction any packages.
     bool group_no_packages{false};
+    /// If set to true, environment operations (install / remove) will only work
+    /// with the environment itself, but will not add to the transaction any groups.
+    bool environment_no_groups{false};
 
     /// Set whether hints should be reported
     bool report_hint{true};
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -35,6 +35,9 @@ namespace libdnf5::rpm {
 class KeyInfo;
 }  // namespace libdnf5::rpm
 
+namespace libdnf5::transaction {
+class TransactionReplay;
+}
 
 namespace libdnf5::base {
 
@@ -168,6 +171,7 @@ private:
     friend class TransactionModule;
     friend class TransactionPackage;
     friend class libdnf5::Goal;
+    friend class libdnf5::transaction::TransactionReplay;
 
     Transaction(const libdnf5::BaseWeakPtr & base);
 
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -84,8 +84,7 @@ public:
     std::vector<std::string> get_resolve_logs_as_strings() const;
 
     /// @return the transaction packages.
-    // TODO(jrohel): Return reference instead of copy?
-    std::vector<libdnf5::base::TransactionPackage> get_transaction_packages() const;
+    std::vector<libdnf5::base::TransactionPackage> & get_transaction_packages() const;
 
     /// @return the number of transaction packages.
     std::size_t get_transaction_packages_count() const;
--- a/libdnf5/transaction/transaction_replay.cpp
+++ b/libdnf5/transaction/transaction_replay.cpp
@@ -347,6 +347,13 @@ TransactionReplay::TransactionReplay(
     }
 };
 
+libdnf5::base::Transaction TransactionReplay::create_transaction() {
+    base::Transaction transaction(this->base);
+    transaction.p_impl->set_transaction(*this);
+
+    return transaction;
+}
+
 void TransactionReplay::fill_goal(libdnf5::Goal & goal) {
     bool skip_unavailable = base->get_config().get_skip_unavailable_option().get_value();
 
--- a/libdnf5/transaction/transaction_replay.cpp
+++ b/libdnf5/transaction/transaction_replay.cpp
@@ -1,0 +1,425 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#include "../base/transaction_impl.hpp"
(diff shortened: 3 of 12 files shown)


libdnf5/transaction/transaction_replay.cpp
```diff
@@ -0,0 +1,432 @@
...
+        throw std::runtime_error(msg);
+    }
+}
+
+}  // namespace
+
+namespace libdnf5::transaction {
+
+TransactionReplay::TransactionReplay(
+    const libdnf5::BaseWeakPtr & base, std::filesystem::path filename, bool ignore_installed)
```
j-mracek, line 51: I am not sure whether it is good to do all of the logic in a constructor. Additionally, it stops on the first issue and this is a difference to goal.resolve().

j-mracek:
The approach described in the PR is OK, but I am thinking about a different API interface. What about to have all classes private. Then we can add in the first part only two methods. Please ignore names.
`libdnf5::Goal::add_transaction_replay(<path>)`
`libdnf5::base::Transaction::save_store_replay(<path>)`.

This will provide a flexibility in future. We can even keep all classes that you already introduced, but they will be used unseen by a user.

In the next step we can even add a public constructor for  `libdnf5::base::Transaction` that will require transaction store and replay file including paths to files. Then DNF5 will not require available repository to run such a transaction, but everything important must be written in the file including check sums and pgp keys or key-ring  location.

@kontura What do you think about?

libdnf5/transaction/transaction_replay.cpp
j-mracek: When there is any problem with file (permissions, format, ...) it is file to throw an error.

include/libdnf5/base/transaction.hpp
```diff
@@ -84,8 +84,7 @@ class Transaction {
     std::vector<std::string> get_resolve_logs_as_strings() const;
 
     /// @return the transaction packages.
-    // TODO(jrohel): Return reference instead of copy?
-    std::vector<libdnf5::base::TransactionPackage> get_transaction_packages() const;
+    std::vector<libdnf5::base::TransactionPackage> & get_transaction_packages() const;
```
j-mracek, line 90: This is a reasonable change, but the change is an API change.

include/libdnf5/base/goal_elements.hpp
```diff
@@ -169,6 +169,9 @@ struct GoalJobSettings : public ResolveSpecSettings {
     /// If set to true, group operations (install / remove) will only work
     /// with the group itself, but will not add to the transaction any packages.
     bool group_no_packages{false};
+    /// If set to true, environment operations (install / remove) will only work
+    /// with the environment itself, but will not add to the transaction any groups.
+    bool environment_no_groups{false};
```
j-mracek, line 174: Please don't take it as no, but I think that this will brake ABI. I think we discussed it recently and my recommendation would be to open a discuss again and discuss all option that we have.

libdnf5/transaction/transaction_replay.cpp
```diff
@@ -0,0 +1,425 @@
...
+            if (!repo_id.empty()) {
+                if (repo_id == "@System") {
+                    pkg_query = installed;
+                } else {
+                    libdnf5::rpm::PackageQuery query_repo(base);
+                    query_repo.filter_repo_id({repo_id});
+                    if (!query_repo.empty()) {
+                        pkg_query = query_repo;
+                        pkg_query |= installed;
+                    }
```
j-mracek, line 276: This is interesting. If repository does not exist then it continues with all packages. May I ask you for an explanation why this behavior is required?

libdnf5/transaction/transaction_replay.cpp
```diff
@@ -0,0 +1,425 @@
...
+                pkg_query_installonly.filter_installonly();
+
+                if (action == "Install" && !pkg_query_na.empty() && pkg_query_installonly.empty()) {
+                    raise_or_warn(
+                        ignore_installed,
+                        fmt::format(
+                            "Package {:s}.{:s} is already installed, it cannot be installed again.",
+                            nevra_pair.second.get_name(),
+                            nevra_pair.second.get_arch()));
+                }
```
j-mracek, line 304: I am quite curious why you run the check for `install` action. If I am not mistaken the request will be validate by libsolv including install-only rule. May I ask you for additional information?

libdnf5/transaction/transaction_replay.cpp
```diff
@@ -0,0 +1,425 @@
...
+            } else if (action == "Reason Change") {
+                pkg_query.filter_installed();
+                if (pkg_query.empty()) {
+                    //TODO(amatej): Unavailable reason changes are just ignored because the reason is fixed in post transaction action anyway (when we resolve goal)
+                    //raise_or_warn(skip_unavailable, fmt::format("Package nevra {:s} not installed for action {:s}.", nevra, action));
+                    continue;
+                }
+                package_actions.push_back({action, reason, *pkg_query.begin()});
+            } else if (
+                action == "Upgraded" || action == "Downgraded" || action == "Reinstalled" || action == "Removed" ||
```
j-mracek, line 323: I believe that we replaced those actions by `Replaced`.

libdnf5/transaction/transaction_replay.cpp
kontura: I have taken this behavior from the dnf4 replay however looking into it I think its there to have a working `--ignore-installed` option.
The option controls whether the command fails if we try to install already installed package.

I see that goal has the same check and it logs `GoalProblem::ALREADY_INSTALLED` but as far as I can see it is only logged it cannot fail the transaction.
I did some testing and it seems libsolv doesn't have a problem with installing already installed pkg, it either reinstalls, upgrades or downgrades the package based on the versions.

libdnf5/transaction/transaction_replay.cpp
kontura: I have taken this behavior from the dnf4 replay however I would be fine if we make it more strict and don't allow running without the repo specified.

kontura:
> The approach described in the PR is OK, but I am thinking about a different API interface. What about to have all classes private. Then we can add in the first part only two methods. Please ignore names. `libdnf5::Goal::add_transaction_replay(<path>)` `libdnf5::base::Transaction::save_store_replay(<path>)`.
> 
> This will provide a flexibility in future. We can even keep all classes that you already introduced, but they will be used unseen by a user.
> 
> In the next step we can even add a public constructor for `libdnf5::base::Transaction` that will require transaction store and replay file including paths to files (I am not sure anymore whether it is a good path (skipping libsolv check, but we can skip loading avalable repositories)). Then DNF5 will not require available repository to run such a transaction, but everything important must be written in the file including check sums and pgp keys or key-ring location.
> 
> @kontura What do you think about?

That does sound like simpler API and I like it.
The only problem is ensuring that the reasons stay the same after `Goal` is resolved.
The `Goal` might have to keep around the `TransactionReplay` reasons and set/check them after resolvement, but it can all be done internally. I think this would also mean we don't have to break API for the `get_transaction_packages()`.

kontura:
I am closing this in favor of: https://github.com/rpm-software-management/dnf5/pull/1005

PR #983, opened 2023-10-19
Fix throwing exceptions, allow only "safe" types

jrohel:
Exceptions were thrown in the code, in which a pointer to text was passed that was no longer in memory at the time the exception was handled.
Fix: Replaced passing `char *` and `const char *` with `std::string` when throwing exception.

To prevent this from happening again, passing only defined "safe" types is now allowed.


jrohel:
Doc text for `concept AllowedErrorArgTypes`. Rebased.

jrohel:
I added a commit fixing `SolvRepo::load_solv_cache` method - Compare string content instead of pointers.

pkratoch:
> I added a commit fixing `SolvRepo::load_solv_cache` method - Compare string content instead of pointers.

Thanks!

Issue #57, opened 2022-09-23
Replace dnf-automatic functionality by DNF5


j-mracek:
dnf-automatic provides essential functionality for the automatic system administration. DNF5 must also provides such a functionality to allow DNF replacement in Fedora 39.

rathann:
Please include https://github.com/rpm-software-management/dnf/issues/1954 when implementing the replacement. Information about pending updates on a system is valuable intel for an attacker and should not be sent over unencrypted SMTP connection.

PR #984, opened 2023-10-19
Implement dnf-automatic functionality for dnf5

m-blaha:
New dnf5 plugin implementing this functionality compatible with what dnf4 does.

Resolves: https://github.com/rpm-software-management/dnf5/issues/57

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1399


dnf5-plugins/automatic_plugin/bin/dnf-automatic
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/sh
+# Compatibility wrapper to handle old "/usr/bin/dnf-automatic" name from dnf4.
+
+dnf5 automatic "$@"
```
j-mracek, line 4: What about to mark it as deprecated from beginning?

There is another question whether we should provide it, because it was used by services that were also provided by original package.

dnf5-plugins/automatic_plugin/config/etc/dnf/automatic.conf
```diff
@@ -0,0 +1,104 @@
+[commands]
```
j-mracek, line 1: May be we should open discussion about location of dnf5 command configurations

dnf5-plugins/automatic_plugin/bin/dnf-automatic
m-blaha: I'm not completely sure we need it either. The services were also adjusted to the new name.

dnf5-plugins/automatic_plugin/config/etc/dnf/automatic.conf
m-blaha: Sure. I wanted to keep automatic as compatible as possible.

j-mracek:
@m-blaha May I ask you to rebase the PR

dnf5-plugins/automatic_plugin/config/etc/dnf/automatic.conf
j-mracek: `[command]` section is somehow redundant with command-line options.
`random_sleep = 0` has an alias `--timer Apply random delay before execution.`
`download_updates = yes/no` has an alias `--downloadupdates Automatically download updated packages` and `--no-downloadupdates Do not automatically download updated packages`
`apply_updates = yes/no` has an alias  `--installupdates Automatically install downloaded updates` and `--no-installupdates Do not automatically install downloaded updates`

The command `automatic` requires path to configuration file and then it uses overrides over the configuration. I know how it was created, but the question is whether should we keep both path?

include/libdnf5-cli/output/smartcols_table_wrapper.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <libsmartcols/libsmartcols.h>
+
+
+namespace libdnf5::cli::output {
+
+class SmartcolsTableWrapper {
+public:
+    SmartcolsTableWrapper();
+    ~SmartcolsTableWrapper();
```
j-mracek, line 32: I think when assign/copy constructors are used, it will result in double de-allocation of memory or access of freed memory. The class contains raw pointers that are only copied as an address to memory. I am not sure whether move constructors also requires additional care.

dnf5/main.cpp
```diff
@@ -933,6 +933,14 @@ int main(int argc, char * argv[]) try {
...
 
     loggers.front()->info("DNF5 start");
+    std::string cmdline;
+    for (int i = 0; i < argc; ++i) {
+        if (i > 0) {
+            cmdline += " ";
+        }
+        cmdline += argv[i];
+    }
+    loggers.front()->info("DNF5 command line: {}", cmdline);
```
j-mracek, line 943: What about to merge both logging events? "DNF5 start" plus "DNF5 command line: {}".
What about new message: `DNF5 performing command: {}`. Or it would be possible to remove logging "DNF5 start" and keep the new message `DNF5 command line: {}`. I see only one reason to keep both - when formatting of  message `"DNF5 command line: {}"` fails. What do you think?

dnf5/main.cpp
```diff
@@ -933,6 +933,14 @@ int main(int argc, char * argv[]) try {
     loggers.emplace_back(std::make_unique<libdnf5::MemoryBufferLogger>(max_log_items_to_keep, prealloc_log_items));
 
     loggers.front()->info("DNF5 start");
+    std::string cmdline;
```
j-mracek, line 936: If I am not mistaken method `std::string Context::get_cmdline()` format the same string.

include/libdnf5-cli/output/smartcols_table_wrapper.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef LIBDNF5_CLI_OUTPUT_SMARTCOLS_TABLE_WRAPPER_HPP
+#define LIBDNF5_CLI_OUTPUT_SMARTCOLS_TABLE_WRAPPER_HPP
+
+#include <libsmartcols/libsmartcols.h>
+
+
+namespace libdnf5::cli::output {
+
+class SmartcolsTableWrapper {
```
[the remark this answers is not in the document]
m-blaha: added.

dnf5/main.cpp
m-blaha: Makes sense. I'll refactor the code to write only one message. I will also remove the code duplicity with command line arguments formatting.

m-blaha: Done.

dnf5-plugins/automatic_plugin/automatic.cpp
```diff
@@ -0,0 +1,468 @@
...
+    std::vector<std::string> urls{};
+    for (const auto & repo : repos) {
+        std::string proxy{};
+        try {
+            proxy = repo->get_config().get_proxy_option().get_value();
+            if (!proxy.empty()) {
+                urls.emplace_back(proxy);
+                continue;
+            }
+        } catch (const libdnf5::OptionValueNotSetError & e) {
```
j-mracek, line 126: Only note: It looks strange that you have to handle `libdnf5::OptionValueNotSetError` and `(!proxy.empty())`

dnf5-plugins/automatic_plugin/automatic.cpp
```diff
@@ -0,0 +1,468 @@
...
+    // parse url to [(host, service(port number or scheme))] using libcurl
+    std::vector<std::tuple<std::string, std::string>> addresses;
+    CURLUcode rc;
+    CURLU * url = curl_url();
+    unsigned int set_flags = CURLU_NON_SUPPORT_SCHEME;
+    unsigned int get_flags = 0;
+    for (auto & u : urls) {
+        rc = curl_url_set(url, CURLUPART_URL, u.c_str(), set_flags);
+        if (!rc) {
+            std::string host{};
```
j-mracek, line 161: Cosmetic - I think that `std::string host{};` could be replaced by `std::string host;`

dnf5-plugins/automatic_plugin/transaction_callbacks_simple.cpp
```diff
@@ -0,0 +1,109 @@
...
+            output_stream << "  Installing ";
+            break;
+        case libdnf5::transaction::TransactionItemAction::REMOVE:
+        case libdnf5::transaction::TransactionItemAction::REPLACED:
+            break;
+        case libdnf5::transaction::TransactionItemAction::REASON_CHANGE:
+        case libdnf5::transaction::TransactionItemAction::ENABLE:
+        case libdnf5::transaction::TransactionItemAction::DISABLE:
+        case libdnf5::transaction::TransactionItemAction::RESET:
+            throw std::logic_error(fmt::format(
```
j-mracek, line 57: Do I understand it correctly that when the logic_error is triggered here it will terminate RPM transaction? The consequence would be a broken system or did I overlooked something?

include/libdnf5-cli/output/transaction_table.hpp
```diff
@@ -580,18 +559,34 @@ bool print_transaction_table(Transaction & transaction) {
...
         scols_cell_set_color(scols_line_get_cell(ln, COL_NAME), action_color(tsmodule.get_action()));
     }
 
-    scols_print_table(tb);
-    scols_unref_symbols(sb);
-    scols_unref_table(tb);
+    return tb;
+}
+
+template <class Transaction>
```
j-mracek, line 574: May I ask you why you prefer to use a template?

doc/dnf5_plugins/automatic.8.rst
```diff
@@ -0,0 +1,238 @@
...
+=========
+
+``<config file>``
+    Path to configuration file. Defaults to ``/etc/dnf/automatic.conf``.
+
+
+Description
+=============
+
+Alternative CLI to ``dnf upgrade`` with specific facilities to make it suitable to be executed automatically and regularly from systemd timers, cron jobs and similar.
```
j-mracek, line 60: Note: I believe that it is an alternative not only t ``dnf upgrade``, but also to ``check-updates``

j-mracek:
I am not sure about:

- Keep `dnf-automatic` and `dnf-automatic-<action>` but ignore part of configuration file
  - configuration (dnf-automatic.conf) vs. command line overrides
- name of services `dnf-automatic` or `dnf5-automatic` with alias `dnf-automatic`
- Default location of DN5 automatic configuration file
  - `/etc/dnf/dnf-automatic.conf`
  - `/etc/dnf/dnf5-.../dnf-automatic.conf`

include/libdnf5-cli/output/transaction_table.hpp
m-blaha: The method is used by both dnf5 and dnf5daemon-client. The transaction transferred via dbus is a different class (DbusGoalWrapper), although it supports the same interface. Thus template is an efficient way how to support it.

dnf5-plugins/automatic_plugin/transaction_callbacks_simple.cpp
m-blaha: I've added a commit which removes throwing exception in transaction callbacks.

dnf5-plugins/automatic_plugin/emitters.cpp
```diff
@@ -0,0 +1,241 @@
...
+                curl_easy_setopt(curl, CURLOPT_URL, email_host.c_str());
+
+                curl_easy_setopt(curl, CURLOPT_MAIL_FROM, from.c_str());
+
+                for (const auto & eml : to) {
+                    recipients = curl_slist_append(recipients, eml.c_str());
+                }
+                curl_easy_setopt(curl, CURLOPT_MAIL_RCPT, recipients);
+
+                FILE * payload_file = fmemopen(payload.data(), payload.size(), "r");
```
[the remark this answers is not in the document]
m-blaha: Fixed.

doc/dnf5_plugins/automatic.8.rst
m-blaha: Well, it kind of is. But I still consider documenting it like an `upgrade` alternative clearer. Automatic have basically three modes:

1. only check - this is like `dnf upgrade` when the user aborts the transaction
2. download upgrades - this is like `dnf upgrade --downloadonly`
3. and finally do the upgrades - alternative to `dnf upgrade` or `dnf upgrade --security`

dnf5-plugins/automatic_plugin/automatic.cpp
```diff
@@ -0,0 +1,468 @@
...
+    std::random_device rd;
+    std::mt19937 rng(rd());
+    std::uniform_int_distribution<int> distribution(0, max_value);
+
+    sleep(distribution(rng));
+}
+
+static bool reboot_needed(const libdnf5::base::Transaction & transaction) {
+    static const std::set<std::string> need_reboot = {
+        "kernel", "kernel-rt", "glibc", "linux-firmware", "systemd", "dbus", "dbus-broker", "dbus-daemon"};
```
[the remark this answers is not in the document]
m-blaha: needs-restarting plugin did not exists back then. But now it does so I'll use the shared code.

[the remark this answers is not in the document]
m-blaha: Fixed, now the same code as needs-restarting uses is used.

dnf5-plugins/automatic_plugin/automatic.cpp
m-blaha: Done.

dnf5/main.cpp
j-mracek: May be I am mistaken but following code might be redundant:
```
    for (int i = 0; i < argc; ++i) {
        if (i > 0) {
            cmdline += " ";
        }
        cmdline += argv[i];
    }
```
and could be replaced by `std::string cmdline = context.get_cmdline()`.

m-blaha: The Context is not yet instantiated. But the Context does not need to have argv/argc attributes, so I pass only concatenated command line to it using `contect.set_cmdline()`. And context.get_cmdline() is now simplified to plain `return cmdline;`.

j-mracek: LGTM, thank you

jan-kolarik:
> I am not sure about:
> 
> * Keep `dnf-automatic` and `dnf-automatic-<action>` but ignore part of configuration file
>   
>   * configuration (dnf-automatic.conf) vs. command line overrides
> * name of services `dnf-automatic` or `dnf5-automatic` with alias `dnf-automatic`
> * Default location of DN5 automatic configuration file
>   
>   * `/etc/dnf/dnf-automatic.conf`
>   * `/etc/dnf/dnf5-.../dnf-automatic.conf`

My thoughts:
- config file location
  - it is a plugin, so in the default `dnf5-plugins` place
  - but I'd first read the conf from the `/etc/dnf/dnf-automatic.conf` if it exists and override it with the values from the new place
- binary wrapper
  - I vote for keeping `/usr/bin/dnf-automatic` wrapper together with some deprecation note
- config file vs CLI overrides
  - why not have config parameters which are overridden by CLI parameters? I feel it consistent with how dnf5 conf is working
- timers
  - not sure about these, who are the users?

j-mracek:
Tested also locally

PR #985, opened 2023-10-23
readthedocs: install missing dependency `sphinx_rtd_theme`

kontura:
According to: https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html the dependency should be added to requirements.txt.

It was working previously becuase it was automatically installed by readthedocs, this has been changed: https://blog.readthedocs.com/python-core-requirements-changed/


jan-kolarik (approved):
Thank you for investigating and fixing this!

PR #986, opened 2023-10-23
key_value_table: Method to drop empty group

jan-kolarik:
Allow dropping the table group without any children to `KeyValueTable` API.

Created as a follow-up to the https://github.com/rpm-software-management/dnf5/pull/896 being closed.


inknos:
lgtm, thanks

PR #988, opened 2023-10-26
Fix and use `ModuleProfile::is_default()` method

pkratoch:


libdnf5/module/module_profile.cpp
```diff
@@ -64,16 +64,20 @@ bool ModuleProfile::is_default() const {
     if (!profile) {
         return {};
     }
-    return modulemd_profile_is_default(profile);
+    return is_default_profile;
```
j-mracek, line 67: I believe that following code could be removed (lane 64 - 66)
```
    if (!profile) {
        return {};
    }
```

j-mracek: Be honest I don't like for a method that return `bool` the following construct - `return {}`; I am not sure whether it returns `true`, `false` or uninitialized value (random).

libdnf5/module/module_item.cpp
```diff
@@ -187,9 +187,14 @@ std::vector<std::string> ModuleItem::get_demodularized_rpms() const {
...
     std::vector<ModuleProfile> result_profiles;
     GPtrArray * profiles = modulemd_module_stream_v2_search_profiles((ModulemdModuleStreamV2 *)md_stream, name);
+    const auto & default_profiles = module_sack->get_default_profiles(get_name(), get_stream());
 
     for (unsigned int i = 0; i < profiles->len; i++) {
-        result_profiles.push_back(ModuleProfile(static_cast<ModulemdProfile *>(g_ptr_array_index(profiles, i))));
+        const auto & modulemd_profile = static_cast<ModulemdProfile *>(g_ptr_array_index(profiles, i));
+        result_profiles.push_back(ModuleProfile(
+            modulemd_profile,
+            std::find(default_profiles.begin(), default_profiles.end(), modulemd_profile_get_name(modulemd_profile)) !=
```
j-mracek, line 196: This is not wrong,  but  written code in that way might be less readable (this is a personal opinion). 
```
std::find(default_profiles.begin(), default_profiles.end(), modulemd_profile_get_name(modulemd_profile)) !=
                default_profiles.end()));
```
It is possible to use temporal variable
```
bool is_default  = std::find(default_profiles.begin(), default_profiles.end(), modulemd_profile_get_name(modulemd_profile)) !=
                default_profiles.end()));
```

libdnf5/module/module_profile.cpp
pkratoch: It was a remnant from dnf4. I am not sure why it was used like that there, as it probably makes sense only for more complicated types. But anyway, you are right it can be removed.

libdnf5/module/module_item.cpp
pmatilai: Indeed, temporals are a simple yet powerful way to make complex conditions far more readable.

pkratoch: Sure, fixed.

j-mracek:
LGTM

j-mracek:
@pkratoch May I ask you whether there is any blocker why CI tests for module list cannot be enabled `dnf/module/list.feature`?

pkratoch:
> @pkratoch May I ask you whether there is any blocker why CI tests for module list cannot be enabled `dnf/module/list.feature`?

Oh, good point. I'll take a look at which tests could be enabled now.

Issue #267, opened 2023-02-06
build: Fails to build due to under-linking


mcrha:
When using stricter options for linking the project fails to build due to undefined symbols. Steps:

```
   export LDFLAGS='-Wl,--as-needed -Wl,-z -Wl,relro -Wl,-z -Wl,now -Wl,-z -Wl,defs'
   cmake .....
   make
```

current result is tons of undefined references, beginning with:

```
/usr/bin/ld: CMakeFiles/actions.dir/actions.cpp.o: in function `(anonymous namespace)::Actions::substitute(libdnf::base::TransactionPackage const*, libdnf::rpm::Package const*, std::basic_string_view<char, std::char_traits<char> >, std::filesystem::__cxx11::path, int)':
..../dnf5/libdnf-plugins/actions/actions.cpp:223: undefined reference to `libdnf::OptionBinds::Item::get_value_string[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:239: undefined reference to `libdnf::rpm::Package::get_name[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:241: undefined reference to `libdnf::rpm::Package::get_arch[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:243: undefined reference to `libdnf::rpm::Package::get_version[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:245: undefined reference to `libdnf::rpm::Package::get_release[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:247: undefined reference to `libdnf::rpm::Package::get_epoch[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:249: undefined reference to `libdnf::rpm::Package::get_na[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:251: undefined reference to `libdnf::rpm::Package::get_evr[abi:cxx11]() const'
/usr/bin/ld: ..../dnf5/libdnf-plugins/actions/actions.cpp:253: undefined reference to `libdnf::rpm::Package::get_nevra[abi:cxx11]() const'
[...]

PR #994, opened 2023-11-03
Fix building with -z defs

ppisar:
This patchset fixes issue #267. First commit fixes a transitive reliance on directly used librpm fucntions in builddep_plugin dnf plugin. The second commit disables -z defs for program plugins. As a result, it is possible to build DNF5 with "-z defs" among system-wide linker flags.


ppisar:
I changed the order to included headers to be in alphabetical order as recommended by the clang linter.
I only express my disappointment because header inclusion is significant. I think it should not be subject of linting.

jrohel:
The inclusion of the missing headers and linking to rpm libraries in the builddep plugin is a bug fix.
And I hope that explicitly disabling "-z defs" in CMakeLists.txt will solve the reported issue.
Thanks for the patch.

PR #995, opened 2023-11-03
Fix typo in github workflows

inknos:


ppisar (approved):
Makes sens. Now it matches restore-keys.

PR #997, opened 2023-11-04
Disable the RHSM plugin by default and enable it in the RPM spec

Conan-Kudo:
Most Linux distributions shipping DNF do not need the RHSM plugin, and it really is only useful for the Red Hat Enterprise Linux family of distributions. Given that the RHSM stuff is often not shipped elsewhere, it's better to have this off by default and only turn it on when needed.

This is done by ensuring that the RPM spec file (which is both the reference spec file and the spec file used for RH/Fedora) turns it on and builds it.


ppisar (approved):
This makes sense.

Issue #999, opened 2023-11-06
Transaction Store and Load API


kontura:
Operations such as `System Upgrade`, `Offline Upgrade`, `History Replay`, `History Redo`, `History Store` and others require an API for storing and loading transactions.

- [x] Add an API to store (serialize) a transaction into a JSON data format. We will need to be able to serialize both `base::transaction` and `transaction::transaction` (history transaction).
- [x] Add API to load the JSON serialized transaction into a Goal.

The first part is done but we still need:
- [x] Fix environments handling - requires ABI break
- [x] Add groups/environments replaying without loaded repos
- [x] Verify functionality without loading repos
- [x] Enable/Add CI tests
- [x] Integrate running without loading repos into offline command (system-upgrade)

The API should support packages (RPMs), groups, environments and potentially also modules.

kontura:
The first draft of the API is in PR: https://github.com/rpm-software-management/dnf5/pull/978

evan-goode:
Second draft of the API is PR https://github.com/rpm-software-management/dnf5/pull/1005

jan-kolarik:
Related issue to the mentioned ABI break: https://github.com/rpm-software-management/dnf5/issues/1025.

kontura:
Comps and running without loading repos PR: https://github.com/rpm-software-management/dnf5/pull/1264

jan-kolarik:
Make sure these scenarios are covered: 
- https://github.com/rpm-software-management/dnf/issues/2031
- https://issues.redhat.com/browse/RHEL-17494

Related PR in libdnf: 
- https://github.com/rpm-software-management/libdnf/pull/1644.

kontura:
The API is finished.

PR #1005, opened 2023-11-09
Add capabilities for storing and replaying a transaction

kontura:
This is a second iteration for: https://github.com/rpm-software-management/dnf5/pull/978 I have made a new PR because it nearly completely changed and the old comments could be confusing.

Now there is only one approach and replaying a transaction must go through a `libdnf5::Goal`. The new API is:
- `void libdnf5::Goal::add_transaction_replay(const std::string & json_serialized_transaction, libdnf5::GoalJobSettings & settings_p);`
- `std::string libdnf5::base::transaction::serialize_to_json();`
- `std::string libdnf5::transaction::transaction::serialize_to_json();`

For: https://github.com/rpm-software-management/dnf5/issues/999

### DIFF
--- a/dnf5/commands/history/history_store.cpp
+++ b/dnf5/commands/history/history_store.cpp
@@ -19,14 +19,65 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "history_store.hpp"
 
+#include "commands/history/transaction_id.hpp"
+
+#include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
+#include <libdnf5/utils/fs/temp.hpp>
+
 namespace dnf5 {
 
 using namespace libdnf5::cli;
 
 void HistoryStoreCommand::set_argument_parser() {
-    get_argument_parser_command()->set_description("Store transaction to a file");
+    auto & cmd = *get_argument_parser_command();
+    cmd.set_description("[experimental] Store transaction to a file");
+    auto & ctx = get_context();
+    auto & parser = ctx.get_argument_parser();
+
+    output_option = dynamic_cast<libdnf5::OptionString *>(
+        parser.add_init_value(std::make_unique<libdnf5::OptionString>("./transaction.json")));
+    auto query_format = parser.add_new_named_arg("output");
+    query_format->set_long_name("output");
+    query_format->set_short_name('o');
+    query_format->set_description("File path for storing the transaction, default is \"./transaction.json\"");
+    query_format->set_has_value(true);
+    query_format->set_arg_value_help("PATH");
+    query_format->link_value(output_option);
+    cmd.register_named_arg(query_format);
+
+    transaction_specs = std::make_unique<TransactionSpecArguments>(*this);
 }
 
-void HistoryStoreCommand::run() {}
+void HistoryStoreCommand::run() {
+    const auto ts_specs = transaction_specs->get_value();
+    auto & history = *get_context().base.get_transaction_history();
+    std::vector<libdnf5::transaction::Transaction> transactions;
+    auto logger = get_context().base.get_logger();
+
+    if (ts_specs.empty()) {
+        transactions = list_transactions_from_specs(history, {"last"});
+    } else {
+        transactions = list_transactions_from_specs(history, ts_specs);
+    }
+
+    if (transactions.empty()) {
+        throw libdnf5::cli::CommandExitError(1, M_("No transactions selected for storing, exactly one required."));
+    }
+    if (transactions.size() != 1) {
+        throw libdnf5::cli::CommandExitError(1, M_("Multiple transactions selected for storing, only one allowed."));
+    }
+
+    const std::string json = transactions[0].serialize();
+
+    std::filesystem::path tmp_path(output_option->get_value());
+    auto tmp_file = libdnf5::utils::fs::TempFile(tmp_path.parent_path(), tmp_path.filename());
+    auto & file = tmp_file.open_as_file("w+");
+    file.write(json);
+    file.close();
+
+    //TODO(amatej): add warning if file already exists?
+    std::filesystem::rename(tmp_file.get_path(), output_option->get_value());
+    tmp_file.release();
+}
 
 }  // namespace dnf5
--- a/dnf5/commands/history/history_store.cpp
+++ b/dnf5/commands/history/history_store.cpp
@@ -21,6 +21,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "commands/history/transaction_id.hpp"
 
+#include <libdnf5-cli/utils/userconfirm.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 #include <libdnf5/utils/fs/temp.hpp>
 
@@ -54,6 +55,17 @@ void HistoryStoreCommand::run() {
     std::vector<libdnf5::transaction::Transaction> transactions;
     auto logger = get_context().base.get_logger();
 
+    std::filesystem::path tmp_path(output_option->get_value());
+
+    if (std::filesystem::exists(tmp_path)) {
+        std::cout << libdnf5::utils::sformat(
+            _("File \"{}\" already exists, it will be overwritten.\n"), tmp_path.string());
+        // ask user for the file overwride confirmation
+        if (!libdnf5::cli::utils::userconfirm::userconfirm(get_context().base.get_config())) {
+            throw libdnf5::cli::AbortedByUserError();
+        }
+    }
+
     if (ts_specs.empty()) {
         transactions = list_transactions_from_specs(history, {"last"});
     } else {
@@ -69,13 +81,11 @@ void HistoryStoreCommand::run() {
 
     const std::string json = transactions[0].serialize();
 
-    std::filesystem::path tmp_path(output_option->get_value());
     auto tmp_file = libdnf5::utils::fs::TempFile(tmp_path.parent_path(), tmp_path.filename());
     auto & file = tmp_file.open_as_file("w+");
     file.write(json);
     file.close();
 
-    //TODO(amatej): add warning if file already exists?
     std::filesystem::rename(tmp_file.get_path(), output_option->get_value());
     tmp_file.release();
 }
--- a/include/libdnf5/base/goal.hpp
+++ b/include/libdnf5/base/goal.hpp
@@ -354,6 +354,13 @@ public:
     void add_provide_install(
         const std::string & spec, const libdnf5::GoalJobSettings & settings = libdnf5::GoalJobSettings());
 
+    /// @warning This method is experimental/unstable and should not be relied on. It may be removed without warning
+    /// Add serialized transaction request to the goal. Only one serialized transaction can be added per goal.
+    ///
+    /// @param json_serialized_transaction      A string of JSON serialized transaction
+    /// @param settings                         A structure to override default goal settings.
+    void add_serialized_transaction(const std::string & serialized_transaction, libdnf5::GoalJobSettings & settings);
+
     /// When true it allows to remove installed packages to resolve dependency problems
     void set_allow_erasing(bool value);
 
--- a/include/libdnf5/base/goal_elements.hpp
+++ b/include/libdnf5/base/goal_elements.hpp
@@ -166,7 +166,7 @@ public:
         return group_package_types ? &group_package_types.value() : nullptr;
     }
 
-    /// If set to true, group operations (install / remove) will only work
+    /// If set to true, group operations (install / remove / upgrade) will only work
     /// with the group itself, but will not add to the transaction any packages.
     bool group_no_packages{false};
 
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -163,6 +163,11 @@ public:
     /// Retrieve a list of the problems that occurred during `check_gpg_signatures` procedure.
     std::vector<std::string> get_gpg_signature_problems() const noexcept;
 
+    /// @warning This method is experimental/unstable and should not be relied on. It may be removed without warning
+    /// Serialize the transaction into a json data format which can be later loaded
+    /// into a `libdnf5::Goal` and replayed.
+    std::string serialize();
+
 private:
     friend class TransactionEnvironment;
     friend class TransactionGroup;
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -29,6 +29,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "rpm/solv/goal_private.hpp"
 #include "solv/id_queue.hpp"
 #include "solv/pool.hpp"
+#include "transaction/transaction_sr.hpp"
 #include "transaction_impl.hpp"
 #include "utils/string.hpp"
 #include "utils/url.hpp"
@@ -104,6 +105,7 @@ public:
     void add_resolved_group_specs_to_goal(base::Transaction & transaction);
     void add_resolved_environment_specs_to_goal(base::Transaction & transaction);
     GoalProblem add_module_specs_to_goal(base::Transaction & transaction);
+    GoalProblem add_transaction_replay_specs_to_goal(base::Transaction & transaction);
     GoalProblem add_reason_change_specs_to_goal(base::Transaction & transaction);
 
     std::pair<GoalProblem, libdnf5::solv::IdQueue> add_install_to_goal(
@@ -203,6 +205,8 @@ private:
 
     void install_group_package(base::Transaction & transaction, libdnf5::comps::Package pkg);
     void remove_group_packages(const rpm::PackageSet & remove_candidates);
+
+    std::unique_ptr<std::pair<transaction::TransactionReplay, GoalJobSettings>> serialized_transaction;
 };
 
 Goal::Goal(const BaseWeakPtr & base) : p_impl(new Impl(base)) {}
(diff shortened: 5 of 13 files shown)


dnf5/commands/history/history_store.cpp
```diff
@@ -19,14 +19,58 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+    if (transactions.size() != 1) {
+        throw libdnf5::cli::CommandExitError(1, M_("Multiple transactions selected for storing, only one allowed."));
+    }
+
+    std::filesystem::path path(output_option->get_value());
+    std::ofstream serialized_trans;
+    //TODO(amatej): add warning if file already exists?
+    serialized_trans.open(path);
+    serialized_trans << transactions[0].serialize_to_json();
+    serialized_trans.close();
```
j-mracek, line 73: I would recommend to perform the operation atomically .
1. Create a temp file
2. Generate content an save it in temp unique file
3. Move temp file to the new location

This approach will ensure that the output will be 100% correct and not only partially written or corrupted.

include/libdnf5/base/goal_elements.hpp
```diff
@@ -166,9 +166,12 @@ struct GoalJobSettings : public ResolveSpecSettings {
         return group_package_types ? &group_package_types.value() : nullptr;
     }
 
-    /// If set to true, group operations (install / remove) will only work
+    /// If set to true, group operations (install / remove / upgrade) will only work
     /// with the group itself, but will not add to the transaction any packages.
     bool group_no_packages{false};
+    /// If set to true, environment operations (install / remove / upgrade) will only work
+    /// with the environment itself, but will not add to the transaction any groups.
+    bool environment_no_groups{false};
```
j-mracek, line 174: This change will brake ABI compatibility (C++).

ppisar: I agree that we should add a pimpl member to every public class on the first need of breaking ABI. So that we can survive without breaking ABI later.

kontura: I am also in favor of pimpl.

I have a question if we should do it one by one, only when its needed or if we should add the pimpls to all API classes at once preemptively to break the ABI the least amount of times?

jan-kolarik: > I am also in favor of pimpl.
> 
> I have a question if we should do it one by one, only when its needed or if we should add the pimpls to all API classes at once preemptively to break the ABI the least amount of times?

I like the preemption idea. If we can detect all potential places now, I am for preparing the codebase for this purpose in advance.

dnf5/commands/history/history_store.cpp
kontura: I have changed it, please take a look.

include/libdnf5/base/goal_elements.hpp
kontura: (I have removed the problematic part from the PR so it doesn't block it, there is only a TODO for now.)

j-mracek: If we will create `pimpl`, how we can keep API compatibility? The structure does not have methods -> getters, setters

j-mracek:
I am not sure whether `dnf history replay` is the best command. We can keep it as a compatibility alias, but I guess we could look for a better command. We already discussed to create `do` command to replace `shell`. What do you think?

include/libdnf5/base/transaction.hpp
```diff
@@ -163,6 +163,10 @@ class Transaction {
     /// Retrieve a list of the problems that occurred during `check_gpg_signatures` procedure.
     std::vector<std::string> get_gpg_signature_problems() const noexcept;
 
+    /// Serialize the transaction into a json data format which can be later loaded
+    /// into a `libdnf5::Goal` and replayed.
+    std::string serialize_to_json();
```
j-mracek, line 168: The method is ok, but I have realized that there might be a white spot in our workflow if we want to download, test and store rpm path in the file.
Before download we can call `Command::goal_resolved()`, but there is no hook later.

Do you think whether it would be possible to return a different set of information if `serialize_to_json()`  as called before downloading of packages (without path to rpm) or after download (include paths of rpm)?

include/libdnf5/base/goal.hpp
```diff
@@ -354,6 +354,12 @@ class Goal {
     void add_provide_install(
         const std::string & spec, const libdnf5::GoalJobSettings & settings = libdnf5::GoalJobSettings());
 
+    /// Add transaction replay request to the goal.
+    ///
+    /// @param json_serialized_transaction      A string of JSON serialized transaction
+    /// @param settings                         A structure to override default goal settings.
+    void add_transaction_replay(const std::string & json_serialized_transaction, libdnf5::GoalJobSettings & settings);
```
j-mracek, line 361: What about to replace `replay` by `from_file`? The current PR allows not only replay the transaction but also add additional requests.

include/libdnf5/base/transaction.hpp
kontura: > Do you think whether it would be possible to return a different set of information if `serialize_to_json()` as called before downloading of packages (without path to rpm) or after download (include paths of rpm)?

Yes, the `package_path` should be included only if the package is available locally.

include/libdnf5/base/goal.hpp
kontura: It is added from string so perhaps something like `add_transaction_actions_from_string()`?

Although I kind of like the `replay` word because it points to the format of the loaded string.

kontura:
> I am not sure whether `dnf history replay` is the best command. We can keep it as a compatibility alias, but I guess we could look for a better command. We already discussed to create `do` command to replace `shell`. What do you think?

Sure, I think we will also need some command for the offline actions. I would do that in a different PR thought.
I can remove the `history replay` or we can keep it if it is going to stay as an alias.

kontura:
I have marked the new API as `experimental` and also extended the `history store/replay` commands description with `[experimental]` keyword.

include/libdnf5/base/goal.hpp
j-mracek: The method that creates the string is called `serialize_to_json`. What about to add to the name both worlds. `add_serialized_transaction_from_json`  or shorter `add_serialized_transaction` with skipping type of information.
Anyway `add_transaction_actions_from_string()` is also ok.

dnf5/commands/history/history_store.cpp
```diff
@@ -19,14 +19,75 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+    int fd = mkstemp(dest.data());
+    if (fd == -1) {
+        throw std::filesystem::filesystem_error(
+            "cannot create temporary file", dest, std::error_code(errno, std::system_category()));
+    }
+
+    if (write(fd, json.data(), json.size()) == -1) {
+        throw std::filesystem::filesystem_error(
+            "cannot write to temporary file", dest, std::error_code(errno, std::system_category()));
+    }
```
j-mracek, line 82: I believe that the code will keep open the file descriptor and also the temp file will remain on the system. What about following code?
```
close(fd);
std::error_code ec;
std::filesystem::remove(tmpfilepath, ec);
```

dnf5/commands/history/history_store.cpp
```diff
@@ -19,14 +19,75 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+
+    if (close(fd) != 0) {
+        throw std::filesystem::filesystem_error(
```
j-mracek, line 85: The TMP file will remain on the system

dnf5/commands/history/history_store.cpp
```diff
@@ -19,14 +19,75 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+            "cannot close temporary file", dest, std::error_code(errno, std::system_category()));
+    }
+
+    //TODO(amatej): add warning if file already exists?
+    std::filesystem::rename(dest, output_option->get_value());
```
j-mracek, line 90: This might also fail - raises an error and tmp file will remain. I checked our code and we never handle that situation.

libdnf5/base/goal.cpp
```diff
@@ -203,6 +205,9 @@ class Goal::Impl {
 
     void install_group_package(base::Transaction & transaction, libdnf5::comps::Package pkg);
     void remove_group_packages(const rpm::PackageSet & remove_candidates);
+
+    transaction::TransactionReplay transaction_replay;
+    GoalJobSettings transaction_replay_settings;
```
j-mracek, line 210: What about to replace it by a vector with pair type inside. An empty vector should be cheaper then two default structures. What do you think?

libdnf5/base/goal.cpp
```diff
@@ -2252,11 +2396,21 @@ base::Transaction Goal::resolve() {
...
+
     transaction.p_impl->set_transaction(p_impl->rpm_goal, module_sack, ret);
 
     return transaction;
 }
 
+
+void Goal::add_transaction_replay(
+    const std::string & json_serialized_transaction, libdnf5::GoalJobSettings & settings) {
+    //TODO(amatej): allow only one replay per goal?
```
j-mracek, line 2409: I guess that we should add a limitation (exception)  or support more transactions. The current implementation performs the last provided transaction, and I don't think that this is right.
If we will add support only for one file per goal we can anytime extend the support to multi transactions support without changing API, ABI, or introducing a new exception. If we will start with a support of multiple files then adding the restriction might be painful. May I ask you for your preferences?

libdnf5/base/transaction.cpp
```diff
@@ -448,6 +449,22 @@ void Transaction::Impl::set_transaction(
...
         packages.emplace_back(std::move(tspkg));
     }
+
+    // After all packages were added check rpm reason overrides
+    //TODO(amatej): Should these reason overrides be done via a REASON_CHANGE action? This would show up in the transaction table
+    if (!rpm_reason_overrides.empty()) {
+        for (auto & pkg : packages) {
+            const auto reason_override = rpm_reason_overrides.find(pkg.get_package().get_nevra());
+            if (reason_override != rpm_reason_overrides.end()) {
+                // For all other actions we don't want to lower the reason
```
j-mracek, line 459: I am not sure about the condition, may be after finishing the review I will have a better picture.

j-mracek: I will test some edge cases

j-mracek:
I discovered confusing output
I did `dnf5 install acpi`   then `dnf5 mark weak acpi`, then stored the transaction with reason change, then `dnf5 history replay transaction.json`
```
dnf5 history replay transaction.json
Repositories loaded.
Argument 'acpi-0:1.7-20.fc38.x86_64' matches only excluded packages.
```
`acpi` was not excluded, because I was able to install it.

j-mracek:
@kontura May I ask you to verify following scenario
`sudo dnf5 history store 57`
`dnf5 history store 56`
Expected fail - unable to replace the file

j-mracek:
Similar to https://github.com/rpm-software-management/dnf5/pull/1005#issuecomment-1824327195
When I have installed a different version when transaction is a reason change.
```
sudo dnf5 history replay transaction.json
Repositories loaded.
Argument 'iftop-0:1.0-0.30.pre4.fc38.x86_64' matches only excluded packages.
```

libdnf5/base/goal.cpp
```diff
@@ -596,6 +601,130 @@ GoalProblem Goal::Impl::add_module_specs_to_goal(base::Transaction & transaction
     return ret;
 }
 
+GoalProblem Goal::Impl::add_transaction_replay_specs_to_goal(base::Transaction & transaction) {
+    libdnf5::GoalJobSettings settings = transaction_replay_settings;
+    bool skip_unavailable = settings.resolve_skip_unavailable(base->get_config());
+
+    for (const auto & package_replay : transaction_replay.packages) {
+        libdnf5::GoalJobSettings settings_per_package = settings;
```
j-mracek, line 609: May I ask you why you create a copy of settings here?

libdnf5/base/goal.cpp
```diff
@@ -596,6 +601,130 @@ GoalProblem Goal::Impl::add_module_specs_to_goal(base::Transaction & transaction
...
+            transaction.p_impl->rpm_reason_overrides[package_replay.nevra] = package_replay.reason;
+        } else if (package_replay.action == transaction::TransactionItemAction::REMOVE) {
+            rpm_specs.emplace_back(GoalAction::REMOVE, package_replay.nevra, settings_per_package);
+            transaction.p_impl->rpm_reason_overrides[package_replay.nevra] = package_replay.reason;
+        } else if (package_replay.action == transaction::TransactionItemAction::REPLACED) {
+            if (!skip_unavailable) {
+                rpm_specs.emplace_back(GoalAction::REMOVE, package_replay.nevra, settings_per_package);
+            }
+        } else if (package_replay.action == transaction::TransactionItemAction::REASON_CHANGE) {
+            rpm_reason_change_specs.emplace_back(
```
j-mracek, line 644: Shell we pass full NEVRA here or only name.arch of the package?

libdnf5/base/transaction.cpp
j-mracek: `// For all other actions we don't want to lower the reason`
May I ask you why?

What do you think  is more important  -  reason that is listed in transaction file or reason in system-state (internal DNF5 DB)?

Note:
Upgrade is an operation that normally does not change reason,  right? But when upgrading  package also obsoletes another package than the final reason is the strongest of both.

include/libdnf5/base/goal.hpp
kontura: I like the shorter `add_serialized_transaction()` which drops the JSON. I am thinking about dropping it also from the `serialize_to_json()`.
So that we have just:
- `libdnf5::base::transaction::serialize()`
- `libdnf5::transaction::transaction::serialize()`
- `libdnf5::goal::add_serialized_transaction()`

libdnf5/base/goal.cpp
kontura: I think limiting to one is the right approach. You are right we can always extend it later if there is a use case.

dnf5/commands/history/history_store.cpp
kontura: I think it would be best to use the `libdnf5::utils::fs::temp.hpp` to handle this. I made a PR to expose it: https://github.com/rpm-software-management/dnf5/pull/1050

libdnf5/base/goal.cpp
kontura: I think vector might be a little bit confusing if we only allow one replay per goal. I will try `std::unique_ptr`.

kontura:
> I discovered confusing output I did `dnf5 install acpi` then `dnf5 mark weak acpi`, then stored the transaction with reason change, then `dnf5 history replay transaction.json`
> 
> ```
> dnf5 history replay transaction.json
> Updating and loading repositories:
> Repositories loaded.
> Failed to resolve the transaction:
> Argument 'acpi-0:1.7-20.fc38.x86_64' matches only excluded packages.
> ```
> 
> `acpi` was not excluded, because I was able to install it.

I am not able to reproduce this, for me it says: 
```
Repositories loaded.
Package "acpi-1.7-21.fc39.x86_64" is already installed with reason "Weak Dependency".

Nothing to do.
```
Which seems reasonable. 
Is there some step missing in the reproducer?

libdnf5/base/goal.cpp
kontura: The `settings` are used to select the repository which can be different for each action. I wanted to avoid the actions affecting each other.
But then it is copied again into the particular vector of specs so it would be possible to have the settings just once and make sure we always overwrite the `to_repo_ids` or other used values.

I can change it if you prefer that? It would be more efficient.

libdnf5/base/goal.cpp
kontura: It would make the `REASON_CHANGE` work even if a different version than expected is installed.
But why should it work differently than all other actions?

libdnf5/base/transaction.cpp
kontura: I think the more important reason is the stronger one. Just like with the obsoletes.

Specifically there is a [test scenario](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/transaction-sr/replay.feature#L1228-L1282) where the replayed transaction has an upgrade with `Dependency` but on host the package has `User` reason. I think it would be wrong to lower the reason to `Dependency`. The package is user-installed.

So I think we don't want to lower the reason because some user installed packages could get marked as dependencies.

libdnf5/base/goal.cpp
j-mracek: Thank you for explanation. I am fine with that.

j-mracek:
> > I discovered confusing output I did `dnf5 install acpi` then `dnf5 mark weak acpi`, then stored the transaction with reason change, then `dnf5 history replay transaction.json`
> > ```
> > dnf5 history replay transaction.json
> > Updating and loading repositories:
> > Repositories loaded.
> > Failed to resolve the transaction:
> > Argument 'acpi-0:1.7-20.fc38.x86_64' matches only excluded packages.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `acpi` was not excluded, because I was able to install it.
> 
> I am not able to reproduce this, for me it says:
> 
> ```
> Updating and loading repositories:
> Repositories loaded.
> Package "acpi-1.7-21.fc39.x86_64" is already installed with reason "Weak Dependency".
> 
> Nothing to do.
> ```
> 
> Which seems reasonable. Is there some step missing in the reproducer?

The problem was fixed just recently (After my testing) - see https://github.com/rpm-software-management/dnf5/commit/763f9ccdfccc3c970603da1f7f5d9ebabd861b08

libdnf5/base/goal.cpp
j-mracek: This is a good question with no good answer.

I can argue that:
`dnf5 mark` with argument `dnf` will be impossible to replay when a different version is installed. We also know, that internal implementation does not care about version and only stores `name.arch` information.
In the following example the original version was not in any repository

```
$ sudo dnf5 mark dependency yum
$ dnf5 history store 61
$ sudo dnf5 upgrade yum
$ sudo dnf5 mark user  yum
$ dnf5 history replay transaction.json
Repositories loaded.
No match for argument: yum-0:4.18.1-20231110004827.0.g41a287e2.fc38.noarch
```

When I replay reason change (to dep) after I changed installed version and also after I changed reason (from dep to user) I got following problem.

```
dnf5 history replay transaction.json
Repositories loaded.
Argument 'yum-0:4.18.1-20231130005028.8.g6229c1d4.fc38.noarch' matches only excluded packages.
```

I think that the implementation in PR is more strict then it has to be. Then it is not robust to any changes of installed package when the reason change is replayed.

libdnf5/base/goal.cpp
j-mracek: Perfect.

libdnf5/base/transaction.cpp
j-mracek: Thank you for the explanation. Looks good to me

j-mracek:
I think there are 3 remaining items:

- Behavior related to reason change
- name of command `history replay`
- allow silent file override for `transaction.json`
  - Personally I have no problem with that, but it would be good that DNF5 will behave uniformly in those situations

I consider only `- name of command 'history replay'` as a item that we should address in this PR.

j-mracek:
LGTM

PR #1007, opened 2023-11-09
modules: Allow globs in module_spec arguments

pkratoch:


test/libdnf5/module/test_module.cpp
```diff
@@ -669,6 +669,55 @@ void ModuleTest::test_module_enable() {
...
+    std::vector<std::string> expected_active_module_specs{
+        "berries:main:4:6c81f848:x86_64",
+        "fruit-salad:main:12:2241675a:x86_64",
+        "gooseberry:5.5:2:72aaf46b6:x86_64",
+        "gooseberry:5.5:3:72aaf46b6:x86_64",
+        "vegetable-salad:latest:1:aaa456b:x86_64"};
+    std::vector<std::string> active_module_specs;
+    for (auto & module_item : base.get_module_sack()->get_active_modules()) {
+        active_module_specs.push_back(module_item->get_full_identifier());
+    }
```
ppisar, line 691: A suggestion: I think the test for globs could end here because following code does not deal with globs in any way. A shorter test is faster to run and easier to understand. What's your opinion?

ppisar (commented):
Otherwise, it looks good.

test/libdnf5/module/test_module.cpp
ppisar: This was meant right before "// Run the transaction" comment.

pkratoch: Makes sense, I shortened the test.

ppisar (approved):

PR #1008, opened 2023-11-09
Remove all copr builds since tmt will take care of them

inknos:
Wait some time before merging. We want to catch false positives in tmt wrapper before removing the old stack.

This PR will remove the actions:

- DNF 5 CI / Copr Build (pull_request_target)
- DNF 5 CI / DNF Integration Tests (--tags dnf5 --command dnf5) (pull_request_target)
- DNF 5 CI / DNF Integration Tests (--tags dnf5daemon --command dnf5daemon-client) (pull_request_target)


jan-kolarik:
Sure, let's drop 'em.

Issue #419, opened 2023-03-31
Translations infrastructure for dnf5


m-blaha:
- [x] prepare cmake targets for generating fresh *.pot files
- [x] prepare dnf5-l10n repo for exchanging source strings and translations with weblate (weblate creates a lot of commits and we do not want to flood dnf5 git history with them)
- [x] create a new dnf5 project in weblate
- [x] translations exchange dnf5 <-> weblate using github actions

PR #1009, opened 2023-11-10
Workflow for *.pot synchronization to weblate

m-blaha:
This workflow runs periodically once a day, but it's also possible to
run it ad hoc from the GitHub web interface. It consists of the
following steps:

- Clone the dnf5 repository.
- Build fresh *.pot files using the gettext-potfiles CMake target.
- Clone the dnf5-l10n synchronization repository.
- Each *.pot file represents a Weblate component. For each file, a
  directory is created (if it doesn't already exist), the *.pot file is
  copied there, and a PATH file with the actual location of the component
  in the dnf5 source tree is created.
- The new version is then directly pushed into the dnf5-l10n repository.

The dnf5-l10n repository serves the purpose of reducing commit traffic
with new translations coming from Weblate. This approach helps us
maintain a cleaner dnf5 Git history. New translations from Weblate will
be pulled automatically as part of the prepare-release GitHub workflow.

Thanks to cockpit team for inspiration!

Closes https://github.com/rpm-software-management/dnf5/issues/419


kontura (approved):
Looks good!

Issue #1014, opened 2023-11-14
Improve rpmlint checks in PRs


inknos:
Currently, the action used to run rpmlint checks is custom. https://github.com/rpm-software-management/dnf5/blob/892e9bb0c7a720ea6aea581caa6f608ba9dc0edc/.pre-commit-config.yaml#L24

We could provide an action as rpm-software-management project, or at least write a [mirror action](https://github.com/pre-commit/pre-commit-mirror-maker) for pre-commit that would follow rpmlint tags.

PR #1015, opened 2023-11-14
use rpmlint pre-commit

inknos:
Closes https://github.com/rpm-software-management/dnf5/issues/1014


ppisar (approved):
The change is reasonable. It replaces an intermediary git repository installing an rpmlint Python module from Python Index with using rpmlint directly from upstream.

ppisar:
The test failure is about disappearing coloring an output. Only in Fedora ≥ 39. This failure seems unrelated.

PR #1016, opened 2023-11-15
modules: Always print "[d]" in module list for default streams

pkratoch:


ppisar (approved):
Good. Disabled default streams are now reported both as default and disabled.
The failed tests are an unrelated issue #1020.

PR #1017, opened 2023-11-15
modules: Fix detecting module dependencies when no stream is specified

pkratoch:


ppisar (approved):
It looks good.
Failed tests are unrelated issue #1020.

PR #1010, opened 2023-11-12
dnf5: Implement "check" command

jrohel:

PR #1023, opened 2023-11-20
dnf5: Implement "check" command using librpm

jrohel:
Unlike the https://github.com/rpm-software-management/dnf5/pull/1010 implementation, this implementation uses librpm.

CI tests https://github.com/rpm-software-management/ci-dnf-stack/pull/1411

Example output of the command "check":
```
pipewire-pulseaudio-0.3.84-3.fc39.x86_64
 missing require "libpipewire-0.3.so.0()(64bit)"
 missing require "pipewire-libs(x86-64) = 0.3.84-3.fc39"
 installed conflict "pulseaudio-daemon" from "pipewire-pulseaudio-0.3.85-1.fc39.x86_64"
 installed conflict "pulseaudio-daemon" from "pulseaudio-16.1-5.fc39.x86_64"
 duplicate with "pipewire-pulseaudio-0.3.85-1.fc39.x86_64"
```


dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,311 @@
...
+        qva->qva_mode = 'q';
+        qva->qva_flags = VERIFY_DEPS;
+        qva->qva_source = RPMQV_ALL;
+        qva->qva_sourceCount = 1;
+        qva->qva_showPackage = group_pkgs_same_name_arch;
+
+        struct rpmInstallArguments_s * ia = &rpmIArgs;
+        rpmtsSetFlags(ts, rpmtsFlags(ts) | (ia->transFlags & RPMTRANS_FLAG_NOPLUGINS));
+
+        rpmcliQuery(ts, qva, NULL);
```
pmatilai, line 261: I'm a bit surprised to see rpmcliVerify() and rpmcliQuery() used, these cli-variants of the API aren't particularly popular amongst developers...

Not that it's wrong, but if it was me, I would've probably just used something like

```
Header h = NULL;
rpmdbMatchIterator mi = rpmtsInitIterator(ts, RPMDBI_PACKAGES, NULL, 0);
while ((h = rpmdbNextIterator(mi))) {
    find_problems(ts, h);
}
rpmdbFreeIterator(mi);
```

...and similarly for group_pkgs_same_name_arch.

jrohel: @pmatilai To be honest, I don't like using `rpmcliVerify` and `rpmcliQuery`. But those functions do more than just iterate.
Example from `rpmciVerify`:
```
    rpmtsOpenDB(ts, O_RDONLY);
    rpmdbOpenAll(rpmtsGetRdb(ts));
    if (rpmChrootSet(rpmtsRootDir(ts)) || rpmChrootIn()) {
```
and also calls `rpmtsSetVSFlags`.

I don't know why all this is being done. That's why I've used higher functions so far. And in the future, when I understand it, I will rewrite this piece of code. I assume when adding file checking support. But if it's as easy as you suggest, I can do it now.

pmatilai: Those functions do all manner of "generic" stuff that's probably not relevant to you. But hmm, rpmtsCheck() can end up looking into the filesystem so if you need "check" to work with --installroot then (at least some of) that chroot dance done by rpmcliVerify() is necessary :thinking:

jrohel:
I fixed the `Problem` structure comment. (Comment lines were swapped).
PR is ready to merge. If you agree with the output.
I would like to add file checking in the next PR.

jrohel:
I added documentation.

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,315 @@
...
+    option->set_description(descr);
+    option->set_parse_hook_func([&opt_to_modify](cli::ArgumentParser::NamedArg *, const char *, const char *) {
+        opt_to_modify = true;
+        return true;
+    });
+    return option;
+}
+
+
+inline bool is_package_installonly(Header h) noexcept {
```
j-mracek, line 92: I guess that `libsolv` is able to identify installony packages faster. Libsolv uses internally integer instead of string and int comparison is significantly faster.

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,315 @@
...
+    rpmtd td = rpmtdNew();
+    headerGet(h, RPMTAG_PROVIDES, td, 0);
+    while (auto provide = rpmtdNextString(td)) {
+        for (const auto & installonly_pkg_provide : installonly_pkgs_provides) {
+            if (installonly_pkg_provide == provide) {
```
j-mracek, line 97: I believe that the behavior might differs from libsolv based provide query. On both side it might appear a string with version `kernel < 4`

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,315 @@
...
+                        }
+                        break;
+                    case RPMPROB_OBSOLETES:
+                        if (all || obsoleted) {
+                            problems[std::string{problem_str}].insert(Problem{
+                                .type = ProblemType::OBSOLETED, .nevra = std::string{alt_nevr}, .file_or_provide = ""});
+                        }
+                        break;
+                    default:;
+                        //std::cout << fmt::format("{} problem {}: {}", nevr, problem_str, alt_nevr) << std::endl;
```
j-mracek, line 165: May I ask you for the purpose of the comment?

jrohel: The remainder after debugging. I deleted it.

dnf5/commands/check/check.cpp
jrohel: But that means using libsolv. This implementation of the "check" command only uses librpm. The installed packages database is not loaded by libsolv.
Additionally, the `is_package_installonly` function is only called when a duplicate is found. Usually twice on a standard Fedora. When found installed second and third kernel versions.
Two calls of `is_package_installonly` function should be faster than having libsolv load the entire rpmdb and do the string indexing.

j-mracek: OK, thank you for explanation. I mark it as reslved

dnf5/commands/check/check.cpp
j-mracek: I am not sure about format of the string (`auto provide = rpmtdNextString(td)`), whether it returns name, or  the full string of provide. In most of cases it will not make any difference because we use a lot of provide names and one of them might be not versioned.

pmatilai: RPMTAG_PROVIDES returns just the provide names. If you want NEVR strings, use RPMTAG_PROVIDENEVRS instead (and pass HEADERGET_EXT to headerGet() flags)

pmatilai: Alternatively you could use the rpmds API so you don't need to care about the low-level header details.

j-mracek: `rpm -q --whatprovides 'kernel'` finds packages that provide `kernel = 6.7`

jrohel:
I fixed/improved reporting of obsoleted packages and add CI tests https://github.com/rpm-software-management/ci-dnf-stack/pull/1411.

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,323 @@
...
+    return rc;
+}
+
+
+// Groups packages with the same "name" and "architecture".
+// For "installonly" packages, only one "nevra" is inserted.
+int group_pkgs_same_name_arch([[maybe_unused]] QVA_t qva, [[maybe_unused]] rpmts ts, Header h) {
+    auto * arch = headerGetString(h, RPMTAG_ARCH);
+    if (!arch) {
+        // skip special package representing key installed in the rpm database
```
j-mracek, line 188: Interesting way of detecting gpg keys. Libsolv uses `memcmp(key, "gpg-pubkey", 10)` for the same purpose. @pmatilai Is there any chance to handle GPG keys differently in RPM? Or is there any smart or recommended detection in RPM whether element is package or a key?

pmatilai: A more accurate comment would be: "skip any non-packages (such as gpg-pubkey) in the database"

dnf5/commands/check/check.cpp
j-mracek: I have a good news. I discovered that the current implementation will work according expectations. Our documentation says `List of provide names of packages`. Our config parser is splitting `--setopt=installonlypkgs=' ',"kernel < 5"` to 3 values. It means that there is no requirement of support of versioned provides.
The second part is also perfect -> libsolv job does not support globs `job multiversion provides kernel*` will not match anything. I only discover a minor issue with `dnf5 repoquery --duplicates` where globs are incorrectly supported. I will provide a patch.

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,323 @@
...
+        qva->qva_source = RPMQV_ALL;
+        qva->qva_sourceCount = 1;
+        qva->qva_showPackage = group_pkgs_same_name_arch;
+
+        struct rpmInstallArguments_s * ia = &rpmIArgs;
+        rpmtsSetFlags(ts, rpmtsFlags(ts) | (ia->transFlags & RPMTRANS_FLAG_NOPLUGINS));
+
+        rpmcliQuery(ts, qva, NULL);
+
+        for (const auto & [na, nevras] : installed_na_nevras) {
```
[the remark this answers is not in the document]
jrohel: > In some very rare condition - problems[nevra] map item may refer to multiple object...

I improved the logic. Now only the first identical nevra is skipped as its own duplicate.

jrohel:
New changes:
Improved output consistency. Package names are always printed including nevra.
Improved the logic. Now only the first identical nevra is skipped as its own duplicate.

j-mracek:
Only note to performance. The new implementation is slower then dnf4 implementation. This is not a problem for check command, but if we would like to inform a user that the system is somehow broken before every transaction by default than this would be a problem.
```
time dnf5 check

real	0m12.816s
user	0m11.777s
sys	0m0.939s
```

```
$ time dnf check

real	0m0.772s
user	0m0.683s
sys	0m0.077s

```

dnf5/commands/check/check.cpp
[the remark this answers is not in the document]
j-mracek: Looks good

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,336 @@
...
+                switch (rpmProblemGetType(p)) {
+                    case RPMPROB_REQUIRES:
+                        if (all || dependencies) {
+                            problems[ensure_full_nevra(alt_cnevr)].insert(Problem{
+                                .type = ProblemType::MISSING_REQUIRE, .nevra = "", .file_or_provide = problem_cstr});
+                        }
+                        break;
+                    case RPMPROB_CONFLICT:
+                        if (all || dependencies) {
+                            auto nevr = ensure_full_nevra(cnevr);
```
j-mracek, line 153: After our discussion it was reveals that `nevr` variable contains full nevra string including EPOCH. May I ask you for renaming the variable?

dnf5/commands/check/check.cpp
```diff
@@ -0,0 +1,336 @@
...
+        rc = rpmpsNumProblems(ps);
+
+        if (rc > 0) {
+            rpmpsi psi = rpmpsInitIterator(ps);
+            while (auto p = rpmpsiNext(psi)) {
+                auto problem_cstr = rpmProblemGetStr(p);
+                libdnf_assert(problem_cstr != NULL, "command \"check\": rpmProblemGetStr() returns NULL");
+                auto cnevr = rpmProblemGetPkgNEVR(p);
+                libdnf_assert(cnevr != NULL, "command \"check\": rpmProblemGetPkgNEVR() returns NULL");
+                auto alt_cnevr = rpmProblemGetAltNEVR(p);
```
j-mracek, line 141: If `rpmProblemGetAltNEVR(p)` returns `nevra` formatted string, may I ask you to rename the variable according the format of the string?

jrohel: Renamed: `cnevr` -> `cnevra`, `alt_cnevr` -> `alt_cnevra`, `nevr` -> `nevra`, `alt_nevr` -> `alt_nevra`
Added `const` to more places.

j-mracek:
> > Only note to performance. The new implementation is slower then dnf4 implementation. This is not a problem for check command, but if we would like to inform a user that the system is somehow broken before every transaction by default than this would be a problem.
> 
> Yeah this isn't something you'd want to do before every transaction. Rpm needs to juggle around hundreds of megabytes of data in loading and reloading headers to do this, whereas libsolv has, or actually _is_ a specialized data structure optimized for the job.
> 
> So comparing to #1010 this actually looks like more code, and (much) slower too. While utilizing librpm for stuff is nice, libsolv is the far better tool for dependency work.

I did not compare this implementation with #1010 that uses libdnf5 queries. `Libsolv` offers to create one transaction and only verify installed packages (`SOLVER_VERIFY` flag).
This approach is used in repoquery `--unsatisfied` and it is slightly slower then `dnf check` - 1.7s vs. 0.7s.
```
        elif self.opts.pkgfilter == "unsatisfied":
            rpmdb = dnf.sack.rpmdb_sack(self.base)
            rpmdb._configure(self.base.conf.installonlypkgs, self.base.conf.installonly_limit)
            goal = dnf.goal.Goal(rpmdb)
            goal.protect_running_kernel = False
            solved = goal.run(verify=True)
            if not solved:
                print(dnf.util._format_resolve_problems(goal.problem_rules()))
            return
```
If we would try to verify  each package separately using libsolv transaction, then it would require much more time.

j-mracek:
LGTM

PR #1035, opened 2023-11-22
github: Do not run scheduled workflow on forks

m-blaha:
It would most likely fail anyway.


kontura:
The failing test is unrelated.
Thanks!

PR #1036, opened 2023-11-22
Support skip-broken option for the upgrade command

### LINKED ISSUE #1033: dnf5 upgrade does not recognize --skip-broken option
dnf5(8) manual lists `--skip-broken` among global options. Yet dnf5 does not recognize it:

~~~~
# dnf5 --skip-broken upgrade
Unknown argument "--skip-broken" for command "dnf5". Add "--help" for more information about the arguments.
# rpm -qf /usr/bin/dnf5
dnf5-5.1.7-20231116005017.18.g2d77952d.fc40.x86_64
~~~~

Personally I would expect the option to be specific to upgrade and install commands, but it's not accepted there either:

~~~~
# dnf5 upgrade --skip-broken 
Unknown argument "--skip-broken" for command "upgrade". Add "--help" for more information about the arguments.
~~~~

jan-kolarik:
- Document dropped support of the `--skip-broken` option from the `upgrade` command
- Exclude the `--skip-broken` option from the global options documentation, restricting only to related commands
- Add missing man pages documentation for options related to resolving missing dependencies

Closes #1033.

### DIFF
--- a/dnf5/commands/group/group_upgrade.cpp
+++ b/dnf5/commands/group/group_upgrade.cpp
@@ -38,7 +38,6 @@ void GroupUpgradeCommand::set_argument_parser() {
     group_specs = std::make_unique<GroupSpecArguments>(*this, ArgumentParser::PositionalArg::AT_LEAST_ONE);
 
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
-    auto skip_broken = std::make_unique<SkipBrokenOption>(*this);
     create_allow_downgrade_options(*this);
 }
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -172,6 +172,10 @@ Upgrade command
  * When any argument does not match any package or it is not installed, DNF5 fail. The behavior can be modified by
    the ``--skip-unavailable`` option.
  * Dropped upgrade command aliases ``upgrade-to`` and ``localupdate``.
+ * Dropped ``--skip-broken`` option, as it was already available in DNF4 only for compatibility reasons with YUM,
+   but it has no effect. Instead, the decision about selecting the newer version of a package into the transaction
+   and skipping possible dependency issues is based on the :ref:`best <best_option_ref-label>` or
+   :ref:`no-best <no_best_option_ref-label>` option.
 
 Updateinfo command
 ------------------
--- a/doc/commands/distro-sync.8.rst
+++ b/doc/commands/distro-sync.8.rst
@@ -47,6 +47,9 @@ Options
 ``--skip-broken``
     | Resolve any dependency problems by removing packages that are causing problems from the transaction.
 
+``--skip-unavailable``
+    | Allow skipping build dependencies not available in repositories. All available build dependencies will be installed.
+
 
 Examples
 ========
--- a/doc/commands/distro-sync.8.rst
+++ b/doc/commands/distro-sync.8.rst
@@ -44,6 +44,9 @@ Options
 ``--allowerasing``
     | Allow erasing of installed packages to resolve any potential dependency problems.
 
+``--skip-broken``
+    | Resolve any dependency problems by removing packages that are causing problems from the transaction.
+
 
 Examples
 ========
--- a/doc/commands/downgrade.8.rst
+++ b/doc/commands/downgrade.8.rst
@@ -42,6 +42,9 @@ Options
 ``--allowerasing``
     | Allow erasing of installed packages to resolve any potential dependency problems.
 
+``--skip-broken``
+    | Resolve any dependency problems by removing packages that are causing problems from the transaction.
+
 
 Examples
 ========
--- a/doc/commands/downgrade.8.rst
+++ b/doc/commands/downgrade.8.rst
@@ -45,6 +45,9 @@ Options
 ``--skip-broken``
     | Resolve any dependency problems by removing packages that are causing problems from the transaction.
 
+``--skip-unavailable``
+    | Allow skipping build dependencies not available in repositories. All available build dependencies will be installed.
+
 
 Examples
 ========
--- a/doc/commands/group.8.rst
+++ b/doc/commands/group.8.rst
@@ -102,6 +102,9 @@ Options
 ``--contains-pkgs``
     | Show only groups containing packages with specified names. List option, supports globs.
 
+``--skip-broken``
+    | Used with ``install`` command to resolve any dependency problems by removing packages that are causing problems from the transaction.
+
 
 Examples
 ========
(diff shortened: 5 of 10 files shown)


jan-kolarik:
I'll try to add also the corresponding CI test.

jan-kolarik:
Hmm, although I've added the option to the `upgrade` command, upon inspecting the `Goal` class, it appears that `skip_broken` is ignored for the `upgrade` action. Instead, the `best` option seems to handle the decision regarding missing dependencies in this case. This behavior aligns with the [dnf documentation](https://github.com/rpm-software-management/dnf/blob/master/doc/cli_vs_yum.rst#--skip-broken), and I assume it's intentional. Therefore, it seems unnecessary to include the option for the `upgrade` command, and we can consider removing it from the `group upgrade` command as well. Alternatively, we could leave it there for compatibility reasons, even though it won't affect behavior. What are your thoughts, @j-mracek? 🙂

ppisar:
Interesting, but it makes sense. Do I understand correctly that --best is dual to --skip-broken now? If so we should amend compatibility.conf to prevent people from reporting issues like #1033.

According to the linked document, --skip-broken is default behavior now. However, right now I observe a different behavior: Fedora 40 has a broken dependency like this:

~~~~
# dnf5 --disable-repo=rawhide --enable-repo=f40-build upgrade 
Repositories loaded.
Problem: package rpmlint-2.5.0-1.fc40.noarch requires python3.12dist(python-magic), but none of the providers can be installed
  - package python3-file-magic-5.45-1.fc40.noarch conflicts with python3-magic < 5.45-1.fc40 provided by python3-magic-0.4.27-5.fc39.noarch
  - installed package python3-file-magic-5.45-1.fc40.noarch obsoletes python3-magic < 5.45-1.fc40 provided by python3-magic-0.4.27-5.fc39.noarch
  - cannot install the best update candidate for package rpmlint-2.4.0-11.fc40.noarch
  - cannot install the best update candidate for package python3-file-magic-5.45-1.fc40.noarch
~~~~

Adding --no-best reveals many independent updates:

~~~~
# dnf5 --disable-repo=rawhide --enable-repo=f40-build --no-best upgrade
Repositories loaded.
Problem: package rpmlint-2.5.0-1.fc40.noarch requires python3.12dist(python-magic), but none of the providers can be installed
  - package python3-file-magic-5.45-1.fc40.noarch conflicts with python3-magic < 5.45-1.fc40 provided by python3-magic-0.4.27-5.fc39.noarch
  - installed package python3-file-magic-5.45-1.fc40.noarch obsoletes python3-magic < 5.45-1.fc40 provided by python3-magic-0.4.27-5.fc39.noarch
  - cannot install the best update candidate for package rpmlint-2.4.0-11.fc40.noarch
  - cannot install the best update candidate for package python3-file-magic-5.45-1.fc40.noarch

Package                                   Arch    Version                        Repository        Size
Upgrading:
 binutils                                 x86_64  2.41-14.fc40                   f40-build     27.2 MiB
[...]
Transaction Summary:
 Upgrading:        36 packages
 Replacing:        36 packages
~~~~

A config dump confims --best is in effect:

~~~~
# dnf5 --disable-repo=rawhide --enable-repo=f40-build --dump-main-config  | grep -E 'best|skip|strict'
best = 1
multilib_policy = best
skip_broken = 0
skip_if_unavailable = 0
skip_unavailable = 0
strict = 1
~~~~

I don't have any of them defined in /etc and /usr configuration.

j-mracek:
> Hmm, although I've added the option to the `upgrade` command, upon inspecting the `Goal` class, it appears that `skip_broken` is ignored for the `upgrade` action. Instead, the `best` option seems to handle the decision regarding missing dependencies in this case. This behavior aligns with the [dnf documentation](https://github.com/rpm-software-management/dnf/blob/master/doc/cli_vs_yum.rst#--skip-broken), and I assume it's intentional. Therefore, it seems unnecessary to include the option for the `upgrade` command, and we can consider removing it from the `group upgrade` command as well. Alternatively, we could leave it there for compatibility reasons, even though it won't affect behavior. What are your thoughts, @j-mracek? 🙂

This is tough question.  From libsolv implementation - upgrade job is always optional. If a package is installed then upgrade don't need to be performed. `upgrade` job is weaker then `install` job. To make `upgrade` job strict, it requires `--best` to force solver to pick the best candidates for operation. 

`Install acpi`
```
sudo dnf shell --disablerepo=* --enablerepo=updates
> install acpi
> remove acpi
> run
```
is failing
vs:
```
sudo dnf shell --disablerepo=* --enablerepo=updates
> upgrade acpi
> remove acpi
> run
```
is removing `acpi`

I would suggest to not add the option to command because it does not make any difference -  it would be there only for compatibility reason and does nothing. But it requires a good documentation in DNF4 vs DNF5 section. This approach is painful. If the option will be implemented then we have to mark it as dummy option that is added only for a compatibility reasons.

j-mracek:
@ppisar It might be surprising, but behavior of DNF and DNF5 does not different much for upgrade.

What you reported in https://github.com/rpm-software-management/dnf5/pull/1036#issuecomment-1822959038 is mostly related to difference in configuration files. DNF is shipped with https://github.com/rpm-software-management/dnf/blob/master/etc/dnf/dnf.conf (Fedora) or with https://github.com/rpm-software-management/dnf/blob/master/etc/dnf/dnf-strict.conf (RHEL). Without configuration overrides, DNF behaves something between (best is in code `false`). DNF5 is not shipped with any configuration overrides (Fedora 41) and behaves like with RHEL configuration.

The default setting for configuration option `skip_if_unavailable` and `best` is very important for distribution and I would suggest that we should open a FESCO ticket about defaults. Be honest I would prefer if configuration overrides will be delivered by distribution packages then DNF packages.

For security reasons we have a requirements to set `best=true` and also `best=false`. The first group argue that the transaction must fail to create attention to start resolve the issue. The second group thinks that partially updated system is better then not updated or broken packages might create a negative image for Fedora.

May be we can do something in the middle - when best=false, install possible updates,  but when a warning is created return value could be not 0.

ppisar:
I see my confusion came from dnf.conf(5) which documents `best=false`, but DNF5 has best=true` in code. Once we or Fedora deliver the override to `best=false`, the manual will agree with the user experience.

Question is whether the manual should document a program default, or the distribution override. Because that or the the other way the manual will disagree with the effective setting either on RHEL or on Fedora.

jan-kolarik:
I've dropped the commit introducing `--skip-broken` for the `upgrade` command. The option was also removed from the `group upgrade` command. Documentation was added to the DNF4 vs DNF5 changes section (see [here](https://github.com/rpm-software-management/dnf5/blob/ae08b940e308c32be54841acaf7dedac27f33b60/doc/changes.rst#upgrade-command) - although the introduced section links are only correctly visible in the generated HTML docs later).

j-mracek:
LGTM

ppisar:
Thanks. The documentation is now much better.

PR #1037, opened 2023-11-22
Implement command `provides`

inknos:


dnf5/main.cpp
```diff
@@ -72,6 +73,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <cstring>
 #include <filesystem>
 #include <iostream>
+#include <memory>
```
jan-kolarik, line 76: I believe this is unneeded.

dnf5/commands/provides/provides.cpp
```diff
@@ -0,0 +1,127 @@
...
+    auto & context = get_context();
+    context.set_load_system_repo(true);
+    context.base.get_config().get_optional_metadata_types_option().add_item(
+        libdnf5::Option::Priority::RUNTIME, libdnf5::METADATA_TYPE_FILELISTS);
+    context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
+}
+
+void ProvidesCommand::run() {
+    auto & ctx = get_context();
+    bool one_match = false;
```
jan-kolarik, line 72: Could you please consider a different name like `any_match`? It is a bit confusing to me, suggesting a meaning "only one match"...

dnf5/main.cpp
inknos: whops, my IDE put it there

dnf5/commands/provides/provides.cpp
inknos: makes sense

jan-kolarik:
> Hmm, not sure what's going on, but I cannot compile it locally:
> 
> ```
> /usr/bin/ld: CMakeFiles/dnf5.dir/main.cpp.o: in function `dnf5::ProvidesCommand::ProvidesCommand(dnf5::Context&)':
> /home/kolage/Projects/swm-devel/dnf5/dnf5/commands/provides/provides.hpp:32: undefined reference to `vtable for dnf5::ProvidesCommand'
> collect2: error: ld returned 1 exit status
> ```
> 
> I've already tried to cleanup and wipe my ccache, but still it remains unbuildable for me.
> 
> Looking into the overridden methods for the `ProvidesCommand`, everything looks good to me.

So it's just a problem on my local machine, I've tried compiling with 1MT VM and it went OK.

dnf5/commands/provides/provides.cpp
```diff
@@ -0,0 +1,127 @@
...
+        if ((spec.rfind("/bin/", 0) == 0) || (spec.rfind("/sbin/", 0) == 0)) {
+            spec.insert(0, "/usr");
+        }
+        filename_query.filter_file(std::vector<std::string>({spec}), libdnf5::sack::QueryCmp::GLOB);
+        if (!filename_query.empty()) {
+            full_package_query = filename_query;
+            match = libdnf5::cli::output::ProvidesMatchedBy::FILENAME;
+        }
+
+        result_pset |= full_package_query;
```
jan-kolarik, line 99: Sorry, but this is also confusing. Because here it doesn't have to be a full query, but just the provides or filenames list. Do we actually need to analyze the `result_pset` further if we know nothing was matched?

dnf5/commands/provides/provides.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+        for (auto package : result_pset) {
+            if (match != 0) {
+                libdnf5::cli::output::print_provides_table(package, spec.c_str(), match);
+                any_match = true;
+            } else {
+                unmatched_specs.insert(spec);
+            }
+        }
+        if (!unmatched_specs.empty() && any_match) {
```
jan-kolarik, line 109: So, what if we have a list with several specs, and only the last one is actually matched, but the previous ones having unmatched specs. In this case, the error output is not printed for the unmatched ones. Additionally, if you were to reorder that list of arguments with the matched one first, then the output would be present. However, the ordering should not affect this, right?

inknos:
```
$ ./build/dnf5/dnf5 -q provides asd asdd wget 
wget-1.21.3-7.fc39.aarch64 : A utility for retrieving files using the HTTP or FTP protocols
Repo         : @System
Matched From : 
Provide      : wget = 1.21.3-7.fc39

wget-1.21.3-7.fc39.aarch64 : A utility for retrieving files using the HTTP or FTP protocols
Repo         : fedora
Matched From : 
Provide      : wget = 1.21.3-7.fc39

No matches found for asd.
No matches found for asdd.
If searching for a file, try specifying the full path or using a wildcard prefix ("*/") at the beginning.
$ echo $?
1
```

It shouldn't. this is what is happening in your example. did I get it right?

dnf5/commands/provides/provides.cpp
inknos: that's right, I was stealing code for `repoquery --whatprovides` that actually needs this step. I can remove it

dnf5/commands/provides/provides.cpp
inknos: this should be fixed, I'll update the tests to track the wanted behavior

dnf5/commands/provides/provides.cpp
inknos: @jan-kolarik should be good now. I fixed the logic and added a case for binary-match, which I forgot

dnf5/commands/provides/provides.cpp
```diff
@@ -0,0 +1,145 @@
...
+                binary_query.filter_file(
+                    std::vector<std::string>({bin_spec, sbin_spec, usr_bin_spec, usr_sbin_spec}),
+                    libdnf5::sack::QueryCmp::GLOB);
+                if (!binary_query.empty()) {
+                    full_package_query = binary_query;
+                    match = libdnf5::cli::output::ProvidesMatchedBy::BINARY;
+                    break;
+                }
+            }
+        } while (match);  // do
```
jan-kolarik, line 118: Hmm, I find the `do {} while;` approach confusing because it's not a typical loop and involves many jumps (breaks), making the workflow hard to follow. Though I eventually understand it, it's a bit painful 😃

There are three blocks representing different match types—`filter_by_provides`, `filter_by_filename`, and `filter_by_binary`. Consider combining them into a single method, like `filter_spec`, which internally handles each condition. This way, the `run()` function can simply call `filter_spec`, making the code more straightforward. What do you think? 🙂

inknos: I thought about the chunkiness too, but I wanted to avoid `goto`s and similar stuff. I   chose do-while because it's kinda better imho to skip to some execution. My idea for the implementation was, you match -> you break.

One problem I had to face while writing this code is that the matching becomes gradually more expensive. so match-by-provides takes less time than match-by-filename which again takes less than match-by-binary. so, yes, it's a bit convoluted but if we want to have the same behavior like dnf4 and try to do things smart I think we have to cut the niceness somewhere. still not sure if there are better ways to achieve the desired results.

I like the `filter_spec` approach, though, to make `run` much more readable. I'll do it.

inknos: strongly agree

jan-kolarik: But still, if you're not convinced about this approach, feel free to keep it as it is. I might be biased due to my experience with TDD :P

PR #1041, opened 2023-11-24
transaction: Fix reporting spec matches only source

m-blaha:
Correctly report "Argument matches only source packages." in case user
tries to install a source rpm.
Current code misleadingly reports "Argument matches only excluded
packages."


j-mracek (approved):
LGTM

j-mracek:
Thank you for the patch

j-mracek:
The problem in CI is not related to PR

PR #1042, opened 2023-11-24
buid: fix untracked *.mo files

inknos:


m-blaha (approved):
Thanks! Let's just wait for a successful build.

PR #1044, opened 2023-11-24
action: Only commit translations on change

m-blaha:
When there are no changes in weblate, do not attempt to add a commit.


inknos:
lgtm. no need to wait for the build since it's an action

PR #1049, opened 2023-11-27
Modify installonly filter

j-mracek:
Libsolv is unable to match globs in installonly set `job multiversion provides kernel*` will not mach with anything.

The `Query::filter_installonly() should provide the same behavior therefore query filter must not support globs.


jrohel:
Cool. With this patch `dnf5 repoquery --duplicates` returns the same result as dnf4 `dnf repoquery --duplicates` in my tests.

I'm thinking. How about extending the CI tests for `repoquery --duplicates` to cover the case with globs in the installonly provide configuration?

j-mracek:
CI-test - https://github.com/rpm-software-management/ci-dnf-stack/pull/1412

jrohel:
LGTM. Thanks

PR #1050, opened 2023-11-28
Expose utis/fs/file.hpp and temp.hpp on API 

kontura:
It can be used in `copr_plugin/json.cpp` and I would like to use it in `history store` command.

Made as a part of: https://github.com/rpm-software-management/dnf5/issues/999


kontura:
> I have nothing against making the code public, but I would prefer to revisit exceptions and logging when something fails.
> 
> It `throw std::filesystem::filesystem_error()`, and the question is whether it should inherit from libdnf exceptions to prevent core dumps.

That sounds like a good idea to me, I will do the changes.

kontura:
@j-mracek I have added a commit with the new exception.

kontura:
Test update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1413

j-mracek:
LGTM

PR #1056, opened 2023-11-30
dnf5: "check" command - optimize, don't use rpmcli

jrohel:
It uses a loop to go through the rpmdb database instead of using the `rpmcliVerify` and `rpmcliQuery` functions (which then called back our code).

Why this PR:
- The original version uses the `rpmcliVerify` and `rpmcliQuery` functions from the librpm library. These functions, from C library, go through the rpmdb database and call back a C++ function from our code for each element. Our C++ function may throw an exception. The result may work, but it's generally better to avoid this situation (passing C++ exceptions through the C library.). In the new version, our code goes through the database and C callbacks are not used.
- librpm does not pass "user" data in callbacks. Therefore, they were passed through global variables (ok, they are in an anonymous namespace, available only in the given object file) but the code is not reentrant. The new version also uses global variables, but this is no longer necessary, we are not limited by callbacks in librpm. If we need to get rid of global variables (move the code to libdnf, make it reentrant, ...) it is now an easy change.
- The new code is simpler.
- The processor will do a few less instructions. But overall, the difference in speed will probably be negligible. That's not why I did this PR.

### DIFF
--- a/dnf5/commands/check/check.cpp
+++ b/dnf5/commands/check/check.cpp
@@ -79,7 +79,6 @@ struct Problem {
 };
 
 // what the program will check
-bool all{true};
 bool dependencies{false};
 bool duplicates{false};
 bool obsoleted{false};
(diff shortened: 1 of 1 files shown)


j-mracek:
I am sorry, I am not sure what it optimaze, because it requires the same time. Is it only a code simplification?

j-mracek:
LGTM

PR #1061, opened 2023-12-04
modules: Add module info command

pkratoch:


dnf5/commands/module/module_info.hpp
```diff
@@ -20,15 +20,19 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #ifndef DNF5_COMMANDS_MODULE_MODULE_INFO_HPP
 #define DNF5_COMMANDS_MODULE_MODULE_INFO_HPP
 
+#include "module_list.hpp"
+
 #include <dnf5/context.hpp>
+#include <libdnf5/module/module_query.hpp>
```
jan-kolarik, line 26: Minor: just that both `dnf5/context.hpp` and `libdnf5/module/module_query.hpp` are already included within the `module_list.hpp`.

dnf5/commands/module/module_list.cpp
```diff
@@ -64,8 +74,20 @@ void ModuleListCommand::run() {
...
+    print(query);
+    if (!query.empty()) {
+        output::print_modulelist_table_hint();
+    }
+
+    if (!unmatched_module_spec.empty()) {
+        for (auto const & module_spec : unmatched_module_spec) {
+            std::cerr << libdnf5::utils::sformat(_("No matches found for \"{}\"."), module_spec) << std::endl;
+        }
+    }
```
jan-kolarik, line 86: This reminds me of our discussion on the consistency of exit codes across various `info` and `list` commands. I've tasked myself with creating a follow-up on this.

include/libdnf5-cli/output/moduleinfo.hpp
```diff
@@ -0,0 +1,123 @@
...
+public:
+    template <typename ModuleItem>
+    void add_module_item(ModuleItem & module_item);
+
+private:
+    void add_multiline_value(const char * key, const std::vector<std::string> & multiline_value);
+};
+
+
+void ModuleInfo::add_multiline_value(const char * key, const std::vector<std::string> & values) {
```
j-mracek, line 49: May I ask you why the code is not in `CPP`?

jan-kolarik: Isn't it because it's used in the template below?

j-mracek:
May I ask you to provide CI tests for the command?

include/libdnf5-cli/output/moduleinfo.hpp
j-mracek: Function `add_line()` has a code in `cpp`

dnf5/commands/module/module_info.hpp
pkratoch: Thanks, removed.

pkratoch:
Yes, I plan to enable the CI tests that use `info` and not `install` or other unimplemented commands (though these tests mostly test something else and just happen to use `info`) and I will create an issue for those that use the `install` command.

But I could also create at least one test specifically for `info` that doesn't use `install` until the `install` command is implemented, if you think it would be better...

dnf5/commands/module/module_list.cpp
jan-kolarik: Nice, I haven't started yet, I'll pass it over to you then. Thanks!

jan-kolarik:
> Yes, I plan to enable the CI tests that use `info` and not `install` or other unimplemented commands (though these tests mostly test something else and just happen to use `info`) and I will create an issue for those that use the `install` command.
> 
> But I could also create at least one test specifically for `info` that doesn't use `install` until the `install` command is implemented, if you think it would be better...

The issue is fine for now I'd say. And related to the missing `install` command, I believe we can resolve it by utilizing the `Given I set dnf command to ...` construct in the tests. This way, we can set up the preparation phase using dnf4 and then use dnf5 to test the `info` or `list` output. I've tested it briefly locally, and I can prepare a PR later against the issue you'll create.

dnf5/commands/module/module_list.cpp
pkratoch: Issue: https://github.com/rpm-software-management/dnf5/issues/1075

PR #1063, opened 2023-12-05
vars: Use uname() for architecture autodetection

m-blaha:
The %_host_cpu macro is not runtime detected, it's only usable to detect
which architecture rpm was built for.
To detect that dnf5 is running on 32bit personality on x86_64 we need to
use uname().
The code is taken from libdnf (see libdnf/hy-util.cpp).

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2252671

### DIFF
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -33,7 +33,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <rpm/rpmmacro.h>
 #include <rpm/rpmts.h>
 #include <stdlib.h>
+#include <sys/auxv.h>
 #include <sys/types.h>
+#include <sys/utsname.h>
 
 #include <algorithm>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


j-mracek:
I believe that we can modify `init_lib_rpm(arch);` call. It was required to call it prior arch detection, but the new implementation does not require it. I would suggest to call it after the last step of detection is performed.

m-blaha:
> I believe that we can modify `init_lib_rpm(arch);` call. It was required to call it prior arch detection, but the new implementation does not require it. I would suggest to call it after the last step of detection is performed.

I've moved librpm initialization after arch detection.

j-mracek:
LGTM, There are some failing tests, but it looks like that two PRs were somehow mixed -> ready to merge

praiskup:
Awesome.  Any ETA for when this will be added to F40 (Rawhide) repos?  In particular this is blocking https://fedoraproject.org/wiki/Changes/BuildWithDNF5

PR #1068, opened 2023-12-06
Implement and use RotatingFileLogger

jrohel:
The DNF5 application uses the `logdir`, `log_size` and `log_rotate` configuration options to be compatible with DNF4.

Unit tests for the `RotatingFileLogger` class are included.

### DIFF
--- a/test/libdnf5/logger/test_rotating_file_logger.cpp
+++ b/test/libdnf5/logger/test_rotating_file_logger.cpp
@@ -1,0 +1,109 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "test_rotating_file_logger.hpp"
+
+#include "libdnf5/utils/fs/file.hpp"
+#include "libdnf5/utils/fs/temp.hpp"
+
+#include <libdnf5/logger/rotating_file_logger.hpp>
+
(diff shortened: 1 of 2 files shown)


jrohel:
I renamed the `MultiprocessRotatingFileLogger` class to `RotatingFileLogger`. I think that the original name was unnecessarily long.
Especially the filenames "multiple_rotating_file_logger.[ch]pp" were long. Now they are named "rotating_file_logger.[ch]pp"

PR is ready for review.

jrohel:
To the reviewer: Sorry. When renaming the `MultiprocessRotatingFileLogger` class to `RotatingFileLogger`, I forgot to edit one comment in the unit tests. So I re-pushed the last commit containing the unit tests. I only replaced `MultiprocessRotatingFileLogger` with `RotatingFileLogger` in one comment.

libdnf5/logger/rotating_file_logger.cpp
```diff
@@ -0,0 +1,163 @@
...
+        while (true) {
+            // open (create) the base log file and lock it
+            auto fd = ::open(
+                base_file_path.c_str(),
+                O_WRONLY | O_APPEND | O_CREAT | O_CLOEXEC,
+                S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
+            if (fd == -1) {
+                return;
+            }
+            ::flock(fd, LOCK_EX);
```
jan-kolarik, line 79: What about to use/extend the existing `libdnf5::utils::Locker`?

jan-kolarik:
It LGTM and the code is very well understandable, I like that. I just wondered, if we could use or extend our existing library functions for file operations. Or maybe at least create a follow-up ticket to unify the methods for file operations across the codebase.

libdnf5/logger/rotating_file_logger.cpp
jrohel: I thought about it. But our `libdnf5::utils::Locker` works differently. If the file cannot be locked, an exception is thrown. We use that elsewhere. But for logs, I want the logger to wait a moment and then write the message instead of generating an exception. So either I would call `libdnf5::utils::Lock::write_lock` in a cycle (polling) or extend our Locker. In addition, it opens a file with different attributes than I would like in the case of a logger.
And now I look in the source codes and there is a bug in the implementation of `libdnf5::utils::Locker`.

- So `libdnf5::utils::Locker` would be best fixed
- and then maybe expand it so that it can be better used in more places, for example even for a logger.

jan-kolarik: Got it. Please create an issue for the bug you've found. We can do the potential refactorings later.

jan-kolarik (approved):

PR #1069, opened 2023-12-06
Fix compiler choice in github workflows

inknos:


inknos:
the fix should be ok however github still uses the old workflow for some reason. I suspect it's not going to change until it's merged

image

PR #1071, opened 2023-12-07
dnf5daemon-client: use libdnf5 `am_i_root()`

kontura:
This is a small follow up to https://github.com/rpm-software-management/dnf5/pull/1051
I have noticed too late that the moved `am_i_root()` can be reused.


jan-kolarik (approved):
LGTM

PR #1078, opened 2023-12-09
dnf5: bash-completion: Use Command substitution instead of Process subst

jrohel:
The original version used "Process Substitution". dnf5 was run in another process and the output was referenced by filename (usually /dev/fd/63). That tends to be a problem in the chroot.

The new version uses "Command Substitution". dnf5 is run and its output is passed using "Here Strings".

The patch resolves the issue:
`dnf5 bash: /dev/fd/63: No such file or directory`


jan-kolarik (approved):
LGTM, thanks for a quick fix!

PR #1081, opened 2023-12-11
state: Fix catching fs exception

### LINKED ISSUE #1080: `Base.setup()` cannot create `/usr/lib/sysimage/libdnf5` with non-root user
As of 5.1.9 in Fedora Rawhide, libdnf5 fails to `.setup()` a Base object on a fresh system as a non-root user. It apparently tries to create `/usr/lib/sysimage/libdnf5` which does not exist on a fresh system and which a regular user does not have permissions to create:

```
>>> import libdnf5
>>> base = libdnf5.base.Base()
>>> base.load_config_from_file()
>>> base.setup()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.12/site-packages/libdnf5/base.py", line 492, in setup
    return _base.Base_setup(self)
           ^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: filesystem error: cannot create directories: Permission denied [/usr/lib/sysimage/libdnf5]
```

Other `dnf5` commands fail with the same error:

``` console
$ dnf5 makecache
filesystem error: cannot create directories: Permission denied [/usr/lib/sysimage/libdnf5]
```

jan-kolarik:
This is just a quick fix for #1080. But I believe a better way is needed, like wrapping the `create_directories` and throwing our `FileSystemError` exception. Also a CI test should be provided.

Closes #1080.

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -716,6 +716,8 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
             base->p_impl->get_system_state().save();
         } catch (const FileSystemError & ex) {
             logger->error("Cannot save system state: {}", ex.what());
+        } catch (const std::filesystem::filesystem_error & ex) {
+            logger->error("Cannot save system state: {}", ex.what());
         }
     }
 
--- a/libdnf5/system/state.cpp
+++ b/libdnf5/system/state.cpp
@@ -603,6 +603,7 @@ void State::reset_packages_states(
         save();
     } catch (const FileSystemError & e) {
         // TODO(mblaha) - log this? (will need access to the base)
+    } catch (const std::filesystem::filesystem_error & e) {
     }
 }
 


j-mracek:
Somehow I have a bad feeling about the proposed patch. 

1. `dnf5 makecache` should not try to create `/usr/lib/sysimage/libdnf5`. 
2. When the system state is reading the state it should pass silently when `/usr/lib/sysimage/libdnf5` and toml files are not present, but I believe it should not pass when cannot read the file due to missing permissions (this is tricky).

jan-kolarik:
> I believe that the problem is in test `system_state.packages_import_required()` (base.cpp:204). It detect empty state and starts to convert data and then save the state. I believe that the condition should change and additional test whether is it possible to create directory. If not than it should return `false`.

It makes sense. My main point was to highlight the problem here by basically reverting the state in the sources to the previous version when it was still working.

kontura:
As we have discussed I have created an alternative PR https://github.com/rpm-software-management/dnf5/pull/1097.

I think checking the permissions is not perfect as they might change before the actual write but I did include it in the PR.

Issue #1084, opened 2023-12-11
builddep cannot handle python3dist(build[virtualenv])


xsuchy:
Builddep command has a problem with BuildRequires containing square brackets:
```
$ sudo dnf5 builddep ~/rpmbuild/SRPMS/reproducer-brackets-0-0.src.rpm
...
Failed to resolve the transaction:
No match for argument: python3dist(build[virtualenv])
```
This breaks Python dependency generators and is a regression from DNF.

reproducer-brackets.spec:
```
Name:           reproducer-brackets
Version:        0
Release:        0
Summary:        ...
License:        MIT
BuildArch:      noarch

# Requirements like this are generated by %%pyproject_buildrequires
# here we add it manually for simplicity of this reproducer
BuildRequires:  python3dist(build[virtualenv])

%description
...
```

Original report against Mock https://github.com/rpm-software-management/mock/issues/1267

hroncok:
There are 80 such BuildRequires in Rawhide from ~200 packages:

```
$ repoquery -q --repo=rawhide-source --requires -a | grep -F '[' | wc -l
80
$ repoquery -q --repo=rawhide-source --whatrequires '*\[*' | wc -l
203
```

m-blaha:
Ouch - looks like we incorrectly handle the requirement as a glob...
Thanks for filing the issue, it's definitely valid.

PR #1085, opened 2023-12-12
builddep: Don't try to expand globs in pkg specs

evan-goode:
This commit makes a breaking change adding `expand_globs` setting to the `ResolveSpecSetting`.

Related: https://github.com/rpm-software-management/mock/issues/1267
Resolves https://github.com/rpm-software-management/dnf5/issues/1084

Are there any other places where the `expand_globs` setting should be `false`? Or maybe `false` should be the default and the commands that accept pkg specs on the command line should set it to `true` themselves?

WIP since a ci-dnf-stack PR is still on the way.

### DIFF
--- a/dnf5-plugins/builddep_plugin/builddep.cpp
+++ b/dnf5-plugins/builddep_plugin/builddep.cpp
@@ -219,18 +219,6 @@ bool BuildDepCommand::add_from_pkg(
     }
 }
 
-std::string escape_glob(const std::string & in) {
-    // Escape fnmatch glob characters in a string
-    std::string out;
-    for (const auto ch : in) {
-        if (ch == '*' || ch == '?' || ch == '[' || ch == ']' || ch == '\\') {
-            out += '\\';
-        }
-        out += ch;
-    }
-    return out;
-}
-
 void BuildDepCommand::run() {
     // get build dependencies from various inputs
     std::set<std::string> install_specs{};
@@ -281,7 +269,7 @@ void BuildDepCommand::run() {
     // Don't expand globs in pkg specs. The special characters in a pkg spec
     // such as the brackets in `python3dist(build[virtualenv])`, should be
     // treated as literal.
-    settings.expand_globs = false;
+    settings.set_expand_globs(false);
 
     for (const auto & spec : install_specs) {
         if (libdnf5::rpm::Reldep::is_rich_dependency(spec)) {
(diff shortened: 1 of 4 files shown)


m-blaha:
We definitely need this feature.
But since this is an ABI-breaking change that we're not ready to deliver quickly, I was thinking about working around the problem on the builddep side. Would escaping `[` (and potentially other globs) before calling `goal->add_rpm_install()` help?

praiskup:
> Would escaping [ (and potentially other globs) before calling goal->add_rpm_install() help?

+1

There just seem to be [3 of them](https://github.com/rpm-software-management/dnf5/blob/7ea44d9271ecdf391f1bb0f4661f73b48fdaa6d8/include/libdnf5/utils/patterns.hpp#L29-L31).

evan-goode:
> We definitely need this feature. But since this is an ABI-breaking change that we're not ready to deliver quickly, I was thinking about working around the problem on the builddep side. Would escaping `[` (and potentially other globs) before calling `goal->add_rpm_install()` help?

Yes, that seems to work. I opened a new PR that does it that way: https://github.com/rpm-software-management/dnf5/pull/1088.

ci-dnf-stack PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1425

jan-kolarik:
> > We definitely need this feature. But since this is an ABI-breaking change that we're not ready to deliver quickly, I was thinking about working around the problem on the builddep side. Would escaping `[` (and potentially other globs) before calling `goal->add_rpm_install()` help?
> 
> Yes, that seems to work. I opened a new PR that does it that way: #1088.
> 
> ci-dnf-stack PR: [rpm-software-management/ci-dnf-stack#1425](https://github.com/rpm-software-management/ci-dnf-stack/pull/1425)

So is this PR still needed? Either we should close it or target it against the `dnf-5.2.0.0` branch.

evan-goode:
> So is this PR still needed? Either we should close it or target it against the `dnf-5.2.0.0` branch.

Thanks for the reminder, I think it's still worth doing. Retargeted and rebased against dnf5-5.2.0.0, and got rid of the temporary fix from https://github.com/rpm-software-management/dnf5/pull/1088.

m-blaha:
The PR seems to be unexpectedly big and contains unrelated commits. Maybe caused by 5.2 re-targetting?

kontura:
@evan-goode please rebase the PR.

evan-goode:
> @evan-goode please rebase the PR.

Done

PR #1088, opened 2023-12-12
builddep: Escape glob characters in pkg specs

evan-goode:
Not parsing the pkg specs as globs would require an ABI-breaking change (https://github.com/rpm-software-management/dnf5/pull/1085). A workaround that doesn't involve breaking changes is to escape the glob characters in the pkg specs.

Related: https://github.com/rpm-software-management/mock/issues/1267
Resolves https://github.com/rpm-software-management/dnf5/issues/1084

### DIFF
--- a/dnf5-plugins/builddep_plugin/builddep.cpp
+++ b/dnf5-plugins/builddep_plugin/builddep.cpp
@@ -217,6 +217,18 @@ bool BuildDepCommand::add_from_pkg(
     }
 }
 
+std::string escape_glob(const std::string & in) {
+    // Escape fnmatch glob characters in a string
+    std::string out;
+    for (const auto ch : in) {
+        if (ch == '*' || ch == '?' || ch == '[' || ch == ']' || ch == '\\') {
+            out += '\\';
+        }
+        out += ch;
+    }
+    return out;
+}
+
 void BuildDepCommand::run() {
     // get build dependencies from various inputs
     std::set<std::string> install_specs{};
@@ -272,7 +284,13 @@ void BuildDepCommand::run() {
             // we do not download filelists and some files could be explicitly mentioned in provide section. The best
             // solution would be to merge result of provide and file search to prevent problems caused by modification
             // during distro lifecycle.
-            goal->add_rpm_install(spec, settings);
+
+            // TODO(egoode) once we have a setting to disable expanding globs
+            // in resolve_pkg_spec, escaping the glob characters will no longer
+            // be needed:
+            // https://github.com/rpm-software-management/dnf5/pull/1085
+            const auto & escaped_spec = escape_glob(spec);
+            goal->add_rpm_install(escaped_spec, settings);
         }
     }
 


dnf5-plugins/builddep_plugin/builddep.cpp
```diff
@@ -217,6 +217,16 @@ bool BuildDepCommand::add_from_pkg(
     }
 }
 
+std::string escape_glob(const std::string & in) {
+    // Escape fnmatch glob characters in a string
+    std::string out = std::regex_replace(in, std::regex("\\\\"), "\\\\");
+    out = std::regex_replace(out, std::regex("\\["), "\\[");
+    out = std::regex_replace(out, std::regex("\\]"), "\\]");
+    out = std::regex_replace(out, std::regex("\\*"), "\\*");
+    out = std::regex_replace(out, std::regex("\\?"), "\\?");
```
praiskup, line 226: Such strings are going to stay in the "glob pattern" category because of the [simple check](https://github.com/rpm-software-management/dnf5/blob/7ea44d9271ecdf391f1bb0f4661f73b48fdaa6d8/include/libdnf5/utils/patterns.hpp#L29-L31).  Is that OK for `is_glob_pattern` callers?

praiskup: I assume yes... just that DNF5/libdnf goes through more expensive code paths...  but long-term, #1085 looks like semantically more correct.  Perhaps "the" issue is about to be fixed in two phases?

evan-goode: Yes, the code path is a bit more expensive, but anyway I think the escaped strings _need_ to be handled as globs or else the escaping backslashes (`\[`, etc) will be treated as literal and not as escaping the following glob character.

#1085 is definitely the better fix in the long term, but I think both patches are small and tidy enough that we can do this for now, and possibly revert it in favor of #1085 later.

dnf5-plugins/builddep_plugin/builddep.cpp
```diff
@@ -272,7 +282,8 @@ void BuildDepCommand::run() {
             // we do not download filelists and some files could be explicitly mentioned in provide section. The best
             // solution would be to merge result of provide and file search to prevent problems caused by modification
             // during distro lifecycle.
-            goal->add_rpm_install(spec, settings);
+            const auto & escaped_spec = escape_glob(spec);
```
m-blaha, line 285: Would you mind adding a TODO to remove the escaping once we have a setting to switch off the globs expanding? Maybe also add a link to the issue.

dnf5-plugins/builddep_plugin/builddep.cpp
m-blaha: Regular expressions seem to be a bit of overkill here. How about doing the escaping in a single loop?
```
  std::string out;
  for (const auto ch : str) {
    if (ch == '*' || ch == '?' || ch == '[' || ch == ']') {
      out += '\\';
    }
    out += ch;
  }
  return out;
```
But I'm not blocking the merge on it if you prefer regex.

dnf5-plugins/builddep_plugin/builddep.cpp
evan-goode: Done!

dnf5-plugins/builddep_plugin/builddep.cpp
evan-goode: Done, that's saner and I think still plenty readable.

m-blaha (approved):
Thank you!

PR #1089, opened 2023-12-12
bindings: Fix and enable builds with Go bindings

### LINKED ISSUE #719: interface with Go programming language
The README states:

> Note, however, that DNF5 cannot yet interface with the following programming languages:
>
> - Go - does not work, we are looking for contributors.

I would potentially like to work on this. Can someone explain on a high level what steps were required to add language bindings for other languages?

hellozee:
Fixes #719 

Ok, disclaimer, I am neither a SWIG expert nor a dnf one but I was just poking around the codebase.

With that being clear, its a rough take on exposing the API. 
 - Disables a lot of the callbacks relying on `std::unique_ptr`, which I don't know if important or not. 
 - Changed the API a bit on the libdnf/conf/vars, there is one more copy down the line but we avoid exposing unique_ptr to the API
 - SWIG/Go doesn't like `constexpr const char*` which is a bit annoying but can be worked around
 - the only thing I couldn't figure out was the Config::opt_bind, SWIG thinks it returns a pointer and not a reference

Definitely need some help.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ option(WITH_DNF5DAEMON_TESTS "Build dnf5daemon tests (requires configured dbus)"
 option(WITH_SANITIZERS "Build with address, leak and undefined sanitizers (DEBUG ONLY)" OFF)
 
 # build options - bindings
-option(WITH_GO "Build Go bindings" OFF)
+option(WITH_GO "Build Go bindings" ON)
 option(WITH_PERL5 "Build Perl 5 bindings" ON)
 option(WITH_PYTHON3 "Build Python 3 bindings" ON)
 option(WITH_RUBY "Build Ruby bindings" ON)
--- a/include/libdnf5/conf/config.hpp
+++ b/include/libdnf5/conf/config.hpp
@@ -33,8 +33,9 @@ namespace libdnf5 {
 /// Base class for configurations objects
 class Config {
 public:
+#ifndef SWIGGO
     OptionBinds & opt_binds() noexcept { return binds; }
-
+#endif
     virtual ~Config() = default;
 
     virtual void load_from_parser(
--- a/include/libdnf5/conf/vars.hpp
+++ b/include/libdnf5/conf/vars.hpp
@@ -101,7 +101,7 @@ public:
     /// @param name Name of the variable
     const Variable & get(const std::string & name) const { return variables.at(name); }
 
-    static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
+    static std::string detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
 
 private:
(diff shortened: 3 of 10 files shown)


CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ option(WITH_DNF5DAEMON_TESTS "Build dnf5daemon tests (requires configured dbus)"
 option(WITH_SANITIZERS "Build with address, leak and undefined sanitizers (DEBUG ONLY)" OFF)
 
 # build options - bindings
-option(WITH_GO "Build Go bindings" OFF)
+option(WITH_GO "Build Go bindings" ON)
```
kontura, line 53: I think we probably don't want to turn it on by default just yet.

I have spent some time trying it out and I don't think the bindings actually work.
Were you able to use them?

I would like to set up some tests the way we have for Python or Perl before we enabled them.

include/libdnf5/conf/vars.hpp
```diff
@@ -101,7 +101,7 @@ struct Vars {
     /// @param name Name of the variable
     const Variable & get(const std::string & name) const { return variables.at(name); }
 
-    static std::unique_ptr<std::string> detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
+    static std::string detect_release(const BaseWeakPtr & base, const std::string & install_root_path);
```
kontura, line 104: We cannot do that, it would break API.
Either we could target this PR to the `dnf5-5.2.0.0` branch (which is a new major release with breaking changes) or we have to find another way. Perhaps we could simply ignore the function in go bindings for now.

include/libdnf5/conf/config.hpp
```diff
@@ -33,8 +33,9 @@ namespace libdnf5 {
 /// Base class for configurations objects
 class Config {
 public:
+#ifndef SWIGGO
```
kontura, line 36: I would rather keep these`SWIGGO` conditionals only in the interface files.
I think it could look something like:
```
#if defined(SWIGGO)
%ignore libdnf5::Config::opt_binds;
#endif
```

include/libdnf5/repo/repo_cache.hpp
```diff
@@ -53,8 +53,13 @@ class RepoCache {
 public:
     using RemoveStatistics = RepoCacheRemoveStatistics;
 
-    /// The name of the attribute used to mark the cache as expired.
+/// The name of the attribute used to mark the cache as expired.
+#ifndef SWIGGO
```
kontura, line 57: Here as well, please keep the `SWIGGO` conditions only in the `*.i` files.

kontura (changes requested):

hellozee:
@kontura thanks for the review, but this PR just went out of my head and I forgot about it, sorry for that. I will close it, since I can't recall what was I doing in the first place.

PR #1092, opened 2023-12-13
Improve Python binding of Reldep and PAckage Classes

j-mracek:
I believe we should directly set Python slot using SWIG feature (see bellow), but I was unable to make it working.

`%feature("python:slot", "tp_str", functype="reprfunc") libdnf5::rpm::Reldep::to_string;`

`set_property(SOURCE rpm.i PROPERTY SWIG_FLAGS "-builtin"`

### DIFF
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -154,6 +154,16 @@ del ClassName##__iter__
 #endif
 %enddef
 
+%define add_hash(ClassName)
+#if defined(SWIGPYTHON)
+%extend ClassName {
+    int __hash__() const {
+        return $self->get_hash();
+    }
+}
+#endif
+%enddef
+
 %{
     #include "libdnf5/common/sack/query.hpp"
     #include "libdnf5/common/sack/query_cmp.hpp"
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -144,6 +144,16 @@ del ClassName##__iter__
 #endif
 %enddef
 
+%define add_repr(ClassName)
+#if defined(SWIGPYTHON)
+%extend ClassName {
+    std::string __repr__() const {
+        return $self->to_string_description();
+    }
+}
+#endif
+%enddef
+
 %{
     #include "libdnf5/common/sack/query.hpp"
     #include "libdnf5/common/sack/query_cmp.hpp"
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -134,6 +134,15 @@ del ClassName##__iter__
 #endif
 %enddef
 
+%define add_str(ClassName)
+#if defined(SWIGPYTHON)
+%extend ClassName {
+    std::string __str__() const {
+        return $self->to_string();
+    }
+}
+#endif
+%enddef
 
 %{
     #include "libdnf5/common/sack/query.hpp"
--- a/include/libdnf5/rpm/package.hpp
+++ b/include/libdnf5/rpm/package.hpp
@@ -515,6 +515,15 @@ public:
     /// @since 5.0.5
     libdnf5::BaseWeakPtr get_base() const;
 
+    /// Return NEVRA -> 0 epoch is not shown in string
+    std::string to_string() const { return get_nevra(); };
+
+    /// Provide descriptive information about instance including NEVRA and ID
+    std::string to_string_description() const;
+
+    /// Return unique ID representing Package
+    int get_hash() const { return get_id().id; };
+
 protected:
     // @replaces libdnf:libdnf/dnf-package.h:function:dnf_package_new(DnfSack *sack, Id id)
     Package(const BaseWeakPtr & base, PackageId id);
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -64,6 +64,7 @@
 
 add_str(libdnf5::rpm::Reldep)
 add_repr(libdnf5::rpm::Reldep)
+add_hash(libdnf5::rpm::Reldep)
 %include "libdnf5/rpm/reldep.hpp"
 
 %rename(next) libdnf5::rpm::ReldepListIterator::operator++();
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -71,6 +71,10 @@ add_hash(libdnf5::rpm::Reldep)
 %rename(value) libdnf5::rpm::ReldepListIterator::operator*();
 %include "libdnf5/rpm/reldep_list_iterator.hpp"
 %include "libdnf5/rpm/reldep_list.hpp"
+
+add_str(libdnf5::rpm::Package)
+add_repr(libdnf5::rpm::Package)
+add_hash(libdnf5::rpm::Package)
 %include "libdnf5/rpm/package.hpp"
 
 %template(VectorPackage) std::vector<libdnf5::rpm::Package>;
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -29,6 +29,7 @@
     }
 }
 
+
 %{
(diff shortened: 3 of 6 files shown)


include/libdnf5/rpm/package.hpp
```diff
@@ -515,6 +515,15 @@ class Package {
     /// @since 5.0.5
     libdnf5::BaseWeakPtr get_base() const;
 
+    /// Return full NEVRA
+    std::string to_string() const { return get_full_nevra(); };
```
gotmax23, line 519: `hawkey.Package`'s `__str__()` returned the equivalent of `get_nevra()`. Can we keep that behavior here?

```suggestion
    std::string to_string() const { return get_nevra(); };
```

include/libdnf5/rpm/package.hpp
```diff
@@ -487,7 +487,7 @@ class Package {
     libdnf5::repo::RepoWeakPtr get_repo() const;
 
     /// @return Id of the repository the package belongs to.
-    /// @since 5.0
+    /// @since 5.0Reldep
```
gotmax23, line 490:
```suggestion
    /// @since 5.0
```

include/libdnf5/rpm/package.hpp
j-mracek: May I ask you why you would prefer the DNF4 behavior? DNF4 already uses for repoquery full NEVRA.

With optional epoch there is a hidden problem when string is parsed again by dnf and then used for query.

Package `docker-0:1.1.x86_64 ` is referenced by str() as `docker-1.1.x86_64`. Then when you filter_nevra() then it find not only original package but also `docker-1:1.1.x86_64 `.

include/libdnf5/rpm/package.hpp
j-mracek: Thank you for catching it.

include/libdnf5/rpm/package.hpp
gotmax23: I'm not really sure what would be the usecase for passing the string representation of a Package to `filter_nevra()` if you already have the Package object. I find the optional epoch form easier to read and understand. Also, fedrq's libdnf5 backend shadows the `libdnf5.rpm.Package` with a subclass that provides a `__str__()` that returns `.get_nevra()`[<sup>1</sup>](https://git.sr.ht/~gotmax23/fedrq/tree/main/item/src/fedrq/backends/libdnf5/backend.py#L718-719), so I'd prefer not to have different semantics between the two.

jan-kolarik: If I understand correctly, `filter_nevra()` accepts both nevra and full nevra forms. So, if the user explicitly passes the string representation, they can decide what to use. Currently, I don't see a strong reason for returning the full nevra, and it might be better to maintain compatibility if possible.

j-mracek: The problem is that missing epoch in NEVRA is handled by `filter_nevra()` as any epoch. It means that 

`query.filter_nevra("pkg-0:1-1.noarch")` and `query.filter_nevra("pkg-1-1.noarch")` might provide a different result when `pkg-0:1-1.noarch` and `pkg-1:1-1.noarch` is in set.

j-mracek: There is a recent use-case - https://bugzilla.redhat.com/show_bug.cgi?id=2254341#c3.

Reporter uses multiple sacks and want to perform query between them. The only way how to do it is to use a string equivalent of the package object when you want to query packages from a different sack.

jan-kolarik: One can still use `query.filter_nevra(pkg.get_full_nevra())` to get desired results, right? But I don't have a strong opinion about this. Not sure how frequently there are more packages with the same nvr, but different epoch in the repository...

bindings/libdnf5/common.i
```diff
@@ -134,6 +134,35 @@ del ClassName##__iter__
 #endif
 %enddef
 
+%define add_str(ClassName)
```
jan-kolarik, line 137: I also didn't figure out how to make this working using the `%feature("python:slot" ...`, so let's keep it as it is. I consider this implementation simple enough.

include/libdnf5/rpm/package.hpp
j-mracek: I believe that In Python you will call `query.filter_nevra(str(pkg))` by default. The problem that I described is rare but not impossible (I remember one case with `docker` package in copr repository that have a different EPOCH than the package in Fedora, but the version and release was identical. In most of cases we will even not get report because it is hard to detect and even worse to debug.

ppisar: I find `filter_nevra()` to ignore the epoch (if not explicitly passed) very confusing.
I find `to_string()` to explicitly output the 0 epoch obnoxious.

If `filter_nevra()` cannot be fixed, I would create a new function respecting the implicit 0 epoch, I would deprecate `filter_nevra()`, and I would keep to_string() not to format 0 epoch.

gotmax23: I agree with @jan-kolarik and @ppisar. `__str__()` is supposed to be a friendly, human representation; the optional epoch form is easier to understand and also keeps the previous behavior. If API users are using two different package sacks at once (which seems like an anti-pattern to me), they can pass `package.get_full_nevra()` instead of `str(package)`.

j-mracek: My apology. I checked the code of `filter_nevra(string)` and it requires `get_nevra()` string type. The problem that I mentioned is possible to reproduce with filter_nevra(Nevra class) where nevra string is parsed by Nevra class. This way is used when spec is parsed. Anyway I will modify the PR according to your suggestion.

include/libdnf5/rpm/package.hpp
```diff
@@ -515,6 +515,15 @@ class Package {
     /// @since 5.0.5
     libdnf5::BaseWeakPtr get_base() const;
 
+    /// Return NEVRA -> 0 epoch is not shown in string
+    std::string to_string() const { return get_nevra(); };
+
+    /// Provide descriptive information about instance including full NEVRA and ID
```
jan-kolarik, line 521: > including full NEVRA
This is not true anymore, right?

include/libdnf5/rpm/package.hpp
j-mracek: Interesting things about epoch

RPM does not show 0 epoch for `rpm -q B`
RPM does show 0 epoch  `rpm -q B --qf "%{name}-%{evr}\n" when `Epoch: 0` in in spec.
`rpmbuild` does not add epoch to rpm file name.
DNF/DNF5 and command repoquery always shows 0 epoch.
DNF/DNF5 transaction table does not show 0 epoch.
DNF transaction summary does not show 0 epoch.
DNF history info does not show 0 epoch for transaction packages.
DNF5 history info shows 0 epoch for transaction packages.

Be honest it is mess and I would love to somehow unify it.

include/libdnf5/rpm/package.hpp
j-mracek: Fixed

include/libdnf5/rpm/package.hpp
jan-kolarik: @j-mracek That would be great, but rather out-of-scope of this PR. Could you create a follow-up issue please?

j-mracek: Not really. The PR sets another difference. It means unifying after merge of this PR will be more painful. Anyway I will create the issue.

j-mracek: Issue - https://github.com/rpm-software-management/dnf5/issues/1175

jan-kolarik: > Not really. The PR sets another difference. It means unifying after merge of this PR will be more painful. Anyway I will create the issue.

It also depends which side we will choose to unify with. Thanks for creating the issue.

PR #1094, opened 2023-12-13
modules: Change module dependency string to be the same as in dnf4

pkratoch:
Add ":[]" when any stream is required instead of keeping only the module name. This makes it consistent with `modulemd_stream` yaml and with dnf4 output of the module info command.


kontura:
Tests pass locally.
LGTM.

PR #1096, opened 2023-12-13
Fix `utils::fs::File` class and use it for loading compressed modules metadata

jrohel:


kontura:
LGTM.
Tests pass locally.
Thank you.

PR #1101, opened 2023-12-14

j-mracek:
This is required for Python binding, because when string argument is provided, binding convert it to vector of characters.


j-mracek:
Closing if favor of https://github.com/rpm-software-management/dnf5/pull/1102

Issue #1091, opened 2023-12-13
Release dnf5-5.2.0.0


kontura:
Since we need to do API and ABI changes we will have to make a new major version release update.

This issue should track all the required/planned changes:
 - [x] https://github.com/rpm-software-management/dnf5/issues/1025
 - [x] https://github.com/rpm-software-management/dnf5/pull/1085
 - [x] https://github.com/rpm-software-management/dnf5/pull/1087
 - [x] https://github.com/rpm-software-management/dnf5/pull/1216
 - [x] https://github.com/rpm-software-management/dnf5/issues/1189
 - [x] https://github.com/rpm-software-management/dnf5/issues/1072
 - [x] Drop deprecated API
   - [x] https://github.com/rpm-software-management/dnf5/pull/1294
   - [x] https://github.com/rpm-software-management/dnf5/issues/1322

Follow-ups:
- [x] https://github.com/rpm-software-management/dnf5/pull/1421

voxik:
"Dnf5-6.0.0" sorry, I can't help, but this won't stop fascinating me.

j-mracek:
I am proposing to add `use_glob{true}` to `ResolveSpecSettings`. Such a change will brake ABI.

Conan-Kudo:
> "Dnf5-6.0.0" sorry, I can't help, but this won't stop fascinating me.

We should have just renamed the old project to "dnf4" and kept calling this "dnf" but here we are. :)

kontura:
We are also considering changing the versioning to add one more number and keep the 5.
So the new breaking version would be `dnf5-5.2.0.0`.

The nevras still do compare correctly: `dnf5-5.2.0.0` > `dnf5-5.1.0`.

jan-kolarik:
OK, I believe we can proceed with merging this to main.

Conan-Kudo:
Do you think we have sufficient API coverage for me to implement a PackageKit backend using libdnf5 based on what we have for the current dnf backend? https://github.com/PackageKit/PackageKit/tree/main/backends/dnf

jan-kolarik:
> Do you think we have sufficient API coverage for me to implement a PackageKit backend using libdnf5 based on what we have for the current dnf backend? https://github.com/PackageKit/PackageKit/tree/main/backends/dnf

[...]

PR #1102, opened 2023-12-14

j-mracek:
This is required for Python binding, because when string argument is provided, binding convert it to vector of characters. The change might be problematic because filter_provides({"dnf"}) stop to work.

### DIFF
--- a/include/libdnf5/rpm/package_query.hpp
+++ b/include/libdnf5/rpm/package_query.hpp
@@ -776,13 +776,24 @@ public:
         const std::vector<std::string> & patterns, libdnf5::sack::QueryCmp cmp_type = libdnf5::sack::QueryCmp::EQ);
 
     /// Filter packages by their `location`.
+    /// @param pattern          A string the filter is matched against.
+    /// @param cmp_type         A comparison (match) operator, defaults to `QueryCmp::EQ`.
+    ///                         Supported values: `EQ`, `NEQ`.
+    /// @since 5.2
+    //
+    // TODO(dmach): enable glob match to enable filename matching: {nevra.rpm, */nevra.rpm}
+    // @replaces libdnf/sack/query.hpp:method:addFilter(int keyname, int cmp_type, const char *match) - cmp_type = HY_PKG_LOCATION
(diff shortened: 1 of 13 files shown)


include/libdnf5/rpm/package_query.hpp
```diff
@@ -281,6 +281,18 @@ class PackageQuery : public PackageSet {
     /// @since 5.0
     //
     // @replaces libdnf/sack/query.hpp:method:addFilter(int keyname, int cmp_type, const char *match) - cmp_type = HY_PKG_PROVIDES
+    void filter_provides(const std::string & pattern, libdnf5::sack::QueryCmp cmp_type = libdnf5::sack::QueryCmp::EQ) {
```
kontura, line 284: If we go with this version the docs need an update, it says `A vector of strings ..` but it takes just a string.

j-mracek:
@jan-kolarik This PR does not change API,  but some project might got an issue with compilation. What do you suggest with this PR?

jan-kolarik:
> @jan-kolarik This PR does not change API, but some project might got an issue with compilation. What do you suggest with this PR?

Hmm, could you please specify the exact issue with compiling the project after these changes are merged?

jan-kolarik:
> The problem is described in commit message: The change might be problematic because filter_arch({"noarch"}) and filter_arch({"src", "nosrc"}) stop to work. It is a reason why commit [8492018](https://github.com/rpm-software-management/dnf5/commit/8492018084fcafe3d1c18f64d0eb1927667d6b2b) also modify additional lines (e.g. `query.filter_provides({"prv-all"});`). The additional overload method cause that compiler is unable to pick the correct methods because both std::string and std::vectorstd::string accept `char *`.

Thanks, I see now. Well, I feel that all changes that could result in breaking the existing client code should be delivered with the major version bump. Therefore, I'd like to bundle this together with other changes in https://github.com/rpm-software-management/dnf5/issues/1091. Is it OK like that?

jan-kolarik:
@j-mracek Could you please target the PR against the https://github.com/rpm-software-management/dnf5/tree/dnf5-5.2.0.0 branch?

j-mracek:
Opened - https://github.com/rpm-software-management/dnf5/pull/1216

PR #1105, opened 2023-12-14
Improve RTD docs build process

gotmax23:
- doc: call Python build script as a pre_build job 
- doc rtd_build: use proper main() function


doc/rtd_build.py
```diff
@@ -0,0 +1,75 @@
...
+        DIR + "/../bindings/libdnf5"), os.path.join(DIR + "/../bindings/python3/libdnf5"))
+
+    # libdnf5-cli
+    generate_bindings_from_dir(os.path.join(DIR + "/../bindings/libdnf5_cli"),
+                               os.path.join(DIR + "/../bindings/python3/libdnf5_cli"))
+
+
+if __name__ == "__main__":
+    if os.environ.get("READTHEDOCS") != "True":
+        print("This setup.py is supposed to be used only to build documentation in Read the Docs.")
```
kontura, line 72: Can you please amend this to say something like: `This script is supposed to be used only...`

kontura (changes requested):

doc/rtd_build.py
gotmax23: @kontura, done!

kontura: Thank you!

kontura (approved):

Issue #332, opened 2023-03-05
how to add "version lock"  with dnf5 ...


Dutchglory:
how do i  add a "version lock"  with dnf5  ...  or add a version-lock plugin to dnf5

```
Package "epson-inkjet-printer-escpr-1.7.25-1lsb3.2.x86_64" is already installed with reason "User".

[verwijs@fedora ~]$ dnf5 upgrade
Updating and loading repositories:
Repositories loaded.
Package                                      Arch         Version                        Repository       Size
Upgrading:                                                                                                                                    
 epson-inkjet-printer-escpr                 x86_64       1:1.7.21-3.1lsb3.2.fc38          fedora        5.3 MiB
 replacing epson-inkjet-printer-escpr       x86_64       1.7.25-1lsb3.2                @commandline     5.4 MiB

Transaction Summary:
 Upgrading:         1 packages
 Replacing:         1 packages

```

j-mracek:
Version lock was implemented in DNF as a plugin using rpm excludes.

The question is whether we should integrate it as a core functionality of libdnf5 and whether we cold use libsolv locking solvable.

KraXen72:
Hello, there is a bug with the libdnf/dnf5 versionlock plugin:
versionlocked packages are still updated through packagekit (e.g. Gnome Software or KDE Discover)

relevant links:
https://bugs.kde.org/show_bug.cgi?id=473607 -> told to file in packagekit
https://github.com/PackageKit/PackageKit/issues/677 -> told to file in libdnf
https://github.com/rpm-software-management/libdnf/issues/1627 -> told that it's not going to be implemented in dnf4 since dnf5 is soon to be released.

dnf5 is now released and this bug still persists.

Please let me know if i should open a new issue and if so, where, since on all the different bug trackers i checked, everyone is just saying to report this issue somewhere else..

Thanks!

PR #1121, opened 2024-01-02
Versionlock functionality for dnf5

m-blaha:
Replacement of dnf4 versionlock plugin.
The config file format changed from ambiguous plain text with globs to TOML format.

PR contains also commands for config file manipulation (dnf5 versionlock add/exclude/delete/clear). More sophisticated manipulations (e.g. "lock to major version 5", "lock to epoch 1") are currently not supported from commandline - the user needs to edit the config file manually to achieve this.

- [x] Setup `since` fields to 5.1.13 for new API methods before the merging

Adjustments of current CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1434
Closes: https://github.com/rpm-software-management/dnf5/issues/332.

### DIFF
--- a/dnf5/commands/list/list.cpp
+++ b/dnf5/commands/list/list.cpp
@@ -145,8 +145,8 @@ void ListCommand::run() {
     auto & ctx = get_context();
     auto & config = ctx.base.get_config();
 
-    libdnf5::rpm::PackageQuery full_package_query(ctx.base);
-    libdnf5::rpm::PackageQuery base_query(ctx.base);
+    libdnf5::rpm::PackageQuery full_package_query(ctx.base, libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK);
+    libdnf5::rpm::PackageQuery base_query(ctx.base, libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK);
 
     // pre-select by patterns
     if (!pkg_specs.empty()) {
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -553,6 +553,8 @@ void RepoqueryCommand::run() {
     libdnf5::sack::ExcludeFlags flags = disable_modular_filtering->get_value()
                                             ? libdnf5::sack::ExcludeFlags::IGNORE_MODULAR_EXCLUDES
                                             : libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES;
+    // XXX handle --upgrades
+    flags = flags | libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK;
     libdnf5::rpm::PackageQuery base_query(ctx.base, flags, false);
     libdnf5::rpm::PackageQuery result_query(ctx.base, flags, true);
 
--- a/dnf5/commands/repoquery/repoquery.cpp
+++ b/dnf5/commands/repoquery/repoquery.cpp
@@ -45,7 +45,7 @@ libdnf5::rpm::PackageQuery repeat_filter(
     // Create source query of all considered packages.
     // To match dnf4 take arch filter into account.
     // (filtering by repo and available/installed is done implicitly by loading only the required metadata)
-    libdnf5::rpm::PackageQuery all_considered(candidates.get_base());
+    libdnf5::rpm::PackageQuery all_considered(candidates.get_base(), libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK);
     if (!arches.empty()) {
         all_considered.filter_arch(arches, libdnf5::sack::QueryCmp::GLOB);
     }
@@ -553,8 +553,9 @@ void RepoqueryCommand::run() {
     libdnf5::sack::ExcludeFlags flags = disable_modular_filtering->get_value()
                                             ? libdnf5::sack::ExcludeFlags::IGNORE_MODULAR_EXCLUDES
                                             : libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES;
-    // XXX handle --upgrades
-    flags = flags | libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK;
+    if (!upgrades->get_value()) {
+        flags = flags | libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK;
+    }
     libdnf5::rpm::PackageQuery base_query(ctx.base, flags, false);
     libdnf5::rpm::PackageQuery result_query(ctx.base, flags, true);
 
--- a/dnf5/commands/versionlock/versionlock_add.cpp
+++ b/dnf5/commands/versionlock/versionlock_add.cpp
@@ -90,10 +90,10 @@ void VersionlockAddCommand::run() {
     auto & ctx = get_context();
     auto package_sack = ctx.base.get_rpm_package_sack();
     auto vl_config = package_sack->get_versionlock_config();
+    auto orig_size = vl_config.get_packages().size();
 
     const auto comment = format_comment("add");
 
-    bool changed{false};
     const libdnf5::ResolveSpecSettings settings{
         .with_nevra = true, .with_provides = false, .with_filenames = false, .with_binaries = false};
     for (const auto & spec : pkg_specs) {
@@ -118,7 +118,6 @@ void VersionlockAddCommand::run() {
             }
             versions.emplace(evr);
             if (lock_version(vl_config, pkg, comment)) {
-                changed = true;
                 std::cout << libdnf5::utils::sformat(_("Adding versionlock on \"{0} = {1}\"."), pkg.get_name(), evr)
                           << std::endl;
             } else {
@@ -126,7 +125,7 @@ void VersionlockAddCommand::run() {
             }
         }
     }
-    if (changed) {
+    if (vl_config.get_packages().size() != orig_size) {
         vl_config.save();
     }
 }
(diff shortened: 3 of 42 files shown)


dnf5/commands/list/list.cpp
```diff
@@ -145,8 +145,8 @@ void ListCommand::run() {
     auto & ctx = get_context();
     auto & config = ctx.base.get_config();
 
-    libdnf5::rpm::PackageQuery full_package_query(ctx.base);
-    libdnf5::rpm::PackageQuery base_query(ctx.base);
+    libdnf5::rpm::PackageQuery full_package_query(ctx.base, libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK);
```
jan-kolarik, line 148: Probably needed also for advisory list, info? And what about plugins like changelog, needs-restarting? Or maybe, if I understand it correctly, we should make sure all existing package queries related to non-transactional operations have the flag present?

include/libdnf5/rpm/versionlock_config.hpp
```diff
@@ -0,0 +1,144 @@
...
+    /// Get the value.
+    std::string get_value() const { return value; }
+
+    /// Get the key as a string.
+    std::string get_key_str() const { return key_str; }
+    /// Get the comparison operator as a string.
+    std::string get_comparator_str() const { return comparator_str; }
+
+    /// Get list of errors found during parsing the entry from configuration file.
+    std::vector<std::string> get_errors() const { return errors; }
```
jan-kolarik, line 60: Or `const &` ?

include/libdnf5/rpm/versionlock_config.hpp
```diff
@@ -0,0 +1,144 @@
...
+    /// Get the package name.
+    std::string get_name() const { return name; }
+
+    /// Get the comment for this entry.
+    std::string get_comment() const { return comment; }
+    /// Set comment for this entry.
+    void set_comment(std::string_view comment);
+
+    /// Get the list of conditions configured for the package.
+    std::vector<VersionlockCondition> get_conditions() const { return conditions; }
```
jan-kolarik, line 100: Same as before, maybe `const &` would be better? Also for the `get_errors()`.

include/libdnf5/rpm/versionlock_config.hpp
```diff
@@ -0,0 +1,144 @@
...
+    std::vector<VersionlockCondition> conditions{};
+    std::vector<std::string> errors{};
+};
+
+
+/// Class contains parsed versionlock configuration file.
+class VersionlockConfig {
+public:
+    /// Get list of configured versionlock entries.
+    std::vector<VersionlockPackage> & get_packages() { return packages; }
```
jan-kolarik, line 125: `const std::vector<VersionlockPackage> &`?

dnf5/commands/versionlock/versionlock_add.cpp
```diff
@@ -0,0 +1,134 @@
...
+    auto package_sack = ctx.base.get_rpm_package_sack();
+    auto vl_config = package_sack->get_versionlock_config();
+
+    const auto comment = format_comment("add");
+
+    bool changed{false};
+    const libdnf5::ResolveSpecSettings settings{
+        .with_nevra = true, .with_provides = false, .with_filenames = false, .with_binaries = false};
+    for (const auto & spec : pkg_specs) {
+        libdnf5::rpm::PackageQuery query(ctx.base);
```
jan-kolarik, line 100: Shouldn't we NOT exclude versionlocks for versionlock commands? :)

jan-kolarik (commented):
I did an initial review and plan to revisit it, likely tomorrow, to perform testing on the commands.

include/libdnf5/rpm/package_sack.hpp
```diff
@@ -28,6 +28,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/base/base_weak.hpp"
 #include "libdnf5/common/exception.hpp"
 #include "libdnf5/common/weak_ptr.hpp"
+#include "libdnf5/rpm/versionlock_config.hpp"
```
jan-kolarik, line 31: Or just `#include "versionlock_config.hpp"` on L27 as others from the same folder.

jan-kolarik:
I'd consider adding the docs for manual editing of the toml config file or track it as the follow-up issue. [This commit message](https://github.com/rpm-software-management/dnf5/pull/1121/commits/994e95431b6d92060b895869be30c84364440673) is a good start.

jan-kolarik:
I've needed to patch the `rpm.i` bindings file to compile with bindings:
[0001-swig-Add-new-versionlock-bindings.patch](https://github.com/rpm-software-management/dnf5/files/14097450/0001-swig-Add-new-versionlock-bindings.patch)

dnf5/commands/versionlock/versionlock_add.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+bool lock_version(
+    libdnf5::rpm::VersionlockConfig & vl_config, const libdnf5::rpm::Package & pkg, const std::string & comment) {
+    auto evr = pkg.get_evr();
+    auto name = pkg.get_name();
+    auto & vl_packages = vl_config.get_packages();
+    // check whether a rule for this version is already present
+    for (const auto & vl_package : vl_packages) {
+        if (!vl_package.is_valid() || vl_package.get_name() != name) {
```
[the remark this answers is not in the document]
m-blaha: It's definitely a possibility.
The reason why I introduced the `is_valid` method was to keep also invalid entries when the configuration is written back to the toml file (e.g. after adding a new entry using `dnf5 versionlock add xxx` command).
The invalid entries are most likely a result of manual versionlock.toml editing and I did not want their changes to vanish. Basically, I wanted to keep the door open for users to fix any mistakes they made.

dnf5/commands/list/list.cpp
m-blaha: Well not all - for example `check-upgrade` does not create a transaction but we still want to keep the results in sync with what `dnf upgrade` would do.
But generally you are right - I'll add the flag to missing places.

dnf5/commands/repoquery/repoquery.cpp
```diff
@@ -553,6 +553,8 @@ void RepoqueryCommand::run() {
     libdnf5::sack::ExcludeFlags flags = disable_modular_filtering->get_value()
                                             ? libdnf5::sack::ExcludeFlags::IGNORE_MODULAR_EXCLUDES
                                             : libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES;
+    // XXX handle --upgrades
```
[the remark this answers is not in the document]
m-blaha: Actually `--upgrades` filter needs to be handled - with this switch repoquery needs to apply versionlock excludes. Thanks for spotting!

include/libdnf5/rpm/package_sack.hpp
```diff
@@ -145,6 +145,24 @@ class PackageSack {
     /// @since 5.0
     void clear_user_includes();
 
+    /// Returns versionlock excluded package set
+    const PackageSet get_versionlock_excludes();
```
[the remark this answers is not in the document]
m-blaha: We should. But what should be the value? Given that it's not clear in which version the PR's gonna be released.

include/libdnf5/rpm/versionlock_config.hpp
m-blaha: Not that easily. Given that returned reference is used to directly manipulate the vector (add new items, delete some items, even clear the vector completely), it would require more API on VersionlockPackage class.

dnf5/commands/versionlock/versionlock_add.cpp
```diff
@@ -0,0 +1,134 @@
...
+    context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
+}
+
+bool lock_version(
+    libdnf5::rpm::VersionlockConfig & vl_config, const libdnf5::rpm::Package & pkg, const std::string & comment) {
+    auto evr = pkg.get_evr();
+    auto name = pkg.get_name();
+    auto & vl_packages = vl_config.get_packages();
+    // check whether a rule for this version is already present
+    for (const auto & vl_package : vl_packages) {
```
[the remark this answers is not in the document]
m-blaha: Indeed it's possible. I tried to rewrite the loops using `std::find_if`. The resulting code was a bit shorter, but I didn't find it more readable (especially due to nested lambdas for find_if predicates). I would prefer the current state.

m-blaha:
@jan-kolarik I've resolved most of you comments. To simplify the review process I added the changes as new commits. But I can squash them after the review if you want me to.

include/libdnf5/rpm/package_sack.hpp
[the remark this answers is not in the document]
jan-kolarik: I'd change it right before the merge, I'll update the description of the PR if you agree...

dnf5/commands/versionlock/versionlock_add.cpp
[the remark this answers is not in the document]
jan-kolarik: Sure, I just kinda wanted to think about and consider other options.

dnf5/commands/versionlock/versionlock_add.cpp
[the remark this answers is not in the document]
jan-kolarik: I see. Let's keep it then.

jan-kolarik (approved):

jan-kolarik:
LGTM, I believe if there is no other complaint or request, we can prepare the since tags to `5.1.13` and merge ...

m-blaha:
Cool, thanks. I'm gonna add a new commit with `Since 5.1.13` later today.

PR #1127, opened 2024-01-04
package_query: Fix filter_version with non EQ comparator

m-blaha:
The current implementation of the `filter_version` and `filter_release` filters
always uses the `cmp_eq` comparator. This means that these filters only work
with EQ and NEQ comparisons. The patch fixes these filters to work
correctly with LT, LTE, GT, and GTE comparisons as well.


j-mracek:
LGTM, Thank you for the patch

m-blaha:
Actually if you check testing farm results (e.g. here https://artifacts.dev.testing-farm.io/70790912-a7e2-41e2-8d9e-758da7d0ef10/) and switch on "show passed tests" on top of the page, you will see that only failing tests are dnf4 ones.
But I agree, it's PITA and we need to fix it ASAP.

j-mracek:
LGTM

Issue #1011, opened 2023-11-13
Additional features for config-manager


j-mracek:
Remaining things from - https://github.com/rpm-software-management/dnf5/issues/405

Requested in: https://github.com/rpm-software-management/dnf5/pull/907

- [ ] Add warning when request to unset config option but configuration file (dnf.conf) is not found
- [ ] Add warning when request to unset repository config option but repo override file is not found
- [ ] Add warning when request to unset repository config option but repository id is not found in repo override file
- [ ] Add warning when unset config option is not present dnf.conf
- [ ] Add warning when unset repository config option is not present repo override file
- [ ] Add warning when request to unset variable but vars direcory is not found
- [ ] Add warning when request to unset variable but variable file is not found
- [ ] Generate repository name by combining repoid and information that repository was created by config-manager

PR #1128, opened 2024-01-04
Additional features for config-manager

jrohel:
It solves tasks in the https://github.com/rpm-software-management/dnf5/issues/1011 issue except for documentation and does not replace crc32 with library functions.

Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1429 adjustment of CI tests. 

Closes: #1011 .

### DIFF
--- a/dnf5-plugins/config-manager_plugin/unsetopt.cpp
+++ b/dnf5-plugins/config-manager_plugin/unsetopt.cpp
@@ -32,10 +32,17 @@ using namespace libdnf5;
 
 namespace {
 
-bool remove_from_config(ConfigParser & parser, const std::string & section_id, const std::set<std::string> & keys) {
+bool remove_from_config(
+    ConfigParser & parser,
+    const std::string & section_id,
+    const std::set<std::string> & keys,
+    std::set<std::string> & used_keys) {
     bool removed = false;
     for (const auto & key : keys) {
-        removed |= parser.remove_option(section_id, key);
+        if (parser.remove_option(section_id, key)) {
+            removed = true;
+            used_keys.insert(key);
+        }
     }
     return removed;
 }
@@ -82,7 +89,7 @@ void ConfigManagerUnsetOptCommand::set_argument_parser() {
                         tmp_repo_conf.opt_binds().at(repo_key);
                     } catch (const OptionBindsOptionNotFoundError & ex) {
                         ctx.base.get_logger()->warning(
-                            "config-manager: Request to remove unknown repository option from config file: {}", key);
+                            "config-manager: Request to remove unsupported repository option: {}", key);
                     }
 
                     in_repos_opts_to_remove[repo_id].insert(repo_key);
@@ -92,7 +99,7 @@ void ConfigManagerUnsetOptCommand::set_argument_parser() {
                         tmp_config.opt_binds().at(key);
                     } catch (const OptionBindsOptionNotFoundError & ex) {
                         ctx.base.get_logger()->warning(
-                            "config-manager: Request to remove unknown main option from config file: {}", key);
+                            "config-manager: Request to remove unsupported main option: {}", key);
                     }
 
                     // Save the global option for later removing from the file.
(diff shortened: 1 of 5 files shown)


dnf5-plugins/config-manager_plugin/unsetopt.cpp
```diff
@@ -107,48 +114,88 @@ void ConfigManagerUnsetOptCommand::configure() {
...
+
+            parser.read(cfg_filepath);
+
+            changed |= remove_from_config(parser, "main", main_opts_to_remove, used_keys);
+
+            // Generate warning for unused options
+            for (const auto & key : main_opts_to_remove) {
+                if (!used_keys.contains(key)) {
+                    ctx.base.get_logger()->warning(
+                        "config-manager: Request to remove not present main option from config file: {}", key);
```
pkratoch, line 132: Just one small thing - this seems a bit hard to read. What about to unite it with the other warnings to e.g.: "Request to remove main option but it is not present in the config file:"

dnf5-plugins/config-manager_plugin/unsetopt.cpp
```diff
@@ -107,48 +114,88 @@ void ConfigManagerUnsetOptCommand::configure() {
...
+                if (const auto used_repoid_opts = used_repos_opts.find(in_repoid);
+                    used_repoid_opts == used_repos_opts.end()) {
+                    ctx.base.get_logger()->warning(
+                        "config-manager: Request to remove repository option but repoid not found in overrides: {}",
+                        in_repoid);
+                } else {
+                    for (const auto & key : keys) {
+                        if (!used_repoid_opts->second.contains(key))
+                            ctx.base.get_logger()->warning(
+                                "config-manager: Request to remove not present repository option from overrides: {}.{}",
```
pkratoch, line 179: The same as the above. I suggest: "Request to remove repository option but it is not present in the overrides:"

pkratoch:
I am also wondering if the messages should be only written to logs or if there also shouldn't be a warning on stderr.

dnf5-plugins/config-manager_plugin/unsetopt.cpp
```diff
@@ -107,48 +114,88 @@ void ConfigManagerUnsetOptCommand::configure() {
...
 
-    auto repos_override_file_path = get_config_manager_repos_override_file_path(config);
-
     // Remove options from repositories overrides configuration file, remove empty sections.
-    if (!in_repos_opts_to_remove.empty() && std::filesystem::exists(repos_override_file_path)) {
-        ConfigParser parser;
-        bool changed = false;
-        parser.read(repos_override_file_path);
+    if (!in_repos_opts_to_remove.empty()) {
+        const auto & repos_override_file_path = get_config_manager_repos_override_file_path(config);
```
pkratoch, line 147: This is not a problem of this PR, but I encountered this:

When I run `dnf5 config-manager unsetopt fedora.gpgcheck` (or any other option), there is no output, but nothing gets changed. Only when I finally looked at logs, I noticed there is `config-manager: Request to remove repository option but config file not found: /etc/dnf/repos.override.d/99-config_manager.repo`.

Shouldn't this file get created if it doesn't exist yet? Or am I missing something?

Moreover, this makes me think even more that the warnings should get also on the stderr, because it took me some time to figure out why nothing is happening.

jrohel:
I wanted to unify logging to the logger and to the terminal. It is not yet implemented. So I will modify this PR to directly print the warning to stderr.

jan-kolarik:
> I unlinked the issue #1011 , so that it's not closed upon merging this, because it wouldn't be fully resolved. We could either split the issue and link the newly created one or leave it as it is and then manually mark the individual items as resolved.

I relinked it again as I created separate issues for the remaining tasks and mentioned them in the original issue.

dnf5-plugins/config-manager_plugin/unsetopt.cpp
jrohel: @pkratoch 
>Shouldn't this file get created if it doesn't exist yet?

No. The file contains overrides of repository configuration. A non-existent file has the same meaning as an empty file - no overrides are defined.

pkratoch: Ah, right. It doesn't make sense to create the file for unsetting the options. I was confused only because I initially expected the original repo file to be modified and seemingly nothing happened when I run the `unsetopt` command. But now I see it only works over the overrides and the error message makes that clear.

rffontenelle:
Translation-related doubt: "repository" and "main" in the warnings below should be kept as in English (e.g. are they keywords in the code or command-line option), or they can be translated?

> config-manager: Request to remove repository option but repoid is not present in the overrides: {}

> config-manager: Request to remove unsupported main option: {}

PR #1135, opened 2024-01-05
Disable zchunk on RHEL

yselkowitz:
zchunk is not supported in RHEL, and other related packages also disable their zchunk dependency on RHEL.


j-mracek:
LGTM

PR #1149, opened 2024-01-08
Workaround for swig-4.2.0 missing fragment dependency

wsfulton:
Fixes error:
repoPYTHON_wrap.cxx:5656:56: error: expected template-name before '<' token
 5656 |     struct SwigPyMapIterator_T : SwigPyIteratorClosed_T<OutIterator, ValueType, FromOper>

due to missing SWIG %fragment dependency on SwigPyIterator_T which provides the missing base class SwigPyIteratorClosed_T.

See https://github.com/swig/swig/issues/2744

### DIFF
--- a/bindings/libdnf5/repo.i
+++ b/bindings/libdnf5/repo.i
@@ -15,6 +15,11 @@
 %import "common.i"
 %import "conf.i"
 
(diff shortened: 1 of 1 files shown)


kontura:
Thank you.

Issue #1048, opened 2023-11-27
Respect defaults when enabling multiple streams of a module


pkratoch:
When enabling a module with more than one stream and it has a default stream defined, dnf errors with "Cannot enable multiple streams", but it should enable the default stream instead.

PR #1152, opened 2024-01-09
modules: Respect defaults when enabling multiple streams of a module

pkratoch:
Resolves: https://github.com/rpm-software-management/dnf5/issues/1048


pkratoch:
Related CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1433

libdnf5/module/module_sack.cpp
```diff
@@ -777,14 +778,90 @@ bool ModuleSack::Impl::enable(const std::string & name, const std::string & stre
...
 
+    // For the given module_spec, create a dict { name : {stream : [solvable id] }
+    auto module_dict = create_module_dict(module_spec_to_query(base, module_spec));
+    // Keep only enabled or default streams if possible
+    auto multiple_stream_modules = prune_module_dict(module_dict);
+    // If there are any modules with multiple streams to be enabled, throw an error
+    if (!multiple_stream_modules.empty()) {
+        throw EnableMultipleStreamsError(
+            M_("Cannot enable multiple streams of one module at the same time. Affected modules: {}"),
+            utils::string::join(multiple_stream_modules, ", "));
```
j-mracek, line 849: I believe that in some aspects DNF4 provided a better message.
```
sudo dnf module enable '*' '*m*'
[sudo] password for jmracek: 
Last metadata expiration check: 2:04:24 ago on Wed 17 Jan 2024 12:17:10 PM CET.
Argument '*' matches 5 streams ('10.5', '10.6', '10.7', '10.8', '10.9') of module 'mariadb', but none of the streams are enabled or default
Unable to resolve argument *
Argument '*m*' matches 5 streams ('10.5', '10.6', '10.7', '10.8', '10.9') of module 'mariadb', but none of the streams are enabled or default
Unable to resolve argument *m*
Error: Problems in request:
broken groups or modules: *, *m*
```
Your code is better because it provides information about all modules with multiple stream for one spec.

```
sudo dnf5 module enable '*' '*m*'
Repositories loaded.
Cannot enable multiple streams of one module at the same time. Affected modules: avocado, avocado-vt, cri-o, mariadb, nextcloud, nginx, perl, perl-DBD-SQLite, perl-libwww-perl, postgresql, swig
```
The new code does not show information about available streams available streams. It means that user has to run additional command to list all streams for spec.

The critical problem - DNF5 does not show error message for all arguments but only for the first one.
Important problem - if I am not mistaken (I am guessing, following error catch path is quite demanding) - problem is not logged into event log. It means API users will have a difficulty to understand the issue. I am investigation it and I will add additional information in future.

libdnf5/module/module_sack_impl.hpp
```diff
@@ -129,11 +129,11 @@ class ModuleSack::Impl {
...
     /// @param module_spec module to be enabled.
     /// @param count if `true`, count the change towards the limit of module status modifications.
-    /// @return `true` if requested change really triggers a change in the ModuleDB, `false` otherwise.
-    /// @throw EnableMultipleStreamsError in case of conflicting enable requests.
+    /// @return `true` if requested change realy triggers a change in the ModuleDB, `false` otherwise,
+    ///         and a set of module:stream strings for modules with multiple streams and no enabled or default one.
     /// @throw NoModuleError if the module doesn't exist.
     /// @since 5.0.14
-    bool enable(const std::string & module_spec, bool count = true);
+    std::pair<bool, std::set<std::string>> enable(const std::string & module_spec, bool count = true);
```
j-mracek, line 136: Due to what add_resolve_log accepts, it is a good option. :+1:

libdnf5/base/goal.cpp
```diff
@@ -567,11 +567,27 @@ GoalProblem Goal::Impl::add_module_specs_to_goal(base::Transaction & transaction
...
+                    if (!enable_ret.second.empty()) {
+                        transaction.p_impl->add_resolve_log(
+                            action,
+                            GoalProblem::MULTIPLE_STREAMS,
+                            GoalJobSettings(),
+                            libdnf5::transaction::TransactionItemType::MODULE,
+                            spec,
+                            enable_ret.second,
+                            log_level);
+                        if (!skip_unavailable) {
```
j-mracek, line 585: Are you sure that you want to use `skip_unavailable` option for result when request is conflicting? What about `skip_broken` option.

j-mracek: I have to test how code handles both situation - speck does not match any module and `GoalProblem::MULTIPLE_STREAMS`

libdnf5/module/module_sack.cpp
```diff
@@ -777,18 +777,94 @@ bool ModuleSack::Impl::enable(const std::string & name, const std::string & stre
...
+
+
+std::pair<bool, std::set<std::string>> ModuleSack::Impl::enable(const std::string & module_spec, bool count) {
     module_db->initialize();
 
+    // For the given module_spec, create a dict { name : {stream : [solvable id] }
+    auto module_dict = create_module_dict(module_spec_to_query(base, module_spec));
+    // Keep only enabled or default streams if possible
+    auto multiple_stream_modules = prune_module_dict(module_dict);
+    // If there are any modules with multiple streams to be enabled, throw an error
```
j-mracek, line 848: I think that the commend is not valid anymore. May I ask you to update it?

libdnf5/base/goal.cpp
j-mracek: After testing - scenario with no match works correctly. Only what works not nicely is a case where GoalProblem::MULTIPLE_STREAMS is triggered and option --skip-unavailable ignores the issue.

j-mracek: This is a minor issue, not something significant, but if it will be not addressed in the PR, then issue should be created.

pkratoch: I added the `skip_broken` option.

j-mracek:
LGTM

PR #1155, opened 2024-01-10
Url option

### LINKED ISSUE #497: dnf download `--url`

derickdiaz:
Fixes #497

Changes get the urls for the package specified by the --url option and returns the first url found as discussed.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -64,6 +64,9 @@ void DownloadCommand::set_argument_parser() {
     alldeps_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
+    url_option = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+
     auto resolve = parser.add_new_named_arg("resolve");
     resolve->set_long_name("resolve");
     resolve->set_description("Resolve and download needed dependencies");
@@ -77,10 +80,17 @@ void DownloadCommand::set_argument_parser() {
     alldeps->set_const_value("true");
     alldeps->link_value(alldeps_option);
 
+    auto url = parser.add_new_named_arg("url");
+    url->set_long_name("url");
+    url->set_description("Print the list of urls where the rpms can be downloaded instead of downloading");
+    url->set_const_value("true");
+    url->link_value(url_option);
+
     cmd.register_named_arg(alldeps);
     create_destdir_option(*this);
     cmd.register_named_arg(resolve);
     cmd.register_positional_arg(keys);
+    cmd.register_named_arg(url);
 }
 
 void DownloadCommand::configure() {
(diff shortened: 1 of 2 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -146,20 +155,34 @@ void DownloadCommand::run() {
...
 
-        for (auto & [nevra, pkg] : download_pkgs) {
-            downloader.add(pkg);
+    if (url_option->get_value()) {
+        for (auto & [nerva, pkg] : download_pkgs) {
+            auto urls = pkg.get_remote_locations();
+            if (urls.empty()) {
+                std::cout << "Failed to get mirror for package: " << pkg.get_name() << std::endl;
+                continue;
+            }
```
kontura, line 168: I don't think this should go to `stdout`. Either just log it like old dnf:
```cpp
ctx.base.get_logger()->warning("Failed to get mirror for package: \"{}\"", pkg.get_name());
```

or maybe even better we could try making it strict and throw an exception. 
```cpp
libdnf_assert(urls.empty(), "Failed to get mirror for package: \"{}\"", pkg.get_name());
```
I think this basically shouldn't happen but if it turns out to be a problem we can change it later.

kontura (changes requested):
Thank you for working on this.

dnf5/commands/download/download.cpp
derickdiaz: Used the libdnf_assert option as suggested

dnf5/commands/download/download.cpp
```diff
@@ -146,20 +156,31 @@ void DownloadCommand::run() {
...
+    if (download_pkgs.empty()) {
+        return;
+    }
 
-        for (auto & [nevra, pkg] : download_pkgs) {
-            downloader.add(pkg);
+    if (url_option->get_value()) {
+        for (auto & [nerva, pkg] : download_pkgs) {
+            auto urls = pkg.get_remote_locations();
+            libdnf_assert(urls.empty(), "Failed to get mirror for package: \"{}\"", pkg.get_name());
```
kontura, line 166: Oh, the condition has to be reversed: `!urls.empty()`
This way it always fails.

kontura (changes requested):
Can you squash the `Changed urloption to url ` commit into the first one?
We prefer to have the history as clean as possible.

dnf5/commands/download/download.cpp
derickdiaz: No problem on the squash. I'm confused on the url.empty part. Wouldn't we want it to trigger an assert if the command when the vector is emptyl? Thank you

kontura (changes requested):
Thank you, it looks good.
One last thing can you remove the `Changed urloption to url` and `reversed assertion condition` from the commit message? It doesn't really make sense when the commits are squashed.

kontura (approved):
Thank you!

Issue #1159, opened 2024-01-11
Make dnf5daemon-client values placeholder instead of hardcoding them


rffontenelle:
I assume these values must not be translated (by translators, in Fedora Weblate). In this case, my suggestion is to make these values placeholders instead of hardcoding in the string, to avoid it getting translated by mistake. (I would propose a PR, but I'm not sure what is the proper syntax.)

https://github.com/rpm-software-management/dnf5/blob/91114403b68a7f45e5fc455d79bb0051f6a66cbf/dnf5daemon-client/commands/advisory/arguments.hpp#L154

PR #1161, opened 2024-01-12
I18N: Annotate literals in advisory command

ppisar:
The comment is extracted by "xgettext -c" and placed to gettext catalogs. It helps the translators not know which words not to translate:

    #. Note for translators: "critical" etc. quoted words are
    #. literals that should not be translated.
    #: commands/advisory/arguments.hpp:155
    msgid ""
    "Limit to packages in advisories with specified severity. List option. Can be "
    "\"critical\", \"important\", \"moderate\", \"low\", \"none\"."
    msgstr ""

Fixes: #1159

### DIFF
--- a/dnf5daemon-client/commands/advisory/arguments.hpp
+++ b/dnf5daemon-client/commands/advisory/arguments.hpp
@@ -150,6 +150,8 @@ public:
               command,
               "advisory-severities",
               '\0',
+              /* Note for translators: "critical" etc. quoted words are
+                 literals that should not be translated. */
               _("Limit to packages in advisories with specified severity. List option. Can be "
                 "\"critical\", \"important\", \"moderate\", \"low\", \"none\"."),
               _("ADVISORY_SEVERITY,..."),


rffontenelle:
How about making them placeholders `{}` instead?

j-mracek:
LGTM

ppisar:
Removing the literals from the message and substituting them dynamically would be bad because:

* The message would have to be built at run-time. A static string as it is now is much faster and saves a memory.
* The data structure the message is stored now does not support run-time computations. The text must be ready when compiling the code. Using a formating string would require changing the code.
* The translators could not explain a meaning of the literals in their translations, like:

~~~~
msgid ""
"Limit to packages in advisories with specified severity. List option. Can be "
"\"critical\", \"important\", \"moderate\", \"low\", \"none\"."
msgstr ""
"Limit to packages in advisories with specified severity. List option. Can be "
"\"critical\" [very serious], \"important\" [serious], \"moderate\" [normal] , \"low\" [unimportant], \"none\" [undefined]."
~~~

Here in English to English translation it does not highlight the problem boldly, but I believe you know what I mean.

dnf5daemon-client/commands/advisory/arguments.hpp
```diff
@@ -150,6 +150,8 @@ class AdvisorySeverityOption : public libdnf5::cli::session::AppendStringListOpt
               command,
               "advisory-severities",
               '\0',
+              /* Note for translators: "critical" etc. quoted words are
+                 literals that should not be translated. */
               _("Limit to packages in advisories with specified severity. List option. Can be "
```
j-mracek, line 155: I am curios ehrther we should use `C_()` to add a context for translators?

ppisar: Translation contexts are for distinguishing two identical messages for a need of different translations. This is not the case. This message only exists at two places dnf5/commands/advisory_shared.hpp and dnf5daemon-client/commands/advisory/arguments.hpp. The messages has the same meaning and use (their code and method name is actually a duplicate), so no artificial context is needed here.

Issue #1158, opened 2024-01-11
dnf5-5.1.11 FTBFS in openSUSE due to msgmerge failure


Conan-Kudo:
I tried to build dnf5-5.1.11, but it fails due to a msgmerge failure:

```
[   24s] [  1%] Generating sv.gmo
[   24s] cd /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build/libdnf5-cli/po && /usr/bin/msgmerge --quiet --update --backup=none -s /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build/libdnf5-cli/po/sv.po /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/libdnf5-cli/po/libdnf5-cli.pot
[   24s] /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build/libdnf5-cli/po/sv.po: warning: Charset "CHARSET" is not a portable encoding name.
[   24s]                                                                              Message conversion to user's charset might not work.
[   24s] /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build/libdnf5-cli/po/sv.po:124: duplicate message definition...
[   24s] /home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build/libdnf5-cli/po/sv.po:3: ...this is the location of the first definition
[   24s] /usr/bin/msgmerge: found 1 fatal error
[   24s] make[2]: *** [libdnf5-cli/po/CMakeFiles/translations_2.dir/build.make:84: libdnf5-cli/po/sv.gmo] Error 1
[   24s] make[2]: Leaving directory '/home/abuild/rpmbuild/BUILD/dnf5-5.1.11/build'
[   24s] make[1]: *** [CMakeFiles/Makefile2:1608: libdnf5-cli/po/CMakeFiles/translations_2.dir/all] Error 2
```

Conan-Kudo:
Full build log: [dnf5-5.1.11-openSUSE_Factory-build.log.txt](https://github.com/rpm-software-management/dnf5/files/13907517/dnf5-5.1.11-openSUSE_Factory-build.log.txt)

Sources: https://build.opensuse.org/package/show/system:packagemanager:dnf/dnf5

ppisar:
I confirm that libdnf5-cli/po/sv.po file is invalid since:

~~~~
commit 22be22bdf2296446db4ee59a5146f3188136d17d
Author: GitHub Workflow <github-actions@github.com>
[...]

PR #1162, opened 2024-01-12
I18N: Do not mark empty strings for a translation

ppisar:
An empty string has a special meaning in gettext and marking such string produces invalid catalogs with a duplicate message ID entry.

Fixes: #1158

### DIFF
--- a/include/libdnf5/utils/bgettext/bgettext-mark-domain.h
+++ b/include/libdnf5/utils/bgettext/bgettext-mark-domain.h
@@ -26,6 +26,10 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "bgettext-mark-common.h"
 
+// clang-format off
+// Use EMPTY_MESSAGE instead of M_(""). Otherwise, xgettext breaks catalogs.
+#define EMPTY_MESSAGE \
+    { .bgettextMsg = "\004" GETTEXT_DOMAIN "\00" "" }
 #define M_(msgId) \
     { .bgettextMsg = "\004" GETTEXT_DOMAIN "\00" msgId }
 #define MP_(msgId, msgIdPlural) \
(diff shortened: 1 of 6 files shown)


ppisar:
CI fails on the duplicate msgids in translations. I will also fix the catalogs in this pull request.

ppisar:
The removal of M_() macro use actually breaks types:

~~~~
/builddir/build/BUILD/dnf5-5.1.11/libdnf5-cli/exception.cpp: In constructor ‘libdnf5::cli::SilentCommandExitError::SilentCommandExitError(int)’:
/builddir/build/BUILD/dnf5-5.1.11/libdnf5-cli/exception.cpp:49:95: error: no matching function for call to ‘libdnf5::cli::Error::Error(const char [1])’
   49 | SilentCommandExitError::SilentCommandExitError(int exit_code) : Error(""), exit_code(exit_code) {}
~~~~

ppisar:
I will try to use a new macro for  `{ .bgettextMsg = "\004" GETTEXT_DOMAIN "\00" "" }`  instead of `M_("")`. But I wonder whether then Error() does not print a translation for an msgid "", i.e metadata about a translation catalog.

ppisar:
The tests starts looking good. I will put a commit on top to suppress clang-format at two places.

include/libdnf5/utils/bgettext/bgettext-mark-domain.h
```diff
@@ -26,6 +26,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "bgettext-mark-common.h"
 
+// Use EMPTY_MESSAGE instead of M_(""). Otherwise, xgettext breaks catalogs.
+#define EMPTY_MESSAGE \
+    { .bgettextMsg = "\004" GETTEXT_DOMAIN "\00" "" }
```
m-blaha, line 31: Can you please fix the formatting?  See https://github.com/rpm-software-management/dnf5/actions/runs/7503905055/job/20429713582?pr=1162

ppisar: Do you mean suppress clang-format, or format all the macros according to the clang-format's taste? I've just pushed a change which does the former.

m-blaha: Thanks, I also find suppressing clang format better here.

m-blaha (approved):
ccache builds still fail for some reason, but copr builds are OK. Thank you.

rffontenelle:
Thanks for the fix, but the existing translations in the translation repository didn't receive the fix. Weblate still fails to renew the translations files because `msgmerge` is still unable to update due to the duplicate strings present in there. https://github.com/rpm-software-management/dnf5-l10n/pull/2 should fix this for good.

Issue #1165, opened 2024-01-12
The 'dnf5 group install' doesn't support --allowerasing


praiskup:
This is just related to:
https://github.com/rpm-software-management/mock/pull/1286

PR #1174, opened 2024-01-15
group: Fix using allowerasing option

jan-kolarik:
Allow passing `--allowerasing` option in group subcommands.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1165.


m-blaha (approved):
Thank you!

PR #1178, opened 2024-01-15
Deduplicate strings for translations

rffontenelle:
Make similar strings equals by replacing argument names with placeholders `{}`. This benefits translators that will see less strings for translation, but the major advantage is to assure consistency across similar translated strings.


ppisar (approved):
The changes are good. The CI failures are unrelated.

Issue #1025, opened 2023-11-20
Improve ABI stability


j-mracek:
Potential problem

### Adding new items to a structure or class
- using pimpl resolves the issue

in `include/libdnf5/base/goal_elements.hpp`
- `struct ResolveSpecSettings`
- `ResolveSpecSettings`

kontura:
I went through `include/libdnf5` and I tried to identify classes and structs that in my opinion would be a good candidates for pImpl. Please let me know if I have missed something and what is your opinion.

I have removed enum classes, error classes and classes that already have pImpl.

## Add pImpl:
Classes:
```cpp
base/log_event.hpp:36:class LogEvent {
base/solver_problems.hpp:33:class SolverProblems {
repo/repo.hpp:63:class Repo {
repo/repo_sack.hpp:38:class RepoSack : public sack::Sack<Repo> {
rpm/rpm_signature.hpp:48:class KeyInfo {
```
Structs:
```cpp
base/goal_elements.hpp:120:struct ResolveSpecSettings {
base/goal_elements.hpp:200:struct GoalJobSettings : public ResolveSpecSettings {
conf/vars.hpp:46:struct Vars {
conf/config_parser.hpp:85:struct ConfigParser {
```

## No sure?:
```cpp
module/module_item.hpp:56:class ModuleItem {
base/transaction_environment.hpp:34:class TransactionEnvironment {
base/transaction_module.hpp:36:class TransactionModule {
base/transaction_group.hpp:38:class TransactionGroup {
base/transaction_package.hpp:39:class TransactionPackage {
transaction/comps_environment.hpp:43:class CompsEnvironment : public TransactionItem {
transaction/comps_environment.hpp:120:class CompsEnvironmentGroup {
transaction/comps_group.hpp:43:class CompsGroup : public TransactionItem {
transaction/comps_group.hpp:123:class CompsGroupPackage {
transaction/rpm_package.hpp:40:class Package : public TransactionItem {
transaction/transaction_history.hpp:35:class TransactionHistory {
transaction/transaction.hpp:79:class Transaction {
transaction/transaction_item.hpp:44:class TransactionItem {
```

## Leave without pImpl:
Classes:
```
advisory/advisory_query.hpp:39:class AdvisoryQuery : public AdvisorySet {
advisory/advisory_collection.hpp:34:class AdvisoryCollection {
advisory/advisory_reference.hpp:31:class AdvisoryReference {
[...]

PR #1180, opened 2024-01-16
Add pImpl to `libdnf5::repo::repo`

kontura:

### DIFF
--- a/include/libdnf5/repo/repo.hpp
+++ b/include/libdnf5/repo/repo.hpp
@@ -236,14 +236,10 @@ public:
     // @replaces libdnf:repo/Repo.hpp:method:Repo.getExpiresIn()
     int get_expires_in() const;
 
-    /// Gets repository freshness
-    // @replaces libdnf:repo/Repo.hpp:method:Repo.fresh()
-    bool fresh();
-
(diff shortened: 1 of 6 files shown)


include/libdnf5/repo/repo.hpp
```diff
@@ -387,21 +383,17 @@ class Repo {
     /// @return Whether at least the repodata cache cloning was successful.
     bool clone_root_metadata();
 
-    libdnf5::BaseWeakPtr base;
-    ConfigRepo config;
+    RepoDownloader & get_downloader() const;
```
jan-kolarik, line 386: What about moving everything from the private part into the pimpl?

jan-kolarik: I mean, if we have already decided that a class is a candidate for potential future changes, it's worth mitigating any risks by keeping the exposed set of methods and members in public header files to a minimum.

jan-kolarik: After the discussion, private members here are not used in any inline function, so it should be safe to potentially remove them in future.

jan-kolarik (approved):

PR #1183, opened 2024-01-18
Add pImpl to several classes

kontura:

Specifically for `RpmSignature` I am proposing to convert it into a namespace rather than adding a `pImpl` to it, it is described in more details in the commit message.

### DIFF
--- a/include/libdnf5/base/log_event.hpp
+++ b/include/libdnf5/base/log_event.hpp
@@ -44,27 +44,29 @@ public:
         const libdnf5::transaction::TransactionItemType spec_type,
         const std::string & spec);
     LogEvent(libdnf5::GoalProblem problem, const SolverProblems & solver_problems);
-    ~LogEvent() = default;
+    ~LogEvent();
+
+    LogEvent(const LogEvent & src);
+    LogEvent & operator=(const LogEvent & src);
+
+    LogEvent(LogEvent && src) noexcept;
+    LogEvent & operator=(LogEvent && src) noexcept;
 
     /// @return GoalAction for which goal event was created
-    libdnf5::GoalAction get_action() const { return action; };
+    libdnf5::GoalAction get_action() const;
     /// @return GoalProblem that specify the type of report
-    libdnf5::GoalProblem get_problem() const { return problem; };
+    libdnf5::GoalProblem get_problem() const;
     /// @return Additional information (internal), that are required for formatted string
-    const std::set<std::string> get_additional_data() const { return additional_data; };
+    const std::set<std::string> & get_additional_data() const;
     /// @return GoalJobSetting if it is relevant for the particular GoalProblem
-    const libdnf5::GoalJobSettings * get_job_settings() const {
-        return job_settings ? &job_settings.value() : nullptr;
-    };
+    const libdnf5::GoalJobSettings * get_job_settings() const;
     /// @return SPEC if it is relevant for the particular GoalProblem
-    const std::string * get_spec() const { return spec ? &spec.value() : nullptr; };
+    const std::string * get_spec() const;
     /// @return SolverProblems if they are relevant for the particular GoalProblem
-    const SolverProblems * get_solver_problems() const { return solver_problems ? &solver_problems.value() : nullptr; };
+    const SolverProblems * get_solver_problems() const;
 
     /// Convert an element from resolve log to string;
-    std::string to_string() const {
-        return to_string(action, problem, additional_data, job_settings, spec_type, spec, solver_problems);
-    };
+    std::string to_string() const;
 
 private:
     /// Convert an element from resolve log to string;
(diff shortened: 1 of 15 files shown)


include/libdnf5/base/log_event.hpp
```diff
@@ -41,30 +41,32 @@ class LogEvent {
         libdnf5::GoalProblem problem,
         const std::set<std::string> & additional_data,
         const libdnf5::GoalJobSettings & settings,
-        const libdnf5::transaction::TransactionItemType spec_type,
+        libdnf5::transaction::TransactionItemType spec_type,
```
jan-kolarik, line 44: Why there is a different signature than in cpp? Meaning the `const` is missing here. Is it on purpose?

include/libdnf5/base/log_event.hpp
```diff
@@ -41,30 +41,32 @@ class LogEvent {
...
 
     /// @return GoalAction for which goal event was created
-    libdnf5::GoalAction get_action() const { return action; };
+    libdnf5::GoalAction get_action() const;
     /// @return GoalProblem that specify the type of report
-    libdnf5::GoalProblem get_problem() const { return problem; };
+    libdnf5::GoalProblem get_problem() const;
     /// @return Additional information (internal), that are required for formatted string
-    const std::set<std::string> get_additional_data() const { return additional_data; };
+    std::set<std::string> get_additional_data() const;
```
jan-kolarik, line 60: Is the removing of `const` needed or wanted?

include/libdnf5/rpm/rpm_signature.hpp
```diff
@@ -81,41 +81,33 @@ class KeyInfo {
...
+/// Check public key presence in rpm database
+bool key_present(Base & base, const KeyInfo & key);
+
+/// Download the key file if needed and return a vector of keys contained in it
+/// @param key_url: URL of the public key to be imported.
+std::vector<KeyInfo> parse_key_file(Base & base, const std::string & key_url);
+
+/// Return string representation of the CheckResult enum
+std::string check_result_to_string(CheckResult result);
+};  // namespace rpm_signature
```
jan-kolarik, line 110: I am not sure about this. I'd rather leave the class there and maybe rename it to `RpmSignatureVerifier`. I agree pimpl is not needed and currently changes are not expected for this class, right?

include/libdnf5/base/log_event.hpp
kontura: It is only wanted.
That is clang-tidy complained [readability-const-return-type](https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html).

include/libdnf5/base/log_event.hpp
kontura: I also got this from clang-tidy: [readability-avoid-const-params-in-decls](https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html)

The way I understand it is that since the `spec_type` is passed by value marking it `const` in the declaration has no effect, the passed value is only copied.
On the other hand in the definition it ensures that the copied valued is not modified.

I would be fine leaving both cases unmodified, reverting my changes, if you prefer.

include/libdnf5/rpm/rpm_signature.hpp
kontura: Actually this might be changing in future.
`rpm` provides ugly API for checking signatures: https://github.com/rpm-software-management/dnf5/blob/8f46d64fb1a44b307bae5fed7ce5ba233ea0ab0e/libdnf/rpm/rpm_signature.cpp#L107-L143

And there is an `RFE` for something better: https://github.com/rpm-software-management/rpm/issues/2041

Perhaps it would behoove us to add the `pImpl` to be ready for anything.

include/libdnf5/base/log_event.hpp
jan-kolarik: I don't have a problem with either way, just I'd be for having the hpp and cpp in sync.

include/libdnf5/rpm/rpm_signature.hpp
jan-kolarik: OK makes sense then. So class + pimpl then?

include/libdnf5/base/log_event.hpp
jan-kolarik: Oh I maybe see now the reason. It is returned by value, maybe we can change it to `const &` ?

include/libdnf5/base/log_event.hpp
kontura: Ok, I have added it back to the hpp.

include/libdnf5/base/log_event.hpp
kontura: I like that, updated.

kontura:
Since we are changing the API anyway I have also changed
```cpp
RpmSignature::check_package_signature(rpm::Package pkg)
```
to
```cpp
RpmSignature::check_package_signature(const rpm::Package & pkg)
````

jan-kolarik (approved):

jan-kolarik:
@kontura Please rebase, it's needed after merging other pimpl PRs.

kontura:
> @kontura Please rebase, it's needed after merging the other pimpl PRs.

Rebased.

PR #1190, opened 2024-01-21
Urlprotocol option

### LINKED ISSUE #950: Implement dnf download `--urlprotocol`
Requires https://github.com/rpm-software-management/dnf5/issues/497

derickdiaz:
Fixes #950

- Implements --urlprotocols to the download command and limits the valid package urls to the set url protocols.
- If no protocols are specified, all protocols are considered valid.

This PR is based on my url PR #1155


derickdiaz:
Note: Ignore first line "Values: url protocol..." <- this was debugging.

Screenshot 2024-01-22 at 10 08 33 PM
Current state is able to filter out by url protocol. If an invalid protocol is found, the following error will occur:
image

Issue #750, opened 2023-07-20
Add descriptive output for sticky vendor options


inknos:
See:
https://github.com/rpm-software-management/dnf5/pull/745#issuecomment-1643740753
https://github.com/rpm-software-management/dnf5/issues/738#issuecomment-1643795151

Issue #326, opened 2023-02-28
reducing duplicate information in upgrade output


carlwgeorge:
I noticed in the output of `dnf5 upgrade` there is quite a bit of duplicated information as compared to the output of `dnf upgrade`.

* example dnf output

```
=======================================================================================================
 Package                  Architecture         Version                     Repository             Size
=======================================================================================================
Upgrading:
 filesystem               x86_64               3.18-4.fc39                 rawhide               1.1 M
```

* example dnf5 output

```
Package                           Arch      Version                          Repository            Size
Upgrading:                                                                                             
 filesystem                       x86_64    3.18-4.fc39                      rawhide          106.0   B
   replacing filesystem           x86_64    3.18-3.fc38                      rawhide          106.0   B
```

Having two lines for every package with (usually) duplicated information is a lot of visual noise.  For updates with many packages it makes it difficult to parse the given information quickly.  While I understand that each of these fields can technically have a different value, most of this information is usually the same and not that useful to always show.

My first instinct is can we make the default dnf5 output more terse, closer to the dnf output, with just information about the new package?  The more detailed output (the "replacing" line) could be available with the verbose flag.  For obsoletes we could always show the "replacing" line, which is similar to what dnf does today.

That said, I do like the idea of always seeing both old and new versions in the output, but that is the only field I feel that way about.  Perhaps we could change the output to be something like this:

```
Package                        Arch       Old Version     New Version        Repository            Size
Upgrading:                                                                                          
 filesystem                    x86_64     3.18-3.fc38     3.18-4.fc39        rawhide          106.0   B
```

This will show the most important changing information (the version) while leaving out the information that is usually the same.  What do you think?

Conan-Kudo:
Does that fit in an 80x24 window? I think that's the major constraint we have with this stuff. Otherwise, yeah, looks good to me.

carlwgeorge:
The constraint of 24 rows doesn't make any sense to me, as the output will routinely go over that when there are multiple packages in the transaction.

At a constraint of 80 characters wide, even the current output will wrap with long package names.  Example:

```
Package                      Arch   Version                 Repository      Size
Installing:
 NetworkManager-config-connectivity-fedora
                             noarch 1:1.42.0-1.fc39         rawhide    310.0   B
```

[...]

PR #1191, opened 2024-01-21
single line transaction list

long-walk:
The transaction list should easy to review. 
Patch Benefits:
- remove replacing line for its package (sub package still listed)
- highlight arch, repo and size to see changes faster
Screenshot_20240121_225751


ppisar:
I like your new format more than the current. But  the old multi-line format was there for some reason. I think @j-mracek should be able to explain the reason.

long-walk:
I know there are lines like this brown and their still present maybe something i not spotted.
image
This time with Standard colors title and major changes are normally white, Sorry I had a modified color theme in my terminal. Maybe we should change this in dnf5 to blue?

long-walk:
It is not possible to put all this output to one line in this small terminals so dual line still need to be support. The actual dnf5 seems also to struggle with the output on so small terminals. I did some testing and this looks mostly good expect some edge cases if the package name or version are to long then they could be truncated at the end.
- [ ]  Adjust output to fit 80 character terminal. I add a limit of 130 char to change between dual line and single line
- [ ]  Color: use lightblue for major/minor changed packages
- [ ]  The size highlighting is adjusted, smaller package need factor 10 change bigger package need 10% to be  highlighted

* comparison previous dnf5 (fc40) with PR 
Screenshot_dnf5-80-compare1
* 129 character
Screenshot_dnf5-129
* 130 character
Screenshot_dnf5-130

long-walk:
Testing: https://copr.fedorainfracloud.org/coprs/tho6677/dnf5_single_line_upgrade/

long-walk:
Hello @j-mracek
Thanks for your replay, abovee is my proposal can you give me a feedback for this.

Conan-Kudo:
Where is `<unknown>` coming from in the repository column?

long-walk:
<unknown>  is based from my testing in a virtual system normally you will see the new source.

j-mracek:
> Where is `<unknown>` coming from in the repository column?

There has been a problem with transfer information from DNF4 but is is fixed now - https://github.com/rpm-software-management/dnf5/pull/1403

m-blaha:
@j-mracek we should decide on this. Recently I've been working on adding skipped packages into transaction table and I'm afraid that PR I plan to submit (currently draft, it needs some testing https://github.com/rpm-software-management/dnf5/pull/1440) will cause bigger conflicts with this one.

Conan-Kudo:
The main thing I think it missing is highlighting vendor changes, which is the main detail missing for #750.

jan-kolarik:
Related issue: https://github.com/rpm-software-management/dnf5/issues/326.

long-walk:
@m-blaha this is the merged version whit the skipped package is this the output you would expect?
 
dnf5_1191

carlwgeorge:
The right-to-left format of the version column is odd to me.  It's rather confusing for the version change to be formatted in that direction.  Left-to-right would make much more sense and match the rest of the output.

I'm not sure there is much value in having old and new values in the repository column.  I think that one should just show the repository that the update is coming from.  In most cases, that is going to be the same name written twice, i.e. `updates -> updates` or `appstream -> appstream`.  Similarly, I don't think it's worth the space taken to show old and new values for the size column.

m-blaha:
> @m-blaha this is the merged version with the #1440 is this the output you would expect?

The first column looks too wide for me - like it contained the section titles. How does it look with 80 cols terminal?
This is a transaction with broken dependency package on narrow terminal using the latest upstream dnf5:

Screenshot_20240516_073556

jan-kolarik:
I tested this a bit, and here are my thoughts:

- The 80-column output is better than the current state in the main branch.
- If there's a hard-coded limit for the version column, it might look odd in some cases. We should check if both versions fit, and if not, place them on separate lines (see image below).
- When a package replaces something with a different name, the "replacing" line should be indented (see also image below).
- I am also in favor of changing to a left-to-right format.

In general, we should decide if there will also be a verbose version of the transaction table (e.g., when -v is passed). In that case, we can include all the info there. Otherwise, if there will only be one version of the output, I'd omit the repo and size changes.

_Overflowed version with COPR packages_
image

_Non-indented replacing lines_
Screenshot from 2024-05-16 13-01-42

long-walk:
- The intention to do the right to left was that the new Version that will installed is the first and line up with the other versions from skipped packages, dependencies packages and Replacing packages. Maybe it could also be the arrow that you have a problem with, would bracket change your mind?
image
With Breakages

- The first column has a dynamic size if possible with a bit of priority to have a small gap after it.

- The size of the column are variable in length and this is done by libsmartcols library. I warp it manually but their is also the option to do it automatic but that can result in multiple lines (Screenshot above) in small windows you have all information but you lose clarity.
With 80 cols it looks not good for me if I warp the text automatic. If I truncate the Version it looks better but you lose some information. The Package name will never warped or truncated because that locks strange. Example:
image
image
Maybe I truncate the data below 130 cols and warp it above.  

- The non-intended rplacing line must be may fault during the merging and is fixed.  

-   The dnf5 Release version shows all the sizes and repository so I tried only to remove the duplicate information for me the -v verbost options sounds good but at the moment I couldn’t figure out how this works I need some more code reading.

Test commands I use:
dnf5 upgrade --releasever=38
dnf5 upgrade --releasever=41

carlwgeorge:
> Maybe it could also be the arrow that you have a problem with, would bracket change your mind?

No, brackets don't change my mind on this.  I think most people will view it in a linear way, left to right, old to new.  Showing the new version first is going to lead to confusion.  If the goal is to vertically align new versions, then I think the best approach is to have an old and new version column, using n/a as needed.  This is what I originally proposed in #326.

When we are restricted to an 80 col terminal, I think the easiest fix is to just suppress the architecture and size fields.  Those are much less important than everything else.

Conan-Kudo:
I think in constrained rendering, we should just drop the old version info or maybe move it to a new line saying it replaces the old version. We also still need some indicator of vendor changes somehow, but that could probably be plugged into the same rendering logic for obsoletes.

jan-kolarik:
This is a complex topic, and not everyone will be satisfied. Our team discussed this and agreed to make minimal changes to the current default transaction table layout. For now, the primary concern mentioned repeatedly is fixing the "replacing" lines for identical package names. Additionally, we need to highlight vendor changes, as Neal mentioned.

For a more verbose version, it would be better to enable it explicitly via a CLI or configuration option, rather than automatically based on window width. This approach will also simplify the fixes in our CI tests stack, where we parse the transaction table output. I will create a separate issue to follow up on the ideas presented here and continue the discussion.

Thanks to everyone, especially @long-walk, for initiating and working on this sensitive topic. :slightly_smiling_face:

long-walk:
Hello @jan-kolarik
I still worked a bit on this PR but as more as I change as more problems it has in edge cases. The last state looked bad with 80 cols because it uses to much space with two Version columns.

Only for reference the last state (code is not in state to push).
not verbose and Truncate data if to long:
image

verbose Warp data:
image


So then I think we can close this PR and I will look in the progress of #1518 

Thanks for your help.

PR #1193, opened 2024-01-22
Add pImpl to `ModuleItem` and remove definitions from header

kontura:

### DIFF
--- a/libdnf5/module/module_item.cpp
+++ b/libdnf5/module/module_item.cpp
@@ -38,58 +38,155 @@ extern "C" {
 
 #include <algorithm>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace libdnf5::module {
 
+class ModuleItem::Impl {
+public:
+    Impl(_ModulemdModuleStream * md_stream, const ModuleSackWeakPtr & module_sack, std::string repo_id);
+
+    ~Impl();
+    Impl(const Impl & mpkg);
+    Impl & operator=(const Impl & mpkg);
+    Impl(Impl && mpkg) noexcept;
+    Impl & operator=(Impl && mpkg) noexcept;
+
+private:
+    friend ModuleItem;
+
+    ModuleSackWeakPtr module_sack;
+
+    ModuleItemId id;
+    std::string repo_id;
+
+    // Corresponds to one yaml document
+    _ModulemdModuleStream * md_stream;
+
+    // For compatibility with older modules that didn't have static contexts
+    std::string computed_static_context;
+};
+
+ModuleItem::Impl::Impl(_ModulemdModuleStream * md_stream, const ModuleSackWeakPtr & module_sack, std::string repo_id)
+    : module_sack(module_sack),
+      repo_id(std::move(repo_id)),
+      md_stream(md_stream) {
+    if (md_stream != nullptr) {
+        g_object_ref(md_stream);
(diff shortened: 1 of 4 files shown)


libdnf5/module/module_item.cpp
```diff
@@ -38,58 +38,149 @@ extern "C" {
...
+
+ModuleItem::Impl & ModuleItem::Impl::operator=(const Impl & mpkg) {
+    if (this != &mpkg) {
+        module_sack = mpkg.module_sack;
+        id = mpkg.id;
+        repo_id = mpkg.repo_id;
+        md_stream = mpkg.md_stream;
+        if (md_stream != nullptr) {
+            g_object_ref(md_stream);
+        }
```
kontura, line 107: I am changing how to copy and move assignment operators handle the `md_stream` pointer because it seemed wrong to me. I know the same implementations is in dnf4 but I think it is not used.

libdnf5/module/module_item.cpp
```diff
@@ -38,58 +38,149 @@ extern "C" {
...
+    if (this != &mpkg) {
+        module_sack = mpkg.module_sack;
+        id = mpkg.id;
+        repo_id = mpkg.repo_id;
+        md_stream = mpkg.md_stream;
+        mpkg.md_stream = nullptr;
+        computed_static_context = mpkg.computed_static_context;
+    }
+    return *this;
+}
```
pkratoch, line 134: Hi, probably a stupid question, but in the previous code, `std::move` was used for some of the attributes. Is it for some reason done automatically in this case? If so, can it be relied on? Or is there some other reason for removing it?

libdnf5/module/module_item.cpp
kontura: I might be wrong but the way I understand it:
- Copy assignment operator (`::operator=(const Impl & mpkg)`) creates a new copy of the `Impl` so now there are two objects that hold the reference to `md_stream` therefore we need to increment the reference counter (`g_object_ref(md_stream)`). The `g_object_unref(md_stream)` is not needed because no object is destroyed.
- Move assignment operator (`::operator=(Impl && mpkg)`) moves the `Impl` so again no objects is destroyed and this time no new copy is created therefore the reference count should stay the same.

libdnf5/module/module_item.cpp
kontura: Oh, I have added it back for the strings (`computed_static_context` and `repo_id`), thanks.
Regarding `id` it is trivially-copyable type (it is basically an `int`) so the move had not effect.
For `module_sack` it also had no effect because `ModuleSackWeakPtr` doesn't have enabled move constructors.

libdnf5/module/module_item.cpp
pkratoch: I think copy constructor creates a new copy, but copy assignment operator just assigns a new value to an already initialized object. So, the already existing `module_item` has an `md_stream` and if it isn't `nullptr`, then we must decrement the reference counter. Because it is then overwritten by the other `md_stream` and the reference counter for the old one would be wrong. Similar for the move assignment operator.

But I am generally not good at this area, so maybe we could ask someone else to take a look at it?

kontura: I have completely confused these, you are correct. I will fix it.
Thank you.

PR #1194, opened 2024-01-23
Add pImpl to `libdnf5::base::transaction_*` classes

kontura:
This includes package, group, environment and module.

### DIFF
--- a/include/libdnf5/base/transaction_environment.hpp
+++ b/include/libdnf5/base/transaction_environment.hpp
@@ -21,7 +21,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #ifndef LIBDNF5_BASE_TRANSACTION_ENVIRONMENT_HPP
 #define LIBDNF5_BASE_TRANSACTION_ENVIRONMENT_HPP
 
-#include "libdnf5/base/goal_elements.hpp"
 #include "libdnf5/base/transaction.hpp"
 #include "libdnf5/comps/environment/environment.hpp"
 #include "libdnf5/transaction/transaction_item_action.hpp"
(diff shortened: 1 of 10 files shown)


include/libdnf5/base/transaction_package.hpp
```diff
@@ -43,70 +42,60 @@ class TransactionPackage {
...
-    State state{State::STARTED};
-    std::optional<std::string> reason_change_group_id;
-
-    std::vector<rpm::Package> replaces;
-    std::vector<rpm::Package> replaced_by;
+        const libdnf5::rpm::Package & pkg, Action action, Reason reason, const std::optional<std::string> & group_id);
+
+    void set_reason(Reason value) noexcept;
+    void set_state(State value) noexcept;
+    void replaced_by_append(rpm::Package && pkg);
```
jan-kolarik, line 94: This (and the one below) is also used only in impl right? Probably not needed in the public API.

libdnf5/base/transaction_environment.cpp
```diff
@@ -0,0 +1,93 @@
...
+            p_impl = std::make_unique<Impl>(*mpkg.p_impl);
+        }
+    }
+
+    return *this;
+}
+TransactionEnvironment & TransactionEnvironment::operator=(TransactionEnvironment && mpkg) noexcept = default;
+
+TransactionEnvironment::Impl::Impl(
+    libdnf5::comps::Environment grp, Action action, Reason reason, const bool with_optional)
```
jan-kolarik, line 59: I know the `const bool` was present in the original code, but the `const` doesn't have any meaning there, or does it? Also on L66.

include/libdnf5/base/transaction_environment.hpp
```diff
@@ -36,46 +35,44 @@ class TransactionEnvironment {
...
-    TransactionEnvironment(
-        const libdnf5::comps::Environment & grp, Action action, Reason reason, const bool with_optional)
-        : environment(grp),
-          action(action),
-          reason(reason),
-          with_optional(with_optional) {}
+    TransactionEnvironment(const libdnf5::comps::Environment & grp, Action action, Reason reason, bool with_optional);
 
-    void set_state(State value) noexcept { state = value; }
+    void set_state(State value) noexcept;
```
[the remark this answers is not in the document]
kontura: It is actually not even used now, I will remove it.

include/libdnf5/base/transaction_package.hpp
kontura: It will require a new private header but I can do that.

kontura: I have moved also the `set_reason(...)` to the Impl.

jan-kolarik (approved):
LGTM, thanks!

Issue #624, opened 2023-06-19
Packages with `file://` protocol are treated as remote


kontura:
Running:
```bash
root@1f0c952f6f82 ~ $ dnf5 repoquery file:///root/htop-3.2.2-2.fc38.x86_64.rpm --disable-repo=*
Updating and loading repositories:
Repositories loaded.
 file:///root/htop-3.2.2-2.fc38.x86_64.rpm        100% |   0.0   B/s | 185.5 KiB |  00m00s
htop-0:3.2.2-2.fc38.x86_64
```
The package is download into cache.

While running
```bash
root@1f0c952f6f82 ~ $ dnf5 repoquery /root/htop-3.2.2-2.fc38.x86_64.rpm --disable-repo=*
Updating and loading repositories:
Repositories loaded.
htop-0:3.2.2-2.fc38.x86_64
```
The package used directly.

jan-kolarik:
Tested with DNF4 having the same behavior.

j-mracek:
The question is whether we should handle both situation as remote or a local. A local path might be mapped remote location.

PR #1197, opened 2024-01-23
Added urlprotocol option

### LINKED ISSUE #950: Implement dnf download `--urlprotocol`

derickdiaz:
Fixes #950 

- Implements --urlprotocols to the download command and limits the valid package urls to the set url protocols.
- If no protocols are specified, all protocols are considered valid.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -82,15 +82,34 @@ void DownloadCommand::set_argument_parser() {
 
     auto url = parser.add_new_named_arg("url");
     url->set_long_name("url");
-    url->set_description("Print the list of urls where the rpms can be downloaded instead of downloading");
+    url->set_description("Print a URL where the rpms can be downloaded instead of downloading");
     url->set_const_value("true");
     url->link_value(url_option);
+
+    urlprotocol_valid_options = {"http", "https", "ftp", "file"};
+    urlprotocol_option = {};
+    auto urlprotocol = parser.add_new_named_arg("urlprotocol");
+    urlprotocol->set_long_name("urlprotocol");
+    urlprotocol->set_description("When running with --url, limit to specific protocols");
+    urlprotocol->set_has_value(true);
+    urlprotocol->set_arg_value_help("{http|https|ftp|file},...");
+    urlprotocol->set_parse_hook_func(
+        [this](
+            [[maybe_unused]] ArgumentParser::NamedArg * arg, [[maybe_unused]] const char * option, const char * value) {
+            if (urlprotocol_valid_options.find(value) == urlprotocol_valid_options.end()) {
+                throw libdnf5::cli::ArgumentParserInvalidValueError(
+                    M_("Invalid urlprotocol option: {}"), std::string(value));
+            }
+            urlprotocol_option.emplace(value);
+            return true;
+        });
 
     cmd.register_named_arg(alldeps);
     create_destdir_option(*this);
     cmd.register_named_arg(resolve);
     cmd.register_positional_arg(keys);
     cmd.register_named_arg(url);
+    cmd.register_named_arg(urlprotocol);
 }
 
 void DownloadCommand::configure() {
@@ -161,9 +180,16 @@ void DownloadCommand::run() {
     }
 
     if (url_option->get_value()) {
+        // If no urlprotocols are specified, all values within the urlprotocol_valid_options will be used
+        if (urlprotocol_option.empty()) {
+            urlprotocol_option = urlprotocol_valid_options;
+        }
         for (auto & [nerva, pkg] : download_pkgs) {
-            auto urls = pkg.get_remote_locations();
-            libdnf_assert(!urls.empty(), "Failed to get mirror for package: \"{}\"", pkg.get_name());
+            auto urls = pkg.get_remote_locations(urlprotocol_option);
+            if (urls.empty()) {
+                ctx.base.get_logger()->warning("Failed to get mirror for package: \"{}\"", pkg.get_name());
+                continue;
+            }
             std::cout << urls[0] << std::endl;
         }
         return;
--- a/doc/commands/download.8.rst
+++ b/doc/commands/download.8.rst
@@ -45,7 +45,14 @@ Options
     | To be used together with ``--resolve``, it downloads all dependencies, not skipping the already installed ones.
 
 ``--destdir=<path>``
-    Set directory used for downloading packages to. Default location is to the current working directory.
+    | Set directory used for downloading packages to. Default location is to the current working directory.
+
+``--url``
+    | Prints the list of URLs where the rpms can be downloaded instead of downloading.
+
+``--urlprotocol``
+    | To be used together with ``--url``. It filters out the URLs to the specified protocols: ``http``, ``https``, ``ftp``, or ``file``. This option can be used multiple times.
+
 
 
 Examples
@@ -62,6 +69,10 @@ Examples
 
 ``dnf5 download --destdir /tmp/my_packages maven-compiler-plugin``
     | Download the ``maven-compiler-plugin`` package to ``/tmp/my_packages`` directory.
+
+``dnf5 download --url --urlprotocol http python``
+    | List the http URL to download the python package
+
 
 See Also
 ========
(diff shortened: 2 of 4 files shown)


derickdiaz:
Works for when users specify multiple url protocols and single url protocol
image
Additionally, if an invalid url protocol is specified the following error will occur
image

dnf5/commands/download/download.cpp
```diff
@@ -161,10 +182,25 @@ void DownloadCommand::run() {
     }
 
     if (url_option->get_value()) {
+        // If no urlprotocols are specified, then any urlprotocol is acceptable
```
ppisar, line 185: Not any protocol. The four protocols listed in `urlprotocol_valid_options`.

dnf5/commands/download/download.cpp
```diff
@@ -161,10 +182,25 @@ void DownloadCommand::run() {
...
+        if (urlprotocol_option.empty()) {
+            urlprotocol_option = urlprotocol_valid_options;
+        }
         for (auto & [nerva, pkg] : download_pkgs) {
             auto urls = pkg.get_remote_locations();
             libdnf_assert(!urls.empty(), "Failed to get mirror for package: \"{}\"", pkg.get_name());
-            std::cout << urls[0] << std::endl;
+            auto valid_url = std::find_if(urls.begin(), urls.end(), [this](std::string url) {
+                for (auto protocol : urlprotocol_option) {
+                    if (url.starts_with(protocol)) {
```
ppisar, line 194: You need to match up to and including a colon charater. Otherwise, "httpnextgen://foo" would match "http".

dnf5/commands/download/download.cpp
```diff
@@ -86,11 +88,30 @@ void DownloadCommand::set_argument_parser() {
     url->set_const_value("true");
     url->link_value(url_option);
 
+    urlprotocol_valid_options = {"http", "https", "rsync", "ftp"};
```
ppisar, line 91: "rsync" should be ignored  according to #497.

dnf5/commands/download/download.cpp
ppisar: Wouldn't be easier to error on `--urlprotocol` option without an argument? If one does not want to limit the protocols, he simply does not pass this option.

ppisar:
Please document this new option in dnf5-download manual page (doc/commands/download.8.rst file).

ppisar:
Could you change the commit message to "Add --urlprotocol option to download command"?

dnf5/commands/download/download.cpp
derickdiaz: It currently does error if there's no value passed into the --urlprotocol option. I believe this is because the set_has_value attribute is set to true
image

derickdiaz:
I added documentation for the url option within this PR as well.

dnf5/commands/download/download.cpp
kontura: It would be better if the filtering logic wasn't duplicated here, you can pass the set of protocols directly to `get_remote_locations()`, it takes it as an optional parameter.

It does have the same problem with the `httpnextgen://foo` but you could fix it there directly.

dnf5/commands/download/download.cpp
```diff
@@ -161,10 +181,27 @@ void DownloadCommand::run() {
...
+                    // Ensure the urls matches the protocol specified by concating the colon.
+                    // i.e https: or http: not httpnextgen:
+                    if (url.starts_with(protocol + ":")) {
+                        return true;
+                    }
+                }
+                return false;
+            });
+            if (valid_url == urls.end()) {
+                libdnf_assert(true, "Failed to get mirror for package: \"{}\"", pkg.get_name());
```
kontura, line 202: With the protocol filtering the asserts are probably too strong. I think it can easily happen that a specific package isn't available over selected protocol.

I would change it to logging:
```cpp
ctx.base.get_logger()->warning("Failed to get mirror for package: \"{}\"", pkg.get_name());
```

dnf5/commands/download/download.cpp
ppisar: Remove "rsync" also from help text in `urlprotocol->set_arg_value_help("{http, https, rsync, ftp},...")`.

derickdiaz: Would "file" also be a valid option?

dnf5/commands/download/download.cpp
derickdiaz: This same solution was proposed for line 190 when I implemented the url option. Should I change it there as well for consistency? Or technically I can remove line 190 because I can catch it in the if condition

dnf5/commands/download/download.cpp
ppisar: Yes, keep "file" among remote protocols. We currently handle file:// schema as a remote location (see issue #624).

dnf5/commands/download/download.cpp
derickdiaz: Added to the get_remote_locations function

doc/commands/download.8.rst
```diff
@@ -45,7 +45,14 @@ Options
...
 
 ``--destdir=<path>``
-    Set directory used for downloading packages to. Default location is to the current working directory.
+    | Set directory used for downloading packages to. Default location is to the current working directory.
+
+``--url``
+    | Prints the list of urls where the rpms can be downloaded instead of downloading.
+
+``--urlprotocol``
+    | To be used together with ``--url``, it filters out the urls to the specified url protocols: {http, https, ftp, file}
```
ppisar, line 54: The option only accepts a single value. On the other hand, the option can be used multiple times. Could you please rewrite this documentation to make it clear. Also please spell URL with capitals. E.g.: 

~~~~
To be used together with ``--url``. It filters out the URLs to the specified protocol: ``http``, ``https``,
``ftp``, or ``file``. This option can be used multiple times.
~~~~

libdnf5/rpm/package.cpp
```diff
@@ -329,7 +329,7 @@ std::vector<std::string> Package::get_remote_locations(const std::set<std::strin
     auto repo_mirrors = get_repo()->get_mirrors();
     for (const auto & mirror : repo_mirrors) {
         for (const auto & protocol : protocols) {
-            if (utils::string::starts_with(mirror, protocol)) {
+            if (utils::string::starts_with(mirror, protocol + ":")) {
```
ppisar, line 332: Here is missing the ``Ensure the urls...`` comment. Either copy it there, or remove it from the other occurance at line 344.

doc/commands/download.8.rst
```diff
@@ -45,7 +45,14 @@ Options
     | To be used together with ``--resolve``, it downloads all dependencies, not skipping the already installed ones.
 
 ``--destdir=<path>``
-    Set directory used for downloading packages to. Default location is to the current working directory.
+    | Set directory used for downloading packages to. Default location is to the current working directory.
+
+``--url``
+    | Prints the list of urls where the rpms can be downloaded instead of downloading.
```
ppisar, line 51: Uppercase "URLs". please.

doc/commands/download.8.rst
```diff
@@ -63,6 +70,10 @@ Examples
 ``dnf5 download --destdir /tmp/my_packages maven-compiler-plugin``
     | Download the ``maven-compiler-plugin`` package to ``/tmp/my_packages`` directory.
 
+``dnf5 download --url --urlprotocol http python``
+    | List the http url to download the python package
```
ppisar, line 74: Uppercase "URL" please.

dnf5/commands/download/download.cpp
```diff
@@ -82,15 +82,34 @@ void DownloadCommand::set_argument_parser() {
 
     auto url = parser.add_new_named_arg("url");
     url->set_long_name("url");
-    url->set_description("Print the list of urls where the rpms can be downloaded instead of downloading");
+    url->set_description("Print a url where the rpms can be downloaded instead of downloading");
```
ppisar, line 85: Could you please capitalize "URL"?

ppisar:
Otherwise, the code looks and works well.

libdnf5/rpm/package.cpp
derickdiaz: Opted to delete the comment. It self explanatory

ppisar (approved):
Now it's good. Thanks.

Issue #1184, opened 2024-01-19
"needs-restarting -r" reports no reboot required if RTC is not in UTC


inknos:
This issue comes from a ticket filled for dnf4
https://issues.redhat.com/browse/RHEL-14900

I am reporting two weird behaviours that I suspect are connected with a single issue.

This [PR](https://github.com/rpm-software-management/dnf-plugins-core/pull/518) fixes the first report partially, but beaker tests revealed that the wanted behavior is inconsistent with repeated testing.

### Reproducer 1
I could reproduce it with `dnf5-5.1.11-1.fc39.x86_64` on F39 with the steps:
1.
```bash
timedatectl set-local-rtc 1
```
2.
```bash
reboot
```
3.
```bash
dnf5 -y reinstall systemd
```
4.
```bash
dnf5 needs-restarting -r
...
Core libraries or services have been updated since boot-up:
  * systemd

Reboot is required to fully utilize these updates.
```
5.
```bash
reboot
```
6.
```bash
dnf5 needs-restarting -r
```

### Actual Results

```bash
Core libraries or services have been updated since boot-up:
  * systemd

Reboot is required to fully utilize these updates.
```

### Expected results
```
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```

---

### Reproducer 2

1.
```bash
timedatectl set-local-rtc 1
timedatectl set-timezone Europe/Paris
```
2.
```bash
reboot
```
3.
```bash
dnf5 reinstall -y systemd
```
4.
```bash
dnf5 needs-restarting -r
```

### Expected results
```bash
Core libraries or services have been updated since boot-up:
  * systemd

Reboot is required to fully utilize these updates.
```

### Actual results
```
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```

evan-goode:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=2237223

I can reproduce with your second reproducer. DNF 5 is getting an incorrect boot time from systemd:

/var/log/dnf5.log:
```
2024-01-22T14:36:31+0000 [3083] DEBUG Got boot time from systemd: 1705937754
```
1705937754 is Mon Jan 22 15:35:54 2024 UTC.

The `btime` in `/proc/stat` that you use in https://github.com/rpm-software-management/dnf-plugins-core/pull/518 seems to be reliable for me, but in containers it will have the host's boot time, not the container's.

[...]

PR #1198, opened 2024-01-23
needs-restarting: get systemd boot time from UnitsLoadStartTimestamp

evan-goode:
For https://github.com/rpm-software-management/dnf5/issues/1184

Following the reproducers in that issue, `UnitsLoadStartTimestamp` seems to consistently have the correct boot time even when `UserspaceTimestamp` is incorrect due to `set-local-rtc 1`. The two times are within a few seconds of each other. I have no idea why `UnitsLoadStartTimestamp` is correct, maybe systemd reads `/etc/adjtime` and adjusts the clock after setting `UserspaceTimestamp` but before setting `UnitsLoadStartTimestamp`.


ppisar:
What does happen if you restart systemd with `systemctl daemon-reexec`. Won't UnitsLoadStartTimestamp reset?

evan-goode:
> What does happen if you restart systemd with `systemctl daemon-reexec`. Won't UnitsLoadStartTimestamp reset?

All the timestamps stay the same for me on systemd 253.

ppisar (approved):
Great.

PR #1199, opened 2024-01-24
Move number placeholder to postposition in copr_repo.cpp

rffontenelle:
Since the first placeholder `{}` is the number of an item (not the number of items), placing the placeholder after the word "item" makes the sentence clearer.

See discussion #1160.


ppisar (approved):
Good.

Issue #358, opened 2023-03-16
Output user name and login at dnf5 history info


geraldosimiao:
Actual dnf 4 history info returns, among other information, user name and login.
Dnf5 instead return only user ID at the same field (as seen on screenshot attached).
It is possible to add again the user name and login? Can be together with user ID.
![Screenshot_20230315_140620-1](https://user-images.githubusercontent.com/72357821/225504810-345d1d81-2e8b-4d42-8a5c-334cb18afbc9.jpg)

derickdiaz:
For this issue, I believe we can use include the <pwd.h> and use the getpwuid() function to resolve the UID to a valid username within in the transaction.cpp file. If the function returns a null, then we just print out the standard uid as it has been doing.

Example: https://linux.die.net/man/3/getpwuid
I tried using this function on my fedora box and it worked.

ppisar:
I believe this issue has been fixed by your PR #1203.

PR #1203, opened 2024-01-25
Change user info display on history command to include display name a…

derickdiaz:
Adds Display name and username to the history info command. This code assumes that the gecos information is in CSV format as normally this is standard. Based on that information it also assumes that the first entry within the gecos information is the display name.

Implements #358

This is an example of without and with the gecos information
Screenshot 2024-01-25 at 6 50 58 AM

Additionally if the user's information is not found, it will only display the user id as it has done originally.

### DIFF
--- a/libdnf5-cli/output/transactioninfo.cpp
+++ b/libdnf5-cli/output/transactioninfo.cpp
@@ -22,8 +22,38 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "utils/string.hpp"
 
+#include <pwd.h>
+
+#include <format>
+#include <sstream>
 
 namespace libdnf5::cli::output {
+
+namespace {
+std::string generate_user_info_str(uint32_t user_id) {
+    auto results = std::to_string(user_id);
+    auto user_info = getpwuid(user_id);
+
+    if (user_info == NULL) {
+        return results;
+    };
+
+    // If gecos information does exists, it is generally CSV formatted.
+    // This section assumes that the first entry in the gecos information is the display name.
+    if (user_info->pw_gecos && strlen(user_info->pw_gecos) > 0) {
+        std::stringstream s_stream(user_info->pw_gecos);
+        std::string display_name;
+        std::getline(s_stream, display_name, ',');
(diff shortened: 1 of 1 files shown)


include/libdnf5-cli/output/transactioninfo.hpp
```diff
@@ -30,6 +30,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace libdnf5::cli::output {
 
 void print_transaction_info(libdnf5::transaction::Transaction & transaction);
+std::string generate_user_info_str(uint32_t user_id);
```
j-mracek, line 33: Is there any reason why you want to add the function into public API? Be honest if there is no good reason I would prefer to move it in cpp to anonymous namespace and remove the declaration from HPP.

libdnf5-cli/output/transactioninfo.cpp
```diff
@@ -45,4 +48,28 @@ void print_transaction_info(libdnf5::transaction::Transaction & transaction) {
...
+    // If gecos information does exists, it is generally CSV formatted.
+    // This section assumes that the first entry in the gecos information is the display name.
+    if (user_info->pw_gecos != NULL && !std::string(user_info->pw_gecos).empty()) {
+        std::stringstream s_stream(user_info->pw_gecos);
+        std::string display_name;
+        std::getline(s_stream, display_name, ',');
+        results += std::format(" {}", display_name);
+    };
+
+    if (user_info->pw_name != NULL && !std::string(user_info->pw_name).empty()) {
```
j-mracek, line 68: The condition `if (user_info->pw_name != NULL && !std::string(user_info->pw_name).empty())` could be replaced by `if (user_info->pw_name && strlen(user_info->pw_name) > 0)` or `if (user_info->pw_name != nullptr && strlen(user_info->pw_name) > 0)`.

It will not create any temporal object (std::string) to only check the length of `char *`

j-mracek:
LGTM

Conan-Kudo:
This broke Mageia builds by silently raising the minimum GCC version to 13, see #1269 for details.

Issue #1167, opened 2024-01-13
FTBFS with gcc 9.x: 'in_range' is not a member of 'std'


shastah:
Trying to compile dnf5 under gcc 9.x (tried 9.2.0, 9.5.0) I'm getting this
```
In file included from /home/rpm/rpmbuild/BUILD/common/utils/string.cpp:21:
/home/rpm/rpmbuild/BUILD/common/utils/string.hpp: In function 'std::string libdnf5::utils::string::format_epoch(T)':
/home/rpm/rpmbuild/BUILD/common/utils/string.hpp:115:14: error: 'in_range' is not a member of 'std'
  115 |     if (std::in_range<time_t>(epoch_num)) {
      |              ^~~~~~~~
/home/rpm/rpmbuild/BUILD/common/utils/string.hpp:115:29: error: expected primary-expression before '>' token
  115 |     if (std::in_range<time_t>(epoch_num)) {
      |                             ^
```
Compiler info
```
# gcc --version
gcc (GCC) 9.2.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

# g++ --version
g++ (GCC) 9.2.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

ppisar:
According to libstdc++-v3/ChangeLog-2020, std::in_range() was added on 2020-07-07. According to <https://gcc.gnu.org/releases.html>, GCC 9.2 was released on 2019-08-12.

According to <https://en.cppreference.com/w/cpp/utility/in_range> std::range is implemented if 
`__cpp_lib_integer_comparison_functions` feature macro is at least `202002L`. According to <https://en.cppreference.com/w/cpp/feature_test#cpp_lib_integer_comparison_functions> the macro was introduced in [P0586R2](https://wg21.link/P0586R2) draft. According to <https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2020> that "Safe integral comparisons" draft was fully implemented in GCC 10.1.

I worry you need to use a newer compiler.

[...]

PR #1204, opened 2024-01-25
Build: Require GCC 10.1 for std::in_range<>()

ppisar:
A GCC 9.2 user reported this failure:

    In file included from /home/rpm/rpmbuild/BUILD/common/utils/string.cpp:21:
    /home/rpm/rpmbuild/BUILD/common/utils/string.hpp: In function 'std::string libdnf5::utils::string::format_epoch(T)':
    /home/rpm/rpmbuild/BUILD/common/utils/string.hpp:115:14: error: 'in_range' is not a member of 'std'
      115 |     if (std::in_range<time_t>(epoch_num)) {
	  |              ^~~~~~~~

The cause is that std::in_range<>() is a C++20 feature added to GCC in version 10.1.

This patch sets the minimal version in dnf5.spec.

Originally, I also wanted to add a check into a build script to fail early. std::in_range<>() is supported if the compiler defines __cpp_lib_integer_comparison_functions >= 202002L feature test macro.

However, checking for it in CMake is very difficult: target_compile_features() does not recognize this specific feature. check_cxx_symbol_exists() is unable to retrieve the macro value. CheckSourceCompiles() requires CMake 3.19 (a year after GCC 10.1).

So I scratched out the idea of adding a check into the build script.

Fixes: #1167

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -136,7 +136,7 @@ BuildRequires:  toml11-static
 %if %{with clang}
 BuildRequires:  clang
 %else
-BuildRequires:  gcc-c++
+BuildRequires:  gcc-c++ >= 10.1
 %endif
 
 %if %{with tests}


jan-kolarik (approved):
Thanks!

Issue #946, opened 2023-10-05
dnf download `--arch <arch>[,<arch>...]`


j-mracek:
Limit the query to packages of given architectures (default is all compatible architectures with
your system). To download packages with arch incompatible with your system use
--forcearch=<arch> option to change basearch.

derickdiaz:
I would like to take this on after #1197 
It looks like we just have to add the option and use the filter_arch on the pkg_query. The dnf4 implementation of this does not validate if it's a valid arch specified.

PR #1206, opened 2024-01-27
Added arch option to the download command

derickdiaz:
Fixes: #946 

Add the --arch option to the download command. Filters queried compatible packages based on the specified architectures. If the architecture needs to be changed, the users can use the global --forcearch to change the base architecture.


kontura (changes requested):

kontura (approved):
Great, thank you!

PR #1207, opened 2024-01-28
download: add `--srpm` option

### LINKED ISSUE #947: Implement dnf download `--srpm`

principis:
Fixes #947

Implements the `--source` option for the download command to download only the source rpm(s).

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -69,6 +69,9 @@ void DownloadCommand::set_argument_parser() {
     url_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
+    srpm_option = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+
     auto resolve = parser.add_new_named_arg("resolve");
     resolve->set_long_name("resolve");
     resolve->set_description("Resolve and download needed dependencies");
@@ -81,6 +84,12 @@ void DownloadCommand::set_argument_parser() {
         "When running with --resolve, download all dependencies (do not exclude already installed ones)");
     alldeps->set_const_value("true");
     alldeps->link_value(alldeps_option);
+
+    auto srpm = parser.add_new_named_arg("srpm");
+    srpm->set_long_name("srpm");
+    srpm->set_description("Download the src.rpm instead");
+    srpm->set_const_value("true");
+    srpm->link_value(srpm_option);
 
     auto url = parser.add_new_named_arg("url");
     url->set_long_name("url");
@@ -140,6 +149,7 @@ void DownloadCommand::set_argument_parser() {
     create_destdir_option(*this);
     cmd.register_named_arg(resolve);
     cmd.register_positional_arg(keys);
+    cmd.register_named_arg(srpm);
     cmd.register_named_arg(url);
     cmd.register_named_arg(urlprotocol);
     cmd.register_named_arg(arch);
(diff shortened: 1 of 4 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -179,6 +194,24 @@ void DownloadCommand::run() {
         return;
     }
 
+    if (source_option->get_value()) {
+        std::map<std::string, libdnf5::rpm::Package> source_pkgs;
+
+        for (auto & [nevra, pkg] : download_pkgs) {
+            libdnf5::rpm::PackageQuery pkg_query(full_pkg_query);
+            pkg_query.resolve_pkg_spec(pkg.get_source_name(), {}, true);
```
kontura, line 202: Resolving just the source name is not enough because the user can specify an exact version to download but this will always download the latest srpm.

You can check out what does old dnf do: https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/download.py#L268-L280

dnf5/commands/download/download.cpp
```diff
@@ -179,6 +194,24 @@ void DownloadCommand::run() {
...
+            pkg_query.filter_arch({"src"});
+            pkg_query.filter_available();
```
kontura, line 204: I think these filters could be done just once outside of the for loop.
You can create your own pre-filtered query and then just copy that instead of the `full_pkg_query`.

dnf5/commands/download/download.cpp
```diff
@@ -80,6 +83,12 @@ void DownloadCommand::set_argument_parser() {
     alldeps->set_const_value("true");
     alldeps->link_value(alldeps_option);
 
+    auto source = parser.add_new_named_arg("source");
+    source->set_long_name("source");
+    source->set_description("Download the src.rpm instead");
+    source->set_const_value("true");
+    source->link_value(source_option);
```
kontura, line 90: Can you please rename the option to `--srpm`?

We are trying to unify  the options in dnf5 and we have decided to use `--srpm` everywhere. Old dnf sometimes used `--source` and sometimes `--srpm` which was confusing.

Sorry about the outdated [issue](https://github.com/rpm-software-management/dnf5/issues/947) I have changed it there to `--srpm` as well.

kontura (changes requested):

principis:
Thanks for the feedback! I updated the PR. I'm not sure if this is the best way to do this, but I couldn't find something else.

dnf5/commands/download/download.cpp
```diff
@@ -179,6 +194,36 @@ void DownloadCommand::run() {
...
+
+        for (auto & [nevra, pkg] : download_pkgs) {
+            auto sourcerpm = pkg.get_sourcerpm();
+
+            if (!sourcerpm.empty()) {
+                libdnf5::rpm::PackageQuery pkg_query(source_pkg_query);
+                // Remove ".rpm" to get sourcerpm nevra
+                sourcerpm.erase(sourcerpm.length() - 4);
+                pkg_query.resolve_pkg_spec(sourcerpm, {}, true);
+
```
kontura, line 212: One last thing, I have noticed the `sourcerpm` doesn't include epoch. Could you add a filter for it?
```cpp
pkg_query.filter_epoch({pkg.get_epoch()});
```
This should resolve a corner case problem when a package is available in multiple epochs.
Otherwise it looks good!

kontura (changes requested):
This loo

kontura:
(I have also prepared a PR with enabled CI tests once we merge this: https://github.com/rpm-software-management/ci-dnf-stack/pull/1443)

kontura:
Could you also add a note to `doc/changes.rst` saying:
```
Download command
----------------
 * Option ``--source`` was renamed to ``--srpm``.
```

principis:
Sorry for the delay, I was on vacation.

I've made the requested changes and rebased the branch, thank you for the review!

kontura:
Thank you for the contribution!

Issue #1106, opened 2023-12-15
Unable to upgrade a single package unless --no-best is used


Lamieur:
F38, dnf5-5.1.7-1.fc38.x86_64

When running `dnf5 upgrade <package>`, dnf5 wants to upgrade any dependency of any package being upgraded.

The same happens with other methods of limiting the upgrade's scope, like `dnf5 upgrade --security` additionally pulls all dependencies of all packages with security updates.

Old dnf 4 works as expected:
```
# LANG=C dnf upgrade discord
(...)
Upgrading:
 discord        x86_64        0.0.38-1.fc38        rpmfusion-nonfree-updates         70 M

Transaction Summary
==========================================================================================
Upgrade  1 Package
```
dnf5 wants to do much more:
```
# dnf5 upgrade discord
(...)
Upgrading:                                                                                
 discord                        x86_64 0.0.38-1.fc38              rpmfusion-nonf 219.5 MiB
  replacing discord             x86_64 0.0.37-1.fc38              rpmfusion-nonf 219.5 MiB
 glibc                          i686   2.37-16.fc38               updates          5.5 MiB
  replacing glibc               i686   2.37-14.fc38               updates          5.5 MiB
(...)
Transaction Summary:
 Upgrading:        15 packages
 Replacing:        15 packages
```
This is not a full upgrade (that would upgrade 93 packages + kernel stuff) nor an upgrade --security (that would upgrade 59 + kernel - hitting the exact same issue, as these packages do NOT offer security fixes - dnf upgrade --security offers just 35 + kernel).

Looks like dnf upgrade <package> additionally pulls updates for the package's dependencies, even if they would be satisfied without any more updates.

And in fact, I can make it work like dnf4:
```
# dnf5 update discord --no-best
Updating and loading repositories:
Repositories loaded.
Package                    Arch   Version                    Repository               Size
Upgrading:                                                                                
 discord                   x86_64 0.0.38-1.fc38              rpmfusion-nonfree-u 219.5 MiB
   replacing discord       x86_64 0.0.37-1.fc38              rpmfusion-nonfree-u 219.5 MiB

Transaction Summary:
 Upgrading:         1 packages
 Replacing:         1 packages
```

dnf5 man page doesn't say --best is the default (and its description covers a very different case), and even if it was, same man page specifically says "Note  that the use of the newest available version is only guaranteed for the packages directly requested (e.g. as a command line  arguments),  and the  solver may use older versions of dependencies to meet their requirements."

[...]

PR #1211, opened 2024-01-29
Revert "Use focusbest: prefer latest deps versions over smaller transactions

kontura:
This reverts commit 2d9df1c43b0b2f7027a6366dc1d64c0e646d1e2a.

The current behavior of `SOLVER_FLAG_FOCUS_BEST` turned out not be viable at this time, more details in: https://github.com/rpm-software-management/dnf5/issues/1106

Depending on https://github.com/openSUSE/libsolv/issues/549 we can either make the `SOLVER_FLAG_FOCUS_BEST` functionality configurable with a new option or use some different approach.

Closes https://github.com/rpm-software-management/dnf5/issues/1106

### DIFF
--- a/libdnf5/rpm/solv/goal_private.cpp
+++ b/libdnf5/rpm/solv/goal_private.cpp
@@ -407,9 +407,6 @@ libdnf5::GoalProblem GoalPrivate::resolve() {
     libsolv_solver.set_flag(SOLVER_FLAG_ALLOW_VENDORCHANGE, vendor_change);
     libsolv_solver.set_flag(SOLVER_FLAG_DUP_ALLOW_VENDORCHANGE, vendor_change);
 
-    // Ensure the solver tries to use the latest versions of dependencies, even if it results in a bigger transaction
-    libsolv_solver.set_flag(SOLVER_FLAG_FOCUS_BEST, 1);
-
     if (libsolv_solver.solve(job)) {
         return libdnf5::GoalProblem::SOLVER_ERROR;
     }


j-mracek:
What about to make the flag optional? In past we've got several reports where users complained that transaction is performed with not latest packages. In case that I have a package `foo` and I want to upgrade or install it with all latest dependencies, such an option would be beneficial. This use case is applicable for containers (build and upgrade). What do you think?

kontura:
> What about to make the flag optional? In past we've got several reports where users complained that transaction is performed with not latest packages. In case that I have a package `foo` and I want to upgrade or install it with all latest dependencies, such an option would be beneficial. This use case is applicable for containers (build and upgrade). What do you think?

Yes as I wrote in the description I would first wait for the response from libsolv and depending on that we can either make it a new option or do something else.

But I think we should do this revert as soon as possible because the behavior is really not great.

libdnf5/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ pkg_check_modules(LIBMODULEMD REQUIRED modulemd-2.0>=2.11.2)
 list(APPEND LIBDNF5_PC_REQUIRES "${LIBMODULEMD_MODULE_NAME}")
 target_link_libraries(libdnf5 PRIVATE ${LIBMODULEMD_LIBRARIES})
 
-pkg_check_modules(LIBSOLV REQUIRED libsolv>=0.7.25)
+pkg_check_modules(LIBSOLV REQUIRED libsolv>=0.7.21)
```
j-mracek, line 54: I do not recommend to lower a minimal version. It might happen that other recently introduced feature requires  newer version then `libsolv>=0.7.21` but at the time of introducing it the minimal require was `libsolv>=0.7.25`. Are you sure that this scenario is not applicable for this case?

j-mracek:
LGTM

libdnf5/CMakeLists.txt
kontura: I am not 100% sure, I have dropped the change.

kontura:
Here is a PR with ci tests update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1441

PR #1216, opened 2024-01-30

j-mracek:


include/libdnf5/rpm/package_query.hpp
```diff
@@ -74,12 +74,23 @@ class PackageQuery : public PackageSet {
 
     /// Filter packages by their `name`.
     ///
-    /// @param patterns         A vector of strings the filter is matched against.
+    /// @param pattern          A strings the filter is matched against.
```
jan-kolarik, line 77:
```suggestion
    /// @param pattern          A string the filter is matched against.
```

jan-kolarik:
It looks good, I'd just squash the commits into a single one, what do you think? I know there are mentioned specific use cases in the commit messages for each of the method, but in the end, it's all about a single behavior change, right?

include/libdnf5/rpm/package_query.hpp
j-mracek: Fixed

j-mracek:
> It looks good, I'd just squash the commits into a single one, what do you think? I know there are mentioned specific use cases in the commit messages for each of the method, but in the end, it's all about a single behavior change, right?

Yes, I can squash it into one commit, but then it will be less readable why additional lines were modified. I would prefer to keep those small commits because each of them represent an atomic change.

Issue #1111, opened 2023-12-19
libdnf5 doesn't pick highest NEVRA when a package is available from 2 arches


bcl:
Anaconda's nightly tests are hitting an odd problem roblem -- packages with different arches are trying to be installed at the same time: https://github.com/rhinstaller/kickstart-tests/actions/runs/7263835386/job/19790019961

```
Problem: cannot install both anaconda-install-img-deps-40.12-1.20231205144801720625.master.3.gd66b9912c5.fc40.noarch and anaconda-install-img-deps-40.15-1.20231219045442936819.master.0.gc6e2e850b8.fc40.x86_64
    - conflicting requests
```

The copr repo at https://download.copr.fedorainfracloud.org/results/%40rhinstaller/Anaconda/fedora-rawhide-x86_64/ seems to have grouped all the different builds together, and it ends up that it has the `anaconda-install-img-deps` package in there as both a noarch and as x86_64. With libdnf5 it is returning BOTH of these, despite the x86_64 version being higher, and then trying to install both and getting a conflict.

This seems to be a filtering issue because with dnf4 I only see it return the higher version:

```pkgnames = [<hawkey.Package object id 137195, anaconda-install-img-deps-40.15-1.20231219045442936819.master.0.gc6e2e850b8.fc40.x86_64, lorax-repo-2>]```

j-mracek:
I guess that the problem is not in query but in way how they are used.
Expectation from the latest filter.

For each `name` and `architecture` keep all package objects with the latest version. It means for each architecture it keeps  the latest version including `noarch`, `x86_64`, `i686` or even non compatible architecture for your system `s390`, ... .

I believe that you have a different expectation therefore you are not satisfy with the query results. May I ask you for clarification of expected results when multiple architecture and multiple version are present in available repositories?
For `x86_64` systems there are following packages available: 

anaconda-1-1.noarch
anaconda-3-3.noarch from repo B
anaconda-3-3.noarch from repo A
anaconda-3-3.noarch from repo A
anaconda-2.2.x86_64
anaconda-2.3.i686
anaconda-3-4.s390

bcl:
I would expect the same behavior as dnf4, which seems to be the highest NEVR irrespective of arch.

j-mracek:
I am suggesting to add a new method to our API

PR #1217, opened 2024-01-30
PackageQuery: Add `filter_{latest,earliest}_evr_ignore_arch`

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/1111.

`filter_latest_evr` and `filter_earliest_evr` group packages by architecture and keep the latest package on each architecture. After `filter_latest_evr`, a PackageQuery might have anaconda-1-1.noarch as well as anaconda-2.2.x86_64.

This commit adds `filter_latest_evr_ignore_arch` and `filter_earliest_evr_ignore_arch` which filter for packages with the latest versions regardless of their architecture. Filtering the above anaconda example with `filter_latest_evr_ignore_arch` would result in just {anaconda-2.2.x86_64} since it is the latest version of the package for any arch.

This is an ABI-breaking change and thus is targeted at dnf5 5.2.

### DIFF
--- a/include/libdnf5/rpm/package_query.hpp
+++ b/include/libdnf5/rpm/package_query.hpp
@@ -618,12 +618,26 @@ public:
     // @replaces libdnf/sack/query.hpp:method:addFilter(int keyname, int cmp_type, int match) - cmp_type = HY_PKG_LATEST_PER_ARCH
     void filter_latest_evr(int limit = 1);
 
+    /// Group packages by `name`. Then within each group, keep packages that correspond with up to `limit` of (all but) latest `evr`s in the group.
+    ///
+    /// @param limit            If `limit` > 0, keep `limit` number `evr`s in each group.
+    ///                         If `limit` < 0, keep all **but** `limit` last `evr`s in each group.
+    /// @since 5.2
+    void filter_latest_evr_any_arch(int limit = 1);
+
     /// Group packages by `name` and `arch`. Then within each group, keep packages that correspond with up to `limit` of (all but) earliest `evr`s in the group.
     ///
     /// @param limit            If `limit` > 0, keep `limit` number `evr`s in each group.
     ///                         If `limit` < 0, keep all **but** `limit` last `evr`s in each group.
     /// @since 5.0
     void filter_earliest_evr(int limit = 1);
+
+    /// Group packages by `name`. Then within each group, keep packages that correspond with up to `limit` of (all but) earliest `evr`s in the group.
+    ///
+    /// @param limit            If `limit` > 0, keep `limit` number `evr`s in each group.
+    ///                         If `limit` < 0, keep all **but** `limit` last `evr`s in each group.
+    /// @since 5.2
+    void filter_earliest_evr_any_arch(int limit = 1);
 
     /// Group packages by `name` and `arch`. Then within each group, keep packages that belong to a repo with the highest priority (the lowest number).
     /// The filter works only on available packages, installed packages are not affected.
--- a/libdnf5/rpm/package_query.cpp
+++ b/libdnf5/rpm/package_query.cpp
@@ -2435,6 +2435,19 @@ static int latest_cmp(const Id * ap, const Id * bp, libdnf5::solv::RpmPool * poo
     return *ap - *bp;
 }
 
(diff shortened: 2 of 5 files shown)


include/libdnf5/rpm/package_query.hpp
```diff
@@ -618,13 +618,27 @@ class PackageQuery : public PackageSet {
     // @replaces libdnf/sack/query.hpp:method:addFilter(int keyname, int cmp_type, int match) - cmp_type = HY_PKG_LATEST_PER_ARCH
     void filter_latest_evr(int limit = 1);
 
+    /// Group packages by `name`. Then within each group, keep packages that correspond with up to `limit` of (all but) latest `evr`s in the group.
+    ///
+    /// @param limit            If `limit` > 0, keep `limit` number `evr`s in each group.
+    ///                         If `limit` < 0, keep all **but** `limit` last `evr`s in each group.
+    /// @since 5.2
+    void filter_latest_evr_ignore_arch(int limit = 1);
```
j-mracek, line 626: I am thinking about naming. 
Important information is `Group packages by `name` or `Group packages by `name` and `arch`. But how it fits with
`ignore_arch`?
`latest` and `latest_evr` -  I think latest is always according `evr`.

libdnf5/rpm/package_query.cpp
```diff
@@ -2471,7 +2505,8 @@ static void filter_first_sorted_by(
     int i;
     for (i = 0; i < samename.size(); ++i) {
         Solvable * considered = pool.id2solvable(samename[i]);
-        if (!highest || highest->name != considered->name || highest->arch != considered->arch) {
+        if (!highest || highest->name != considered->name ||
+            (grouping == FilterGrouping::BY_NAME_ARCH && (highest->arch != considered->arch))) {
```
j-mracek, line 2509: Only  note with no significant impact on the code. It adds one boolean test which is fast,  but for each query this test is performed for each name in sets  (Fedora has something like 40 000 package  names).

include/libdnf5/rpm/package_query.hpp
evan-goode: Yeah, I had `filter_latest_any_arch` in an earlier draft, do you think that's better than `filter_latest_ignore_arch`? I agree that `ignore_arch` is not ideal since `ignorearch` has an established meaning in our project.

I don't like simply `filter_latest` as much since `filter_latest_unresolved_advisories` also groups packages by arch; it seems that grouping by name+arch is common for the two existing `filter_latest_*` functions.

libdnf5/rpm/package_query.cpp
evan-goode: Hmm, the only alternative I can think of is duplicating the function or at least the for-loop. It should be only one additional AND instruction though?

libdnf5/rpm/package_query.cpp
```diff
@@ -2453,11 +2466,32 @@ static int earliest_cmp(const Id * ap, const Id * bp, libdnf5::solv::RpmPool * p
...
+        return r;
+    r = pool->evrcmp(sb->evr, sa->evr, EVRCMP_COMPARE);
+    if (r > 0)
+        return -1;
+    if (r < 0)
+        return 1;
+    return *ap - *bp;
+}
+
+enum class FilterGrouping {
```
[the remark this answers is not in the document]
evan-goode: It probably won't. I tend to be overzealous in favoring enums over booleans. I'll change it to a boolean.

include/libdnf5/rpm/package_query.hpp
jan-kolarik: `filter_earliest_evr_any_arch` sounds good to me. Especially when looking to the current state of sources in the `package_query.hpp` where the docstrings for `filter_latest_evr` are followed by the `filter_earliest_evr_any_arch` that gives a clear and coherent feeling.

kontura:
I may be missing something but how is this breaking ABI? It is adding two new functions to a class that don't overload anything that should be fine no?

evan-goode:
Oh :facepalm: you're right, I was under the impression that adding any new member would break ABI, but it's only virtual functions and member variables. I can retarget this to `main` then.

j-mracek:
LGTM

Issue #1168, opened 2024-01-13
FTBFS with gcc 10.x: 'isatty' was not declared in this scope


shastah:
Trying to build with gcc 10.4.0:
```
[ 48%] Building CXX object libdnf5-cli/CMakeFiles/libdnf5-cli.dir/output/advisorylist.cpp.o
cd /home/rpm/rpmbuild/BUILD/linux-build/libdnf5-cli && /bin/c++ -DGETTEXT_DOMAIN=\"libdnf5-cli\" -DPROJECT_BINARY_DIR=\"/home/rpm/rpmbuild/BUILD/linux-build\" -DPROJECT_SOURCE_DIR=\"/home/rpm/rpmbuild/BUILD\" -DPROJECT_VERSION_MAJOR=5 -DPROJECT_VERSION_MINOR=1 -DPROJECT_VERSION_PATCH=10 -Dlibdnf5_cli_EXPORTS -I/home/rpm/rpmbuild/BUILD/include -I/home/rpm/rpmbuild/BUILD/common -I/home/rpm/rpmbuild/BUILD/libdnf5-cli/. -Wmaybe-uninitialized -fPIC -fmacro-prefix-map=/home/rpm/rpmbuild/BUILD/= -Wall -Wextra -Werror -Wcast-align -Wformat-nonliteral -Wmissing-format-attribute -Wredundant-decls -Wsign-compare -Wtype-limits -Wuninitialized -Wwrite-strings -Werror=unused-result -Wodr -Wconversion -Wl,--as-needed -std=gnu++2a -MD -MT libdnf5-cli/CMakeFiles/libdnf5-cli.dir/output/advisorylist.cpp.o -MF CMakeFiles/libdnf5-cli.dir/output/advisorylist.cpp.o.d -o CMakeFiles/libdnf5-cli.dir/output/advisorylist.cpp.o -c /home/rpm/rpmbuild/BUILD/libdnf5-cli/output/advisorylist.cpp

/home/rpm/rpmbuild/BUILD/libdnf5-cli/output/advisorylist.cpp: In function 'libscols_table* libdnf5::cli::output::create_advisorylist_table(std::string)':
/home/rpm/rpmbuild/BUILD/libdnf5-cli/output/advisorylist.cpp:33:9: error: 'isatty' was not declared in this scope
   33 |     if (isatty(1)) {
      |         ^~~~~~
```
We fix this internally with
```
[...]

PR #1219, opened 2024-01-31
build: Include <unistd.h> for isatty()

ppisar:
A compilation fails on some systems with:

    /home/rpm/rpmbuild/BUILD/libdnf5-cli/output/advisorylist.cpp: In function 'libscols_table* libdnf5::cli::output::create_advisorylist_table(std::string)':
    /home/rpm/rpmbuild/BUILD/libdnf5-cli/output/advisorylist.cpp:33:9: error: 'isatty' was not declared in this scope
       33 |     if (isatty(1)) {
	  |         ^~~~~~

The cause is a missing "#include <unistd.h>" in the compilation unit.

This bug was introduced in commit
d8d9c631faf259d0d4d88671fe2029749a081f66 (Add output formatting for advisory list and info into libdnf-cli).

Fixes: #1168

### DIFF
--- a/libdnf5-cli/output/advisorylist.cpp
+++ b/libdnf5-cli/output/advisorylist.cpp
@@ -22,6 +22,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "utils/string.hpp"
 
 #include <libsmartcols/libsmartcols.h>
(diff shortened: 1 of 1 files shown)


j-mracek:
LGTM

Issue #758, opened 2023-07-21
Aliases for explicit NEVRA matching: remove-n, remove-na, remove-nevra


j-mracek:
Document as dropped

j-mracek:
Changes are tracked in https://github.com/rpm-software-management/dnf5/blob/main/doc/changes.rst

Issue #759, opened 2023-07-21
Deprecated aliases: erase, erase-n, erase-na, erase-nevra


j-mracek:
Document as dropped

j-mracek:
Changes are documented in https://github.com/rpm-software-management/dnf5/blob/main/doc/changes.rst

pkratoch:
Already fixed in https://github.com/rpm-software-management/dnf5/pull/1414

PR #1222, opened 2024-02-01
Document explicit nevra remove commands and aliases dropped

j-mracek:


enmcarv:
Closes #758 and #759?

PR #1227, opened 2024-02-07
Add tmt tests identifiers

jan-kolarik:
To allow manual triggering of test suites on the GitHub PR with `/packit test –identifier <id>`.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -38,6 +38,7 @@ jobs:
       - fedora-all
   - job: tests
     trigger: pull_request
+    identifier: "dnf5-tests"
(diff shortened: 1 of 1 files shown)


jan-kolarik:
OK so now the each tests suite is mentioned separately in the checks list. The `dnf-tests` and `dnf5daemon-tests` are having the "brown dot" as they are meant to be triggered manually only. The only issue left is that when the `dnf5-tests` are executed, the dnf5daemon tests are also included there.

jan-kolarik:
tmt tests looks fixed now. Ready for review 👀

PR #1228, opened 2024-02-07
Silence clang error -Wunneeded-internal-declaration

jan-kolarik:
Temporarily until the changes from https://github.com/rpm-software-management/dnf5/issues/1091 are introduced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,7 +85,15 @@ add_compile_options(-Werror=unused-result -Wodr)
 add_compile_options(-Wconversion)
 
 # linker options
-if(NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    # TODO(jkolarik): Remove when the needed ABI breaking change is merged (see https://github.com/rpm-software-management/dnf5/issues/1091)
+    # This block silence this error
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Follow up on https://github.com/rpm-software-management/dnf5/pull/1130. Locally the build using Clang 17 is passing, we'll see how it is here. We might just need to bump the host containers to F39?

jan-kolarik:
> > It's possible we might need to update the host containers to F39.
> 
> I believe the host container is fine, no? https://github.com/rpm-software-management/ci-dnf-stack/blob/main/.github/workflows/ci-host.yml#L32

Hmm, here it seems so, but in the package build logs I see "F38" and also older Clang 16. But still it might not be related to the failure...

inknos:
I see now. I'll rebuild the image.

https://github.com/rpm-software-management/ci-dnf-stack/pkgs/container/dnf-ci-host

image

inknos:
ok this is awkward... apparently the container is failing to build for months. https://github.com/rpm-software-management/ci-dnf-stack/actions/workflows/ci-host.yml

inknos:
[Rebuilding](https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/rpm-gitoverlay/builds/) rpm-gitoverlay against the correct python version should fix the container build which should also fix the correct fedora which might fix the clang pr. fingers crossed.

PR #1229, opened 2024-02-08
modules: Report module solver problems

pkratoch:


libdnf5/module/module_sack_impl.hpp
```diff
@@ -181,6 +184,7 @@ class ModuleSack::Impl {
     std::unique_ptr<libdnf5::solv::SolvMap> excludes;
     std::map<Id, ModuleItem *> active_modules;
     std::vector<libdnf5::solv::IdQueue> modules_to_enable;
+    std::vector<std::vector<std::tuple<ProblemRules, Id, Id, Id, std::string>>> solver_problems;
```
pkratoch, line 187: I am not sure about this. It would be better to return the problems from `ModuleSack::Impl::module_solve` and then from `ModuleSack::resolve_active_module_items`, but it's on the public API. Hence this workaround. But maybe we could change the return type if we are doing more changes that break the API.

include/libdnf5/base/goal_elements.hpp
```diff
@@ -82,7 +107,10 @@ enum class GoalProblem : uint32_t {
     SOLVER_PROBLEM_STRICT_RESOLVEMENT = (1 << 13),
     WRITE_DEBUG = (1 << 14),
     UNSUPPORTED_ACTION = (1 << 15),
-    MULTIPLE_STREAMS = (1 << 16)
+    MULTIPLE_STREAMS = (1 << 16),
+    MODULE_SOLVER_ERROR = (1 << 17),
```
j-mracek, line 111: This is not a problem but an idea, what about to add a doc strings for each item?

libdnf5/module/module_sack.cpp
```diff
@@ -555,6 +558,27 @@ std::pair<std::vector<std::vector<std::string>>, ModuleSack::ModuleErrorType> Mo
 }
 
 
+GoalProblem ModuleSack::Impl::module_solver_error_to_goal_problem(ModuleSack::ModuleErrorType module_error) {
```
j-mracek, line 561: Subject for a discussion - The last changes makes somehow `ModuleErrorType` redundant because GoalProblem has the same item set. Or did I missed something. Removal of ModuleErrorType would result in change of IP, but there is 5.2 tah introduce significant changes.

Other option would be that GoalProblem might inherit items from `ModuleErrorType`. This option might be a problem in future when module will get optional.

And the last option we can keep it like it is.

libdnf5/module/module_sack_impl.hpp
j-mracek: I am so sorry, but somehow I missed some connection in the comment.

libdnf5/base/goal.cpp
```diff
@@ -2313,15 +2314,33 @@ base::Transaction Goal::resolve() {
...
     ret |= p_impl->add_module_specs_to_goal(transaction);
     // Resolve modules
     auto module_error = module_sack.resolve_active_module_items().second;
-    if (module_error != libdnf5::module::ModuleSack::ModuleErrorType::NO_ERROR) {
-        throw module::ModuleResolveError(M_("Failed to resolve modules."));
+
+    // Report problems from the module solver
+    auto module_solver_problems =
+        base::process_module_solver_problems(module_sack.p_impl->pool, module_sack.p_impl->solver_problems);
+    auto module_goal_problem = module_sack.p_impl->module_solver_error_to_goal_problem(module_error);
```
j-mracek, line 2325: This converter is somehow additional step, but removal  of it would result in modification of enums, changes on API. From my point of view it would require much more energy then benefits.

libdnf5/base/solver_problems.cpp
```diff
@@ -63,7 +67,31 @@ static const std::map<ProblemRules, BgettextMessage> PKG_PROBLEMS_DICT = {
         " the following protected packages: {}")},
     {ProblemRules::RULE_PKG_REMOVAL_OF_RUNNING_KERNEL,
      M_("The operation would result in removing"
-        " of running kernel: {}")}};
+        " of running kernel: {}")},
+    {ProblemRules::RULE_MODULE_DISTUPGRADE, M_("{} does not belong to a distupgrade repository")},
```
j-mracek, line 71: Just a question - what about information about repository?

pkratoch: The changes do not prevent to add it in the future, so I though I would leave it for another PR to not make this one too big. But the disadvantage is that the strings for translations will change. I can take a look.

PR #1236, opened 2024-02-12
dnfdaemon: Limit number of simultaneously active sessions

m-blaha:


dnf5daemon-server/session_manager.cpp
```diff
@@ -26,11 +26,15 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
 
 #include <iostream>
+#include <numeric>
 #include <random>
 #include <sstream>
 #include <string>
 #include <thread>
 
+// TODO(mblaha): Make this constant configurable
+const int MAX_SESSIONS = 10;
```
Conan-Kudo, line 36: This should be lower. It is probably not reasonable to have more than 5. Most things don't even allow more than 2.

Conan-Kudo (changes requested):

dnf5daemon-server/session_manager.cpp
m-blaha: OK, makes sense. I originally had 5 here...

Conan-Kudo (approved):

jan-kolarik:
Thanks!

Issue #1189, opened 2024-01-20
dnfdaemon Missing end transaction signal


anaselli:
As i noticed and said on bug #1188 I cannot find a way to understand that the transaction is finished.

The thing is can seen also during an update:

```
2024-01-20 19:15:51,874 [dnfdaemon.client]{dnfd_client.py:725}(DEBUG) on_TransactionVerifyStart ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.UInt64(2)))
2024-01-20 19:15:51,875 [dnfdaemon.client]{dnfd_client.py:739}(DEBUG) on_TransactionVerifyProgress ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.UInt64(0), dbus.UInt64(2)))
2024-01-20 19:15:51,875 [dnfdaemon.client]{dnfd_client.py:739}(DEBUG) on_TransactionVerifyProgress ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.UInt64(1), dbus.UInt64(2)))
2024-01-20 19:15:51,877 [dnfdaemon.client]{dnfd_client.py:753}(DEBUG) on_TransactionVerifyStop ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.UInt64(2)))
2024-01-20 19:15:52,439 [dnfdaemon.client]{dnfd_client.py:631}(DEBUG) on_TransactionActionStart ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.String('x11-server-common-0:21.1.8-7.2.mga9.x86_64'), dbus.Int32(2), dbus.UInt64(1278280)))
2024-01-20 19:15:52,439 [dnfdaemon.client]{dnfd_client.py:648}(DEBUG) on_TransactionActionProgress ((dbus.String('/org/rpm/dnf/v0/dfd41060d4de750ea084c15dff75e1d5'), dbus.String('x11-server-common-0:21.1.8-7.2.mga9.x86_64'), dbus.UInt64(0), dbus.UInt64(1278280)))
[...]

PR #1250, opened 2024-02-16
rpm: New callbacks to wrap whole rpm transaction

m-blaha:
There is currently no way for dnf5daemon user to find out when the rpm
transaction began and when it finished.
The change introduces two new callbacks on rpm::TransactionCallbacks class:

TransactionCallbacks::before_begin(int total)
- executed right before the rpm transaction is run
- parameter total contains number of elements in the rpm transaction.

TransactionCallbacks::after_complete(bool success)
- executed right after the rpm transaction finished
- parameter success indicates whether the transaction finished successfully.

These callbacks are then used in dnf5daemon-server to send respective signals
to the client.

Resolves https://github.com/rpm-software-management/dnf5/issues/1189

### DIFF
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
@@ -189,6 +189,26 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
     </method>
 
     <!--
+        transaction_before_begin:
+        @total: number of elements in the rpm transaction
+
+        Send right before the rpm transaction is run.
+    -->
+    <signal name="transaction_before_begin">
+        <arg name="total" type="t" />
+    </signal>
+
+    <!--
+        transaction_after_complete:
+        @success: true if the rpm transaction was completed successfully
+
+        Send right after the rpm transaction run finished.
+    -->
+    <signal name="transaction_after_complete">
+        <arg name="success" type="b" />
+    </signal>
+
+    <!--
         transaction_action_start:
         @nevra: full NEVRA of the package
         @action: one of the dnfdaemon::RpmTransactionItem::Actions enum
--- a/include/libdnf5/rpm/transaction_callbacks.hpp
+++ b/include/libdnf5/rpm/transaction_callbacks.hpp
@@ -69,6 +69,13 @@ public:
 
     virtual ~TransactionCallbacks() = default;
 
+    /// Called right before the rpm transaction is run
+    /// @param total Number of elements in the rpm transaction
+    virtual void before_begin(uint64_t total) {}
+    /// Called after the transaction run finished
+    /// @param success Whether the rpm transaction was completed successfully
+    virtual void after_complete(bool success) {}
+
     virtual void install_progress(const TransactionItem & item, uint64_t amount, uint64_t total) {}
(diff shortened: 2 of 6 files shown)


ppisar:
The code looks good, but then new signals are missing a documentation in dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml. Please add the documentation.

ppisar:
The three failing scenarios about installonly packages are not related (a change in libsolv).

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -188,6 +188,26 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
     </method>
 
+    <!--
+        transaction_before_begin:
+        @total: number of elements in the rpm transaction
+
+        Send right before the rpm transaction is run.
+    -->
+    <signal name="transaction_before_begin">
+        <arg name="total" type="i" />
```
ppisar, line 198: I'd like to see there `type="t"` (unsigned 64-bit integer) rather than `"i"` (signed 32-bit integer). I know that `"i"` directly reflects what `rpmtsNElements()` returns on x86_64. However:

* We already use "t" in `transaction_action_start` and `transaction_action_progress` signals for counting transactions. Mixing signed and unsigned values brings a risk for programs (e.g. in C comparison between the two is undefined.)
* Cautious clients would have to handle negative values. What would a negative value mean? I guess people will use that value for rendering progress bars.
* There could be platforms where C `int` is not 32-bit. 64-bit integer buys us some time.

Could you please consider changing it to `t` type?

m-blaha: Yeah, I was thinking about it and eventually went with what rpmtsNElements. But I agree with your signed/unsigned reasoning. I will change it.

m-blaha:
@ppisar Done. Sorry for the rebase, I didn't want to add more noise to git history using new commits.

m-blaha:
Oh, int remained mentioned in the commit message...

ppisar:
Thew new RPM build failure in F41 ("Directory not found: /builddir/build/BUILDROOT/dnf5-5.1.12-1.20240216152113944400.pr1250.22.gd9fb3e11.fc41.x86_64/usr/share/bash-completion") in unrelated.

include/libdnf5/rpm/transaction_callbacks.hpp
```diff
@@ -69,6 +69,13 @@ class TransactionCallbacks {
 
     virtual ~TransactionCallbacks() = default;
 
+    /// Called right before the rpm transaction is run
+    /// @param total Number of elements in the rpm transaction
+    virtual void before_begin(uint64_t total) {}
```
ppisar, line 74: /usr/include/libdnf5/rpm/transaction_callbacks.hpp is a public header. Adding virtual methods breaks its ABI. Can this feature be implemented without virtual methods, or in a private class? Otherwise, we need to postpone this pull request until a window for ABI change opens.

m-blaha: Ech. OK then, moving the PR to 5.2.

m-blaha:
The change breaks ABI, moving to 5.2.

PR #1253, opened 2024-02-19
rpm: New callbacks to wrap whole rpm transaction

m-blaha:
There is currently no way for dnf5daemon user to find out when the rpm
transaction began and when it finished.
The change introduces two new callbacks on rpm::TransactionCallbacks class:

TransactionCallbacks::before_begin(int total)

    executed right before the rpm transaction is run
    parameter total contains number of elements in the rpm transaction.

TransactionCallbacks::after_complete(bool success)

    executed right after the rpm transaction finished
    parameter success indicates whether the transaction finished successfully.

These callbacks are then used in dnf5daemon-server to send respective signals
to the client.

For https://github.com/rpm-software-management/dnf5/issues/1189

Replaces https://github.com/rpm-software-management/dnf5/pull/1250
Originally opened on the main branch, but since it causes ABI breakage by adding new virtual methods to public class, moved here.

### DIFF
--- a/dnf5daemon-server/callbacks.cpp
+++ b/dnf5daemon-server/callbacks.cpp
@@ -125,6 +125,24 @@ sdbus::Signal DbusTransactionCB::create_signal_pkg(
 }
 
 
+void DbusTransactionCB::before_begin(uint64_t total) {
+    try {
+        auto signal = create_signal(dnfdaemon::INTERFACE_RPM, dnfdaemon::SIGNAL_TRANSACTION_BEFORE_BEGIN);
+        signal << total;
+        dbus_object->emitSignal(signal);
+    } catch (...) {
+    }
+}
+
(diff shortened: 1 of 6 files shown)


ppisar (approved):
Fine. ABI break confirmed.

ppisar:
What's the procedure with PRs against non-main? Do you want to merge it into dnf5-5.2.0.0 branch now? Be ware that merging this PR could close the referred issue.

ppisar (approved):
The code is good.
createrepo_c failure is unrelated.
dnf5daemon-tests are stuck somewhere in the infrastructure. Probably unrelated.
dnf5-tests passed.

ppisar:
I recommend you rerunning the tests to see results for dnf5daemon-tests before merging it.

PR #1256, opened 2024-02-20
Typo in dnf5.spec

inknos:
required for https://github.com/rpm-software-management/dnf5/actions/runs/7973173112


ppisar:
Goddammit. I tested my patch on Fedora 41 and Fedora 40. But I did not on RHEL.

inknos:
packit got it. we might be able to run some pre-checks to get such errors

PR #1261, opened 2024-02-20
When limiting installonly pkg handle each provider just once

kontura:
First gather all unique providers of installonly provides and then
handle each set of packages (providers) with the same name.

Previously the handling was done per provide, this could result in repeated
handling of a single package if it provided multiple installonly
provides. Such as `kernel` and `installonlypkg(kernel)`.
This could also lead to problems when multiple versions of installonly
pkgs didn't have the same installonly provides.

https://bugzilla.redhat.com/show_bug.cgi?id=2263675

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1454

### DIFF
--- a/libdnf5/rpm/solv/goal_private.cpp
+++ b/libdnf5/rpm/solv/goal_private.cpp
@@ -128,7 +128,11 @@ bool can_depend_on(Pool * pool, Solvable * sa, Id b) {
     return false;
 }
 
-static void same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue * out) {
+
+/// @brief It takes an `in` Queue which contains packages sorted by name (see `installonly_cmp`).
+///        It selects the first name and moves all packages with that name to the `out` Queue.
+///        Additionally libsolv Id of the selected name is returned.
+static Id same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue * out) {
     Id el = queue_pop(in);
     Id name = pool.id2solvable(el)->name;
     queue_empty(out);
@@ -136,6 +140,7 @@ static void same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue 
     while (in->count && pool.id2solvable(in->elements[in->count - 1])->name == name)
         // reverses the order so packages are sorted by descending version
         queue_push(out, queue_pop(in));
+    return name;
 }
 
 
@@ -228,6 +233,10 @@ inline bool name_solvable_cmp_key(const Solvable * first, const Solvable * secon
     return first->name < second->name;
 }
 
+inline bool name_solvable_id_name_cmp_key(const Solvable * first, const Id second) {
+    return first->name < second;
+}
+
 }  // namespace
 
 
@@ -291,29 +300,29 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
 
         libdnf5::solv::IdQueue same_names;
         while (q.size() > 0) {
-            same_name_subqueue(spool, &q.get_queue(), &same_names.get_queue());
+            Id name = same_name_subqueue(spool, &q.get_queue(), &same_names.get_queue());
             if (same_names.size() <= static_cast<int>(installonly_limit)) {
                 continue;
             }
+
+            // We want to avoid reinstalling packages marked for ERASE, therefore
+            // if some unused provider is also available we need to mark it ERASE as well.
+            auto low = std::lower_bound(
+                available_unused_providers.begin(),
+                available_unused_providers.end(),
+                name,
+                name_solvable_id_name_cmp_key);
+            while (low != available_unused_providers.end() && (*low)->name == name) {
+                job.push_back(SOLVER_ERASE | SOLVER_SOLVABLE, spool.solvable2id(*low));
+                ++low;
+            }
+
             reresolve = true;
             for (int j = 0; j < same_names.size(); ++j) {
                 Id id = same_names[j];
                 Id action = SOLVER_ERASE;
                 if (j < static_cast<int>(installonly_limit)) {
                     action = SOLVER_INSTALL;
-                } else {
-                    // We want to avoid reinstalling packages marked for ERASE, therefore
-                    // if some unused provider is also available we need to mark it ERASE as well.
-                    Solvable * solvable = spool.id2solvable(id);
-                    auto low = std::lower_bound(
-                        available_unused_providers.begin(),
-                        available_unused_providers.end(),
-                        solvable,
-                        nevra_solvable_cmp_key);
-                    while (low != available_unused_providers.end() && (*low)->name == solvable->name) {
-                        job.push_back(SOLVER_ERASE | SOLVER_SOLVABLE, spool.solvable2id(*low));
-                        ++low;
-                    }
                 }
                 job.push_back(action | SOLVER_SOLVABLE, id);
             }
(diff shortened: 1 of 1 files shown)


libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -262,61 +268,55 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
...
             // conflicts = dependencies cannot be met).
             if (libsolv_solver.get_decisionlevel(p) > 0) {
-                q.push_back(p);
+                marked_for_install_providers_map.add_unsafe(p);
             } else {
-                Solvable * solvable = spool.id2solvable(p);
-                if (!spool.is_installed(solvable)) {
-                    available_unused_providers.push_back(solvable);
-                }
+                available_unused_providers_map.add_unsafe(p);
```
j-mracek, line 273: I see that the test (`spool.is_installed(solvable)`) was removed. From the description of `libsolv_solver.get_decisionlevel(p)` it looks like a redundant, but I am not sure.

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -243,16 +248,17 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
     ::Pool * pool = *spool;
     bool reresolve = false;
 
+    // Use maps to handle each package just once
+    libdnf5::solv::SolvMap marked_for_install_providers_map(spool.get_nsolvables());
+    libdnf5::solv::SolvMap available_unused_providers_map(spool.get_nsolvables());
+
     for (int i = 0; i < installonly.size(); ++i) {
```
j-mracek, line 255: Suggestion: I believe that it would be possible to iterate over using  `for (Id install_only_provide : installonly) {`
It should be more clear in macro `FOR_PROVIDES(p, pp, installonly[i]) {` what is used.

libdnf5/rpm/solv/goal_private.cpp
j-mracek: Oh, I see the test was redundant and it is only performed later and once. Good job.

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -262,61 +268,55 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
...
         }
 
-        const InstallonlyCmpData installonly_cmp_data{spool, running_kernel};
-        q.sort(&installonly_cmp, &installonly_cmp_data);
-        std::sort(available_unused_providers.begin(), available_unused_providers.end(), name_solvable_cmp_key);
+        // We want to avoid reinstalling packages marked for ERASE, therefore
+        // if some unused provider is also available we need to mark it ERASE as well.
+        auto low = std::lower_bound(
+            available_unused_providers.begin(), available_unused_providers.end(), name, name_solvable_id_name_cmp_key);
+        while (low != available_unused_providers.end() && (*low)->name == name) {
```
j-mracek, line 306: I believe that there is a long standing problem, because we always calculate the order according to name, but may be we should add architecture. Think about `x86_64` and `i686`. Do you think whether it would be worth to add an architecture in account?

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -262,61 +268,55 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
...
-                        solvable,
-                        nevra_solvable_cmp_key);
-                    while (low != available_unused_providers.end() && (*low)->name == solvable->name) {
-                        job.push_back(SOLVER_ERASE | SOLVER_SOLVABLE, spool.solvable2id(*low));
-                        ++low;
-                    }
-                }
-                job.push_back(action | SOLVER_SOLVABLE, id);
+        reresolve = true;
+        for (int j = 0; j < same_names.size(); ++j) {
```
j-mracek, line 312: I believe that iterator might be modify to not use indexes. What do you think?

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -128,14 +128,15 @@ bool can_depend_on(Pool * pool, Solvable * sa, Id b) {
     return false;
 }
 
-static void same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue * out) {
+static Id same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue * out) {
```
j-mracek, line 131: It would be good to add a doc string.

libdnf5/rpm/solv/goal_private.cpp
j-mracek: This might sound strange, but handling of `available_unused_providers` by name is OK (or more secure), but handling of `marked_for_install_providers` by only name might be an issue.

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -262,61 +268,55 @@ bool GoalPrivate::limit_installonly_packages(libdnf5::solv::IdQueue & job, Id ru
...
         }
-        if (!installing.size()) {
+    }
+
+    const InstallonlyCmpData installonly_cmp_data{spool, running_kernel};
+    marked_for_install_providers.sort(&installonly_cmp, &installonly_cmp_data);
+    std::sort(available_unused_providers.begin(), available_unused_providers.end(), name_solvable_cmp_key);
+
+    libdnf5::solv::IdQueue same_names;
+    while (!marked_for_install_providers.empty()) {
```
j-mracek, line 296: It would be great to add a comment what while loop does.

libdnf5/rpm/solv/goal_private.cpp
kontura: The index `j` is also used as a counter for the `installonly_limit`. I find this approach simpler than having a separate counter variable so I would prefer to keep this.

libdnf5/rpm/solv/goal_private.cpp
kontura: Added.

libdnf5/rpm/solv/goal_private.cpp
kontura: I think that would be another change in behavior so I would rather leave that for a different PR.

But I would be interested to know what kind of issues it might cause, perhaps a new github issue to track that would be nice. :+1:

libdnf5/rpm/solv/goal_private.cpp
j-mracek: Good point that I overlooked. Please keep it like it is.

libdnf5/rpm/solv/goal_private.cpp
```diff
@@ -128,14 +127,19 @@ bool can_depend_on(Pool * pool, Solvable * sa, Id b) {
     return false;
 }
 
-static void same_name_subqueue(libdnf5::solv::RpmPool & pool, Queue * in, Queue * out) {
+
+/// @brief It takes an `in` Queue which contains packages sorted by name (see `installonly_cmp`).
+///        It selects the first name and upon return the `out` Queue is populated with packages
```
j-mracek, line 132: May be I do not understand it correctly. I don't see that it states that it moves all packages (Ids) with the name that is returned from `in` to `out` Queue. Or did I overlooked something?

kontura: Hopefully its better now.

j-mracek (approved):
Perfect, ready to merge after all test finish.

Issue #1224, opened 2024-02-01
Running transactions CLI (storing transactions, replaying transactions, offline transactions)


kontura:
In old [dnf](https://github.com/rpm-software-management/dnf) there was `offline-upgrade`, `history store` and `history replay`.
In dnf5 we would like to extend this functionality and make it more general, namely any transaction could be stored (not just history transactions) and replayed even offline.

The question is how should the CLI look like.

Proposed ideas:
- Any transaction could be stored using a `--store <path-to-file>` option.
- For running offline transactions a convenient `--offline`/`--postponed` option could be used. This would store the transaction internally and it would be automatically used after `dnf5 offline reboot` or perhaps the reboot would be done immediately?
- For replaying transactions stored to a file we could use something like `dnf5 do` or `dnf5 run` (Possibly related: https://github.com/rpm-software-management/dnf5/issues/1223). This command could also serve as a replacement for `dnf shell` as it would allow multiple operations per transaction. Alternatively simple `dnf5 replay <path-to-file>` could replay just a single file.

Any other ideas are welcome.

Related issues: https://github.com/rpm-software-management/dnf5/issues/258 https://github.com/rpm-software-management/dnf5/issues/999

Conan-Kudo:
This is partly what the [txnupd libdnf plugin](https://code.opensuse.org/microos/libdnf-plugin-txnupd) did. It'd be nice to have more of that functionality integrated right into dnf5 so that the only thing that needs to be worked out in a plugin is the btrfs snapshot and retargeting part.

evan-goode:
> Any transaction could be stored using a --store <path-to-file> option.

+1 

> For running offline transactions a convenient --offline/--postponed option could be used. This would store the transaction internally and it would be automatically used after dnf5 offline reboot or perhaps the reboot would be done immediately?

I like `--offline` better; `--postpone`/`--postponed` is a little unclear to me. "Offline updates" is what they are called by [systemd](https://www.freedesktop.org/software/systemd/man/latest/systemd.offline-updates.html).

I think it's also best to split the reboot step to a different command, like how system-upgrade currently works, in case the user doesn't want to update immediately. `--offline` should be roughly equivalent to setting `--store /usr/lib/sysimage/libdnf5/offline/offline-transaction.json` and `--setopt=cachedir=/usr/lib/sysimage/libdnf5/offline`. Any necessary RPMs would get downloaded to that directory and the transaction would be run with the `test` tsflag.

[...]

PR #1221, opened 2024-02-01
Parsing serialized transaction: format validation

kontura:
Parsing serialized transaction: format validation.
I will also provide CI tests once have know how the CLI should look like for commands that will use this API.

For: https://github.com/rpm-software-management/dnf5/issues/999

PR #1264, opened 2024-02-21
Replay3

kontura:
- This contains several commits from: https://github.com/rpm-software-management/dnf5/pull/1221 - these can be merged separately and are not directly related
- Adds comps paths to transaction_sr and goal
- Add support for storing a full transaction into a directory: `transaction::serialize()` now takes an optional path argument, if provided full transaction is stored there in such a way that it can be replayed without loading repositories. It stores packages and required comps metadata.
- Adds global `--store <path>` option
- Add temporary `history replay` command just for testing since the CLI is not clear (https://github.com/rpm-software-management/dnf5/issues/1224)


dnf5/main.cpp
```diff
@@ -616,6 +616,23 @@ void RootCommand::set_argument_parser() {
...
+        store->set_description("Store the current transaction.");
+        store->set_parse_hook_func([&ctx](
+                                       [[maybe_unused]] ArgumentParser::NamedArg * arg,
+                                       [[maybe_unused]] const char * option,
+                                       const char * value) {
+            libdnf5::OptionString transaction_store_path(value);
+            ctx.transaction_store_path = transaction_store_path.get_value_string();
+            return true;
+        });
+        global_options_group->register_argument(store);
```
evan-goode, line 633: Should `--store` be a global argument? It may be better to implement it in `shared_options.cpp` and only register it on the commands that create transactions (`install`, `upgrade`, ... but not `repoquery`, `list`).

kontura:
Rebased.

evan-goode (changes requested):
This could be a separate PR, but I would prefer if the RPM paths in the transaction.json were relative to the destdir (in the case of `--store`, this would be relative to the stored transaction path), rather than being absolute paths. If I store a transaction to `/tmp/store-a` and then move it to `/tmp/store-b`, I would expect it to still be runnable.

dnf5/shared_options.cpp
```diff
@@ -73,5 +73,23 @@ void create_downloadonly_option(dnf5::Command & command) {
...
+    store->set_long_name("store");
+    store->set_has_value(true);
+    store->set_description("Store the current transaction instead of running it.");
+    store->set_arg_value_help("STORED_TRANSACTION_PATH");
+    command.get_argument_parser_command()->register_named_arg(store);
+    store->set_parse_hook_func([&ctx](
+                                   [[maybe_unused]] libdnf5::cli::ArgumentParser::NamedArg * arg,
+                                   [[maybe_unused]] const char * option,
+                                   const char * value) {
+        libdnf5::OptionString transaction_store_path(value);
```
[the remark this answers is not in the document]
kontura: Oh, that is not needed, I will drop that, thanks!

kontura:
> This could be a separate PR, but I would prefer if the RPM paths in the transaction.json were relative to the destdir (in the case of `--store`, this would be relative to the stored transaction path), rather than being absolute paths. If I store a transaction to `/tmp/store-a` and then move it to `/tmp/store-b`, I would expect it to still be runnable.

The paths are now taken as relative. I had to change the `Goal::add_serialized_transaction()` API because it needs to know where the serialized transaction is but since it is marked as experimental I think it is fine. Especially if we make it into the 5.2.0.0 release.

libdnf5/base/goal.cpp
```diff
@@ -625,10 +627,15 @@ GoalProblem Goal::Impl::add_transaction_replay_specs_to_goal(base::Transaction &
     if (!serialized_transaction) {
         return GoalProblem::NO_PROBLEM;
     }
-    libdnf5::GoalJobSettings settings = serialized_transaction->second;
+
+    auto & [replay_path, settings] = *serialized_transaction;
+    utils::fs::File replay_file(replay_path, "r");
+    auto replay_location = replay_path.remove_filename();
+    auto replay = transaction::parse_transaction_replay(replay_file.read());
```
evan-goode, line 634: It's a bit confusing that `--store` takes a directory path but `history replay` needs a path to the `transaction.json`, which must be in a directory along with the `packages` and `comps`.

Currently, if you pass a directory to `history replay` it throws:

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```

IMO `history replay` should only accept a directory, unless we really want to support bare `transaction.json`s.

dnf5/context.cpp
```diff
@@ -404,7 +406,22 @@ void Context::download_and_run(libdnf5::base::Transaction & transaction) {
...
+        constexpr const char * packages_in_trans_dir{"./packages"};
+        constexpr const char * comps_in_trans_dir{"./comps"};
+        auto & destdir_opt = base.get_config().get_destdir_option();
+        destdir_opt.set(transaction_store_path / packages_in_trans_dir);
+        std::filesystem::create_directories(transaction_store_path);
+        transaction.download();
+        std::filesystem::path comps_path = transaction_store_path / comps_in_trans_dir;
+        transaction.store_comps(comps_path);
+        libdnf5::utils::fs::File transfile(transaction_store_path / "transaction.json", "w");
+        transfile.write(transaction.serialize(packages_in_trans_dir, comps_in_trans_dir));
```
evan-goode, line 420: Maybe we should prompt for user confirmation if this would overwrite an existing `transaction.json`.

evan-goode (changes requested):

evan-goode:
> The paths are now taken as relative. I had to change the `Goal::add_serialized_transaction()` API because it needs to know where the serialized transaction is but since it is marked as experimental I think it is fine. Especially if we make it into the 5.2.0.0 release.

Nice, makes sense.

> I would also like to update the `offline` transaction running to use the local paths in the serialized transaction but since this PR is already quite big we can do that separately.

Sounds reasonable, let's do it in a separate PR.

libdnf5/base/goal.cpp
kontura: I agee, but actually I didn't want to include the `history replay` commit :facepalm:, we want to move/rename the command. It was just for testing, I will drop it for now.

dnf5/context.cpp
kontura: That is a very good idea, thank you.

libdnf5/base/goal.cpp
evan-goode: OK gotcha. Another thing I noticed is that the RPMs in the stored transaction get deleted during replay, which may be undesired. We'll figure it out later.

evan-goode:
Everything LGTM now, merging. Thanks!

Issue #771, opened 2023-07-25
dnf5daemon: Add functions to enable/disable repos in org.rpm.dnf.v0.rpm.Repo


mcrha:
In order to be feature complete with PackageKit (on the gnome-software side), it would be good to add functions to enable and disable repositories into the org.rpm.dnf.v0.rpm.Repo D-Bus interface.

jan-kolarik:
More details in https://bugzilla.redhat.com/show_bug.cgi?id=2124496.

mcrha:
Maybe I do something wrong. I've compiled dnf5 at commit 440469681c8ec06884dcf8dbd73c2ab9105eeed1 and I run it (started as a root from a command line `$PREFIX/sbin/dnf5daemon-server --verbose` (the `--verbose` doesn't seem to do anything)), and I see it being used based on the `DEBUG` lines on the dnf5daemon-server terminal.

The response from enable/disable repo call is an error: `Action org.rpm.dnf.v0.rpm.Repo.conf_write is not registered`.

When I try to install an RPM app I'm asked for a wheel user password and the app is installed, from which i suppose my way of running the server is not that wrong. I've not been asked for the wheel user password when doing enable/disable of the repo.

I know I'm running the built version of the dnf5 server, because when the system server is running the error is different: `Unknown method disable or interface org.rpm.dnf.v0.rpm.Repo`, which makes sense, the method had been just added.

ppisar:
Does your $PREFIX/sbin/dnf5daemon-server link to your new libdnf5.so.1?

I tried enabling a repository with "dnf5daemon-client repo enable f41-build" command with DNF5 built from commit c60cd47e5920fe22d06b0e70e96c6d3fa354be72 (later than yours) and it worked for me fine. You can try the nightly builds yourself from <https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/dnf-nightly/fedora-rawhide-x86_64/>.

m-blaha:
I think the problem is that the PR (specifically 57f8f3a61ffc8de47602145bf10dae2cfb57c36d) renames polkit policy org.rpm.dnf.v0.rpm.RepoConf.write -> org.rpm.dnf.v0.rpm.Repo.conf_write.

mcrha:
> Does your $PREFIX/sbin/dnf5daemon-server link to your new libdnf5.so.1?

Yes, it does. Though I see it depends how the terminal is set up, because `ldd` can pick the system lib as well.

[...]

PR #1266, opened 2024-02-21
dnf5daemon: Introduce enabling repositories in Repo interface

jan-kolarik:
Move the existing functionality of enabling and disabling of repositories from the `RepoConf` interface.

Drop the `RepoConf` interface as the remaining list functionality that is required by clients is already provided by the `Repo::list`.

Change the logic of commands to be strict against invalid arguments.

Add support for introduced commands in the dnf5daemon client.

Closes: #771.


m-blaha:
There is a small difference between what `RepoConf.list()` and `Repo.list()` do. The RepoConf interface returns contents of *.repo config files as they are - without any substitutions. OTOH Repo interface returns repositories as they are used - with variables substituted.

So from my point of view unifying all repo methods in one interface (Repo) makes sense. I'm not sure whether we need (have a user for) also RepoConf.list() functionality. But it can also be implemented in Repo interface with new names (like Repo.list_config()). Also we should unify attribute names used in Repo.list() and Repo.list_config(). Currently the former use `patterns` to filter repositories, the latter `ids`.

IIUC the original RepoConf interface was not documented so we should be able to drop it.

m-blaha (approved):

PR #1270, opened 2024-02-23
Replace `std::format` with `fmt::format`

### LINKED ISSUE #1269: `std::format` is unavailable in GCC < 13 (breaks Mageia builds)
Mageia 9 and Cauldron are at GCC 12 currently, and 2427a336a952f633307f4776a875dab06b10b2c6 introduced usage of `std::format`. Prior to inclusion in GCC 13, it was [available through Boost](https://www.boost.org/doc/libs/1_84_0/libs/format/doc/format.html).

If the minimum GCC version is raised, then this becomes unbuildable on Mageia. Alternatively, either the usage of `std::format` is removed or fall back to `boost::format` when `std::format` is not available.

Conan-Kudo:
Fixes #1269

### DIFF
--- a/dnf5-plugins/automatic_plugin/email_message.cpp
+++ b/dnf5-plugins/automatic_plugin/email_message.cpp
@@ -19,6 +19,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "email_message.hpp"
 
+#include <fmt/chrono.h>
 #include <libdnf5/utils/format.hpp>
 
 #include <chrono>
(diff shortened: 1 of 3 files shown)


ppisar:
Thanks. Could you merge the two top commits which remove std::format into a single commit?

Conan-Kudo:
> Thanks. Could you merge the two top commits which remove std::format into a single commit?

As they're in different parts of the codebase, do you think it's a good idea to do that? Eventually we're going to have to split up this monorepo, so I'm treating it as if I'm touching different repos here.

Conan-Kudo:
No, I would expect that we'd use filter-branch to split everything up. At least, that's what I'd do.

ppisar (approved):

Issue #1259, opened 2024-02-20
"dnf5 list" searches case-sensitively while dnf4 does not


ilikelinux69:
Example:

`dnf5 list '*exif*'` does not display `perl-Image-ExifTool.noarch`

Is this intentional, as in dnf4 it gets listed?

I just think we should have a consistent output for same search schemes in all options like dnf search, dnf repolist etc. 

See discussion:
https://discussion.fedoraproject.org/t/dnf-list-howto-display-case-insensitive-packages/105730/4

ppisar:
I would say it's on purpose. "dnf5 list" searches in package names and package names are case sensitive. It's a glob the same as shell glob for file names is also case sensitive. See dnf5-specs(5) manual page. Case insensitive search is provided in "dnf5 search" command which searches among more texts than the package name.

ppisar:
I discussed this issue with other DNF5 developers, and the resolution is that "dnf5 list" should behave insensitively.

PR #1273, opened 2024-02-26
Make info and list commane case insesitive

j-mracek:
It unifies behavior with repoquery command and DNF4

Closes: https://github.com/rpm-software-management/dnf5/issues/1259

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1460

### DIFF
--- a/dnf5/commands/list/list.cpp
+++ b/dnf5/commands/list/list.cpp
@@ -43,10 +43,10 @@ void ListCommand::set_argument_parser() {
     auto & parser = ctx.get_argument_parser();
 
     auto & cmd = *get_argument_parser_command();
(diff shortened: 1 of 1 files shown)


ppisar:
The code looks good. However, I'd like also see documenting the behavior so that users do not have to guess. If there is no manual page, either add one, or amend the usage output like this:

~~~~
$ dnf5 list --help
[...]
Arguments:                      
  specs                         List of keys to match case insensitively
~~~~

ppisar:
Great.
The createrepo_c test failure (createrepo_c/zchunk.feature:43  create repository with zck metadata with bad package) is unrelated.

ppisar:
I tested the code locally.

PR #1275, opened 2024-02-27
Add pImpl to `libdnf5::rpm::*`

kontura:
This PR is into a new dnf5-5.2.0.0 branch to keep the main branch ABI compatible.

Even though the PR contains the changes I am not sure if we should add the pImpls to:
 - `Changelog` - the struct is very simple
 - `Reldep` and `Package` - these are just libsolv `id`s, I find it unlikely that we would add some custom data that is not in libsolv. Additionally there is a lot of them so the overhead of the pImpls might cause some slow down, on the other hand I did some brief testing and wasn't able to detect any.


jan-kolarik (approved):
I am fine with it. Better to have more classes covered. And I don't expect any performance degradation in this manner.

Issue #1052, opened 2023-11-28
Implement system-upgrade and offline commands


evan-goode:
Part of #258.

Use transaction store/replay capabilities (https://github.com/rpm-software-management/dnf5/issues/999) to implement system-upgrade command.

- [x] `system-upgrade download`
  - Downloads everything needed to upgrade to a new major release. Serializes the upgrade transaction to a file to be replayed later by `system-upgrade upgrade`.
  - [x] `--no-downgrade`: do not install packages from the new release if they are older than what is currently installed. Mutually exclusive with `--distro-sync`.
  - [x] `--distro-sync`: always install packages from the new release, even if they are older than the currently-installed version. Mutually exclusive with `--no-downgrade`. This is the default behavior.
  - [x] `--downloaddir=<path>`: Redirect download of packages to provided `<path>`. By default, packages are downloaded into (per repository created) subdirectories of /var/lib/dnf/system-upgrade.
- [x] `system-upgrade clean`
  - Remove previously-downloaded data.
- [x] `system-upgrade reboot`
  - Could be an alias for a more generic offline transaction command.
  - `--poweroff`: system will power off after upgrades are completed, instead of restarting
- [x] `system-upgrade log`
- [x] `offline _execute`
  - If the upgrade succeeds, previously-downloaded data is removed
  - Not meant to be run directly by the user, so this command could be replaced with the more generic transaction replay command.
- [x] `offline log`
  - [x] `--number`: show the log specified by the number
- [x] `offline clean`
- [x] `offline reboot`

PR #1209, opened 2024-01-29
Implement system-upgrade plugin

evan-goode:
For https://github.com/rpm-software-management/dnf5/issues/1052.

The `system-upgrade` plugin is responsible for performing "offline" upgrades of the system. It has three main subcommands:

- `dnf5 system-upgrade download`: Does a dry run of an upgrade to a new `releasever` and stores the resulting transaction for later.
- `dnf5 system-upgrade reboot`: Creates a "magic symlink" at `/system-update` to trigger the start of `dnf5-system-upgrade.service` on the next boot, then reboots. See https://www.freedesktop.org/software/systemd/man/latest/systemd.offline-updates.html. `dnf5-system-upgrade.service` will take the place of the usual boot target so we can preform the upgrade without disturbing too many running processes.
- `dnf5 system-upgrade upgrade`: Not intended to be run by the user directly, but rather by `dnf5-system-upgrade.service`. Loads the transaction saved by `system-upgrade download` and performs the upgrade. Interacts with [Plymouth](https://www.freedesktop.org/wiki/Software/Plymouth/) (if it's available) to show upgrade progress on the animated boot screen. The usual DNF 5 stdout is additionally logged to journald. Reboots or powers off when the upgrade is complete.

 This implementation is, for the most part, a straightforward port of the DNF 4 system-upgrade plugin in dnf-plugins-core (formerly in dnf-plugins-extras). It uses the [DNF 5 transaction load/store API](https://github.com/rpm-software-management/dnf5/issues/999).

It's still a work in progress, but most of the functionality from DNF 4 system-upgrade is ported over and it's been successfully updating F38 to F39 without hiccups. Things left to do:

- [x] Plymouth output
- [ ] Doc page
- [x] `log` subcommand
- [ ] Go back through commit history of `dnf-plugins-extras` to make sure no historical issues with DNF 4 system-upgrade are present in the new implementation
- [ ] Tests: port `test_system_upgrade.py`

In my opinion, each of these (perhaps with the exception of the `log` subcommand) is a high-priority item, but they could be split off into future PRs if desired.

PR #1280, opened 2024-02-27
Implement system-upgrade and arbitrary offline transactions

evan-goode:
This PR supersedes https://github.com/rpm-software-management/dnf5/pull/1209.

Resolves https://github.com/rpm-software-management/dnf5/issues/1052.

Together with https://github.com/rpm-software-management/dnf5/pull/1264, resolves https://github.com/rpm-software-management/dnf5/issues/1224.

This patch implements `dnf5 system-upgrade` as a regular command rather than a plugin. `system-upgrade` is responsible for performing offline upgrades of the system to a new major release. It has three main subcommands:

- `dnf5 system-upgrade download --releasever XX`: Does a dry run of an upgrade to a new `releasever` and stores the resulting transaction for later.
- `dnf5 system-upgrade reboot`: Creates a "magic symlink" at `/system-update` to trigger the start of `dnf5-offline-transaction.service` on the next boot, then reboots. See https://www.freedesktop.org/software/systemd/man/latest/systemd.offline-updates.html. `dnf5-offline-transaction.service` will take the place of the usual boot target so we can preform the upgrade without disturbing too many running processes.
- `dnf5 system-upgrade _execute`: Not intended to be run by the user directly, but rather by `dnf5-offline-transaction.service`. Loads the transaction saved by `system-upgrade download` and performs the upgrade. Interacts with [Plymouth](https://www.freedesktop.org/wiki/Software/Plymouth/) (if it's available) to show upgrade progress on the animated boot screen. The usual DNF 5 stdout is additionally logged to journald. Reboots or powers off when the upgrade is complete.

This PR also adds a new command, `dnf5 offline`, which shares the `reboot`, `log`, and `clean` subcommands with `dnf5 system-upgrade`, and adds a new `--offline` argument that allows any operation to be performed offline, as proposed here: https://github.com/rpm-software-management/dnf5/issues/1224#issuecomment-1936651917. For example, the process for running an offline upgrade would look like:

```
$ sudo dnf5 upgrade --offline mypackage
Repositories loaded.
...
Transaction stored to be performed offline. Run `dnf5 offline reboot` to reboot and run the transaction.
$ sudo dnf5 offline reboot
The system will now reboot to perform the offline transaction initiated by the following command:
	dnf5 upgrade --offline mypackage
Is this ok [y/N]: 
```

The following commands support `--offline`: `autoremove`, `distro-sync`, `downgrade`, `install`, `group install`, `group upgrade`, `reinstall`, `remove`, `swap`, and `upgrade`.

A caveat of making `system-upgrade` and `offline` regular commands instead of plugins is that `libsystemd` and `sdbus-c++` now have to be build dependencies of `dnf5`, since `dnf5 offline log` uses these libraries to display logs from offline transactions.

Remaining tasks:

- [x] Add `dnf5 offline-distrosync` and `dnf5 offline-upgrade` aliases to `dnf5 distro-sync --offline` and `dnf5 upgrade --offline`, respectively
- [x] Add `systemd` build condition so we don't have to always depend on libsystemd+sdbus-c++ 
- [x] `dnf5 offline status` subcommand 
- [x] Documentation
- [x] Go back through commit history of `dnf-plugins-extras` to make sure no historical issues with DNF 4 system-upgrade are present in the new implementation

I'll implement the tests in a separate PR if that works. We have the DNF 4 tests to go from: https://github.com/rpm-software-management/dnf-plugins-core/blob/master/tests/test_system_upgrade.py

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -259,6 +259,7 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %dir %{_datadir}/dnf5/aliases.d
 %config %{_datadir}/dnf5/aliases.d/compatibility.conf
 %config %{_unitdir}/dnf5-offline-transaction.service
+%config %{_unitdir}/dnf5-offline-transaction-cleanup.service
 %dir %{_libdir}/dnf5
 %dir %{_libdir}/dnf5/plugins
 %dir %{_datadir}/dnf5/dnf5-plugins
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -84,6 +84,7 @@ Provides:       dnf5-command(system-upgrade)
 %else
 %bcond_without zchunk
 %endif
+%bcond_without systemd
 
 %bcond_with    html
 %if 0%{?rhel} == 8
@@ -136,9 +137,7 @@ BuildRequires:  pkgconfig(librepo) >= %{librepo_version}
 BuildRequires:  pkgconfig(libsolv) >= %{libsolv_version}
 BuildRequires:  pkgconfig(libsolvext) >= %{libsolv_version}
 BuildRequires:  pkgconfig(rpm) >= 4.17.0
-BuildRequires:  pkgconfig(sdbus-c++) >= 0.8.1
 BuildRequires:  pkgconfig(sqlite3) >= %{sqlite_version}
-BuildRequires:  systemd-devel
 BuildRequires:  toml11-static
 
 %if %{with clang}
@@ -163,6 +162,11 @@ BuildRequires:  pkgconfig(modulemd-2.0) >= %{libmodulemd_version}
 
 %if %{with zchunk}
 BuildRequires:  pkgconfig(zck) >= %{zchunk_version}
+%endif
+
+%if %{with systemd}
+BuildRequires:  pkgconfig(sdbus-c++) >= 0.8.1
+BuildRequires:  systemd-devel
 %endif
 
 %if %{with html} || %{with man}
@@ -750,6 +754,7 @@ automatically and regularly from systemd timers, cron jobs or similar.
     -DWITH_COMPS=%{?with_comps:ON}%{!?with_comps:OFF} \
     -DWITH_MODULEMD=%{?with_modulemd:ON}%{!?with_modulemd:OFF} \
     -DWITH_ZCHUNK=%{?with_zchunk:ON}%{!?with_zchunk:OFF} \
+    -DWITH_SYSTEMD=%{?with_systemd:ON}%{!?with_systemd:OFF} \
     \
     -DWITH_HTML=%{?with_html:ON}%{!?with_html:OFF} \
     -DWITH_MAN=%{?with_man:ON}%{!?with_man:OFF} \
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -262,8 +262,6 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %dir %{_datadir}/dnf5
 %dir %{_datadir}/dnf5/aliases.d
 %config %{_datadir}/dnf5/aliases.d/compatibility.conf
-%config %{_unitdir}/dnf5-offline-transaction.service
-%config %{_unitdir}/dnf5-offline-transaction-cleanup.service
 %dir %{_libdir}/dnf5
 %dir %{_libdir}/dnf5/plugins
 %dir %{_datadir}/dnf5/dnf5-plugins
@@ -316,6 +314,12 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %{_mandir}/man5/dnf5.conf.5.*
 %{_mandir}/man5/dnf5.conf-todo.5.*
 %{_mandir}/man5/dnf5.conf-deprecated.5.*
+
+%if %{with systemd}
+%{_unitdir}/dnf5-offline-transaction.service
+%{_unitdir}/dnf5-offline-transaction-cleanup.service
+%{_unitdir}/system-update.target.wants/dnf5-offline-transaction.service
+%endif
 
 # ========== libdnf5 ==========
 %package -n libdnf5
@@ -806,6 +810,13 @@ done
 # Remove if condition when Fedora 37 is EOL
 %if 0%{?fedora} > 37 || 0%{?rhel} > 10
 ln -sr %{buildroot}%{_bindir}/dnf5 %{buildroot}%{_bindir}/microdnf
+%endif
+
+%if %{with systemd}
+mkdir -p %{buildroot}%{_unitdir}/system-update.target.wants/
+pushd %{buildroot}%{_unitdir}/system-update.target.wants/
+  ln -sr ../dnf5-offline-transaction.service
+popd
 %endif
 
 %find_lang dnf5
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -167,6 +167,10 @@ BuildRequires:  pkgconfig(zck) >= %{zchunk_version}
 %if %{with systemd}
 BuildRequires:  pkgconfig(sdbus-c++) >= 0.8.1
 BuildRequires:  systemd-devel
+
+ # We need to get the SYSTEMD_SYSTEM_UNIT_DIR from
+ # /usr/share/pkgconfig/systemd.pc
+BuildRequires:  systemd
 %endif
 
 %if %{with html} || %{with man}
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -62,6 +62,8 @@ Provides:       dnf5-command(advisory)
 Provides:       dnf5-command(clean)
 Provides:       dnf5-command(download)
 Provides:       dnf5-command(makecache)
+Provides:       dnf5-command(offline)
+Provides:       dnf5-command(system-upgrade)
 
 
 # ========== build options ==========
@@ -134,7 +136,9 @@ BuildRequires:  pkgconfig(librepo) >= %{librepo_version}
 BuildRequires:  pkgconfig(libsolv) >= %{libsolv_version}
 BuildRequires:  pkgconfig(libsolvext) >= %{libsolv_version}
 BuildRequires:  pkgconfig(rpm) >= 4.17.0
+BuildRequires:  pkgconfig(sdbus-c++) >= 0.8.1
 BuildRequires:  pkgconfig(sqlite3) >= %{sqlite_version}
+BuildRequires:  systemd-devel
 BuildRequires:  toml11-static
 
 %if %{with clang}
@@ -254,6 +258,7 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %dir %{_datadir}/dnf5
 %dir %{_datadir}/dnf5/aliases.d
 %config %{_datadir}/dnf5/aliases.d/compatibility.conf
+%config %{_unitdir}/dnf5-offline-transaction.service
 %dir %{_libdir}/dnf5
 %dir %{_libdir}/dnf5/plugins
 %dir %{_datadir}/dnf5/dnf5-plugins
--- a/dnf5/CMakeLists.txt
+++ b/dnf5/CMakeLists.txt
@@ -2,7 +2,8 @@ if(NOT WITH_DNF5)
     return()
 endif()
 
-set(SYSTEMD_SYSTEM_UNIT_DIR /usr/lib/systemd/system)
+pkg_check_modules(SYSTEMD REQUIRED systemd)
+pkg_get_variable(SYSTEMD_SYSTEM_UNIT_DIR systemd systemdsystemunitdir)
 
 find_package(Threads)
 
@@ -56,11 +57,10 @@ install(FILES bash-completion/dnf5 DESTINATION ${BASH_COMPLETION_COMPLETIONSDIR}
 install(FILES "README.plugins" DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/dnf5/plugins" RENAME "README")
 install(DIRECTORY "config/usr/" DESTINATION "${CMAKE_INSTALL_PREFIX}" PATTERN ".gitkeep" EXCLUDE)
 install(DIRECTORY "config/etc/" DESTINATION "${CMAKE_INSTALL_FULL_SYSCONFDIR}" PATTERN ".gitkeep" EXCLUDE)
-install(DIRECTORY "config/systemd/system/" DESTINATION "${SYSTEMD_SYSTEM_UNIT_DIR}" PATTERN ".gitkeep" EXCLUDE)
-install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
-    ${SYSTEMD_SYSTEM_UNIT_DIR}/dnf5-offline-transaction.service
-    ${SYSTEMD_SYSTEM_UNIT_DIR}/system-update.target.wants/dnf5-offline-transaction.service)"
-)
+
+if(WITH_SYSTEMD)
+  install(DIRECTORY "config/systemd/system/" DESTINATION "${SYSTEMD_SYSTEM_UNIT_DIR}" PATTERN ".gitkeep" EXCLUDE)
+endif()
 
 # Makes an empty directory for dnf5-plugins configuration files
 install(DIRECTORY DESTINATION "${CMAKE_INSTALL_FULL_SYSCONFDIR}/dnf/dnf5-plugins")
--- a/dnf5/CMakeLists.txt
+++ b/dnf5/CMakeLists.txt
@@ -34,11 +34,15 @@ target_link_libraries(dnf5 PRIVATE common libdnf5 libdnf5-cli Threads::Threads)
 install(TARGETS dnf5 RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 pkg_check_modules(RPM REQUIRED rpm>=4.17.0)
-pkg_check_modules(SDBUS_CPP REQUIRED sdbus-c++)
-pkg_check_modules(LIBSYSTEMD REQUIRED libsystemd)
 
-target_link_libraries(dnf5 PRIVATE ${RPM_LIBRARIES} ${SDBUS_CPP_LIBRARIES} ${LIBSYSTEMD_LIBRARIES})
-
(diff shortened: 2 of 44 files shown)


kontura:
I see this is still a draft but I would prefer if the commits got restructured.

Specifically I think adding the plugin and then removing it right away is just complicating things, I would squash the commits.

Also I think the `system-upgrade: Refactor, allow any offline transaction` commit could be split it smaller ones.
Basically each bullet point could be a separate commit.

evan-goode:
> I see this is still a draft but I would prefer if the commits got restructured.
> 
> Specifically I think adding the plugin and then removing it right away is just complicating things, I would squash the commits.
> 
> Also I think the `system-upgrade: Refactor, allow any offline transaction` commit could be split it smaller ones. Basically each bullet point could be a separate commit.

Yeah, that's reasonable, I'll reorganize this. Since I'm doing it after the fact, it might not be super tidy (each intermediate commit may not compile), but the history will be more readable.

evan-goode:
OK, I've squashed the commits from the old PR (https://github.com/rpm-software-management/dnf5/pull/1209) and split the large commit into (1) the addition of the `offline` and `system-upgrade` commands and (2) the `--offline` flag. More squashing might be needed as I add more things, but reviews now are appreciated.

Conan-Kudo:
This needs a rebase on current mainline. It does not apply.

include/libdnf5/offline/offline.hpp
```diff
@@ -0,0 +1,91 @@
...
+const std::string STATUS_DOWNLOAD_COMPLETE{"download-complete"};
+const std::string STATUS_READY{"ready"};
+const std::string STATUS_UPGRADE_INCOMPLETE{"upgrade-incomplete"};
+
+const int STATE_VERSION = 0;
+const std::string STATE_HEADER{"offline-transaction-state"};
+
+const std::filesystem::path DEFAULT_DATADIR{std::filesystem::path(libdnf5::SYSTEM_STATE_DIR) / "offline"};
+const std::filesystem::path TRANSACTION_STATE_FILENAME{"offline-transaction-state.toml"};
+const std::filesystem::path TRANSACTION_JSON_FILENAME{"offline-transaction.json"};
```
kontura, line 49: Are all of these values part of the public library API on purpose?

include/libdnf5/offline/offline.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <libdnf5/conf/const.hpp>
+#include <libdnf5/conf/option_bool.hpp>
+#include <libdnf5/conf/option_number.hpp>
+#include <sdbus-c++/sdbus-c++.h>
+#include <toml.hpp>
+
+namespace libdnf5::offline {
+
+const std::string OFFLINE_STARTED_ID{"3e0a5636d16b4ca4bbe5321d06c6aa62"};
+const std::string OFFLINE_FINISHED_ID{"8cec00a1566f4d3594f116450395f06c"};
```
kontura, line 37: The `OFFLINE_FINISHED_ID` doesn't appear to be used.

include/libdnf5/offline/offline.hpp
```diff
@@ -0,0 +1,91 @@
...
+    void write();
+    OfflineTransactionState(std::filesystem::path path);
+    OfflineTransactionStateData & get_data() { return data; };
+    const std::exception_ptr & get_read_exception() const { return read_exception; };
+
+private:
+    void read();
+    std::exception_ptr read_exception;
+    std::filesystem::path path;
+    OfflineTransactionStateData data;
```
kontura, line 77: Could we hide this into a `p_impl`?
Perhaps also the `OfflineTransactionStateData` could store the data in `p_impl` and have a bunch of getters/setters for the values?

This would allow us to do modifications later without breaking ABI.

include/libdnf5/offline/offline.hpp
```diff
@@ -0,0 +1,91 @@
...
+    state_version,
+    status,
+    cachedir,
+    target_releasever,
+    system_releasever,
+    verb,
+    cmd_line,
+    poweroff_after,
+    enabled_repos,
+    disabled_repos)
```
kontura, line 93: Could this also be moved to the cpp file?
The conversions are only used in `offline.cpp` right?

Conan-Kudo:
It probably makes sense to offer a build conditional for this since dnf5 on non-systemd distributions (such as Yocto) is going to be a thing.

dnf5/include/dnf5/context.hpp
```diff
@@ -76,6 +77,10 @@ class Context : public libdnf5::cli::session::Session {
     std::vector<std::string> enable_plugins_patterns;
     std::vector<std::string> disable_plugins_patterns;
 
+    bool should_store_offline = false;
```
jan-kolarik, line 80: I'd rather be for wrapping it into get / set methods instead of having a variable on public API.

include/libdnf5/offline/offline.hpp
```diff
@@ -0,0 +1,95 @@
+/*
```
jan-kolarik, line 1: Are there any plans or considerations for utilizing the `libdnf5::offline` module within the library? Currently, it seems to be associated only with reading/writing offline transaction states, processed only in the `dnf5::Context`, is that correct? Maybe we can consider moving it to the dnf5 part?

dnf5/include/dnf5/context.hpp
jan-kolarik: Also please attach some docstring.

dnf5/commands/system-upgrade/system-upgrade.cpp
```diff
@@ -0,0 +1,158 @@
...
+    register_subcommand(std::make_unique<SystemUpgradeDownloadCommand>(get_context()));
+    register_subcommand(std::make_unique<OfflineCleanCommand>(get_context()));
+    register_subcommand(std::make_unique<OfflineRebootCommand>(get_context()));
+    register_subcommand(std::make_unique<OfflineLogCommand>(get_context()));
+}
+
+SystemUpgradeSubcommand::SystemUpgradeSubcommand(Context & context, const std::string & name)
+    : Command(context, name),
+      datadir(std::filesystem::path{libdnf5::SYSTEM_STATE_DIR} / "system-upgrade"),
+      state(datadir / "state.toml") {}
```
jan-kolarik, line 65: Is this different from the filename defined in the `libdnf5::offline` on purpose?

dnf5/commands/offline/offline.cpp
```diff
@@ -0,0 +1,582 @@
...
+    state->write();
+
+    reboot(poweroff_after->get_value());
+}
+
+void OfflineExecuteCommand::set_argument_parser() {
+    OfflineSubcommand::set_argument_parser();
+    auto & cmd = *get_argument_parser_command();
+    cmd.set_description(
+        "Internal use only, not intended to be run by the user. Executes the transaction in the offline environment.");
```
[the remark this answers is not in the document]
evan-goode: I set `set_complete(false)`. The `visible` flag could be nice, but I don't think it's a problem that it's documented in the help, it could help someone debug `dnf5-offline-transaction.service`. I think it's clear enough that it's not meant to be used.

dnf5/commands/system-upgrade/system-upgrade.hpp
```diff
@@ -0,0 +1,75 @@
...
+class SystemUpgradeCommand : public Command {
+public:
+    explicit SystemUpgradeCommand(Context & context) : Command(context, "system-upgrade") {}
+    void set_parent_command() override;
+    void set_argument_parser() override;
+    void register_subcommands() override;
+    void pre_configure() override;
+};
+
+class SystemUpgradeSubcommand : public Command {
```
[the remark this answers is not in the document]
evan-goode: Yes, thanks!

include/libdnf5/offline/offline.hpp
evan-goode: Thanks, I've added log messages that use it now.

include/libdnf5/offline/offline.hpp
evan-goode: I've moved this whole file out of libdnf5 per @jan-kolarik's suggestion.

include/libdnf5/offline/offline.hpp
evan-goode: Yes, but IMHO it's best to keep it in the header since all users of the struct should agree on how to serialize/deserialize it.

dnf5/include/dnf5/context.hpp
evan-goode: Done, thanks!

include/libdnf5/offline/offline.hpp
evan-goode: Good thought, I've moved it out of libdnf5 to `dnf5/include/dnf5/offline.hpp`.

dnf5/commands/system-upgrade/system-upgrade.cpp
evan-goode: Ah no, this was some old code, thanks.

include/libdnf5/offline/offline.hpp
evan-goode: I think we no longer need to do this if the `offline` module is not in libdnf5, right?

evan-goode:
> It probably makes sense to offer a build conditional for this since dnf5 on non-systemd distributions (such as Yocto) is going to be a thing.

Done in https://github.com/rpm-software-management/dnf5/pull/1280/commits/705de820b103d44ef6237fd3c6ad40141ffa2012, thanks for the reminder.

evan-goode:
I'm marking this as RFR now, I think it's in a good state. I'd like to add the tests in a separate PR.

doc/commands/offline.8.rst
```diff
@@ -0,0 +1,87 @@
...
+Synopsis
+========
+
+``dnf5 offline <subcommand> [options]``
+
+
+Description
+===========
+
+The ``offline`` command is used to manage "offline" transactions---transactions that run when the system is booted into a minimal environment. Running a transaction in this stripped-down environment can be safer than running it when the system is booted normally since the transaction is less likely to interfere with running processes.
```
kontura, line 34: Is the `transactions---transactions` a typo?

dnf5.spec
```diff
@@ -254,6 +262,8 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %dir %{_datadir}/dnf5
 %dir %{_datadir}/dnf5/aliases.d
 %config %{_datadir}/dnf5/aliases.d/compatibility.conf
+%config %{_unitdir}/dnf5-offline-transaction.service
```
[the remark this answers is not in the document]
Conan-Kudo: Uhh, why is something in `/usr` marked as a config file?!

dnf5/commands/offline/offline.cpp
```diff
@@ -0,0 +1,646 @@
...
+    register_subcommand(std::make_unique<OfflineExecuteCommand>(get_context()));
+    register_subcommand(std::make_unique<OfflineLogCommand>(get_context()));
+    register_subcommand(std::make_unique<OfflineStatusCommand>(get_context()));
+}
+
+OfflineSubcommand::OfflineSubcommand(Context & context, const std::string & name) : Command(context, name) {}
+
+void OfflineSubcommand::configure() {
+    auto & ctx = get_context();
+    magic_symlink = "/system-update";
```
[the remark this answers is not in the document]
Conan-Kudo: It should be sufficient to point to `systemd.offline-updates(7)` man page.

dnf5/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ if(NOT WITH_DNF5)
     return()
 endif()
 
+set(SYSTEMD_SYSTEM_UNIT_DIR /usr/lib/systemd/system)
```
Conan-Kudo, line 5: We should get this from pkgconfig by requesting the `systemdsystemunitdir` variable from `systemd` pc file.

```suggestion
pkg_check_modules(SYSTEMD REQUIRED systemd)
pkg_get_variable(SYSTEMD_SYSTEM_UNIT_DIR systemd systemdsystemunitdir)
```

dnf5/CMakeLists.txt
```diff
@@ -47,6 +56,11 @@ install(FILES bash-completion/dnf5 DESTINATION ${BASH_COMPLETION_COMPLETIONSDIR}
 install(FILES "README.plugins" DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/dnf5/plugins" RENAME "README")
 install(DIRECTORY "config/usr/" DESTINATION "${CMAKE_INSTALL_PREFIX}" PATTERN ".gitkeep" EXCLUDE)
 install(DIRECTORY "config/etc/" DESTINATION "${CMAKE_INSTALL_FULL_SYSCONFDIR}" PATTERN ".gitkeep" EXCLUDE)
+install(DIRECTORY "config/systemd/system/" DESTINATION "${SYSTEMD_SYSTEM_UNIT_DIR}" PATTERN ".gitkeep" EXCLUDE)
+install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
+    ${SYSTEMD_SYSTEM_UNIT_DIR}/dnf5-offline-transaction.service
+    ${SYSTEMD_SYSTEM_UNIT_DIR}/system-update.target.wants/dnf5-offline-transaction.service)"
+)
```
Conan-Kudo, line 63: This needs to be wrapped in a WITH_SYSTEMD conditional.

dnf5.spec
```diff
@@ -254,6 +262,8 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
...
+%config %{_unitdir}/dnf5-offline-transaction-cleanup.service
```
Conan-Kudo, line 266: This cannot be a config file, as it's in `/usr`. It also doesn't need to be a config file.

Conan-Kudo (changes requested):

doc/commands/offline.8.rst
evan-goode: That was intended, RST compiles it to an em dash:

> The `offline` command is used to manage “offline” transactions—transactions that run when the system is booted into a minimal environment.

dnf5/commands/offline/offline.cpp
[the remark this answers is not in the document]
evan-goode: Added a comment here, thanks.

dnf5/commands/offline/offline.cpp
```diff
@@ -0,0 +1,646 @@
...
+        << _("Warning: the `_execute` command is for internal use only and is not intended to be run directly by "
+             "the user. To initiate the system upgrade/offline transaction, you should run `dnf5 offline reboot`.")
+        << std::endl;
+
+    if (!std::filesystem::is_symlink(get_magic_symlink())) {
+        throw libdnf5::cli::CommandExitError(0, M_("Trigger file does not exist. Exiting."));
+    }
+
+    const auto & symlinked_path = std::filesystem::read_symlink(get_magic_symlink());
+    if (symlinked_path != get_datadir()) {
```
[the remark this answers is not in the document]
evan-goode: Done, thanks.

dnf5/CMakeLists.txt
evan-goode: Thanks, done!

dnf5/CMakeLists.txt
evan-goode: Thanks!

dnf5.spec
evan-goode: Right, thanks for catching that.

dnf5.spec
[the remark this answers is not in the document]
evan-goode: > I was wondering why the offline execution doesn't work for me when I installed dnf5 from the PR build.

Ah, I had been testing mostly with `make install` and not building the RPMs.

> I think we also need %{_unitdir}/system-update.target.wants/dnf5-offline-transaction.service to enable the service.
I am not sure why the rpm build passed because the symlink is INSTALLed in cmake and I would think its an extra file that is missing in the spec.
Maybe the symlinks are special and we will need something like: https://github.com/rpm-software-management/dnf-plugins-core/blob/master/dnf-plugins-core.spec#L494-L497?

I tried for a bit to have CMake create the symlink and get RPM to package it correctly, but I couldn't get it working, so I've just done it in the spec file, like in dnf-plugins-core.spec as you mentioned. The CMake script doesn't install the service file to `system-update.target.wants`, which is unfortunate since it won't get installed when you do `make install`, but at least the RPM package works now..

> I am also thinking if we could check that the service is present and enabled before doing the reboot?

Good idea, done.

dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
```diff
@@ -131,6 +131,24 @@ descr = "Alias for 'makecache'"
...
+
+['offline-upgrade']
+type = 'command'
+attached_command = 'upgrade'
+group_id = 'commands-compatibility-aliases'
+complete = true
+attached_named_args = [
+  { id_path = 'upgrade.offline' }
+]
+
```
kontura, line 151: I see these aliases don't match the dnf4 commands.
It basically changes `dnf offline-distrosync download` -> `dnf5 offline-distrosync` and discards the rest.

We do have all the functionality so I think we could build the aliases to (more or less) fit.

Something like: 
```
['offline-distrosync']
type = 'command'
attached_command = 'distro-sync'
group_id = 'commands-compatibility-aliases'
complete = true

['offline-distrosync.download']
type = 'command'
attached_command = 'distro-sync'
complete = true
attached_named_args = [
  { id_path = 'distro-sync.offline' }
]

['offline-distrosync.log']
type = 'command'
attached_command = 'offline.log'
complete = true
...
```

(If we decide to do that I would rather do it in a separate PR.)

kontura (commented):
It seems there is some issue with `systemd`, its not building on f38 and f39.

On Monday I would also like to run the CLI by everyone again to make sure they are ok with it. It is a lot of changes.

evan-goode:
> It seems there is some issue with `systemd`, its not building on f38 and f39.

Alright, I think it's fixed. We needed to get the `SYSTEMD_SYSTEM_UNIT_DIR` variable from a pkgconfig file provided by `systemd`, not `libsystemd-devel`.

kontura:
Thank you!

Issue #1139, opened 2024-01-05
dnf5daemon org.rpm.dnf.v0.rpm.Rpm List crash


anaselli:
with the following options:
```
options = {
"package_attrs": [
                "name",
                "epoch",
                "version",
                "release",
                "arch",
                "repo_id",
                "from_repo_id",
                "is_installed",
                "install_size",
                "download_size",
                "sourcerpm",
                "summary",
                "url",
                "license",
                "description",
                "files",
                "changelogs",
                "provides",
                "requires",
                "requires_pre",
                "conflicts",
                "obsoletes",
                "recommends",
                "suggests",
                "enhances",
                "supplements",
                "evr",
                "nevra",
                "full_nevra",
                "reason",
                "vendor",
                "group",
            ],
            "scope": "all",
        }

```
Method List gives the following exception:

```
Traceback (most recent call last):
  File "/home/angelo/src/manatools/dnfdragora/../test-dnf5daemon.py", line 178, in <module>
    test()
  File "/home/angelo/src/manatools/dnfdragora/../test-dnf5daemon.py", line 121, in test
    pkglist = iface_rpm.list(
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib64/python3.10/site-packages/dbus/connection.py", line 634, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
```

m-blaha:
Thanks for noticing!
Looks like there is a problem with large D-Bus replies (and this one is particularly huge - it contains all packages with all their attributes). It needs more investigation to find out whether the root cause of the problem is in dnf5, or in one of underlying libraries.

anaselli:
Well using dnf5-5.1.11-1.fc38.x86_64 i experienced crashes here and there with scope "available" and  packages attrs:
```
"name",
"epoch",
"version",
"release",
"arch",
"repo_id"
"full_nevra"
```
But also with packages attrs:
```
"name",
"epoch",
"version",
"release",
 "arch",
 "repo_id",
 "is_installed",
 "install_size",
 "download_size",
 "summary",
  "url",
[...]

PR #1284, opened 2024-02-28
dnfdaemon: Add new API for requests with large replies

m-blaha:
Adds new API `list_fd()` on `Rpm` interface.
This is an alternative to the original Rpm.list() method. It accepts the same 
input arguments and behaves the same way. The difference is that it uses a
file descriptor to transfer package data to the user.
It enables dnfdaemon to bypass D-Bus message size restrictions. The draw-back
is that the client has to implement reading from a pipe and parsing returned
JSON dictionaries with packages.

For: https://github.com/rpm-software-management/dnf5/issues/1139

### DIFF
--- a/dnf5daemon-client/commands/repoquery/repoquery.cpp
+++ b/dnf5daemon-client/commands/repoquery/repoquery.cpp
@@ -23,14 +23,23 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "wrappers/dbus_package_wrapper.hpp"
 
 #include <dnf5daemon-server/dbus.hpp>
+#include <fcntl.h>
 #include <fmt/format.h>
+#include <json-c/json.h>
+#include <libdnf5-cli/exception.hpp>
 #include <libdnf5-cli/output/package_info_sections.hpp>
+#include <libdnf5/common/exception.hpp>
 #include <libdnf5/conf/option_string.hpp>
 #include <libdnf5/rpm/package.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/rpm/package_set.hpp>
+#include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
+#include <limits.h>
+#include <poll.h>
 
+#include <array>
 #include <iostream>
+#include <string>
 
 namespace dnfdaemon::client {
 
@@ -98,6 +107,64 @@ dnfdaemon::KeyValueMap RepoqueryCommand::session_config() {
     return cfg;
 }
 
+/// Read package objects from json stream. The `json_stream` string is modified, all
+/// successfully parsed json objects are removed from it, possible partial json object
+/// is left for next round of parsing after the rest of the string arrives.
+std::vector<DbusPackageWrapper> json_to_packages(std::string & json_stream) {
+    libdnf_assert(json_stream.size() < INT32_MAX, "can not parse JSON string longer then INT32_MAX");
+    std::vector<DbusPackageWrapper> packages;
+
+    json_tokener * tokener;
+    tokener = json_tokener_new();
+    while (!json_stream.empty()) {
+        json_object * json_pkg =
+            json_tokener_parse_ex(tokener, json_stream.c_str(), static_cast<int>(json_stream.size()));
+        if (json_pkg) {
+            dnfdaemon::KeyValueMap dbuspkg;
+            json_object_object_foreach(json_pkg, key, val) {
+                switch (json_object_get_type(val)) {
+                    case json_type_boolean:
+                        dbuspkg[key] = static_cast<bool>(json_object_get_boolean(val));
+                        break;
+                    case json_type_int:
+                        dbuspkg[key] = static_cast<uint64_t>(json_object_get_int64(val));
+                        break;
+                    default:
+                        dbuspkg[key] = json_object_get_string(val);
+                }
+            }
+            packages.emplace_back(DbusPackageWrapper(dbuspkg));
+
+            auto parse_end = json_tokener_get_parse_end(tokener);
+            json_tokener_reset(tokener);
+            json_object_put(json_pkg);
+            if (parse_end < json_stream.size()) {
+                // more json objects on the line, remove parsed part and continue
+                json_stream = json_stream.substr(parse_end);
+            } else {
+                // everything has been parsed
+                json_stream.clear();
+            }
+        } else {
+            auto err = json_tokener_get_error(tokener);
+            if (err == json_tokener_continue) {
+                // only partial json string in the input line, continue on the next one
+                break;
+            } else {
+                // this is unrecoverable, throw error
+                json_tokener_free(tokener);
+                throw libdnf5::cli::CommandExitError(
+                    1,
+                    M_("Error parsing JSON object \"{}\": {}"),
+                    json_stream,
+                    std::string{json_tokener_error_desc(err)});
+            }
+        }
+    }
+    json_tokener_free(tokener);
+    return packages;
+}
+
 void RepoqueryCommand::run() {
     auto & ctx = get_context();
 
@@ -135,29 +202,86 @@ void RepoqueryCommand::run() {
         options.insert(std::pair<std::string, std::vector<std::string>>("package_attrs", {"full_nevra"}));
     }
 
-    dnfdaemon::KeyValueMapList packages;
-    ctx.session_proxy->callMethod("list")
+    std::array<int, 2> pipefd{};  // File descriptors for the pipe
+    if (pipe2(pipefd.data(), O_NONBLOCK) == -1) {
+        std::cerr << "Error: failed to create a pipe." << std::endl;
+        return;
+    }
+
+    std::string fd_id;
+    ctx.session_proxy->callMethod("list_fd")
         .onInterface(dnfdaemon::INTERFACE_RPM)
         .withTimeout(static_cast<uint64_t>(-1))
         .withArguments(options)
-        .storeResultsTo(packages);
+        .withArguments(sdbus::UnixFd{pipefd[1]})
+        .withArguments(fd_id);
+    close(pipefd[1]);
 
-    auto num_packages = packages.size();
-    for (auto & raw_package : packages) {
-        --num_packages;
-        DbusPackageWrapper package(raw_package);
-        if (info_option->get_value()) {
-            auto out = libdnf5::cli::output::PackageInfoSections();
-            out.setup_cols();
-            out.add_package(package);
-            out.print();
-            if (num_packages) {
-                std::cout << std::endl;
-            }
-        } else {
-            std::cout << package.get_full_nevra() << std::endl;
+    int in_fd = pipefd[0];
+
+    const size_t pipe_buffer_size = fcntl(in_fd, F_GETPIPE_SZ);
+    const int timeout = 30000;
+
+    std::array<pollfd, 1> pfds;
+    pfds[0].fd = in_fd;
+    pfds[0].events = POLLIN;
+
+    std::string message;
+    ssize_t total_bytes_read = 0;
+    ssize_t bytes_read = -1;
+
+    std::vector<char> buffer(pipe_buffer_size);  // Buffer for reading data
+    enum class STATUS { PENDING, FINISHED, FAILED } reading_status;
+    reading_status = STATUS::PENDING;
+    while (reading_status == STATUS::PENDING) {
+        switch (poll(pfds.data(), pfds.size(), timeout)) {
+            case -1:
+                // poll() failed
+                reading_status = STATUS::FAILED;
+                std::cerr << "Error: poll() failed." << std::endl;
+                break;
+            case 0:
+                // timeout was reached
+                reading_status = STATUS::FAILED;
+                std::cerr << "Error: timeout while waiting for data." << std::endl;
+                break;
+            default:
+                if ((pfds[0].revents & POLLIN) != 0) {
+                    bytes_read = read(in_fd, buffer.data(), pipe_buffer_size);
+                    if (bytes_read == -1) {
+                        // read error
+                        if (errno != EAGAIN && errno != EWOULDBLOCK) {
+                            reading_status = STATUS::FAILED;
+                            std::cerr << "Error: reading from the pipe failed." << std::endl;
+                        }
+                    } else if (bytes_read == 0) {
+                        // EOF
+                        reading_status = STATUS::FINISHED;
+                    } else {
+                        // data read
+                        message.append(buffer.data(), bytes_read);
+                        total_bytes_read += bytes_read;
+                        for (const auto & package : json_to_packages(message)) {
+                            if (info_option->get_value()) {
+                                auto out = libdnf5::cli::output::PackageInfoSections();
+                                out.setup_cols();
+                                out.add_package(package);
+                                out.print();
+                                std::cout << std::endl;
+                            } else {
+                                std::cout << package.get_full_nevra() << std::endl;
+                            }
+                        }
+                    }
+                } else if ((pfds[0].revents & POLLHUP) != 0) {
+                    // EOF
+                    reading_status = STATUS::FINISHED;
+                } else {
+                    reading_status = STATUS::FAILED;
+                }
         }
     }
+    close(in_fd);
 }
 
 }  // namespace dnfdaemon::client
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
@@ -87,6 +87,24 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
     </method>
 
     <!--
+        list_fd:
+        @options: an array of key/value pairs
+        @file_descriptor: File descriptor opened for writing used to transfer data
+        @transfer_id: Identifier of this transfer. Used for signaling the end of transfer.
+
+        Get list of packages that match to given filters.
+        Packages are transfered in form of stream of JSON objects. Each JSON
+        object is a dictionary representing one package.
+
+        The same options as in list() method are supported.
+    -->
+    <method name="list">
+        <arg name="options" type="a{sv}" direction="in"/>
+        <arg name="file_descriptor" type="h" direction="in"/>
+        <arg name="transfer_id" type="s" direction="in"/>
+    </method>
+
+    <!--
         install:
         @pkg_specs: an array of package specifications to be installed on the system
         @options: an array of key/value pairs to modify install behavior
@@ -382,6 +400,20 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
     <signal name="transaction_unpack_error">
         <arg name="session_object_path" type="o" />
         <arg name="nevra" type="s" />
+    </signal>
+
+    <!--
+        write_to_fd_finished:
+        @success: true if the tranfer finished successfully
+        @transfer_id: id of the tranfer was finished and its file descriptor closed
+        @error_msg: error message if the transfer was not successful.
+
+        Processing of the item has started.
+    -->
+    <signal name="write_to_fd_finished">
+        <arg name="success" type="b" />
+        <arg name="transfer_id" type="s" />
+        <arg name="error_msg" type="s" />
     </signal>
 </interface>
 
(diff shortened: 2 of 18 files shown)


m-blaha:
Converting back to draft until the new possible approach (https://github.com/rpm-software-management/dnf5/issues/1139#issuecomment-2001923655) is properly investigated/tested.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -86,6 +86,24 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        list_fd:
+        @options: an array of key/value pairs
+        @file_descriptor: File descriptor opened for writing used to transfer data
+        @transfer_id: Identifier of this transfer. Used for signaling the end of transfer.
+
+        Get list of packages that match to given filters.
+        Packages are transfered in form of stream of JSON objects. Each JSON
+        object is a dictionary representing one package.
+
+        The same options as in list() method are supported.
```
j-mracek, line 99: Do you want to deprecate the list option? At least it would be good to mention known issue about large data set.

dnf5daemon-server/main.cpp
```diff
@@ -39,6 +39,8 @@ int main() {
     sigaddset(&mask, SIGINT);
     sigaddset(&mask, SIGTERM);
     pthread_sigmask(SIG_BLOCK, &mask, NULL);
+    // ignore SIGPIPE globally, handle pipe writing errors locally where they occur
```
j-mracek, line 42: Do you mean `locally when they occur`?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
m-blaha: Not sure about deprecating - `list` provides much simpler interface then `list_fd` and should work for most use cases quite well. But I agree with describing possible issues with large data set.

I was also thinking about adding new option to the `list` (e.g. `limit`) to set a maximal number of returned packages. But this is out of scope of this PR.

dnf5daemon-server/main.cpp
m-blaha: Actually I meant `where`. The pipe errors are handled in `write_to_fd()` function during attempt to write.
Without blocking, the `SIGPIPE` signal is emitted instead of returning an errno and the program stops.

dnf5daemon-server/utils.hpp
```diff
@@ -56,4 +57,13 @@ ItemType key_value_map_get(
     }
 }
 
+/// Write the message to the file descriptor opened for writing.
+/// @param Message
+/// @param out_fd Open file descriptor
+/// @param error_msg In case of error this string contains error description
+/// @return Bool whether the write succeeded
```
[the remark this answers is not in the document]
m-blaha: Well, it's a bool, not int. But I see your point and will think about it.

dnf5daemon-server/services/rpm/rpm.cpp
```diff
@@ -392,6 +408,64 @@ sdbus::MethodReply Rpm::list(sdbus::MethodCall & call) {
...
+
+    int out_fd = dbus_unix_fd.get();
+
+    session.fill_sack();
+
+    auto query = filter_packages(options);
+
+    std::vector<std::string> package_attrs =
+        dnfdaemon::key_value_map_get<std::vector<std::string>>(options, "package_attrs", std::vector<std::string>());
+    std::string error_msg{};
```
j-mracek, line 430: What about to use `std::string error_msg;`?

dnf5daemon-server/services/rpm/rpm.cpp
```diff
@@ -392,6 +408,64 @@ sdbus::MethodReply Rpm::list(sdbus::MethodCall & call) {
...
+        std::string jsonstr;
+        try {
+            jsonstr = package_to_json(pkg, package_attrs) + "\n";
+        } catch (const std::exception & ex) {
+            error_msg = fmt::format(
+                "Error serializing package \"{0}\" from repo \"{1}\": {2}",
+                pkg.get_nevra(),
+                pkg.get_repo_id(),
+                ex.what());
+        } catch (...) {
```
j-mracek, line 441: Interesting - Do you have any particular experience that this might happen?

dnf5daemon-server/services/rpm/rpm.cpp
```diff
@@ -392,6 +408,64 @@ sdbus::MethodReply Rpm::list(sdbus::MethodCall & call) {
...
+
+    // signal client that the transfer has finished and the output file descriptor is closed
+    auto dbus_object = get_session().get_dbus_object();
+    auto signal = dbus_object->createSignal(call.getInterfaceName(), dnfdaemon::SIGNAL_WRITE_TO_FD_FINISHED);
+    signal << error_msg.empty();
+    signal << out_fd_id;
+    signal << error_msg;
+    try {
+        dbus_object->emitSignal(signal);
+    } catch (...) {
```
j-mracek, line 465: Shell we log something on the server side?

dnf5daemon-server/package.hpp
```diff
@@ -72,4 +72,6 @@ enum class PackageAttribute {
 dnfdaemon::KeyValueMap package_to_map(
     const libdnf5::rpm::Package & libdnf_package, const std::vector<std::string> & attributes);
 
+std::string package_to_json(const libdnf5::rpm::Package & libdnf_package, const std::vector<std::string> & attributes);
```
j-mracek, line 75: May I ask you to add a doc string. It would be good if it will provide an information about a structure of json.

dnf5daemon-server/package.cpp
```diff
@@ -196,3 +196,158 @@ dnfdaemon::KeyValueMap package_to_map(
...
+void add_string_list(json_object * json_pkg, const char * cattr, const std::vector<std::string> & vector) {
+    json_object * array = json_object_new_array();
+    json_object_object_add(json_pkg, cattr, array);
+    for (const auto & elem : vector) {
+        json_object_array_add(array, json_object_new_string(elem.c_str()));
+    }
+}
+
+std::string package_to_json(
+    const libdnf5::rpm::Package & libdnf_package, [[maybe_unused]] const std::vector<std::string> & attributes) {
```
j-mracek, line 209: May I ask you why `[[maybe_unused]]` is used here?

dnf5daemon-server/services/rpm/rpm.cpp
m-blaha: Nothing in particular. Just didn't want the server to abort.
But the unknown exceptions are already caught one level higher in `ThreadsManager::handle_method_fd()` so I will remove these "catch all" branches here.

dnf5daemon-server/package.cpp
```diff
@@ -196,3 +196,158 @@ dnfdaemon::KeyValueMap package_to_map(
...
+
+    // attributes required by client
+    for (const auto & attr : attributes) {
+        const auto * const cattr = attr.c_str();
+        auto it = package_attributes.find(attr);
+        if (it == package_attributes.end()) {
+            throw std::runtime_error(fmt::format("Package attribute '{}' not supported", attr));
+        }
+        switch (it->second) {
+            case PackageAttribute::name:
```
j-mracek, line 223: Just a question, may I request the same attribute multiple times?

dnf5daemon-server/package.cpp
m-blaha: You may... forgotten from the early stages of development where the function was not fully implemented. Thanks for catching it!

doc/dnf_daemon/examples/rpm_list_fd.py
```diff
@@ -0,0 +1,165 @@
...
+        # wait for data
+        polled_event = poller.poll(timeout)
+        if not polled_event:
+            print("Timeout reached.")
+            break
+        # we know there is only one fd registered in poller
+        descriptor, event = polled_event[0]
+        # read a chunk of data
+        buffer = os.read(descriptor, buffer_size)
+        if buffer:
```
j-mracek, line 72: A suggestion - May be it would be better to switch order

```
if not buffer:
    break
else:
    ...
```

dnf5daemon-server/utils.cpp
```diff
@@ -0,0 +1,84 @@
...
+                            // e.g. the client has closed the read end of the pipe
+                            failed = true;
+                            error_msg = "Error writing to the file descriptor.";
+                        }
+                    } else {
+                        total_bytes_written += bytes_written;
+                        bytes_remaining -= bytes_written;
+                    }
+                } else {
+                    failed = true;
```
j-mracek, line 76: Interesting - in while loops there are four location where `failed = true`. Two of them are followed by break. May I ask you to unify it?

doc/dnf_daemon/examples/rpm_list_fd.py
```diff
@@ -0,0 +1,165 @@
...
+def repoquery(iface_rpm, options):
+    """Generator function that yields packages as they arrive from the server."""
+
+    # create a pipe and pass the write end to the server
+    pipe_r, pipe_w = os.pipe()
+    # transfer id serves as an identifier of the pipe transfer for a signal emitted
+    # after server finish. This example to not use it.
+    iface_rpm.list_fd(options, pipe_w, "the_transfer_id")
+    # close the write end - otherwise poll cannot detect the end of transmission
+    os.close(pipe_w)
```
j-mracek, line 45: What happen when DBus is overloaded and pipe_w is closed before server part receive DBus message? Or file descriptor is already cloned for server?

doc/dnf_daemon/examples/rpm_list_fd.py
```diff
@@ -0,0 +1,165 @@
...
+    # 64k is a typical size of a pipe
+    buffer_size = 65536
+
+    # remaining string to parse (can contain unfinished json from previous run)
+    to_parse = ""
+    # remaining raw data (i.e. data before UTF decoding)
+    raw_data = b""
+    while True:
+        # wait for data
+        polled_event = poller.poll(timeout)
```
j-mracek, line 64: Does it wait until server part calls `close(fd)`?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
j-mracek: I think that data size is not about number of packages. Think that most of RPM string fields has no limitation of size.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -86,6 +86,24 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         <arg name="packages" type="aa{sv}" direction="out"/>
     </method>
 
+    <!--
```
j-mracek, line 89: There is a typo above - "upradable" - line 55

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
j-mracek: Is there a list of allowed package_attrs? May I ask for file lists?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -86,6 +86,24 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        list_fd:
+        @options: an array of key/value pairs
+        @file_descriptor: File descriptor opened for writing used to transfer data
+        @transfer_id: Identifier of this transfer. Used for signaling the end of transfer.
+
+        Get list of packages that match to given filters.
+        Packages are transfered in form of stream of JSON objects. Each JSON
+        object is a dictionary representing one package.
```
j-mracek, line 97: What about to mention time-out to write to a pipe?

dnf5daemon-server/main.cpp
j-mracek: Thank you

dnf5daemon-client/commands/repoquery/repoquery.cpp
```diff
@@ -135,29 +202,86 @@ void RepoqueryCommand::run() {
...
 
-    dnfdaemon::KeyValueMapList packages;
-    ctx.session_proxy->callMethod("list")
+    std::array<int, 2> pipefd{};  // File descriptors for the pipe
+    if (pipe2(pipefd.data(), O_NONBLOCK) == -1) {
+        std::cerr << "Error: failed to create a pipe." << std::endl;
+        return;
+    }
+
+    std::string fd_id{};
```
j-mracek, line 211: What about `std::string fd_id;`?

dnf5daemon-client/commands/repoquery/repoquery.cpp
```diff
@@ -135,29 +202,86 @@ void RepoqueryCommand::run() {
         options.insert(std::pair<std::string, std::vector<std::string>>("package_attrs", {"full_nevra"}));
     }
 
-    dnfdaemon::KeyValueMapList packages;
```
j-mracek, line 138: Note to `package_attrs`  that are defined above - The output is used for  package wrapper that has set of supported attributes. When data are not available then it raises Out of range error. What about to store required attribute directly in the class?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
m-blaha: I added a commit with enhanced documentation covering these comments.

dnf5daemon-server/threads_manager.hpp
```diff
@@ -103,6 +105,63 @@ class ThreadsManager {
...
+                    error_msg = e.what();
+                } catch (...) {
+                    error_msg = "Unknown exception caught";
+                }
+
+                if (error_msg.empty()) {
+                    try {
+                        (service.*method)(call);
+                    } catch (...) {
+                        // TODO(mblaha): log the error
```
j-mracek, line 140: Would be possible to add at least `std::cerr << fmt::format(`, but keep TODO?

dnf5daemon-client/commands/repoquery/repoquery.cpp
m-blaha: I know about this issue. But dnf5daemon-client is kind of an example how to work with D-Bus API and never got priority high enough to work on fixing these edgecases. I can file a low priority issue on it.

dnf5daemon-server/services/rpm/rpm.cpp
m-blaha: Yes. Let me just add a TODO for now, because there are more places which need logging (already marked with TODO). I would like to fix them in follow up PR.

dnf5daemon-server/utils.cpp
m-blaha: Those breaks are part of swich/case statement. They do not break the cycle.

doc/dnf_daemon/examples/rpm_list_fd.py
m-blaha: The `list_fd()` call here is synchronous - it will wait for the reply from D-Bus.

doc/dnf_daemon/examples/rpm_list_fd.py
j-mracek: Thank you for explanation

dnf5daemon-server/threads_manager.hpp
m-blaha: Question is what information could we print here - we do not have an exception (anything could have been thrown), we do not now which method has been called (we only have a pointer to the method).

dnf5daemon-server/package.cpp
j-mracek: It means we can de-duplicate attributes before iteration (Including package IDs), but this is not required for this PR.

dnf5daemon-server/threads_manager.hpp
j-mracek: For now we can keep it with TODO. I suggest to log general unique message that will allow us to identify the correct catch in the code. If I am not mistaken it is possible to get something like a traceback with C++20.

doc/dnf_daemon/examples/rpm_list_fd.py
m-blaha: True, `raw_data` can still contain part of input failing to `decode()`. Also `to_parse` string can contain some unfinished JSON objects. So the `repoquery` generator should probably raise an exception in case these variables are not empty after parsing loop finished.

m-blaha:
Rebased due to conflicts and squashed the fixups.

dnf5daemon-server/package.hpp
```diff
@@ -72,4 +72,10 @@ enum class PackageAttribute {
 dnfdaemon::KeyValueMap package_to_map(
     const libdnf5::rpm::Package & libdnf_package, const std::vector<std::string> & attributes);
 
+/// Convert given libdnf_package to a JSON string using requested attributes.
+/// @param libdnf_package Package to convert to JSON
+/// @param attributes A list of attributes of ligdnf_package that are included in JSON
+/// @return JSON String with the package represented as key:value dictionary.
+std::string package_to_json(const libdnf5::rpm::Package & libdnf_package, const std::vector<std::string> & attributes);
+
```
j-mracek, line 80: I am not sure whether it is important, but `package.hpp` and `cpp` does not use any namespaces.

dnf5daemon-server/package.cpp
```diff
@@ -196,3 +196,157 @@ dnfdaemon::KeyValueMap package_to_map(
     }
     return dbus_package;
 }
+
+void add_string_list(json_object * json_pkg, const char * cattr, const std::vector<std::string> & vector) {
```
j-mracek, line 200: I believe that `add_string_list` is only internally used in this `cpp` therefore it should be in anonymous namespace or parked as static. Please consider move the code to a namespace.

dnf5daemon-server/package.hpp
m-blaha: I know I've already started moving stuff into namespace in this PR and if you want me to, I can continue with `package.[ch]pp`. But personally I'd prefer to create separate issue for move to namespace and resolve it in separate PR. Are you OK with this?

m-blaha: Created an issue for this refactoring: https://github.com/rpm-software-management/dnf5/issues/1389

dnf5daemon-server/package.cpp
m-blaha: Done.

j-mracek (approved):
LGTM

mcrha:
I just noticed a small typos in the `write_to_fd_finished` description in the `org.rpm.dnf.v0.rpm.Rpm.xml` file:
1. there is a `tranfer` word used twice (missing `s`)
2. the `Processing of the item has started.` description looks odd when the signal is about something finished.

mcrha:
A bit bigger problem is that the same file contains:
```
    <method name="list">
        <arg name="options" type="a{sv}" direction="in"/>
        <arg name="file_descriptor" type="h" direction="in"/>
        <arg name="transfer_id" type="s" direction="in"/>
    </method>
```
whose name clashes with the original `list` method, aka there should be `<method name="list_fd">` instead (`list ~> list_fd`).

Compiler catches it when the gdbus-codegen processes the files into the C sources and tries to compile the generated code.

mcrha:
Would it make sense to generate the `transfer_id` on the server side? That can avoid clashes of it when multiple clients try to `list_fd` something, especially generating on the server side can avoid implementing certain techniques to generate truly unique transfer ids. I'd probably use something like `transfer_id = g_strdup_printf ("%d-%p-%p-%d", getpid (), rpm_proxy, g_thread_self (), counter++);`, while the server side can just use `transfer_id = g_strdup_printf ("%p", transfer_object);`, because the server runs as a single process.

m-blaha:
> Would it make sense to generate the `transfer_id` on the server side? That can avoid clashes of it when multiple clients try to `list_fd` something, especially generating on the server side can avoid implementing certain techniques to generate truly unique transfer ids. I'd probably use something like `transfer_id = g_strdup_printf ("%d-%p-%p-%d", getpid (), rpm_proxy, g_thread_self (), counter++);`, while the server side can just use `transfer_id = g_strdup_printf ("%p", transfer_object);`, because the server runs as a single process.

It's definitely possible - it's basically a change if input transfer_id argument to an output one and adding some code to generate the id. @anaselli what do you think about it? You already use this functionality and such change will break your code.

anaselli:
Sorry @m-blaha  i was out some days for my daily job... I don't think it is a big issue for me changing the code... I'm still in a porting phase so no worries...

mcrha:
I opened https://github.com/rpm-software-management/dnf5/issues/1493 for the `transfer_id` argument suggested change.

PR #1277, opened 2024-02-27
modules: Report module solver problems

pkratoch:
Replacement for https://github.com/rpm-software-management/dnf5/pull/1229#pullrequestreview-1888032076 , but opened against dnf5-5.2.0.0 branch.

PR #1285, opened 2024-02-29
modules: Report switching module streams

pkratoch:


libdnf5/base/goal.cpp
```diff
@@ -2316,6 +2316,25 @@ base::Transaction Goal::resolve() {
...
+                    GoalProblem::MODULE_CANNOT_SWITH_STREAMS,
+                    GoalJobSettings(),
+                    libdnf5::transaction::TransactionItemType::MODULE,
+                    item.first,
+                    {"0:" + item.second.first, "1:" + item.second.second},
+                    libdnf5::Logger::Level::ERROR);
+            }
+            ret |= GoalProblem::MODULE_CANNOT_SWITH_STREAMS;
+            transaction.p_impl->problems = ret;
+            return transaction;
```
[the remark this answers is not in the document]
pkratoch: Yes, if it is not possible to switch module streams, but the operation requires it, it's not possible to resolve modules, therefore we don't know what modules are active and what rpms are used, so it doesn't make sense to resolve the rest of the goal.

The same case is when the modules can't be resolved for other reasons, but it's mostly covered in https://github.com/rpm-software-management/dnf5/pull/1277 .

This is consistent with DNF4 behavior.

libdnf5/base/transaction.cpp
```diff
@@ -461,6 +461,14 @@ void Transaction::Impl::set_transaction(
             name, "", transaction::TransactionItemAction::RESET, transaction::TransactionItemReason::USER);
         modules.emplace_back(std::move(tsmodule));
     }
+    for (auto & name_streams : module_db->get_all_newly_switched_streams()) {
+        TransactionModule tsmodule(
+            name_streams.first,
+            name_streams.second.first + " -> " + name_streams.second.second,
```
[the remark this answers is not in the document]
pkratoch: Sure, I can change the `TransactionModule` and move it to the dnf-5.2.0.0.

libdnf5/base/goal.cpp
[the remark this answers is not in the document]
kontura: I see, could you add a comment with that explanation please?

j-mracek, line 2335: There might be a problem because if the operation was called with requesting of a debugsolver data they will be not created.

j-mracek: May I ask why you prefer to not continue and collect all problems?

libdnf5/base/log_event.cpp
```diff
@@ -186,6 +186,28 @@ std::string LogEvent::to_string(
...
+            ret.append(utils::sformat(
+                _("The operation would result in switching of module '{0}' stream '{1}' to stream '{2}'\n"),
+                *spec,
+                original_stream,
+                new_stream));
+            return ret.append(
+                _("Error: It is not possible to switch enabled streams of a module unless explicitly enabled via "
+                  "configuration option module_stream_switch.\nIt is recommended to rather remove all installed "
+                  "content from the module, and reset the module using 'dnf module reset <module_name>' command. After "
+                  "you reset the module, you can install the other stream."));
```
j-mracek, line 209: > It is recommended to rather remove all installed "
                  "content from the module, and reset the module using 'dnf module reset <module_name>' command. After "
                  "you reset the module, you can install the other stream."

What about to remove it. It is a rudiment and after we have module-swich command in DNF4 this approach was obsoleted. Anyway we discovered that this suggestion is even harmful because modules also contains packages from base and they removal would result in a malfunction system.

libdnf5/base/transaction.cpp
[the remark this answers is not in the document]
j-mracek: Please also think that modules might obsolete each other. Therefore module switch might also obsolete something else like RPMs.

libdnf5/base/goal.cpp
pkratoch: The writing of debugsolver data from module solver was not implemented and not called anywhere yet, so this PR doesn't affect that. I believe that this needs to be separate for modules and that it doesn't make sense to continue with resolving the rpm goal if there is a fatal error when resolving modules. But I might be missing something.

pkratoch: I think it would be good to create an issue for the debugsolver data for modules.

libdnf5/base/log_event.cpp
pkratoch: Ah, ok, I didn't know it's not the recommended approach anymore. I'll remove it.

libdnf5/base/goal.cpp
j-mracek: I was thinking about RPM debugsolve code bellow. I completely forget about that modular debugsolv data are not created

j-mracek: DNF tries to report all problems at once. If I am not mistaken the problem will suggest to allow module switching by configuration option modification. Think that user will enable module stream switching for the next run and then it will receive another set problems with RPMs.
The second user is API user - even when DNF5 goal.resolve() returns some error then API user can still perform the transaction. But if we will skip some step then the transaction will be incomplete.
Think that user even don't want to run a transaction - they only want to see what happened - what request requires and so on - then it would be good to provide maximum results if it is possible. Switching stream is bad thing, but not a dead end like solver dependency error.

pkratoch: Ah, ok, I had the assumption that switching streams is a problem like the solver error, so we should not continue. I'll look into it.

pkratoch:
Closing, since I opened a new PR against the dnf-5.2.0.0 branch: https://github.com/rpm-software-management/dnf5/pull/1308

Issue #1251, opened 2024-02-17
dnf5 performance vs dnf: should it be better?


jashank:
Hello!

I've mostly migrated from `dnf` 4 (currently 4.19.0) to `dnf5` (currently 5.11.12) on my rawhide system. I'm sure I've seen references to the design being intentionally faster — and I've been certainly enjoying places where `dnf5` performs way better than `dnf` — e.g., responding to `C-c` pretty rapidly.  But I've found `dnf5` _feels_ slower.

<details>
<summary><code>/etc/dnf/dnf.conf</code></summary>

```
[main]
gpgcheck = True
installonly_limit = 3
clean_requirements_on_remove = True
best = False
skip_if_unavailable = True
keepcache = True

countme = True

fastestmirror = True
max_parallel_downloads = 10

# Try to never, ever remove package management logs.
log_rotate = 64
log_compress = True
```

</details>

<details>
<summary><code>rpm -qa | grep 'dnf|hawkey|libsolv'</code></summary>

```
libsolv-0.7.28-1.fc40.x86_64
libdnf5-5.1.12-1.fc40.x86_64
libdnf5-cli-5.1.12-1.fc40.x86_64
dnf-data-4.19.0-1.fc40.noarch
libdnf-0.73.0-1.fc40.x86_64
python3-libdnf-0.73.0-1.fc40.x86_64
python3-hawkey-0.73.0-1.fc40.x86_64
dnf5-5.1.12-1.fc40.x86_64
dnf5daemon-server-5.1.12-1.fc40.x86_64
python3-dnf-4.19.0-1.fc40.noarch
python3-dnf-plugins-core-4.5.0-1.fc40.noarch
dnf-plugins-core-4.5.0-1.fc40.noarch
dnf-4.19.0-1.fc40.noarch
dnf-utils-4.5.0-1.fc40.noarch
python3-dnf-plugin-versionlock-4.5.0-1.fc40.noarch
python3-dnf-plugin-post-transaction-actions-4.5.0-1.fc40.noarch
python3-dnf-plugin-modulesync-4.5.0-1.fc40.noarch
dnf5daemon-client-5.1.12-1.fc40.x86_64
dnf5-plugins-5.1.12-1.fc40.x86_64
libsolv-demo-0.7.28-1.fc40.x86_64
ruby-libdnf5-cli-5.1.12-1.fc40.x86_64
perl-libdnf5-cli-5.1.12-1.fc40.x86_64
[...]

PR #1286, opened 2024-03-01
Workaround slow scols_table_print_range() for pkg info formatting

kontura:
This is a hacky workaround for a performance problem with `scols_table_print_range()`. Once the table is large (eg. all pkgs in fedora repo) calling `scols_table_print_range()` for each package is very slow. However we cannot print the whole table at once because it isn't possible to add an empty line into the table to separate the packages.

To workaround this print each package right away after it is added and clear the table. This changes the behavior, the printing is done by `add_section()` and the final call to `print()` is a noop.

This should be reworked for 5.2.0.0 where we can break API.

For: https://github.com/rpm-software-management/dnf5/issues/1251

### DIFF
--- a/libdnf5-cli/output/package_info_sections.cpp
+++ b/libdnf5-cli/output/package_info_sections.cpp
@@ -71,6 +71,20 @@ bool PackageInfoSections::add_section(
                 add_package(pkg, tmp_heading, colorizer, {});
             }
             tmp_heading = "";
+            //TODO(amatej): This is a hacky workaround for a performance problem with
+            //`scols_table_print_range()`. Once the table is large (eg. all pkgs in fedora repo)
+            //calling `scols_table_print_range()` for each package is very slow.
+            //However we cannot print the whole table at once because it isn't possible
+            //to add an empty line into the table to separate the packages.
+            //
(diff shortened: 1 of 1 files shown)


kontura:
Tests update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1464

jan-kolarik:
I don't know why yet, but it seems it doesn't work in all terminal emulators. E.g. in GNOME Terminal it seems working correctly, though in the integrated Visual Studio Code terminal, it is still very slow...

kontura:
> I don't know why yet, but it seems it doesn't work in all terminal emulators. E.g. in GNOME Terminal it seems working correctly, though in the integrated Visual Studio Code terminal, it is still very slow...

Interesting, I would think the terminal wouldn't have an effect on this.

ppisar:
In the old code there was ioctl(1, TCGETS, ...) call before printing every package section. If that call propagates to a terminal emulator, then a terminal could have an effect.

ppisar:
However, I believe that kernel answers the queried terminal settings from its own in-kernel structures without bothering the PTY master process.

jan-kolarik (approved):

PR #1287, opened 2024-03-01
Use PR specfile for copr builds

inknos:


kontura:
@inknos I think we can close this right?

inknos:
yup

Issue #1185, opened 2024-01-19
Switch deltarpm=False for Fedora 40+


jan-kolarik:
Based on the system-wide change: https://fedoraproject.org/wiki/Changes/Drop_Delta_RPMs.

This is a follow-up to https://bugzilla.redhat.com/show_bug.cgi?id=2252128.

PR #1291, opened 2024-03-04
Switch off deltarpm support

jan-kolarik:
Aligning upstream code with Fedora 40 changes, where deltarpm support was discontinued.

Fixes: https://github.com/rpm-software-management/dnf5/issues/1185.


kontura:
Thanks, there is one failing createrepo_c tests (issue [here](https://github.com/rpm-software-management/createrepo_c/issues/386)) that is not related.

PR #1294, opened 2024-03-05
Hide/Remove deprecated methods

kontura:
There is still public `libdnf5::repo::Repo::load()` which is marked as deprecated but it
cannot be easily hidden because dnfdaemon uses it as well.

@j-mracek since you originally marked it as deprecated do we want to keep the `load()` method or do we need to do some bigger changes to hide it?

For: https://github.com/rpm-software-management/dnf5/issues/1091

### DIFF
--- a/dnf5/include/dnf5/shared_options.hpp
+++ b/dnf5/include/dnf5/shared_options.hpp
@@ -69,9 +69,6 @@ void create_destdir_option(dnf5::Command & command);
 /// The values are stored in the `downloadonly` configuration option
 void create_downloadonly_option(dnf5::Command & command);
 
-/// Create the `--forcearch` option for a command provided as an argument.
-/// The values are stored in the `forcearch` configuration option
-[[deprecated("--forcearch is now a global argument")]] void create_forcearch_option(dnf5::Command & command);
 
 }  // namespace dnf5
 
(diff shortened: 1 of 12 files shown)


jan-kolarik:
LGTM. 

Regarding the `Repo::load()` issue, we can address it in a follow-up PR if it's too complex. Currently, it's used in the Context and dnf5daemon. Since `update_and_load_enabled_repos` is expected to be the only method for loading repositories, the only way to load just the system repository seems to be manually disabling available repositories before executing that method, is that right? I'm not sure how common is this use case, but it does seem somewhat clumsy.

kontura:
> Regarding the `Repo::load()` issue, we can address it in a follow-up PR if it's too complex. 

A follow-up PR sounds good.

> Currently, it's used in the Context, main and dnf5daemon. Since `update_and_load_enabled_repos` is expected to be the only method for loading repositories, the only way to load just the system repository seems to be manually disabling available repositories before executing that method, is that right? 

Yes, without the `libdnf5::repo::Repo::load()` I think that is the case.

> I'm not sure how common is this use case, but it does feel somewhat clumsy.

Both dnf5 and dnf5daemon do it, I think its pretty common. For example for querying just installed packages.

jan-kolarik (approved):

PR #1297, opened 2024-03-06
Adjust code to new rpm::PackageQuery::filter_* methods after the rebase

jan-kolarik:


jan-kolarik:
The CCache build should be false negative, because the mentioned failure references an old code that was replaced in this PR.

kontura:
Thanks, the failing test is expected because the `main` branch has diverged again.

Issue #1246, opened 2024-02-13
dnf5daemon - revise methods/signals registration and documentation


m-blaha:
- [ ] all methods and signals are correctly registered on interfaces with correct signatures
- [ ] all parameters of methods and signals are named during the registration
- [ ] the documentation of methods and signals in *.xml files are in sync with implementation

Related - https://github.com/rpm-software-management/dnf5/pull/1232

PR #1300, opened 2024-03-07
dnf5daemon: Introspection files revision

m-blaha:
The change makes sure that all dnf5daemon-server methods and signals:
- are documented in <interface>.xml files
- have correct D-Bus signatures
- all their parameters have provided names for better introspection

Fixes: https://github.com/rpm-software-management/dnf5/issues/1246

### DIFF
--- a/dnf5daemon-server/callbacks.cpp
+++ b/dnf5daemon-server/callbacks.cpp
@@ -81,7 +81,7 @@ int DownloadCB::progress(void * user_cb_data, double total_to_download, double d
 int DownloadCB::end(void * user_cb_data, TransferStatus status, const char * msg) {
     try {
         auto signal = create_signal_download(dnfdaemon::SIGNAL_DOWNLOAD_END, user_cb_data);
-        signal << static_cast<int>(status);
+        signal << static_cast<uint32_t>(status);
         signal << msg;
         dbus_object->emitSignal(signal);
     } catch (...) {
@@ -131,7 +131,7 @@ void DbusTransactionCB::install_start(const libdnf5::rpm::TransactionItem & item
         action = dnfdaemon::transaction_package_to_action(item);
         auto signal = create_signal_pkg(
             dnfdaemon::INTERFACE_RPM, dnfdaemon::SIGNAL_TRANSACTION_ACTION_START, item.get_package().get_full_nevra());
-        signal << static_cast<int>(action);
+        signal << static_cast<uint32_t>(action);
         signal << total;
         dbus_object->emitSignal(signal);
     } catch (...) {
@@ -171,7 +171,7 @@ void DbusTransactionCB::script_start(
     try {
         auto signal = create_signal_pkg(
             dnfdaemon::INTERFACE_RPM, dnfdaemon::SIGNAL_TRANSACTION_SCRIPT_START, to_full_nevra_string(nevra));
-        signal << static_cast<int>(type);
+        signal << static_cast<uint32_t>(type);
         dbus_object->emitSignal(signal);
     } catch (...) {
     }
@@ -185,7 +185,7 @@ void DbusTransactionCB::script_stop(
     try {
         auto signal = create_signal_pkg(
             dnfdaemon::INTERFACE_RPM, dnfdaemon::SIGNAL_TRANSACTION_SCRIPT_STOP, to_full_nevra_string(nevra));
-        signal << static_cast<int>(type);
+        signal << static_cast<uint32_t>(type);
         signal << return_code;
         dbus_object->emitSignal(signal);
     } catch (...) {
@@ -211,7 +211,7 @@ void DbusTransactionCB::script_error(
     try {
         auto signal = create_signal_pkg(
             dnfdaemon::INTERFACE_RPM, dnfdaemon::SIGNAL_TRANSACTION_SCRIPT_ERROR, to_full_nevra_string(nevra));
-        signal << static_cast<int>(type);
+        signal << static_cast<uint32_t>(type);
         signal << return_code;
         dbus_object->emitSignal(signal);
     } catch (...) {
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Base.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Base.xml
@@ -101,6 +101,26 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         <arg name="message" type="s" />
     </signal>
 
+    <!--
+        repo_key_import_request:
+        @session_object_path: object path of the dnf5daemon session
+        @key_id: PGP key id
+        @user_ids: User id
+        @key_fingerprint: Fingerprint of the PGP key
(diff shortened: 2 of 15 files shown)


dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.SessionManager.xml
```diff
@@ -49,19 +49,19 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
     -->
     <method name="open_session">
         <arg name="options" type="a{sv}" direction="in"/>
```
ppisar, line 51: "options" argument name does not match documented name "configuration".

ppisar: Because "options" is also used in session_manager.cpp, I recommend renaming "configuration" in the documentation.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Base.xml
```diff
@@ -89,38 +89,18 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         download_end:
         @session_object_path: object path of the dnf5daemon session
         @download_id: unique id of downloaded object (repo or package)
-        @status: libdnf5::repo::DownloadCallbacks::TransferStatus (0 - successful, 1 - already exists, 2 - error)
+        @transfer_status: libdnf5::repo::DownloadCallbacks::TransferStatus (0 - successful, 1 - already exists, 2 - error)
         @error: error message in case of failed download
```
ppisar, line 93: Documented "error" name does match "message" name in XML.

ppisar: A code in base.cpp uses "message".

dnf5daemon-server/services/base/base.cpp
```diff
@@ -34,14 +34,30 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        {"session_object_path", "download_id", "description", "total_to_download"});
+    dbus_object->registerSignal(
+        dnfdaemon::INTERFACE_BASE,
+        dnfdaemon::SIGNAL_DOWNLOAD_PROGRESS,
+        "osxx",
+        {"session_object_path", "download_id", "total_to_download", "downloaded"});
+    dbus_object->registerSignal(
+        dnfdaemon::INTERFACE_BASE,
+        dnfdaemon::SIGNAL_DOWNLOAD_END,
+        "osis",
```
ppisar, line 54: The signature should  be "osus". A code in callbacks.cpp emits uint32_t as well as org.rpm.dnf.v0.Base.xml documents "u" type.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Base.xml
```diff
@@ -89,38 +89,18 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
-        @session_object_path: object path of the dnf5daemon session
-        @key_id: PGP key id
-        @user_ids: User id
-        @key_fingerprint: Fingerprint of the PGP key
-        @key_url: URL of the PGP key
-        @timestamp: timestamp when the key was created
-
-        Request for repository key import confirmation.
-    -->
-    <signal name="repo_key_import_request">
```
ppisar, line 115: There is a bug in callbacks.cpp: KeyImportRepoCB::repokey_import() emits
a dnfdaemon::SIGNAL_REPO_KEY_IMPORT_REQUEST signal in wrong
dnfdaemon::INTERFACE_BASE interface. On the other hand, the signal is correctly registered in services/repo/repo.cpp as dnfdaemon::INTERFACE_REPO.

I guess we want to change callbacks.cpp. Technically, it's will break D-Bus API as we will move the singnal into a different interface. Do you want to resolve the callbacks.cpp bug in a separate issue targetting dnf-5.2.0.0 branch, or do you want to sweep it under the rug in this pull request?

dnf5daemon-server/callbacks.cpp
```diff
@@ -81,7 +81,7 @@ int DownloadCB::progress(void * user_cb_data, double total_to_download, double d
 int DownloadCB::end(void * user_cb_data, TransferStatus status, const char * msg) {
     try {
         auto signal = create_signal_download(dnfdaemon::SIGNAL_DOWNLOAD_END, user_cb_data);
-        signal << static_cast<int>(status);
+        signal << static_cast<uint32_t>(status);
```
ppisar, line 84: This breaks API. Though unless the client performs more than 2^31, ABI will be unchanged. I will accept it into main because of no practical impact.

ppisar:
I'm done with reviewing this patchset (3eb8c34f45cc74069c41706a4651ff7856e7b645). Please look at the in-line comments.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Base.xml
m-blaha: Hm, this is unfortunate. The signal was originally registered on Repo interface, but in base.cpp. So I just moved the registration to repo.cpp, which was a wrong decision. Thanks for noticing it!
I think it's safer to keep the signal on Base interface as the clients need to handle it in order to correctly import repo keys. I'd rather not disrupt them.

dnf5daemon-server/callbacks.cpp
m-blaha: Yeah, I wanted to unify data types used to transfer these "status"-like values. They are basically enums with only couple of available options so I do not expect to exceed maxint.

dnf5daemon-server/session_manager.cpp
```diff
@@ -48,11 +48,23 @@ SessionManager::~SessionManager() {
 void SessionManager::dbus_register() {
     dbus_object = sdbus::createObject(*connection, dnfdaemon::DBUS_OBJECT_PATH);
     dbus_object->registerMethod(
-        dnfdaemon::INTERFACE_SESSION_MANAGER, "open_session", "a{sv}", "o", [this](sdbus::MethodCall call) -> void {
+        dnfdaemon::INTERFACE_SESSION_MANAGER,
+        "open_session",
+        "a{sv}",
+        {"options"},
```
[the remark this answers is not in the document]
m-blaha: Actually we need 0.9.0 because of `signal.setDestination`. I'll bump it.

ppisar:
dnf5daemon-tests has not yet finished.
testing-farm:fedora-rawhide-x86_64:dnf5-tests fails on probably unrelated "dnf/zchunk.feature:141  using mirror wihtout ranges supports and zchunk results in only two GET requests for primary (the first try is with range specified)".

Waiting on the dnf5daemon-tests.

m-blaha:
Oh daemon tests do not run automatically (although they can now, I'll file a ticket).

ppisar (approved):
Thanks. dnf5daemon tests passed.

PR #1301, opened 2024-03-08
upgrade: check that `--destdir` is used only with `--downloadonly`

kontura:
If only `--destdir` is specified the command will download the pkgs to specified directory and then fail because transaction cannot find them (they are not in cache).

### DIFF
--- a/dnf5/commands/upgrade/upgrade.cpp
+++ b/dnf5/commands/upgrade/upgrade.cpp
@@ -68,6 +68,8 @@ void UpgradeCommand::set_argument_parser() {
     create_allow_downgrade_options(*this);
     create_downloadonly_option(*this);
     create_destdir_option(*this);
+    auto & destdir = parser.get_named_arg("upgrade.destdir", false);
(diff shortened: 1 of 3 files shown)


ppisar:
The patch is straightforward, but this new requirement should be documented. Could you please amend a documentation with the --downloadonly option and mention that at --destdir that it requires --downloadonly?

However, all that leads to a doubt: Does the user have to be aware about this peculiarity?  Wouldn't be better if the upgrade command automatically enabled --downloadonly if the user supplies --destdir? If you agree that it would be beneficial, could you reimplement this pull request in that way and document at --destdir that it implies --downloadonly?

ppisar:
By the way your current code fails without the options:

~~~~
# ~test/dnf5/redhat-linux-build/dnf5/dnf5 --disable-repo=rawhide --enable-repo=f41-build upgrade
Option value is not set
~~~~

Maybe one needs to check an OptionString with .get_priority() == Priority::EMPTY to distinguish an empty string from an undefined value.

ppisar:
Great. The code and documentation look and behave well.
Waiting on finish dnf5daemon-tests CI check.

Issue #1304, opened 2024-03-12
Do not export internal symbols in shared object files (libraries)


jrohel:
Now, all symbols that are not marked static or placed in an anonymous namespace are exported in a shared object. Thus, internal symbols are exported as ABI.
- Users can use interfaces they should not use
- Internal library changes change exported symbols - ABI compatibility problem.
- The dynamic symbol table and its string table are available at run-time and therefore must be loaded. This require memory. And CPU time at startup (relocations, ...).

PR #1307, opened 2024-03-12
Do not export internal symbols in shared object files

jrohel:
Closes https://github.com/rpm-software-management/dnf5/issues/1304

PR disables C++ unit tests for libdnf5, libdnf5-cli, dnf5 cpr_plugin C++. These unit tests use private symbols that are no longer accesible. It is a temporarily solution until the unit tests are modified.


include/libdnf5/defs.h
```diff
@@ -0,0 +1,25 @@
...
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef LIBDNF5_DEFS_H
+#define LIBDNF5_DEFS_H
+
+#define PUBLIC_API __attribute__((visibility("default")))
```
ppisar, line 23: __attribute__((visibility())) is a compiler-specific extension. E.g. GCC and Clang have them. Other compilers might not have. Consider guarding the definition with __GNUC__ preprocessor condition.

bindings/libdnf5/advisory.i
```diff
@@ -38,6 +38,9 @@
 
 #define CV __perl_CV
 
+#define LIBDNF_API
+#define LIBDNF_LOCAL
```
jan-kolarik, line 42: Can we define it once, e.g. in `shared.i`?

dnf5/include/dnf5/context.hpp
```diff
@@ -249,25 +250,25 @@ class RpmTransCB : public libdnf5::rpm::TransactionCallbacks {
     void verify_stop([[maybe_unused]] uint64_t total) override;
 
 private:
-    void new_progress_bar(int64_t total, const std::string & descr);
+    DNF_LOCAL void new_progress_bar(int64_t total, const std::string & descr);
```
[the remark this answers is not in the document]
jrohel: For private classes, the symbols are hidden by default. But this is a public API class. It is defined by `class DNF_API class_name`. Such a class has symbols public (exported) by default. That's why I explicitly set DNF_LOCAL for private methods.

bindings/libdnf5/advisory.i
jrohel: OK. I moved the macro definition to `shared.i`.

jan-kolarik:
Thanks, looks good. Just a rebase is probably needed to fix tests.

jrohel:
The Locker class has been moved to the public API in upstream.
So, I did a rebase and modification of this PR.

jan-kolarik:
The current tests issues are connected with the new rpm `4.19.91-6` in Rawhide. New `4.19.91-7` version should fix the issues (tested locally). Waiting to get into compose and we can re-run the tests...

jan-kolarik:
Sorry about that, but could you rebase again please? From the other PRs, it seems that the CI is finally working again...

PR #1308, opened 2024-03-13
modules: Report switching module streams

pkratoch:
Replacement for https://github.com/rpm-software-management/dnf5/pull/1285 , but opened against the dnf5-5.2.0.0 branch.

The tests are still here: https://github.com/rpm-software-management/ci-dnf-stack/pull/1462

### DIFF
--- a/libdnf5/base/transaction_module.cpp
+++ b/libdnf5/base/transaction_module.cpp
@@ -19,23 +19,11 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5/base/transaction_module.hpp"
 
+#include "transaction_module_impl.hpp"
+
 #include <utility>
 
 namespace libdnf5::base {
-
-class TransactionModule::Impl {
-public:
-    Impl(std::string module_name, std::string module_stream, Action action, Reason reason);
-
-private:
-    friend TransactionModule;
-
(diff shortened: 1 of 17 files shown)


kontura:
@pkratoch it seems a rebase is needed.

libdnf5/base/transaction_module.cpp
```diff
@@ -34,7 +34,8 @@ class TransactionModule::Impl {
     Action action;
     Reason reason;
     std::string module_name;
-    std::string module_stream;
+    std::string source_stream;
+    std::string target_stream;
```
kontura, line 38: Looking at https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_obsoletes_v1.yaml I think it would be possible that a `TransactionModule` should change to a different `module:stream` (not only a stream).

What about to use similar approach like `TransactionPackage`, add something like `get_replaces()` or `get_replaced_by()`?

pkratoch: Thanks for the suggestion! I am not quite sure how the obsoletes will be implemented, but I modified the `TransactionModule` similarly to the `TransactionPackage`. The `get_replaced_by()` is not needed for stream switching, but I added it, too, for later use for obsoletes.

kontura:
Thank you

PR #1309, opened 2024-03-13
Don't set obsoletes < 5 for ELN/RHEL yet

sgallagher:
The necessary pieces are not yet in place for RHEL 11 and this results in conflicts.

screenshot

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -2,7 +2,7 @@
 %global project_version_minor 1
 %global project_version_patch 14
 
-%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 41 || 0%{?rhel} > 10]
+%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 41 || 0%{?rhel} > 11]
 
 Name:           dnf5
 Version:        %{project_version_major}.%{project_version_minor}.%{project_version_patch}


ppisar:
Can you tell us what are "The necessary pieces are not yet in place for RHEL 11" so that we know when we can do the switch?

Besides your patch is not syntactically correct and all builds fail ("macro expansion did not return an integer:  (%2)
"), I'd rather prefer changing it into "%[0%{?fedora} > 41 || 0%{?rhel} > 11]". We will decrease it to 10 when "the necessary pieces" are there.

sgallagher:
> Guessing from the screenshot, it's that comps groups list "yum" as a mandatory package?

I honestly don't know what the "necessary pieces" are, but since they're not there for `f40`, they are similarly not ready for ELN/RHEL 11. So if you want to make it `> 11` that works fine for me. I'll push a revision.

sgallagher:
For more context: we want DNF5 to be available in Fedora ELN because we want to switch to it for mock/Koji build purposes. We'll move to it as a replacement for DNF4 at the same time Rawhide does.

ppisar:
I'm unable to reproduce the error with a dnf5 nightly build. The ODCS compose mentioned in the screenshot does not contain dnf5 package. Koji build root for ELN contains dnf5 and it obsoletes yum, but installing "yum", or "yum dnf5" does not fail. If I add a repository with a group requiring "yum", installing that group also pass.

Isn't the error a feature of Anaconda?

ppisar:
The new patch looks good. After finishing rpm-build:fedora-rawhide-x86_64 test, I will merge it.
I guess you want it as soon as possible in Rawhide to propagate it to ELN.

Conan-Kudo:
I know that a bunch of things choke on the fact we're using Provides instead of subpackages. I've been tempted to break out the dnf, microdnf, and yum provides back into subpackages to deal with it... 😕

sgallagher:
I can't find the specific reasons this is failing and we need to get containers building again for ELN. Let's merge this and do a build and we'll figure the rest out alongside Rawhide.

ppisar:
Merged. I'm logging out now. I will do the Fedora build on Monady, unless somebody else does it sooner.

ppisar:
This change will be in dnf5-5.1.15-2.fc41 which is now building.

PR #1310, opened 2024-03-14
PackageQuery: Source rpms do not upgrade binary ones

m-blaha:
Noarch packages can be used to upgrade any binary packages, but not source
packages. Fixes incorrect filter_upgrades() behaviour where src packages
were returned as upgrade candidates for noarch packages.

Tests adjustment: https://github.com/rpm-software-management/ci-dnf-stack/pull/1472

### DIFF
--- a/libdnf5/rpm/package_query.cpp
+++ b/libdnf5/rpm/package_query.cpp
@@ -135,7 +135,8 @@ inline libdnf5::sack::QueryCmp remove_glob_when_unneeded(
  *    other (i.e. i386 package can not be upgraded to x86_64, neither the other
  *    way round). If one of them is noarch and the other one colored then the
  *    pkg is upgradable (i.e. one can upgrade .noarch to .x86_64 and then again
- *    to a new version that is .noarch)
+ *    to a new version that is .noarch). Source rpms cannot be used to upgrade
+ *    any binary rpm.
  * :: is of lower version than pkg.
  * :: if there are multiple packages of that name return the highest version
  *    (implying we won't claim we can upgrade an old package with an already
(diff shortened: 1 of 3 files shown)


libdnf5/rpm/package_query.cpp
```diff
@@ -159,8 +160,14 @@ Id what_upgrades(libdnf5::solv::RpmPool & spool, const Solvable * solvable) {
...
-        if (updated->arch != solvable->arch && updated->arch != ARCH_NOARCH && solvable->arch != ARCH_NOARCH) {
-            continue;
+        if (updated->arch != solvable->arch) {
+            if (solvable->arch == ARCH_SRC || solvable->arch == ARCH_NOSRC || updated->arch == ARCH_SRC ||
+                updated->arch == ARCH_NOSRC) {
+                continue;
+            }
+            if (updated->arch != ARCH_NOARCH && solvable->arch != ARCH_NOARCH) {
+                continue;
+            }
```
j-mracek, line 170: LGTM, this condition is really good one. Interesting thing is that this condition will still create incorrect result for packages with incompatible architecture.

When installed package is `pkgA-1-1.noarch` and available are `pkgA-2-1.x86_64` and `pkgA-2-1.s390` both will be considered as upgrades regardless of system architecture. Fortunately it requires 2 rare conditions. I do not think that we should resolve the issue, because for the final decision we have a solver.

j-mracek (approved):
LGTM

libdnf5/rpm/package_query.cpp
m-blaha: Well, although the `.s390` version is not installable, it still is (unlike the source rpm) a possible upgrade to a `.noarch` package. It depends on what we want from `filter_upgrades()` to exactly do... Btw we need to document this filter (I've just checked the `package_query.hpp` header file and found that the documentation is virtually missing).

PR #1314, opened 2024-03-15
conf: Fix append options overrides

m-blaha:
Fixes the use case when an append option is initialized with higher
priority, and user tries to change it with lower priority. This should
not be allowed.

Example of incorrect behavior without this patch:

$ cat /etc/dnf/dnf.conf:
[main]
excludepkgs=kernel-core,acpi

$ dnf5 install acpi --exclude=
Repositories loaded.
Argument 'acpi' matches only excluded packages.

For: https://github.com/rpm-software-management/dnf5/issues/1331

### DIFF
--- a/libdnf5/conf/config_utils.hpp
+++ b/libdnf5/conf/config_utils.hpp
@@ -28,11 +28,13 @@ namespace libdnf5 {
 
 template <typename T>
 static void option_T_list_append(T & option, Option::Priority priority, const std::string & value) {
+    if (priority < option.get_priority()) {
+        return;
+    }
     if (value.empty()) {
         option.set(priority, value);
         return;
     }
-    auto add_priority = priority < option.get_priority() ? option.get_priority() : priority;
     auto val = option.from_string(value);
     bool first = true;
     for (auto & item : val) {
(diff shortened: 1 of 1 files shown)


m-blaha:
Unfortunately this change is not enough. Although this fixes resetting of append options, it breaks the other functionality of append options - ability to add new items to the configured ones from the command line.

Currently the order of config loading in dnf5 is following:

1. command line options are set to `config` object directly during parsing using COMMANDLINE priority
2. then options from the config files are loaded using MAINCONFIG / REPOCONFIG priority (which is lower then commandline)

So before the fix reset was broken due to not respecting priorities (config options with lower priority were actually added to command line values with higher priority).
And after the priorities fix the adding of new values is broken due to the loading order.

It makes me wonder whether introducing the `append options` was the best choice...

m-blaha:
Well, we have (there are four of five scenarios currently disabled for dnf5 like this one - https://github.com/rpm-software-management/ci-dnf-stack/blob/2c61e632ebaaf4613c6bb2767d38fab3bc87f9c2/dnf-behave-tests/dnf/append_options.feature#L95). But this fix introduces another issue so I tend to close the PR in favor of a proper fix in the future.

Issue #1316, opened 2024-03-15
Unable to build dnf5-5.2 locally when dnf5-5.1 is installed (failing unit-tests)


j-mracek:
How to reproduce the issue on Fedora 38:

Download branch https://github.com/rpm-software-management/dnf5/tree/dnf5-5.2.0.0
run `$ tito build --test --rpm`

Build does not work when dnf5-5.1.14-20240315004753.23.g663a2cfd.fc38 from               copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly is installed from https://copr.fedorainfracloud.org/coprs/jmracek/test-dnf5/

Build worked when dnf5-5.2.0.0-1.git.3087.063032f.fc38 was installed

Interesting thing is that copr builds passes even for Fedora 40 where dnf5-5.1.12-1.fc40.x86_64 was installed (https://download.copr.fedorainfracloud.org/results/jmracek/test-dnf5/fedora-40-x86_64/07163848-dnf5/builder-live.log.gz).
```
est project /tmp/tito/rpmbuild-dnf58n53jvfw/BUILD/dnf5-git-3087.063032f/redhat-linux-build
      Start  1: test_libdnf
 1/15 Test  #1: test_libdnf .................................................***Exception: SegFault  0.21 sec
ConfigParserTest::test_create_simple (duration: 0ms) : OK
ConfigParserTest::test_create_with_comments_header (duration: 0ms) : OK
ConfigParserTest::test_create_crazy (duration: 0ms) : OK
ConfigParserTest::test_parse_check_results_simple (duration: 0ms) : OK
ConfigParserTest::test_parse_check_results_with_comments_header (duration: 0ms) : OK
ConfigParserTest::test_parse_check_results_crazy (duration: 0ms) : OK
ConfigParserTest::test_read_write_simple (duration: 0ms) : OK
ConfigParserTest::test_read_write_with_comments_header (duration: 0ms) : OK
ConfigParserTest::test_read_write_crazy (duration: 0ms) : OK
ConfigParserTest::test_read_modify_write (duration: 0ms) : OK
OptionTest::test_options_bool (duration: 0ms) : OK
OptionTest::test_options_child (duration: 0ms) : OK
OptionTest::test_options_enum (duration: 0ms) : OK
OptionTest::test_options_number (duration: 0ms) : OK
OptionTest::test_options_path (duration: 0ms) : OK
OptionTest::test_options_seconds (duration: 0ms) : OK
[...]

PR #1327, opened 2024-03-19
Prevent loading plugins for unittests

j-mracek:
Loading plugins from host system is risky for both sides.

Closes: https://github.com/rpm-software-management/dnf5/issues/1316

### DIFF
--- a/test/dnf5daemon-server/support.py
+++ b/test/dnf5daemon-server/support.py
@@ -70,10 +70,12 @@ class InstallrootCase(unittest.TestCase):
         self.iface_session = dbus.Interface(
             self.bus.get_object(DNFDAEMON_BUS_NAME, DNFDAEMON_OBJECT_PATH),
             dbus_interface=IFACE_SESSION_MANAGER)
+        # Prevent loading plugins from host by setting "plugins" to False
         self.session = self.iface_session.open_session({
             "config": {
                 "config_file_path": self.config_file_path,
                 "installroot": self.installroot,
+                "plugins": False,
                 "cachedir": os.path.join(self.installroot, "var/cache/dnf"),
                 "reposdir": self.reposdir,
             }
--- a/test/perl5/libdnf5/rpm/test_package_query.t
+++ b/test/perl5/libdnf5/rpm/test_package_query.t
@@ -34,6 +34,9 @@ my $tmpdir = tempdir("libdnf5_perl5_unittest.XXXX", TMPDIR => 1, CLEANUP => 1);
 $base->get_config()->get_installroot_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/installroot");
 $base->get_config()->get_cachedir_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/cache");
 
+# Prevent loading plugins from host
+$base->get_config()->get_plugins_option()->set("False");
+
 # Sets base internals according to configuration
 $base->setup();
 
--- a/test/python3/libdnf5/tutorial/session/create_base.py
+++ b/test/python3/libdnf5/tutorial/session/create_base.py
@@ -11,6 +11,9 @@ base = libdnf5.base.Base()
 base_config = base.get_config()
 base_config.installroot = installroot
 
+# Optionally prevent of loading of plugins. Plugins are loaded by default from the host
+base_config.plugins = False
+
 # Optionally load configuration from the config files.
 #
 # The Base's config is initialized with default values, one of which is
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -133,6 +133,11 @@ void Base::Impl::with_config_file_path(std::function<void(const std::string &)> 
 }
 
 void Base::load_plugins() {
+    // load plugins according to configuration
+    if (!p_impl->config.get_plugins_option().get_value()) {
+        return;
+    }
+
     const char * plugins_config_dir = std::getenv("LIBDNF_PLUGINS_CONFIG_DIR");
     if (plugins_config_dir &&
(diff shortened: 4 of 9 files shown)


test/dnf5daemon-server/support.py
```diff
@@ -70,10 +70,13 @@ def setUp(self):
         self.iface_session = dbus.Interface(
             self.bus.get_object(DNFDAEMON_BUS_NAME, DNFDAEMON_OBJECT_PATH),
             dbus_interface=IFACE_SESSION_MANAGER)
+        # Prevent loading plugins from host by redirecting of pluginconfpath and pluginpath to an empty directory
         self.session = self.iface_session.open_session({
             "config": {
                 "config_file_path": self.config_file_path,
                 "installroot": self.installroot,
+                "pluginconfpath": os.path.join(self.installroot, "pluginconfpath"),
```
ppisar, line 78: If the directory is supposed to be empty, I would call it "empty" instead of "pluginconfpath" or "pluginpath".
Also if you do not create the directory and thus the directory does not exist, I would simply use "/dev/null" instead of an artificial path.

ppisar (approved):
Otherwise, I confirm this patch looks good and indeed fixes the bug.

test/dnf5daemon-server/support.py
ppisar: Unless you want to address this, feel free to merge your current, 95cf7d0121b98e3743867e9e943e2ab909a54716, pull request.

j-mracek: I will modify it and thank you very much for the suggestion.

j-mracek:
@ppisar I've modified the PR. May I ask you to look again to it?

test/dnf5daemon-server/support.py
```diff
@@ -70,10 +70,12 @@ def setUp(self):
         self.iface_session = dbus.Interface(
             self.bus.get_object(DNFDAEMON_BUS_NAME, DNFDAEMON_OBJECT_PATH),
             dbus_interface=IFACE_SESSION_MANAGER)
+        # Prevent loading plugins from host by redirecting of pluginconfpath and pluginpath to an empty directory
         self.session = self.iface_session.open_session({
             "config": {
                 "config_file_path": self.config_file_path,
                 "installroot": self.installroot,
+                "plugins": 0,
```
ppisar, line 78: Later, you use "false" instead of "0". Please unify it.

test/perl5/libdnf5/rpm/test_package_query.t
```diff
@@ -34,6 +34,10 @@ my $tmpdir = tempdir("libdnf5_perl5_unittest.XXXX", TMPDIR => 1, CLEANUP => 1);
 $base->get_config()->get_installroot_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/installroot");
 $base->get_config()->get_cachedir_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/cache");
 
+# Prevent loading plugins from host
+$base->get_config()->get_plugins_option()->set(0);
```
ppisar, line 38: Later you use "false" instead of "0". Please unify it.

test/python3/libdnf5/tutorial/session/create_base.py
```diff
@@ -1,4 +1,5 @@
 import libdnf5
+import os.path
```
ppisar, line 2: Remove importing "os.path". It's unnecessary.

ppisar:
I recommend squashing "Prevent loading plugins for unittests" and "Use a different method for disabling plugins during build" commits into one.

j-mracek:
@ppisar Done

j-mracek:
@ppisar Thank you for the review

test/dnf5daemon-server/support.py
```diff
@@ -70,10 +70,12 @@ def setUp(self):
         self.iface_session = dbus.Interface(
             self.bus.get_object(DNFDAEMON_BUS_NAME, DNFDAEMON_OBJECT_PATH),
             dbus_interface=IFACE_SESSION_MANAGER)
+        # Prevent loading plugins from host by redirecting of pluginconfpath and pluginpath to an empty directory
```
ppisar, line 73: This comment does not match the current code. Please remove it.

test/perl5/libdnf5/rpm/test_package_query.t
```diff
@@ -34,6 +34,10 @@ my $tmpdir = tempdir("libdnf5_perl5_unittest.XXXX", TMPDIR => 1, CLEANUP => 1);
 $base->get_config()->get_installroot_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/installroot");
 $base->get_config()->get_cachedir_option()->set($libdnf5::conf::Option::Priority_RUNTIME, $tmpdir."/cache");
 
+# Prevent loading plugins from host
+$base->get_config()->get_plugins_option()->set("False");
```
ppisar, line 38: Use <code>False</code> without quotes. As a string it evaluates to True.

ppisar:
Otherwise, it's good.

test/perl5/libdnf5/rpm/test_package_query.t
j-mracek: This is funny - set() accepts a string and then it converts it to bool. `OptionBool` config class does that and there is wide range of definition what false means.

ppisar: That's Perl. I thought it's Python. In Perl <code>"false"</code> means true, <code>false</code> is a Perl glob identifier (so called a bare word). In Perl, using <code>0</code> is the best option.

ppisar: I see, OptionBool::get_default_false_values() recognizes "false" string.

Issue #1131, opened 2024-01-04
Print a hint on transaction file dependency failure


evan-goode:
This is the DNF 5 equivalent of https://github.com/rpm-software-management/dnf/pull/2026.

DNF 5 does not load filelist metadata by default, so it will error when trying to install packages that depend on file paths (paths in `/usr/bin`, `/usr/sbin`, and `/etc` are OK, however). To the user, it may not be clear what the underlying issue is, see https://github.com/fedora-copr/copr/issues/3070. The current output includes:

```
Failed to resolve the transaction:
No match for argument: /usr/share/fonts/dejavu-sans-fonts/DejaVuSans-Bold.ttf
No match for argument: /usr/share/fonts/dejavu-sans-fonts/DejaVuSans.ttf
```

If a transaction fails due to a missing file dependency and filelists are not loaded, DNF 5 should print a helpful message to the user suggesting they try again with `--setopt=optional_metadata_types=filelists`.

evan-goode:
Somewhat related: https://github.com/rpm-software-management/dnf5/issues/1053

Issue #548, opened 2023-05-22
Add hints when transaction fails for DNF5 outputs


j-mracek:
In DNF4 there is a hint which option could be used to resolve the situation:

```
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
```

In DNF5 we have a new option `--skip-unavailable`.

The hint is dynamically generated because we should not suggest option that are not relevant for the current situation.

It would be good to suggest or document to not use all option at the same time. When user use `--allow-erasing` with `--skip-broken` and `best` then the valid solution range is from do nothing (sucess) to erase the whole system, or do nothing with error.

evan-goode:
The hint would be great to have in DNF 5. A couple participants in the 2023-08-11 test week ran into "conflicting requests" errors where `--allowerasing` would have helped.

jan-kolarik:
Another fresh request for displaying the `--allowerasing` hint for Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=2250559.

damaestro:
I was glad to see the option was available, but for new users this would just look like a broken transaction.

PR #1328, opened 2024-03-19
dnf5: Print command line hints after resolve failure

m-blaha:
Print to the user relevant hints about what command line options could
be used to fix transaction resolving issues.
It can suggest using --skip-unavailable, --no-best,
--allowerasing, --skip-broken, and --setopt=optional_metadata_types=filelists

Fixes: https://github.com/rpm-software-management/dnf5/issues/548
Fixes: https://github.com/rpm-software-management/dnf5/issues/1131

Because the PR changes the stderr output, some tests need adjustments: https://github.com/rpm-software-management/ci-dnf-stack/pull/1473

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -933,6 +933,106 @@ static void print_new_leaves(Context & context) {
 }  // namespace dnf5
 
 
+// Recursively search command and its parents whether given argument is configured
+static bool has_named_arg(libdnf5::cli::ArgumentParser::Command * command, std::string_view arg_name) {
+    while (command) {
+        for (const auto & arg : command->get_named_args()) {
+            if (arg->get_long_name() == arg_name) {
+                return true;
+            }
+        }
+        command = command->get_parent() == command ? nullptr : command->get_parent();
+    }
+    return false;
+}
+
+static void print_resolve_hints(dnf5::Context & context) {
+    auto & conf = context.base.get_config();
+    std::vector<std::string> hints;
+    auto transaction_problems = context.get_transaction()->get_problems();
+    auto * command = context.get_selected_command()->get_argument_parser_command();
+
+    // hint --skip-unavailable if a package was not found
+    if ((transaction_problems & libdnf5::GoalProblem::NOT_FOUND) == libdnf5::GoalProblem::NOT_FOUND &&
+        !conf.get_skip_unavailable_option().get_value()) {
+        const std::string_view arg{"--skip-unavailable"};
+        if (has_named_arg(command, arg.substr(2))) {
+            hints.emplace_back(libdnf5::utils::sformat(_("{} to skip unavailable packages"), arg));
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


dnf5/main.cpp
```diff
@@ -933,6 +933,107 @@ static void print_new_leaves(Context & context) {
...
+    }
+
+    if ((transaction_problems & libdnf5::GoalProblem::SOLVER_ERROR) == libdnf5::GoalProblem::SOLVER_ERROR) {
+        bool conflict = false;
+        bool broken_file_dep = false;
+        bool best = false;
+        // walk through all solver problem to detect a conflict, missing file dependency and best
+        for (const auto & resolve_log : context.get_transaction()->get_resolve_logs()) {
+            if (resolve_log.get_problem() == libdnf5::GoalProblem::SOLVER_ERROR ||
+                resolve_log.get_problem() == libdnf5::GoalProblem::SOLVER_PROBLEM_STRICT_RESOLVEMENT) {
```
j-mracek, line 971: May I ask you why you would like to analyze `libdnf5::GoalProblem::SOLVER_PROBLEM_STRICT_RESOLVEMENT` and not only `libdnf5::GoalProblem::SOLVER_ERROR`? When `libdnf5::GoalProblem::SOLVER_ERROR` is missing it means that current setting does not require any modification for the success?

m-blaha: Ah, thanks. I wanted to investigate what `STRICT_RESOLVEMENT` actually means before pushing, but I forgot. Will do now.

j-mracek: In some cases (Fedora or best-false) dnf upgrade might provide a warning (problem - libdnf5::GoalProblem::SOLVER_PROBLEM_STRICT_RESOLVEMENT) due to missing file dependencies. This is only case when a hint might make sense. But I am not sure, because the warning should be scary enough but without a hint.

m-blaha: Thanks, you are right. I will remove `SOLVER_PROBLEM_STRICT_RESOLVEMENT` solver problems from analysis.

dnf5/main.cpp
```diff
@@ -933,6 +933,106 @@ static void print_new_leaves(Context & context) {
...
+                                break;
+                            case libdnf5::ProblemRules::RULE_PKG_NOTHING_PROVIDES_DEP:
+                                if (params.size() >= 1) {
+                                    if (params[0].starts_with('/')) {
+                                        broken_file_dep = true;
+                                    }
+                                }
+                                break;
+                            case libdnf5::ProblemRules::RULE_BEST_1:
+                            case libdnf5::ProblemRules::RULE_BEST_2:
```
j-mracek, line 987: Optional: I am proposing to add one hint more for `package dnf-data-4.19.0-20240326004737.13.ga6d82221.fc38.noarch is filtered out by exclude filtering`.

j-mracek:
LGTM

j-mracek (approved):

PR #1329, opened 2024-03-20
Update `package_info_sections` not to use `scols_table_print_range`

kontura:
`scols_table_print_range` is slow when the table is large (when the
table contains info for all available packages). Since we need to print
an empty line between the packages the only option is to print each
package as a separate table.

To achieve that and keep the class compatible with
`package_list_sections` it needed to be changed.

`package_list/info_sections` still take a `PackageSet` when adding a
section, this could be updated to a more general vector of IPackage
which could also be used from dnf5daemon but the sets would require
converting and I am not certain it will be needed.

### DIFF
--- a/dnf5/commands/check-upgrade/check-upgrade.cpp
+++ b/dnf5/commands/check-upgrade/check-upgrade.cpp
@@ -99,7 +99,6 @@ void CheckUpgradeCommand::configure() {
 
 std::unique_ptr<libdnf5::cli::output::PackageListSections> CheckUpgradeCommand::create_output() {
     auto out = std::make_unique<libdnf5::cli::output::PackageListSections>();
-    out->setup_cols();
     return out;
 }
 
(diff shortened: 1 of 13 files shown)


m-blaha (changes requested):
Thank you, the `dnf5 info` is now blazing fast! :)
Only one small thing - the output packages in `dnf5 info` are not separated by an empty line. Could you please fix that?

m-blaha:
It looks like we do not have any tests for `dnf info` in our CI, so the missing newline hasn't been caught.

kontura:
> Thank you, the `dnf5 info` is now blazing fast! :) Only one small thing - the output packages in `dnf5 info` are not separated by an empty line. Could you please fix that?

I have missed that, should be fixed now, thanks!


> It looks like we do not have any tests for `dnf info` in our CI, so the missing newline hasn't been caught.

I have adapted some of the `list` tests for `info`: https://github.com/rpm-software-management/ci-dnf-stack/pull/1476

I have also added a commit with an `update` alias, to match the `list` command.

Issue #1322, opened 2024-03-18
Hide libdnf5::repo::Repo::load() for dnf5-5.2.0.0


jan-kolarik:
This is a follow-up to #1294.

The method is still used by the main, context and dnf5daemon. Also the way how API users should proceed to load only the system repo should be discussed.

kontura:
PR: https://github.com/rpm-software-management/dnf5/pull/1332

PR #1332, opened 2024-03-20
Make `libdnf5::repo::Repo::load()` private

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/1322

### DIFF
--- a/test/libdnf5/comps/test_environment.cpp
+++ b/test/libdnf5/comps/test_environment.cpp
@@ -66,8 +66,8 @@ void CompsEnvironmentTest::test_load_defaults() {
 
 void CompsEnvironmentTest::test_merge() {
     // Load multiple different definitions of the minimal environment
-    add_repo_repomd("repomd-comps-minimal-environment");
-    add_repo_repomd("repomd-comps-custom-environment");
+    add_repo_repomd("repomd-comps-minimal-environment", false);
+    add_repo_repomd("repomd-comps-custom-environment", false);
     add_repo_repomd("repomd-comps-minimal-environment-v2");
 
     // The "Minimal Install v2" is preferred because its repoid is alphabetically higher
@@ -86,8 +86,8 @@ void CompsEnvironmentTest::test_merge() {
 void CompsEnvironmentTest::test_merge_when_different_load_order() {
     // Load multiple different definitions of the minimal environment
     // The order of loading the repositories does not matter
-    add_repo_repomd("repomd-comps-minimal-environment-v2");
-    add_repo_repomd("repomd-comps-custom-environment");
+    add_repo_repomd("repomd-comps-minimal-environment-v2", false);
+    add_repo_repomd("repomd-comps-custom-environment", false);
     add_repo_repomd("repomd-comps-minimal-environment");
 
     // The "Minimal Install v2" is preferred because its repoid is alphabetically higher
@@ -106,8 +106,8 @@ void CompsEnvironmentTest::test_merge_when_different_load_order() {
 
 void CompsEnvironmentTest::test_merge_with_empty() {
     // Load minimal environment and another definition with all attributes empty
-    add_repo_repomd("repomd-comps-minimal-environment");
-    add_repo_repomd("repomd-comps-custom-environment");
+    add_repo_repomd("repomd-comps-minimal-environment", false);
+    add_repo_repomd("repomd-comps-custom-environment", false);
     add_repo_repomd("repomd-comps-minimal-environment-empty");
 
     // All the attributes are taken from the non-empty definition
@@ -127,8 +127,8 @@ void CompsEnvironmentTest::test_merge_with_empty() {
 
 void CompsEnvironmentTest::test_merge_empty_with_nonempty() {
     // Load minimal environment and another definition with all attributes empty
-    add_repo_repomd("repomd-comps-minimal-environment-empty");
-    add_repo_repomd("repomd-comps-custom-environment");
+    add_repo_repomd("repomd-comps-minimal-environment-empty", false);
+    add_repo_repomd("repomd-comps-custom-environment", false);
     add_repo_repomd("repomd-comps-minimal-environment");
 
     // All the attributes are taken from the non-empty definition
@@ -148,7 +148,7 @@ void CompsEnvironmentTest::test_merge_empty_with_nonempty() {
 
 void CompsEnvironmentTest::test_merge_different_translations() {
     // Load different definitions of the minimal environment with different set of translations
-    add_repo_repomd("repomd-comps-minimal-environment");
+    add_repo_repomd("repomd-comps-minimal-environment", false);
     add_repo_repomd("repomd-comps-minimal-environment-different-translations");
 
     EnvironmentQuery q_minimal_env(base);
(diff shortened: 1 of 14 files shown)


jan-kolarik:
As we are already introducing breaking changes, what about changing the signature to accept a single parameter with an enum specifying which repositories we want to load: only system / only available / all (default)? This way, users or other methods don't have to execute the repos disable method in case only the system repo is wanted.

dnf5/context.cpp
```diff
@@ -560,36 +560,29 @@ std::vector<std::string> match_specs(
...
             for (auto & repo : enabled_repos.get_data()) {
                 repo->set_sync_strategy(libdnf5::repo::Repo::SyncStrategy::ONLY_CACHE);
                 repo->get_config().get_skip_if_unavailable_option().set(libdnf5::Option::Priority::RUNTIME, true);
             }
-
-            ctx.load_repos(false);
-        } catch (...) {
-            // Ignores errors when completing available packages, other completions may still work.
+        } else {
+            std::for_each(enabled_repos.begin(), enabled_repos.end(), [](auto & repo) { repo->disable(); });
```
jan-kolarik, line 580: As we are already introducing breaking changes, what about changing the signature of RepoSack::update_and_load_enabled_repos to accept a single parameter with an enum specifying which repositories we want to load: only system / only available / all (default)? This way, users or other methods don't have to execute the repos disable method in case only the system repo is wanted.

kontura:
The PR got quite a bit bigger, in addition to hiding `libdnf5::repo::Repo::load()` it also hides `libdnf5::repo::RepoSack::update_and_load_repos(...)` and it adds a new overload for `update_and_load_enabled_repos` that takes a vector of `Repo::Type`.

I have also added some unit tests. It is working in bindings as well but currently it has to be called like this:
```python
repo_sack.update_and_load_enabled_repos(libdnf5.repo.VectorRepoType(1, libdnf5.repo.Repo.Type_AVAILABLE))
```

If this gets merged I would like to create a follow up issue for the bindings to allow automatic conversion from a list:
```python
repo_sack.update_and_load_enabled_repos([libdnf5.repo.Repo.Type_AVAILABLE])
```
(This is already an issue for https://github.com/rpm-software-management/dnf5/blob/dnf5-5.2.0.0/test/python3/libdnf5/rpm/test_package_query.py#L127-L128)

jan-kolarik:
> I have also added some unit tests. It is working in bindings as well but currently it has to be called like this:

Have you tried `repo_sack.update_and_load_enabled_repos((libdnf5.repo.Repo.Type_AVAILABLE,))`? I remember dealing with a similar problem when implementing `optional_metadata_types`. See also [here](https://github.com/rpm-software-management/dnf5/blob/5.1.15/test/python3/libdnf5/conf/test_option.py#L42).

kontura:
> > I have also added some unit tests. It is working in bindings as well but currently it has to be called like this:
> 
> Have you tried `repo_sack.update_and_load_enabled_repos((libdnf5.repo.Repo.Type_AVAILABLE,))`? I remember dealing with a similar problem when implementing `optional_metadata_types`. See also [here](https://github.com/rpm-software-management/dnf5/blob/5.1.15/test/python3/libdnf5/conf/test_option.py#L42).

I tried it now and unfortunately it doesn't work either.

jan-kolarik:
> > > I have also added some unit tests. It is working in bindings as well but currently it has to be called like this:
> > 
> > 
> > Have you tried `repo_sack.update_and_load_enabled_repos((libdnf5.repo.Repo.Type_AVAILABLE,))`? I remember dealing with a similar problem when implementing `optional_metadata_types`. See also [here](https://github.com/rpm-software-management/dnf5/blob/5.1.15/test/python3/libdnf5/conf/test_option.py#L42).
> 
> I tried it now and unfortunately it doesn't work either.

Or maybe `[libdnf5.repo.Repo.Type_AVAILABLE,]`? The comma at the end is important in this case.

kontura:
> Also I see that the current approach is not using the enum as we've discussed, but a vector. I don't see it's a bad idea, but from the first sight, it seems quite a lot of additional code is needed because of that. Have you considered using new enum for this purpose or extending the `Repo::Type` with e.g. `ANY` value which could be the default one?

I did start with a new enum but it didn't seem right to me because it was duplicating the meaning of the existing `Repo::Type` and `update_and_load_enabled_repos` basically needed to convert between them in order to know which repos to select for `update_and_load_repos`.

At the same time I am not sure about adding `ANY`, that is not really a `Repo::Type`, If we decide to add such value I would rather create a new enum.

I was also considering making the `Repo::Type` a bitwise enum so that the values could be ORed but I thought this vector of enums approach is already used in the API however I can't find it anywhere now (apart from the `filter_type` I added)., so perhaps I was mistaken.

jan-kolarik:
> > Also I see that the current approach is not using the enum as we've discussed, but a vector. I don't see it's a bad idea, but from the first sight, it seems quite a lot of additional code is needed because of that. Have you considered using new enum for this purpose or extending the `Repo::Type` with e.g. `ANY` value which could be the default one?
> 
> I did start with a new enum but it didn't seem right to me because it was duplicating the meaning of the existing `Repo::Type` and `update_and_load_enabled_repos` basically needed to convert between them in order to know which repos to select for `update_and_load_repos`.
> 
> At the same time I am not sure about adding `ANY`, that is not really a `Repo::Type`, If we decide to add such value I would rather create a new enum.
> 
> I was also considering making the `Repo::Type` a bitwise enum so that the values could be ORed but I thought this vector of enums approach is already used in the API however I can't find it anywhere now (apart from the `filter_type` I added)., so perhaps I was mistaken.

And what about creating two overloads:
- `update_and_load_enabled_repos` with no argument for loading all enabled repos
- `update_and_load_enabled_repos` with the `Repo::Type` argument to filter only available repos or system repo

jan-kolarik:
> And what about creating two overloads:
> 
> * `update_and_load_enabled_repos` with no argument for loading all enabled repos
> * `update_and_load_enabled_repos` with the `Repo::Type` argument to filter only available repos or system repo

Trying to kickoff the discussion :slightly_smiling_face: 
So as this is one of the main use cases, what about to also simplify things and use a new name `load_repos`?

m-blaha:
My thoughts about it:

- `update_and_load_enabled_repos` really feels like unnecessarily complicated name. I concur renaming it to something simpler, more straightforward, and more memorable like `load_repos`
- btw, is there a use-case to load disabled repos?
- I'd like to have an overload without any parameter (or with a default value) which would load all enabled repos and a system repo
- I'd prefer to convert Repo::Type to a bitwise enum and pass the ORed value instead of a list/tuple of values. Of course if this approach doesn't complicate things on bindings side. It's quite commonly used across the code (e.g. GoalProblem, ExcludeFlags... - just search for `<< 1` in the `include` dir)

kontura:
> And what about creating two overloads:
> 
>     * `update_and_load_enabled_repos` with no argument for loading all enabled repos
> 
>     * `update_and_load_enabled_repos` with the `Repo::Type` argument to filter only available repos or system repo

I have updated it according to your specification.

dnf5/main.cpp
```diff
@@ -1103,9 +1103,7 @@ int main(int argc, char * argv[]) try {
                 if (context.get_load_available_repos() != dnf5::Context::LoadAvailableRepos::NONE) {
                     context.load_repos(context.get_load_system_repo());
                 } else if (context.get_load_system_repo()) {
-                    repo_sack->get_system_repo()->load();
-                    // TODO(lukash) this is inconvenient, we should try to call it automatically at the right time in libdnf
-                    context.base.get_rpm_package_sack()->load_config_excludes_includes();
```
jan-kolarik, line 1108: We don't need this now?

jan-kolarik: Oh I see, it's moved to another place.

jan-kolarik (approved):
Great and thanks for covering and extending the tests!

Issue #1337, opened 2024-03-21
Document --version option


ppisar:
dnf5(8) manual is missing --version option. It should be mention that it lists successfully loaded modules (discovered in isse #1335). It should also be mention that it is a terminating option, i.e. the program exits without performing any commands.

jan-kolarik:
Hi Petr, I have already covered this during working on #644. Will link this issue when publishing the PR, which should be tomorrow.

glum23:
In addition to that, `dnf5 --version ` significantly differs from `dnf --version` so it'd be nice to document it in https://dnf5.readthedocs.io/en/latest/changes.html

Issue #644, opened 2023-06-22
Doc: check that all command line options are documented in respective man pages


m-blaha:
Discovered via https://bugzilla.redhat.com/show_bug.cgi?id=2216660 (man dnf5-install lacks `--skip-unavailable` and `--skip-broken` description, but I suppose there is more missing.

glum23:
There is a lot of such missing options and the list of options is inconsistent in `man dnf5-<command>` and `dnf5 <command> --help`. E.g. in `man dnf5-upgrade` in comparison with `dnf5 upgrade --help` there are missing the following options:
~~~
--advisory=ADVISORY_NAME,...
--allow-downgrade
--bz=BUGZILLA_ID,...
--cve=CVE_ID,...
--downloadonly
--no-allow-downgrade
--skip-unavailable
~~~

jan-kolarik:
> (man dnf5-install lacks --skip-unavailable and --skip-broken description, but I suppose there is more missing.

This is already fixed by https://github.com/rpm-software-management/dnf5/pull/1036.

hedayat:
I just stumbled upon the missing `list` command in `man dnf5`!

PR #1342, opened 2024-03-25
Review of CLI man pages

jan-kolarik:
- Review and fix of missing commands and parameters in man pages
- Sync man pages with the `help` command output
- Improve `--version` parameter documentation

Closes https://github.com/rpm-software-management/dnf5/issues/644.
Closes https://github.com/rpm-software-management/dnf5/issues/1337.


doc/commands/module.8.rst
```diff
@@ -46,31 +48,44 @@ Subcommands
...
 
 Options
 =======
 
+``--skip-broken``
+    | Used with ``enable`` command to resolve any dependency problems by removing items that are causing problems from the transaction.
+
+``--skip-unavailable``
+    | Used with ``enable``, ``disable`` and ``reset`` commands to allow skipping modular dependencies that are not possible to resolve.
+    | All remaining items will be processed.
```
pkratoch: This sounds like this would allow skipping available modules with unavailable dependencies. But it should be the same as for RPM package: --skip-unavailable should skip modules that are not available in the repositories and --skip-broken should skip modules with dependency problems.

doc/commands/module.8.rst
```diff
@@ -28,13 +25,18 @@
 Synopsis
 ========
 
-``dnf5 module <subcommand> [options] [<module-spec>]``
+``dnf5 module <subcommand> [options] [<module-spec>..]``
```
pkratoch, line 28: There are 3 dots everywhere else (well, for `<package-spec>`, but I'd say `<module-spec>` is similar enough :D)

doc/commands/check-upgrade.8.rst
```diff
@@ -0,0 +1,93 @@
...
+========
+
+``dnf5 check-upgrade [options] [<package-spec>...]``
+
+
+Description
+===========
+
+Non-interactively checks for available updates of specified packages. If no ``<package-spec>``
+is provided, it verifies updates for the entire system.
```
pkratoch, line 35: To "verify updates" sounds very strange to me, so I'd suggest to use "check for" again. I can't think of any synonym that would fit well enough. But it you are sure about the "verify", feel free to keep it.

jan-kolarik: I agree, I also cannot think of anything else better than "check for" :)

doc/commands/module.8.rst
```diff
@@ -46,31 +48,57 @@ Subcommands
...
+``--skip-broken``
+    | Used with ``enable`` command to resolve any dependency problems by removing items that are causing problems from the transaction.
 
-Options
+``--skip-unavailable``
+    | Used with ``enable``, ``disable`` and ``reset`` commands to allow skipping modules that are not available in repositories.
+    | All remaining items will be processed.
 =======
+Options for ``list`` and ``info``
+=================================
```
pkratoch, line 70: The overline (`=======`) probably shouldn't be there. It results in "Title overline & underline mismatch." error when building the documentation.

pkratoch: Oh, and the `--skip-broken` and `--skip-unavailable` options are both in the "Subcommands" section and the "Options for ``enable``, ``disable``, ``reset``" section now. Looks like it's all a result of a merge conflict resolving?

jan-kolarik: Right, some oversight during merging. Should be OK now.

PR #1346, opened 2024-03-26
libdnf5::OptionEnum: remove template, add pImpl

kontura:

In next PRs I would like to add pImpl also to other `Option`* classes.

### DIFF
--- a/libdnf5/conf/option_enum.cpp
+++ b/libdnf5/conf/option_enum.cpp
@@ -22,169 +22,89 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/utils/bgettext/bgettext-mark-domain.h"
 
 #include <algorithm>
-#include <sstream>
 
 namespace libdnf5 {
 
-template <typename T>
-bool from_string(T & out, const std::string & in, std::ios_base & (*manipulator)(std::ios_base &)) {
-    std::istringstream iss(in);
-    return !(iss >> manipulator >> out).fail();
-}
+class OptionEnum::Impl {
+public:
+    Impl(std::string && default_value, std::vector<std::string> && enum_vals)
+        : enum_vals(std::move(enum_vals)),
+          default_value(default_value),
+          value(std::move(default_value)) {}
 
-template <typename T>
-OptionEnum<T>::OptionEnum(ValueType default_value, const std::vector<ValueType> & enum_vals)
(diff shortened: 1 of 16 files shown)


libdnf5/conf/option_enum.cpp
```diff
@@ -26,165 +26,86 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
-bool from_string(T & out, const std::string & in, std::ios_base & (*manipulator)(std::ios_base &)) {
-    std::istringstream iss(in);
-    return !(iss >> manipulator >> out).fail();
-}
-
-template <typename T>
-OptionEnum<T>::OptionEnum(ValueType default_value, const std::vector<ValueType> & enum_vals)
+class OptionEnum::Impl {
+public:
+    Impl(std::string default_value, const std::vector<std::string> & enum_vals)
```
jrohel, line 31: Replace `const std::vector<std::string> & enum_vals` -> `std::vector<std::string> && enum_vals`
`const` cannot be moved out.

libdnf5/conf/option_enum.cpp
```diff
@@ -26,165 +26,86 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        : enum_vals(std::move(enum_vals)),
+          default_value(default_value),
```
jrohel, line 33: Replace `std::string default_value` by `const std::string & default_value`. Pass by reference. no need to copy.
It is copied aggain on the following lines.

kontura: Thanks, updated.

libdnf5/conf/option_enum.cpp
```diff
@@ -22,169 +22,89 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+    std::string value;
+};
+
+OptionEnum::OptionEnum(std::string default_value, std::vector<std::string> enum_vals)
     : Option(Priority::DEFAULT),
-      enum_vals(enum_vals),
-      default_value(default_value),
-      value(default_value) {
+      p_impl(new Impl(std::move(default_value), std::move(enum_vals))) {
     test(default_value);
```
jrohel, line 53: Thanks for the changes.
Please replace `test(default_value);` by `test(p_impl->value);` as it is in my code. `default_value` has been moved to `p_impl->value` and no longer heva a valid value.

kontura: Thanks, I have updated it.

PR #1348, opened 2024-03-27
dnfdaemon: Add NEVRA field to advisory packages

m-blaha:
To improve user experience add also complete NEVRAs to returned
packages in advisory collections.

### DIFF
--- a/dnf5daemon-server/advisory.cpp
+++ b/dnf5daemon-server/advisory.cpp
@@ -76,6 +76,7 @@ KeyValueMapList collections_to_list(
             package["v"] = pkg.get_version();
             package["r"] = pkg.get_release();
             package["a"] = arch;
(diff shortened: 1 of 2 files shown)


m-blaha:
For the record - we also discussed another variant how to solve it - adding a new `packages_format` option to `Advisory.list()` call with allowed values `split` (default) and `combined`. With `split` the behavior remains the same - packages are returned as a dictionary with separate elements (`n`, `e`, `v`, `r`, `a`). With `combined`, there would be only `nevra` key in the dictionary.
But I eventually found this overly complicated and went with this simple solution.

mcrha:
Works for me (tested as well).

inknos:
>  But I eventually found this overly complicated and went with this simple solution.

yes

lgtm

inknos (approved):

Issue #1057, opened 2023-12-01
Deployment issues to be solved before switching the default from DNF to DNF5


jan-kolarik:
A follow-up to https://github.com/rpm-software-management/dnf5/issues/635. This is a tracking issue for unresolved problems discovered during previous switch to DNF5 in Fedora Rawhide:

- [dnf5 seems to default to skip_if_unavailable=False behavior](https://bugzilla.redhat.com/show_bug.cgi?id=2216205)
  - deliver releng config overrides
  - announce on devel list
- [`man dnf` does not work](https://bugzilla.redhat.com/show_bug.cgi?id=2216850)

Conan-Kudo:
We now use KIWI for building cloud images starting with F40, so now for cloud images and base containers to change, a pull request will be required for [fedora-kiwi-descriptions](https://pagure.io/fedora-kiwi-descriptions).

jan-kolarik:
> We now use KIWI for building cloud images starting with F40, so now for cloud images and base containers to change, a pull request will be required for [fedora-kiwi-descriptions](https://pagure.io/fedora-kiwi-descriptions).

Thanks for info, noted.

evan-goode:
`skip_if_unavailable=True` and `best=False` PR: https://src.fedoraproject.org/rpms/fedora-release/pull-request/331

PR #1350, opened 2024-03-27
Prepare bash-completion for switch to dnf5 as the default package manager

jan-kolarik:
For: https://bugzilla.redhat.com/show_bug.cgi?id=2223324.

Connected with:
- https://github.com/rpm-software-management/dnf/pull/2070

Related:
- https://github.com/rpm-software-management/dnf5/issues/1057

### DIFF
--- a/dnf5/bash-completion/dnf5
+++ b/dnf5/bash-completion/dnf5
@@ -7,4 +7,11 @@ _do_dnf5_completion()
     mapfile -t COMPREPLY <<<$("${1}" "--complete=${cword}" "${words[@]}")
 }
 
-complete -F _do_dnf5_completion -o nosort -o nospace dnf5
+complete_cmds="dnf5"
+dnf_target=$(readlink -f "/usr/bin/dnf")
+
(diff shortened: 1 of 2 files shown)


dnf5/bash-completion/dnf5
```diff
@@ -7,4 +7,11 @@ _do_dnf5_completion()
     mapfile -t COMPREPLY <<<$("${1}" "--complete=${cword}" "${words[@]}")
 }
 
-complete -F _do_dnf5_completion -o nosort -o nospace dnf5
+complete_cmds="dnf5"
+dnf_target=$(readlink -f "/usr/bin/dnf")
```
ppisar, line 11: Add "Requires: coreutils" into dnf5.spec for the readlink tool.

ppisar:
Otherwise, it looks good.

dnf5/bash-completion/dnf5
jan-kolarik: Thanks, I did the same change for the dnf pull request.

PR #1352, opened 2024-03-28
dnfdaemon: Enhance comps.Group.list() method

m-blaha:
New accepted options:

- with_hidden - boolean, defaults to false
  whether include hidden (uservisible=false) groups into output
- scope - string, accepted values "all" (default), "installed",
"available"
- group_with_id - boolean, defaults to true
  whether patterns should be matched against group ids
- group_with_name - boolean, defaults to false
  whether patterns should be matched against group names
- contains_pkgs - list of strings
  only return groups containing given packages

Change in behavior:

- in case the scope is "all" (default), duplicated groups are removed
from the output. Duplicities basically mean that the group is
installed and previously both installed and available groups were
returned.

Also adds comps.Groups interface to the D-Bus API documentation and adjusts dnf5daemon-client accordingly.

Tests: https://github.com/rpm-software-management/dnf5/pull/1352

### DIFF
--- a/dnf5daemon-client/commands/group/group_list.hpp
+++ b/dnf5daemon-client/commands/group/group_list.hpp
@@ -22,12 +22,44 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "commands/command.hpp"
 
+#include <libdnf5-cli/session.hpp>
 #include <libdnf5/conf/option_enum.hpp>
+#include <libdnf5/utils/bgettext/bgettext-lib.h>
 
 #include <memory>
 #include <vector>
 
 namespace dnfdaemon::client {
+
+class GroupAvailableOption : public libdnf5::cli::session::BoolOption {
+public:
+    explicit GroupAvailableOption(libdnf5::cli::session::Command & command)
+        : BoolOption(command, "available", '\0', _("Show only available groups."), false) {}
+};
+
+class GroupInstalledOption : public libdnf5::cli::session::BoolOption {
+public:
+    explicit GroupInstalledOption(libdnf5::cli::session::Command & command)
+        : BoolOption(command, "installed", '\0', _("Show only installed groups."), false) {}
+};
+
+class GroupHiddenOption : public libdnf5::cli::session::BoolOption {
+public:
+    explicit GroupHiddenOption(libdnf5::cli::session::Command & command)
+        : BoolOption(command, "hidden", '\0', _("Show also hidden groups."), false) {}
+};
+
+class GroupContainPkgsOption : public libdnf5::cli::session::AppendStringListOption {
+public:
+    explicit GroupContainPkgsOption(libdnf5::cli::session::Command & command)
+        : AppendStringListOption(
+              command,
+              "contains-pkgs",
+              '\0',
+              _("Show only groups containing packages with specified names. List option, supports globs."),
+              "PACKAGE_NAME,...") {}
+};
+
 
 class GroupListCommand : public DaemonCommand {
 public:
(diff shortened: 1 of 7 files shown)


dnf5daemon-client/wrappers/dbus_group_wrapper.hpp
```diff
@@ -51,7 +51,7 @@ class DbusGroupWrapper {
     std::string get_order() const { return rawdata.at("order"); }
     std::string get_langonly() const { return rawdata.at("langonly"); }
     // TODO(mblaha) proper installed value
-    bool get_installed() const { return false; }
+    bool get_installed() const { return rawdata.at("installed"); }
```
pkratoch, line 54: Can the TODO note be removed now?

dnf5daemon-client/commands/group/group_list.hpp
```diff
@@ -22,20 +22,56 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+};
+
+class GroupContainPkgsOption : public libdnf5::cli::session::AppendStringListOption {
+public:
+    explicit GroupContainPkgsOption(libdnf5::cli::session::Command & command)
+        : AppendStringListOption(
+              command,
+              "contains-pkgs",
+              '\0',
+              _("Show only groups containing packages with specified names. List option  , supports globs."),
```
pkratoch, line 59: There are two extra spaces in the description.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.comps.Group.xml
```diff
@@ -0,0 +1,61 @@
...
+        Get list of groups that match to given filters.
+
+        Following options and filters are supported:
+
+            - attributes: list of strings
+                list of group attributes that are returned
+            - group_with_id: bool (default true)
+                match patterns against ids of the groups
+            - group_with_name: bool (default false)
+                match patterns against names of the groups
```
pkratoch, line 42: I am a bit puzzled by the `group_with_id` and `group_with_name` option names, as I wouldn't intuitively know what they mean. The only thing similar I found were options for rpm filtering (`with_src`, `with_provides`, `with_filenames`...), but then there are similar ones for advisories that mean something different (`with_bz`, `with_cve`). Plus the `group_` prefix could be seen as "related to groups" or as a verb. What was the intention behind the naming?

pkratoch:
I have just three small notes above, otherwise this looks good, thanks!

dnf5daemon-client/commands/group/group_list.hpp
m-blaha: Thanks.

dnf5daemon-client/wrappers/dbus_group_wrapper.hpp
m-blaha: Sure, thanks.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.comps.Group.xml
m-blaha: Thanks for the feedback!
The option names originate in `ResolveSpecSettings` which has the same attributes. But thinking about it again... you are right that the names in this context are puzzling.
How about renaming `group_with_id` -> `match_group_id`, `group_with_name`->`match_group_name` ? If you have another/better idea, it'd be appreciated.

pkratoch: `match_group_id` and `match_group_name` sound great, I don't have anything better.

m-blaha:
@pkratoch I've committed the changes you requested. Hopefully you don't mind rebase...

PR #1362, opened 2024-04-02
Vars::unset: API method for removing variable

### LINKED ISSUE #1360: API: libdnf5::Vars: Missing method for variable removal
A new variable can be created by the API method `set`.
But the possibility to delete a variable is missing.

jrohel:
Closes #1360


m-blaha:
Would be nice to also add a unit test for it. Currently, the unit tests only cover reading variables from files and environment.

jrohel:
@m-blaha 
> Would be nice to also add a unit test for it. Currently, the unit tests only cover reading variables from files and environment.

Good idea. I added a commit with unit tests.

m-blaha (approved):
Thank you!

PR #1366, opened 2024-04-02
Prepare man pages for switch to dnf5 as the default package manager

jan-kolarik:
For: https://bugzilla.redhat.com/show_bug.cgi?id=2216850.

Requires:
- https://github.com/rpm-software-management/dnf/pull/2072
- https://github.com/rpm-software-management/dnf-plugins-core/pull/529

Related:
- https://github.com/rpm-software-management/dnf5/issues/1057

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -327,6 +327,52 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %{_mandir}/man5/dnf5.conf.5.*
 %{_mandir}/man5/dnf5.conf-todo.5.*
 %{_mandir}/man5/dnf5.conf-deprecated.5.*
+%if %{with dnf5_obsoletes_dnf}
+%{_mandir}/man8/dnf.8.*
+%{_mandir}/man8/dnf-advisory.8.*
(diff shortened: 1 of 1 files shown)


inknos:
tricky. one more thing: would it be ok to have `man dnf` and `man dnf5` pointing at the same manpage?

Because now the PR aims at:
- `man dnf` -> DNF5 man pages
- `man dnf5` -> nothing
- `man dnf4` -> DNF man pages
- `man dnf-copr` -> DNF5 plugins man pages
- `man dnf5-copr` -> nothing
- `man dnf4-copr` -> DNF plugins man pages

But what about having also:
- `man dnf5` -> DNF5 man pages
- `man dnf5-copr` -> DNF5 plugins man pages


we are providing a simlink anyway 🤔

inknos:
> Install dnf5 plugins dnf5 install -y dnf5-plugins

This step is faulty, I should install plugins from the built rpms, which have the `dnf5_obsoletes_dnf` flag on. It's not an issue

inknos:
good, we can think about [this comment](https://github.com/rpm-software-management/dnf5/pull/1366#issuecomment-2034019634) in the next step

PR #1369, opened 2024-04-03
dnf5daemon: Remove reposdir from allowed config overrides

m-blaha:
The option is potentially dangerous and can cause dnf5daemon-server to
block on malicious reposdir.


jan-kolarik (approved):
Thanks!

PR #1372, opened 2024-04-03
Add pImpl to libdnf5::Options*

kontura:

### DIFF
--- a/include/libdnf5/conf/option_string_list.hpp
+++ b/include/libdnf5/conf/option_string_list.hpp
@@ -45,11 +45,7 @@ public:
     OptionStringContainer(const ValueType & default_value, std::string regex, bool icase, std::string delimiters);
 
     OptionStringContainer(const OptionStringContainer & src);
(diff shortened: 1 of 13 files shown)


include/libdnf5/conf/option_string_list.hpp
```diff
@@ -117,47 +113,13 @@ class OptionStringContainer : public Option {
...
-    return " ,\n";
-}
-
-template <typename T>
-inline const char * OptionStringContainer<T>::get_delimiters() const noexcept {
-    return delimiters ? delimiters->c_str() : get_default_delimiters();
-}
+    class Impl;
+    ImplPtr<Impl> p_impl;
+};
```
jan-kolarik, line 122: I guess we can drop now the `<optional>` and `<regex>` includes?

jan-kolarik (approved):
LGTM, except the mentioned includes ...

kontura:
> It seems the `regex` reference is missing on more `copr`-related places...

It should be good to go now. :+1:

PR #1376, opened 2024-04-04
config: add search (se) and info (if) aliases

pennbauman:
Adds two command aliases present in dnf4:

- `se` for the `search` command
- `if` for the `info` command

### DIFF
--- a/dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
+++ b/dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
@@ -110,6 +110,13 @@ descr = "Alias for 'group'"
 group_id = 'commands-compatibility-aliases'
 complete = true
 
+['if']
+type = 'command'
+attached_command = 'info'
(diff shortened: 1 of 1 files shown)


ppisar:
What's the reason for adding the aliases?

ppisar:
> What's the reason for adding the aliases?

I see dnf4.

j-mracek:
The question is whether compatibility aliases with zypper should be maintained in the upstream of DNF5. DNF5 introduce aliases defined in config files to allow customization on distribution sides.

pennbauman:
These are aliases I personally use regularly, and I assume there are others who use them too. 

dnf5 already has several aliases in this style (eg `in` for `install`, `rm` for `remove`), so I would think it makes sense to include these too.

Conan-Kudo:
I agree as well. I personally regularly use the aliases we introduced in dnf4 and I would prefer that they be retained by default in dnf5.

Conan-Kudo (approved):

PR #1377, opened 2024-04-04
Add pImpl to the remaining various classes

kontura:
I think this could be the last PR for https://github.com/rpm-software-management/dnf5/issues/1025

In my option the rest of the classes are not suitable for pImpl. 
Please let me know if we need to add it on some additional place.

### DIFF
--- a/include/libdnf5/conf/config.hpp
+++ b/include/libdnf5/conf/config.hpp
@@ -33,9 +33,10 @@ namespace libdnf5 {
 /// Base class for configurations objects
 class Config {
 public:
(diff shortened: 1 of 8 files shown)


jan-kolarik (approved):

jan-kolarik:
Nice, LGTM! I briefly went through [the list](https://github.com/rpm-software-management/dnf5/issues/1025#issuecomment-2006405581) of remaining classes without pimpl and the rest looks fine by me. I believe we've covered everything important. 

Thanks for working on this! I understand this may have been quite a tedious task.

Issue #1349, opened 2024-03-27
dnf5daemon: Advisory::list D-Bus method does not filter


mcrha:
I'd like to get only advisories, which are updates for the user, severity is critical and I do not care of bz nor CVE. In python (modified upgrades_with_severities.py):
```py
options = {
    "advisory_attrs": [
                "severity",
                "collections",
            ],
    "availability": "updates",
    "severity": "critical",
    "with_bz": False,
    "with_cve": False,
}
``` 

Issues:
* I get the full list, not only critical advisories
* the returned `severity` has upper-case first letter, aka `None`, `Low`, `Moderate`,... - it's confusing, because the filter value, which is "all lower case"
* there are also a lot of `'applicability': <'unrelated'>` attributes on the packages (maybe all of them, I do not know, my list is huge) - no idea what they mean, I thought it's a typo from the `availability` filter, but probably not.
* I also tried `"availability": "installed",` with the same effect; is this one meant to show already locally installed advisories? That's something I do not care of at the moment (thus I might misuse it) ;)

m-blaha:
The `availability` filter basically follows `--all`, `--available`, `--installed`, and `--updates` options of `dnf5 advisory list` command. So see man dnf5-advisory for details.

There is a bug in the Advisory documentation - the option is actually called `severities` (in plural, as it accepts a list of strings). I will update the documentation accordingly.

Additionally, there is another bug related to case sensitivity - the severity value is stored in advisories in capitalized form. Although the filter is intended to accept lowercase values, it seems that it doesn't. I'll address this issue to ensure the filter works with any capitalization of the values.

To prevent similar issues in the future, I'll also add more tests for dnfdamon advisories.

m-blaha:
So to get what you wanted, you need `severities: ["Critical"]` for now.

mcrha:
It seems the `--updates` is the default. Maybe drop a note about it into the --help and the man page.

m-blaha:
Default availability value is "available", and `--available` is also default for `dnf5 advisory list` command. I've added a note to the documentation.
I fixed the case sensitive filtering and enhanced the documentation in PR https://github.com/rpm-software-management/dnf5/pull/1380.

PR #1380, opened 2024-04-05
dnfdaemon: Enhance Advisory interface

m-blaha:
- fix the documentation to contain correct option names
- make filtering by severities case insensitive

Fixes: https://github.com/rpm-software-management/dnf5/issues/1349


kontura:
> By the way, looking into the failed tests in [the testing farm](https://artifacts.dev.testing-farm.io/d47b70a8-9a7c-4374-8673-b2693873c250/), the `Scenario: create repository with zck metadata with bad package` test suggests there's a use-after-free involved, which would be good to fix (the `free(): unaligned chunk detected in tcache 2` error is suspicious).

Indeed, for the other failing test we have some more details in https://github.com/rpm-software-management/createrepo_c/issues/386.

dnf5daemon-server/services/advisory/advisory.cpp
```diff
@@ -73,7 +74,7 @@ libdnf5::advisory::AdvisoryQuery Advisory::advisory_query_from_options(
         // Filter by advisory severity
         if (!opt_severities.empty()) {
             auto advisories_severities = libdnf5::advisory::AdvisoryQuery(base);
-            advisories_severities.filter_severity(opt_severities);
+            advisories_severities.filter_severity(opt_severities, libdnf5::sack::QueryCmp::IEXACT);
```
pkratoch, line 77: This fixes the first letter capitalization requirement, however, it is now a bit inconsistent that the "severities" are case insensitive, while "availability" and "types" remain case sensitive. My question is if it shouldn't be all case insensitive. But it's just a minor thing.

pkratoch (approved):
Looks good.

Also, I saw that some CI tests that are still disabled for dnf5daemon are already passing, but your last commit says they were unstable and I am not sure what's the status now. Just mentioning it in case the unstability was resolved.

dnf5daemon-server/services/advisory/advisory.cpp
m-blaha: Thinking about it once more - what do you think about setting `libdnf5::sack::QueryCmp::IEXACT` as the default comparator directly in `include/libdnf5/advisory/advisory_query.hpp`? Could be done also for types.
That would fix also future potential issues with case sensitivity for API users.

pkratoch: I think it would make sense for those filters that accept values from predefined lists, which is the case for severities and types.

m-blaha: I changed default comparator in filter_severity and filter_type to IEXACT. Also availability is now handled as case insensitive.

pkratoch: Thank you!

PR #1384, opened 2024-04-06
dnf5daemon-server/dbus: Install config files into /usr

Conan-Kudo:
We should prefer to install DBUS policy configuration files into /usr rather than /etc.


m-blaha (approved):
Thank you!

PR #1386, opened 2024-04-06
Give inline methods hidden visibility by default

jrohel:
Hidden inline methods will not appear in the export table of a DSO (dynamic shared object) and will not require a PLT indirection when used within the DSO. 

According to the GCC man page:
Enabling  this  option  can have a dramatic effect on load and link times of a DSO as it massively reduces the size of the dynamic export table when the library  makes heavy use of templates.

So:
- it is optimization
- also hides our internal inline methos (Related to https://github.com/rpm-software-management/dnf5/issues/1304

And we can do this because we do not attempt to compare pointers to inline functions or methods where the  addresses  of  the two functions are taken in different shared objects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,9 @@ find_package(PkgConfig REQUIRED)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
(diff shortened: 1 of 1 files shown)


j-mracek:
May I ask you to add to the commit message why you want that change?

j-mracek (approved):
LGTM

Issue #883, opened 2023-09-11
Document new repository and configuration paths


Conan-Kudo:
With #813, #820, and #882 landing, we need to update our documentation on how configuring DNF works now, and what the purpose for various directories are.

PR #1356, opened 2024-03-28
doc: document config-manager plugin

jrohel:
Closes #1282

PR #1393, opened 2024-04-09
Document repo dirs

inknos:
resolves: https://github.com/rpm-software-management/dnf5/issues/883

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -1168,6 +1168,69 @@ the user file /etc/dnf/libdnf5.conf.d/60-something.conf):
 5. /usr/share/dnf5/libdnf.conf.d/90-something.conf
 6. /etc/dnf/dnf.conf
 
+Drop-in repo directories
+========================
+
+After the repository configurations are loaded other repo configurations can be overloaded from the directories
+:ref:`user repos override directory <user_repos_override_dir-label>`
+and :ref:`distribution repos override directory <distro_repos_override_dir-label>`.
+
+The format of the files inside the directories is the same as the format of the repository configuration files.
+The options in the overridden files can modify existing repos but cannot create new repositories.
+
+Override files support globs in the repository ID in order to support bulk modifications of repository parameters.
+
+The repository overrides are processed following this order:
+
+1. Files from ``/usr/share/dnf5/repos.override.d/`` and ``/etc/dnf5/repos.override.d/`` are loaded in an alphabetically
+   sorted list. In case files have the same name, the file from ``/etc/dnf5/repos.override.d/`` is used.
+   This implies the list has only unique filenames. This also implies that the repository configuration files can be
+   simply masked by creating a file with the same name in the ``/etc`` override directory.
+
+2. The options from the files are applied in the order they are loaded. The last option wins.
+
+
+Example configuration
+---------------------
+
+.. code-block::
+
+   # Enable `skip_if_unavailable` for all repositories
+   [*]
+   skip_if_unavailable = true
+
+   # And then disable `skip_if_unavailable` for repositories with id prefix "fedora"
+   [fedora*]
+   skip_if_unavailable = false
+
+Example of configuration files
+------------------------------
+
+This example shows the order in which override files are processed.
+
+Files with user repos overrides:
+
+- /etc/dnf/repos.overide.d/20-user-overrides.repo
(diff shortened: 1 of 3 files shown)


doc/dnf5.conf.5.rst
```diff
@@ -1141,32 +1147,150 @@ Users can define custom config options in this way.
...
+
+After the repository configurations are loaded other repo configurations can be overloaded from the directories :ref:`user repos override directory <user_repos_override_dir-label>` and :ref:`distribution repos override directory <distro_repos_override_dir-label>`.
+
+The format of the files inside the directories is the same as the format of the repository configuration files. The options in the overridden files can modify existing repos but cannot create new repositories.
+
+Override files support globs in the repository ID in order to support bulk modifications of repository parameters.
+
+The repository overrides are processed following this order:
+
+1. Files from ``/usr/share/dnf5/repos.override.d/`` and ``/etc/dnf5/repos.override.d/`` are loaded in an alphabetically sorted list. In case files have the same name, the file from ``/usr/share/dnf5/repos.override.d/`` is used. This implies the list has only unique filenames. This also implies that the repository configuration files can be simply masked by creating a file with the same name in the ``/usr`` override directory.
```
ppisar, line 1161: The order does not match "Example configuration" section. In the example 60-something2.repo is taken from /etc, while this paragraph reads "file from /usr[…] is used".

ppisar:
Otherwise, I did not spot any mistake. However, I don't feel qualified to approve it as I don't know whether it indeed matches the implemented, or at least the expected behavior.

doc/dnf5.conf.5.rst
jrohel: I have a PR https://github.com/rpm-software-management/dnf5/pull/1356 with the config-manager plugin documentation. I also mention there the order of application of repositories configuration overrides. Maybe it will help here.

inknos: Thanks. I'll take example from that PR. But overall, I think the content is right regarding the implementation. Am I saying wrong stuff? @jrohel

jrohel: > In case files have the same name, the file from ``/usr/share/dnf5/repos.override.d/`` is used.

No. The file from ``/etc/dnf5/repos.override.d/`` is used.

>  can be simply masked by creating a file with the same name in the ``/usr`` override directory.

No. ...in the ``/etc`` override directory.

The idea is that system configuration (in `/etc`, created by the system administrator) can mask distribution's default configuration (in the `/usr`).

Look at the CI test https://github.com/rpm-software-management/ci-dnf-stack/blob/6d80409e967b67d601d73223c0f39ee2e703dd50/dnf-behave-tests/dnf/config-repos-overrides.feature#L35

inknos: so, I fixed the docs, and I cross-referenced the plugin and this page.


A note on the general status of our docs: there is some duplicate information, which is not necessarily bad but it can be improved.

I would ask, in the future the doc team to review our dnf5 docs to uniform the style and give guidance.

inknos: For example, we have repetitions in the heading of some sections, or some subsections are not in the same place for plugins/commands. Also, the docs are missing cross-references in general, which brings me to another point that is inconsistency in the label naming. I feel like all the docs should be reviewed once dnf5 lands in rawhide and there is enough time to prioritize it.

Just a general thought, unrelated to this specific PR

doc/dnf5.conf.5.rst
```diff
@@ -1261,6 +1261,43 @@ Files
...
 
+
+Directories
+===========
+
+Directories are listed in the opposite order in which they are
+processed meaning that the first directory in each list
+has the highest priority in its category.
+
+Repository Directories
```
jrohel, line 1280: The repository configuration directories are processed in the following order: `/etc/yum.repos.d/`, `/etc/distro.repos.d/`, `/usr/share/dnf5/repos.d`.

The important thing is that, unlike overrides that modify an existing repository configuration, the configuration in these directories defines new repositories. And we cannot define multiple repositories with the same ID. If multiple repositories with the same ID are defined only the first one will be loaded. Loading the next one will log an error.

inknos: updated the docs

Conan-Kudo (approved):

PR #1399, opened 2024-04-11
dnfdaemon: Missing signal registration

m-blaha:
- correctly register "transaction_before_begin" and
  "transaction_after_complete" signals on Rpm interface
- add missing argument to the signals in introspection file


ppisar (approved):
Good. The failing dnf5 tests about modules are unrelated.

Issue #1187, opened 2024-01-20
dnfdaemon do_transaction crash


anaselli:
It seems that do_transaction does not return. In my test i was installing one package only (atanks)
```
Traceback (most recent call last):
  File "/usr/lib64/python3.10/pdb.py", line 1723, in main
    pdb._runscript(mainpyfile)
  File "/usr/lib64/python3.10/pdb.py", line 1583, in _runscript
    self.run(statement)
  File "/usr/lib64/python3.10/bdb.py", line 598, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/home/angelo/src/manatools/dnfdragora/bin/dnfdragora", line 85, in <module>
    main_gui.handleevent()
  File "/home/angelo/src/manatools/venv/lib/python3.10/site-packages/dnfdragora/ui.py", line 1583, in handleevent
    rebuild_package_list = self._manageDnfDaemonEvent()
  File "/home/angelo/src/manatools/venv/lib/python3.10/site-packages/dnfdragora/ui.py", line 2155, in _manageDnfDaemonEvent
    self._OnBuildTransaction(info)
  File "/home/angelo/src/manatools/venv/lib/python3.10/site-packages/dnfdragora/ui.py", line 1991, in _OnBuildTransaction
    self.backend.RunTransaction(sync=True)
  File "/home/angelo/src/manatools/venv/lib/python3.10/site-packages/dnfdragora/dnfd_client.py", line 1237, in RunTransaction
    self._run_dbus_sync('RunTransaction', options)
  File "/home/angelo/src/manatools/venv/lib/python3.10/site-packages/dnfdragora/dnfd_client.py", line 512, in _run_dbus_sync
    return func(*args)
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib64/python3.10/site-packages/dbus/connection.py", line 634, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

```

m-blaha:
Hi, would it be possible to have more details on this? It's not completely clear to me how to reproduce  the issue.

anaselli:
[...]

PR #1400, opened 2024-04-11
dnfdaemon: Catch timeout during CheckAuthorization

m-blaha:
- set a timeout for the user to enter the password (default 25 seconds
  seems to be too short)
- catch timeout error - otherwise the error is returned to the client
  causing the original call to fail with timeout

For example:

The user invokes the do_transaction() method on the Goal interface. The
first action of the do_transaction handler is to verify that the user is
authorized to run an RPM transaction. If an authentication agent is
available, this may result in prompting the user to enter their
password. If the user ignores the prompt, after a timeout period, the
result should be a "Not authorized" error.

However, without this patch, the result would be a "Method call timed
out," which is misleading. While the user called do_transaction(), the
timeout did not occur at this stage.

For: https://github.com/rpm-software-management/dnf5/issues/1187

### DIFF
--- a/dnf5daemon-server/session.cpp
+++ b/dnf5daemon-server/session.cpp
@@ -245,7 +245,6 @@ bool Session::check_authorization(
     const std::string destination_name = "org.freedesktop.PolicyKit1";
     const std::string object_path = "/org/freedesktop/PolicyKit1/Authority";
     const std::string interface_name = "org.freedesktop.PolicyKit1.Authority";
-    // allow polkit to ask user to enter root password
     auto polkit_proxy = sdbus::createProxy(connection, destination_name, object_path);
     polkit_proxy->finishRegistration();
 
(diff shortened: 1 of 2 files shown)


ppisar:
The code looks reasonable. Though I'd like to see documented the hard-coded 2-minute timeout in dnf5daemon_server(8) or dnf5daemon-dbus-api(8).

Also I'd like to try the patch, but I don't know how to configure polkit to ask a user for a password. Now it immediately rejects a non-superuser.

ppisar:
Put it short, please amend dnf5daemon_server(8) with a notice about the 2-minute timeout and I will approve this pull request.

anaselli:
Just a question, have you experienced that after a timeout the transaction would fail if the user gives the right password? Because I'm almost sure it worked for me, but the call returned "Method call timed out,", IIUC now it fails with "Not authorized", but if the transaction worked, that does not have a real sense to me...

anaselli:
> Just a question, have you experienced that after a timeout the transaction would fail if the user gives the right password? Because I'm almost sure it worked for me, but the call returned "Method call timed out,", IIUC now it fails with "Not authorized", but if the transaction worked, that does not have a real sense to me...

I just tested, waited a lot given the right password and the package had not installed, that means that probably it is not the same issue as reported in  issue #1187, I would like to test my issue with this patch to be sure, but since it is in mageia i need @Conan-Kudo help...

ppisar:
> have you experienced that after a timeout the transaction

I was unable to get polkit to ask for a password. Maybe in requires some fancy graphical application I don't have installed. I always got immediate rejection.

m-blaha:
Although I think this fix is correct, it apparently does not resolve the original issue. I need to re-investigate.

m-blaha:
@ppisar I've added a commit with the documentation improvement.

ppisar (approved):
Good.

anaselli:
> Although I think this fix is correct, it apparently does not resolve the original issue. I need to re-investigate.

agreed

Issue #1379, opened 2024-04-04
`dnf5 list` shows unknown repo for packages installed with DNF 4


evan-goode:
`dnf5 list` shows the repo as `<unknown>` for packages installed with DNF 4, and vice versa: `dnf4 list` shows the repo as `@System` for packages installed with DNF 5.

On a fresh Fedora 39 container:

```
[root@2d825ae35a5f /]# dnf4 install dnf5 -y
[root@2d825ae35a5f /]# dnf5 install hello -y

[root@2d825ae35a5f /]# dnf4 list hello
Last metadata expiration check: 0:08:35 ago on Thu Apr  4 16:48:05 2024.
Installed Packages
hello.x86_64                             2.12.1-2.fc39                              @System

[root@2d825ae35a5f /]# dnf5 list hello
Updating and loading repositories:
Repositories loaded.
Installed packages
hello.x86_64 2.12.1-2.fc39 fedora

[root@2d825ae35a5f /]# dnf4 list --installed
Installed Packages
alternatives.x86_64                      1.25-1.fc39                @anaconda       
audit-libs.x86_64                        3.1.2-5.fc39               @koji-override-1
...

[root@2d825ae35a5f /]# dnf5 list --installed
Installed packages
alternatives.x86_64                      1.25-1.fc39                <unknown>
audit-libs.x86_64                        3.1.2-5.fc39               <unknown>
...
```

Reported by multiple users during the [2023-08-11 DNF 5 test day](https://fedoraproject.org/wiki/Test_Day:2023-08-11_Fedora_39_DNF_5) and the [2024-03-15 DNF 5 test day](https://fedoraproject.org/wiki/Test_Day:2024-03-15_Fedora_40_DNF_5).

j-mracek:
DNF and DNF5 does not share internal information from which repository a package was installed. DNF5 uploads those information from DNF4 once and then it keeps own independent copy.

The information about repository is restored in DNF5 after a package is modified by DNF5.
But this is not true for package reason. When new package is installed by DNF4, DNF5 will not known a reason for this package.

I am really sorry but we cannot do much here, therefore I am closing the issue.

jan-kolarik:
> DNF and DNF5 does not share internal information from which repository a package was installed. DNF5 uploads those information from DNF4 once and then it keeps own independent copy.
> 
[...]

PR #1403, opened 2024-04-11
Enable import data from DNF4 for systems without state dir

j-mracek:
The original test for import failed because tmp file cannot be created when target directory is not present.

Closes: https://github.com/rpm-software-management/dnf5/issues/1379


jan-kolarik (approved):
LGTM, thank you!

PR #1409, opened 2024-04-12
Bump libdnf5/libdnf5-cli so version

m-blaha:
This should have been done during 5.2 branch merge.


ppisar (approved):
Good.

ppisar:
The only test failure (" Policy rejects 5752ABEB74C53D9D: No binding signature at time") on Fedora 39 is strange, but definitely unrelated.

Issue #1407, opened 2024-04-12
dnf5.spec marks /usr/share/dnf5/aliases.d/compatibility.conf as a configuration file


ppisar:
/usr/share/dnf5/aliases.d/compatibility.conf is supposed to deliver a distribution-wide configuration. I touched that file when doing some experiments and after updating dnf5 Fedora RPM package, rpmconf tool caught that file as a change in a configuration file. That's because that files is marked with %config attribute in %files section of dnf.spec.

I think that file should not be handled as a configuration file. A reason is that upstream or distribution won't be able to change that file during after user installing the package. Compare to /usr/share/polkit-1/actions/ files which are also not handled as configuration files. I propose removing the %config attribute from that file.

The attribute was there from the very beginning, see f8e861b56a3e7c1b2b81a45f8dec86c94b404c1a commit.

j-mracek:
I think that issue is valid

PR #1413, opened 2024-04-15
Not handle compatibility.conf as configuration file

j-mracek:
Distribution configuration files in /usr/share/dnf5 are not supposed to be modified by user and should be always overriden by the next update. This behavior allows to modify distribution configuration without modifying DNF5 code.

Closes: https://github.com/rpm-software-management/dnf5/issues/1407


ppisar (approved):
Good. I confirm that that file disappeared from "rpm -q --configfiles -p" output.

Issue #1290, opened 2024-03-04
Review and finalize the DNF vs DNF5 changes documentation page


jan-kolarik:
Focusing mainly on the CLI part.

- [x] `--version` difference (from https://github.com/rpm-software-management/dnf5/issues/1337)
- [x] short options not working with `=` (from https://github.com/rpm-software-management/dnf5/issues/1353)
- [x] https://github.com/rpm-software-management/dnf5/issues/1289
- [x] https://github.com/rpm-software-management/dnf5/issues/812
- [x] https://github.com/rpm-software-management/dnf5/issues/1965

Doc pages here: https://dnf5.readthedocs.io/en/latest/changes.html.

jan-kolarik:
Also we should think about generalizing the current `dnf4 vs dnf5 changes` doc section to include also details how the systems using dnf4 will migrate to dnf5, like explicitly describing if there are any manual steps they can take to preserve existing configuration or system state, what is done automatically, what is not supported, etc.

PR #1414, opened 2024-04-15
doc: Review of DNF4 vs DNF5 CLI and configuration changes

### LINKED ISSUE #940: doc: dnf5 mark documentation issues
* `dnf5 mark` is significantly different from `dnf4 mark` and it should be properly documented in the section "Changes on the command line" of https://dnf5.readthedocs.io/en/latest/changes.html
* in dnf5-mark man page there's missing --skip-unavailable option (which is not a global option and it is provided by `dnf5 mark --help`)

jan-kolarik:
Related to #1290.
Closes #940.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -1,12 +1,302 @@
-====================================
-Changes in DNF5 in comparison to DNF
-====================================
+.. _changes_ref-label:
 
-The chapter describe differences between ``DNF5`` (https://github.com/rpm-software-management/dnf5) and ``DNF``
-(https://github.com/rpm-software-management/dnf)
+#############################
+ Changes between DNF and DNF5
+#############################
+
+This chapter describes the differences between `DNF5 <https://github.com/rpm-software-management/dnf5>`_ and `DNF <https://github.com/rpm-software-management/dnf>`_.
+
+.. _cli_changes_ref-label:
+
+Changes on the CLI
+==================
+
+Options
+-------
+
+Global options scoping
+^^^^^^^^^^^^^^^^^^^^^^
+Options that cannot be applied to all commands or may be applicable but have no effect are removed from general options and implemented only for related commands.
+
+Examples: ``--best``, ``--no-best`` are only relevant to several transaction commands.
+
+
+Options renaming
+^^^^^^^^^^^^^^^^
+Renaming boolean options to the following formats:
+
+  * ``--<option>`` and ``--no-<option>``
+  * ``--enable-<option>`` and ``--disable-<option>``
+
+The options with original names are retained for now as compatibility aliases.
+
+Examples: ``--best`` and ``--no-best``.
+
+
+Strict behavior
+^^^^^^^^^^^^^^^
+  * Options ``--disable-repo=REPO_ID`` and ``--setopt=[REPO_ID.]OPTION=VALUE`` now consistently result in an error when provided with an invalid ``REPO_ID``.
+  * The behavior is now aligned with the ``--repo=REPO_ID`` and ``--enable-repo=REPO_ID``.
+  * The ``strict`` configuration option is no longer considered, see the :ref:`strict option deprecation <strict_option_conf_changes_ref-label>` for more information.
+
+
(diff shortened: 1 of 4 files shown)


doc/changes.rst
```diff
@@ -1,12 +1,302 @@
...
+  * Now a DNF5 plugin.
+  * Different configuration file locations and some changes in format.
+  * See the :ref:`Automatic command <automatic_plugin_ref-label>` for more information.
+
+``autoremove``
+  * Dropped the ``<spec>`` positional argument since the use case is sufficiently covered by the ``remove`` command.
+  * Specific variants of the command, ``autoremove-n``, ``autoremove-na``, and ``autoremove-nevra``, are not supported anymore.
+
+``builddep``
+  * Dropped ``--spec`` and ``--srpm`` arguments as automatic detection from file extensions is implemented now.
```
Conan-Kudo, line 160: This is probably a bad idea because there is no guarantee we can detect from file extensions. I've seen spec files with all kinds of crazy file names, and while it's not super-common anymore to see `.spm` or similar weird things for SRPMs, it's still possible since RPM allows reconfiguring the package filename template.

Conan-Kudo (changes requested):

doc/changes.rst
```diff
@@ -1,12 +1,302 @@
...
+  * Dropped ``--updates`` option, only ``--upgrades`` is available now.
+
+``list``
+  * Dropped ``--all`` option since this behavior is the default one.
+  * Changed the behavior of the ``--available`` option.
+    * In DNF4, only packages not installed or with higher versions were listed. This behavior remains unchanged when the option is not used, reducing duplications in the "Installed Packages" section.
+    * When using the ``--available`` option, DNF5 considers all versions available in enabled repositories, irrespective of the installed version.
+
+``makecache``
+  * Metadata is now stored in different directories, see the ``cachedir`` configuration option :ref:`changes <cachedir_option_conf_changes_ref-label>` for more details.
```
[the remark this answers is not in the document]
kontura: The `makecache` command still exists I think this is just explanation that `dnf5` stores cache at different location.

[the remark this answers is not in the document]
jan-kolarik: Yep, in this section there is always a list of changes related to an individual command. If the command is entirely dropped, it's explicitly stated here. I hope it will be better visible in the final HTML format of the docs, though there was already some confusion mentioned on the proposal discussion related to this "changes doc", so if you have a specific tip how to improve the situation, please share it with us :slightly_smiling_face:

doc/changes.rst
jan-kolarik: Hmm, interesting, I wouldn't think of that. This implementation of the `builddep` plugin is there from the beginning and I don't see any explicit request to do it in the automatic way, so we can definitely think about bringing back the original behavior. Could you create an issue for that please?

doc/changes.rst
```diff
@@ -1,12 +1,302 @@
...
+    * Examples:
+
+      * Before: ``--add-repo`` Now: ``addrepo``
+      * Before: ``--save --setopt`` Now: ``setopt``
+
+  * Existing repository files are not modified; drop-in override files are created instead.
+  * See the :ref:`config-manager documentation <config_manager_plugin_ref-label>` for more information.
+
+``distro-sync``
+  * Now when any argument does not match any package or is not installed, DNF5 fail. The behavior can be modified by the ``--skip-unavailable`` option.
```
kontura, line 175: This sentence feels clunky to me and I am not sure I understand it. I think that at least it should be `..., DNF5 fails`.

What about: `DNF5 fails when an argument doesn't match any installed package.` or `Now when any argument does not match an installed package DNF5 fails.`?

It is repeated several times.

doc/changes.rst
kontura: There already is https://github.com/rpm-software-management/dnf5/issues/799 and the associated PR https://github.com/rpm-software-management/dnf5/pull/810

kontura (approved):

PR #1415, opened 2024-04-16
[API] Base::enable_disable_plugins

### LINKED ISSUE #391: enable/disable plugins via API

jrohel:
Adds a request to enable/disable plugins that match the names (glob patterns) in the list.

Closes #391


include/libdnf5/base/base.hpp
```diff
@@ -86,6 +86,15 @@ class Base {
...
     rpm::PackageSackWeakPtr get_rpm_package_sack();
 
+    /// Adds a request to enable/disable plugins that match the names (glob patterns) in the list.
+    /// Can be called multiple times. Requests (`plugin_names` and `enable` state) are queued.
+    /// The enable state of a plugin is set according to the last matching request.
+    /// Must be called before the Base::setup.
+    /// @param plugin_names Plugin names (glob patterns) to enable/disable
+    /// @param enable Request: true - enable plugins, false - disable plugins
+    /// @exception libdnf5::UserAssertionError When called after Base::setup
+    void enable_disable_plugins(const std::vector<std::string> & plugin_names, bool enable);
```
jan-kolarik, line 96: I was considering splitting it to `enable_plugins` and `disable_plugins`, but probably it's ok like this as we have now like one or two plugins and this is going to be used as just an override of the configured values. 

But anyway, I think it should be explictily stated here that this is related to **libdnf5** plugins.

Also I was checking that `--enable-plugin` and `--disable-plugin` is apparently not doing anything now, is that right? We should probably create a follow-up issue for that and ideally introduce it in the same release as this feature.

jrohel: > I was considering splitting it to enable_plugins and disable_plugins, but probably it's ok like this as we have now like one or two plugins and this is going to be used as just an override of the configured values.

I started with two methods `enable_plugins` and `disable_plugins` in the API. However both methods had the same priority and added requests to the same queue. Then when I documented this, I redesigned the API. Yes, the name `enable_disable_plugins` is not that nice, but I don't have to write to the two methods that they put requests in the same queue just with a different attribute. The order in which requests are queued is important. Although we don't currently have many libdnf5 plugins.

> But anyway, I think it should be explictily stated here that this is related to libdnf5 plugins.

OK. I added "libdnf5 library" to the commit message. But it was obvious in the commit message and the code anyway. `Base::enable_disable_plugins` is a method of the `Base` class provided by the libdnf5 library.

> Also I was checking that --enable-plugin and --disable-plugin is apparently not doing anything now, is that right?

Yes you're right. There were 2 reasons for this.
- Missing API to enable/disable listed libdnf5 plugins. This PR solves that.
- Missing agreement which plugins will be enabled/disabled by these arguments - libdnf5, dnf5 or both?
  dnf5 plugins currently add new commands to the dnf5 application, but do not change the behavior of existing ones. So in theory there is no need to disable them, the user simply won't use the command provided by the plugin. On the other hand, there is a risk that a bad dnf5 plugin will do something in its init phase.
libdnf5 plugins do background work and it doesn't matter which program uses libdnf5 - dnf5 application or some other. These should be controlled by --enable-plugin and --disable-plugin. But generally disabling plugins is for troubleshooting. It is probably not appropriate if, for example, we disable the libdnf5 plugin, while it was required by the system's security policy (auditing, creating snapshots to restore the system to its previous state, ...)

> We should probably create a follow-up issue for that and ideally introduce it in the same release as this feature.
Yes

jan-kolarik: > OK. I added "libdnf5 library" to the commit message. But it was obvious in the commit message and the code anyway. Base::enable_disable_plugins is a method of the Base class provided by the libdnf5 library.

All right, I was rather thinking about the docstring here as it is also explicitly mentioned in the `setup()` method below, but if you think it's clear, we can keep it as it is.

> We should probably create a follow-up issue for that and ideally introduce it in the same release as this feature.

Here is the follow-up: https://github.com/rpm-software-management/dnf5/issues/1422

PR #1416, opened 2024-04-16
Resolve goal according to strict setting first

j-mracek:
It should help to force solver to prefer latest solution rather then minimalistic solution skipping something.

Performance improvement for operation that does not generate any problem, because transaction is resolved only once with strict setting.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2272257

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -373,32 +373,38 @@ std::vector<std::string> Transaction::get_gpg_signature_problems() const noexcep
     return p_impl->signature_problems;
 }
 
-void Transaction::Impl::set_transaction(
-    rpm::solv::GoalPrivate & solved_goal, module::ModuleSack & module_sack, GoalProblem problems) {
-    auto solver_problems = process_solver_problems(base, solved_goal);
-    if (!solver_problems.empty()) {
-        add_resolve_log(GoalProblem::SOLVER_ERROR, solver_problems);
-    } else {
-        // TODO(jmracek) To improve performance add a test whether it make sense to resolve transaction in strict mode
-        // Test whether there were skipped jobs or used not the best candidates due to broken dependencies
-        rpm::solv::GoalPrivate solved_goal_copy(solved_goal);
-        solved_goal_copy.set_run_in_strict_mode(true);
-        solved_goal_copy.resolve();
-        auto solver_problems_strict = process_solver_problems(base, solved_goal_copy);
-        if (!solver_problems_strict.empty()) {
-            add_resolve_log(GoalProblem::SOLVER_PROBLEM_STRICT_RESOLVEMENT, solver_problems_strict);
-        }
-    }
+void Transaction::Impl::resolve_and_set_transaction(
+    rpm::solv::GoalPrivate & goal_to_resolve, module::ModuleSack & module_sack, GoalProblem problems) {
     this->problems = problems;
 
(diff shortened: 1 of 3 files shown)


libdnf5/base/transaction.cpp
```diff
@@ -373,32 +373,37 @@ std::vector<std::string> Transaction::get_gpg_signature_problems() const noexcep
...
-        solved_goal_copy.set_run_in_strict_mode(true);
-        solved_goal_copy.resolve();
-        auto solver_problems_strict = process_solver_problems(base, solved_goal_copy);
-        if (!solver_problems_strict.empty()) {
+void Transaction::Impl::resolve_and_set_transaction(
+    rpm::solv::GoalPrivate & goal_to_resolve, module::ModuleSack & module_sack, GoalProblem problems) {
+    this->problems = problems;
+
+    // TODO(jmracek) To improve performance add a test whether it make sense to resolve transaction in strict mode
+    // Test whether there were skipped jobs or used not the best candidates due to broken dependencies
```
m-blaha, line 381: Does this comment make sense with reverted order of strict resolution?

libdnf5/base/transaction.cpp
```diff
@@ -373,32 +373,37 @@ std::vector<std::string> Transaction::get_gpg_signature_problems() const noexcep
...
+    goal_to_resolve.set_run_in_strict_mode(true);
+    GoalProblem resolve_problems = goal_to_resolve.resolve();
+
+    auto solver_problems_strict = process_solver_problems(base, goal_to_resolve);
+    if (!solver_problems_strict.empty()) {
+        goal_to_resolve.set_run_in_strict_mode(false);
+        resolve_problems = goal_to_resolve.resolve();
+
+        auto solver_problems = process_solver_problems(base, goal_to_resolve);
+        if (solver_problems.empty()) {
```
m-blaha, line 391: Shouldn't be `solver_problems_strict` added to resolve log unconditionally? Regardless non-strict problems exist?

j-mracek: I think that when we add both problem sets to event log it will result in situations when both error messages will appear in terminal. Strict as warning and second one as an error. Or have I understood you correctly?

j-mracek: Or shall we use always the message from strict transaction because it provides more information? I think this is a good idea.

m-blaha: I was thinking about different messages from strict and non-strict solvers (if there is a scenario when they can occur). But having the same message printed twice is strange and will probably cause some confusion to the users.

libdnf5/base/transaction.cpp
m-blaha: OK, thanks. Let's keep the comment then.

libdnf5/base/transaction.cpp
m-blaha: Thinking about it once more - let's keep it as it is. It's compatible with original behavior.

m-blaha (approved):
Thank you!

m-blaha:
Hm, it looks like some tests are failing - e.g. if you try to install already installed package name and newer version is available, the result is upgrade instead of "nothing to do".

j-mracek:
Closing the PR, because the new behavior will make no-best option useless and will trigger upgrades in situation when request should be satisfied by installed version.

PR #1417, opened 2024-04-16
Fix: implicit conversion changes signedness, unused value

jrohel:
Fixes clang compilation errors:
4 x error: implicit conversion changes signedness
1 x error: variable 'total_bytes_read' set but not used

### DIFF
--- a/dnf5daemon-client/commands/repoquery/repoquery.cpp
+++ b/dnf5daemon-client/commands/repoquery/repoquery.cpp
@@ -219,7 +219,7 @@ void RepoqueryCommand::run() {
 
     int in_fd = pipefd[0];
 
-    const size_t pipe_buffer_size = fcntl(in_fd, F_GETPIPE_SZ);
+    const auto pipe_buffer_size = static_cast<size_t>(fcntl(in_fd, F_GETPIPE_SZ));
     const int timeout = 30000;
 
     std::array<pollfd, 1> pfds;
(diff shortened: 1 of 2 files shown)


jrohel:
I have fixed this type of error several times.
Can we re-enable compilation using clang? So it will probably be necessary to include this https://github.com/rpm-software-management/dnf5/pull/1418 to pass the unit tests with clang compiled code.

dnf5daemon-client/commands/repoquery/repoquery.cpp
```diff
@@ -219,16 +219,14 @@ void RepoqueryCommand::run() {
 
     int in_fd = pipefd[0];
 
-    const size_t pipe_buffer_size = fcntl(in_fd, F_GETPIPE_SZ);
+    const auto pipe_buffer_size = static_cast<size_t>(fcntl(in_fd, F_GETPIPE_SZ));
```
m-blaha, line 222: We should handle situation when `fcntl` fails and return -1 and add some reasonable fallback value - 4096?

jrohel: @m-blaha Good idea. Moreover, F_GETPIPE_SZ is Linux specific, so a fallback would be useful.

m-blaha: Maybe we can just go with a constant value. Looks like on my system the default value is 65536 and do not try to detect the size.

m-blaha: Anyway, it's out of this PR scope, I filed a new issue for it - https://github.com/rpm-software-management/dnf5/issues/1419

m-blaha (approved):
Thank you!

m-blaha:
The test failures are unrelated.

Issue #1317, opened 2024-03-15
dnf5 copr_plugin: Unit test failed. Probably unitialized variable (class member) in code


jrohel:
failed:
test: CoprRepoTest::test_enable (F) line: 146 test/dnf5-plugins/copr_plugin/CoprRepoTest.cpp

The problem seems to be caused by an uninitialized `enabled` class member in the `CoprRepoPart` class.
All non-zero values mean true. The test will only fail if `enabled` is false. That is, only if there is zero in uninitialized memory.
```
private:
    std::string id;
    std::string name;
    bool enabled;        <-  not initialized here, only initialized in two of the five constructors
    std::string baseurl;
    std::string gpgkey;
    int priority = 99;
    int cost = 0;
    bool module_hotfixes = false;
};
```

jrohel:
@praiskup Can you take a look at this?

PR #1418, opened 2024-04-16
Disable unit tests for Copr dnf5 plugin

jrohel:
The Copr dnf5 plugin is faulty, issue https://github.com/rpm-software-management/dnf5/issues/1317

It uses an uninitialized variable. Because of the random value of this variable, unit tests fail randomly (when it has a value of 0, which is whenever we allow the compiler to initialize memory to 0, or we use clang).

I'm in favor of temporarily disabling the Copr unit tests until the plugin is fixed, so we can re-enable compilation with clang.

### DIFF
--- a/test/dnf5-plugins/CMakeLists.txt
+++ b/test/dnf5-plugins/CMakeLists.txt
@@ -1,1 +1,5 @@
-add_subdirectory(copr_plugin)
+# Temporarily disable Copr unit tests until the plugin is fixed.
+# The Copr dnf5 plug-in is broken. It uses an uninitialized variable. Because of the random value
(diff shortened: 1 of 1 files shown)


test/dnf5-plugins/CMakeLists.txt
```diff
@@ -1 +1 @@
-add_subdirectory(copr_plugin)
+#add_subdirectory(copr_plugin)
```
m-blaha, line 1: We should add a comment describing why the copr_plugin test is disabled.

jrohel: Okay, I'll add it.

m-blaha (approved):
Thank you!

PR #1424, opened 2024-04-18
spec: Simplify man page files

jan-kolarik:
Use globs to cover both dnf5 man files and dnf symlinks.


ppisar (approved):
It looks good. Though I did not try building it with %dnf5_obsoletes_dnf.

PR #1425, opened 2024-04-18
Prepare for switch of dnf5 in Rawhide

jan-kolarik:
Related:
- https://github.com/rpm-software-management/dnf/pull/2082
- https://github.com/rpm-software-management/dnf-plugins-core/pull/531

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -3,7 +3,7 @@
 %global project_version_minor 0
 %global project_version_micro 0
 
-%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 41 || 0%{?rhel} > 11]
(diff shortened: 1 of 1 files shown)


dnf5.spec
```diff
@@ -3,7 +3,7 @@
 %global project_version_minor 0
 %global project_version_micro 0
 
-%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 41 || 0%{?rhel} > 11]
+%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 40 || 0%{?rhel} > 11]
```
j-mracek, line 6: May I ask you why `0%{?rhel} > 11` and not `0%{?rhel} > 10`?

j-mracek: LETS merge the Fedora change first. RHEL condition may be modify next time

j-mracek (approved):

PR #1427, opened 2024-04-18
base: Make get_transaction_history unstable

jan-kolarik:


test/libdnf5/transaction/test_transaction.cpp
```diff
@@ -52,7 +52,8 @@ create_getter(new_transaction, &libdnf5::transaction::TransactionHistory::new_tr
 }  //namespace
 
 static Transaction create_transaction(libdnf5::Base & base, int nr) {
-    auto trans = (*(base.get_transaction_history()).*get(new_transaction{}))();
+    auto history = std::make_unique<libdnf5::transaction::TransactionHistory>(base);
```
kontura, line 55: Is the `unique_ptr` needed here?

jan-kolarik: As these are tests, I just applied the first idea that came to my mind. Is there a better alternative?

jan-kolarik: Oh I see what you mean, just create an object on stack right?

PR #1428, opened 2024-04-19
Set `group` reason for packages removed by a group removal

kontura:
It also renames `add_transaction_group_installed` since it is now used to store removed packages as well.

I have discovered this while working on `undo` command, when undoing a `group remove` transaction the packages need to have a `group` reason.


ppisar (approved):
Good. It does not break public API or ABI.

Issue #1335, opened 2024-03-21
Listing libdnf5 plugins


glum23:
Is there an easy way how to list all currently enabled plugins? If not, it should be implemented and documented.
For dnf4, it was possible to list them e.g. via `dnf -v repolist`.
(Going through configs in /etc/dnf/libdnf5-plugins/ is not supposed to be the requested easy solution.)

ppisar:
Loaded plugins are logged into /var/log/dnf5.log:

~~~~
2024-03-21T10:13:25+0000 [16009] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/builddep_cmd_plugin.so"
2024-03-21T10:13:25+0000 [16009] INFO Loaded dnf5 plugin "builddep" ("/usr/lib64/dnf5/plugins/builddep_cmd_plugin.so") version "1.0.0"
2024-03-21T10:13:25+0000 [16009] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/changelog_cmd_plugin.so"
2024-03-21T10:13:25+0000 [16009] INFO Loaded dnf5 plugin "changelog" ("/usr/lib64/dnf5/plugins/changelog_cmd_plugin.so") version "1.0.0"
2024-03-21T10:13:25+0000 [16009] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/config-manager_cmd_plugin.so"
2024-03-21T10:13:25+0000 [16009] INFO Loaded dnf5 plugin "config-manager" ("/usr/lib64/dnf5/plugins/config-manager_cmd_plugin.so") version "0.1.0"
2024-03-21T10:13:25+0000 [16009] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/copr_cmd_plugin.so"
2024-03-21T10:13:25+0000 [16009] INFO Loaded dnf5 plugin "copr" ("/usr/lib64/dnf5/plugins/copr_cmd_plugin.so") version "0.1.0"
2024-03-21T10:13:25+0000 [16009] DEBUG Loading plugin file="/usr/lib64/dnf5/plugins/needs_restarting_cmd_plugin.so"
2024-03-21T10:13:25+0000 [16009] INFO Loaded dnf5 plugin "needs_restarting" ("/usr/lib64/dnf5/plugins/needs_restarting_cmd_plugin.so") version "1.0.0"
[...]

PR #1430, opened 2024-04-19
[libdnf5 API] Base::get_plugins_info

### LINKED ISSUE #1429: Missing API providing information about available/loaded libdnf5 plugins
Needed for #1335

jrohel:
Introduces the `Base::get_plugins_info` method, which returns a list of information about plugins found during `Base::setup`.

It also introduces the `plugin::PluginInfo` class, which provides information about one libdnf5 plugin.

Closes #1429
Related to #1335

### DIFF
--- a/include/libdnf5/plugin/plugin_info.hpp
+++ b/include/libdnf5/plugin/plugin_info.hpp
@@ -1,0 +1,75 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of li
+bdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef LIBDNF5_PLUGIN_PLUGIN_INFO_HPP
+#define LIBDNF5_PLUGIN_PLUGIN_INFO_HPP
+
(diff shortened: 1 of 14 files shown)


jan-kolarik:
Why can't we just use and extend the existing `libdnf5::Base::Impl::get_plugins()`?

include/libdnf5/plugin/plugin_info.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+    /// @return a nullptr terminated array of attributes supported by the plugin or nullptr if the plugin is not loaded
+    const char * const * get_attributes() const noexcept;
+
+    /// Gets the value of the attribute from the plugin.
+    /// Returns nullptr if the attribute does not exist or plugin is not loaded.
+    /// @return the value of the `name` attribute or nullptr
+    const char * get_attribute(const char * name) const noexcept;
+
+    class Impl;
```
jan-kolarik, line 65: Is there a need to have it in the `public` section?

include/libdnf5/plugin/plugin_info.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+private:
+    explicit PluginInfo(Impl & p_ipml);
```
jan-kolarik, line 68:
```suggestion
    explicit PluginInfo(Impl & p_impl);
```

jan-kolarik (commented):
Otherwise, LGTM.

include/libdnf5/plugin/plugin_info.hpp
jrohel: > Is there a need to have it in the public section?

Yes, it is. Class `Impl` is used by the private class `libdnf5::plugin::Plugins`. I feel it is better to expose only the existence of the type, rather than listing the internal class as fried in the public interface (and thus giving the `libdnf5::plugin::Plugins` class full access to `PluginInfo`, which was not the purpose. I just need access to the type).

include/libdnf5/plugin/plugin_info.hpp
jrohel: Thanks. I fixed the typo.

include/libdnf5/plugin/plugin_info.hpp
jan-kolarik: Thx!

jan-kolarik (approved):

Issue #1401, opened 2024-04-11
Advisory `--advisory-severities` and type filters produce union instead of intersection


pkratoch:
For example, `dnf5 advisory list --security --advisory-severities=critical` shows all advisories that are either or type security or have critical severity (so even e.g. moderate securities or critical bugfixes are shown).

I believe that other filter combinations produce intersection (except when specifying multiple type filters, but that makes sense if viewed as if it was just one list option), so this is inconsistent. If there is a reason for this behavior, it should be at least documented.

ppisar:
I confirm the bug. --advisory-severities option is document as "limit…", but it does not do an intersection:

~~~~
# dnf5 advisory list --security --advisory-severities=critical
Updating and loading repositories:
Repositories loaded.
Name                   Type     Severity                           Package              Issued
FEDORA-2024-43a0920f12 security Low      perl-Clipboard-0.29-1.fc39.noarch 2024-04-10 14:44:52
~~~~

kontura:
I think the reason is compatibility with old dnf which is unfortunately also doing an union.

I agree that the docs should definitely be fixed (both man pages and `--help`).
I have also noticed that the `--help` of advisory commands refers to packages which is confusing because its filtering advisories. This is probably caused by the fact that the options are shared with other commands.

j-mracek:
I have created a patch, but I am not sure whether we should modify the behavior or modify DNF5 in more complex way.

I think that both following command should target the same set of advisories.

`dnf5 advisory list --security --advisory-severities=critical`

`dnf5 upgrade --security --advisory-severities=critical`

What do you think?

ppisar:
Both commands should behave the same.

I'm for changing it to an intersection because users who want an union can use the options in separate calls. Contrary with the union behavior, users cannot not achieve the intersection.

kontura:
I agree that all the commands should behave the same.

I also prefer intersection but changing it now could be quite unexpected for users.

j-mracek:
> I agree that all the commands should behave the same.
> 
> I also prefer intersection but changing it now could be quite unexpected for users.

@kontura May I ask you when you would prefer that the change happen?

kontura:
> @kontura May I ask you when you would prefer that the change happen?

I think the best time was when we introduced the advisory options to dnf5 but since the change was rejected at that time due to compatibility concerns I wanted to mention it.

j-mracek:
I am still not sure whether the change make sense from `upgrade` command usages.

[...]

PR #1433, opened 2024-04-22
Filter advisories using AND operator and not OR

j-mracek:
This is a difference in comparison to DNF4 and it also intorduce a different behavior than DNF5 upgrade command.

Closes: https://github.com/rpm-software-management/dnf5/issues/1401


m-blaha:
The DNF team has decided not to change the command behavior at this stage and will only clarify the documentation instead.

PR #1435, opened 2024-04-22
repo_sack: Treat all repos with solv_repo created as loaded

jan-kolarik:
Fix the implementation of the `is_loaded()` flag, as it's used for initializing includes in the solver.

This issue arose in the 5.2.0.0 branch. Before that, the check relied on whether the `solv_repo` was null.

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1489.
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2275530.

### DIFF
--- a/libdnf5/repo/repo.cpp
+++ b/libdnf5/repo/repo.cpp
@@ -85,7 +85,6 @@ private:
     ConfigRepo config;
 
     SyncStrategy sync_strategy{SyncStrategy::TRY_CACHE};
-    bool loaded{false};
     bool expired{false};
     std::unique_ptr<SolvRepo> solv_repo;
     std::unique_ptr<RepoDownloader> downloader;
@@ -219,7 +218,7 @@ void Repo::download_metadata(const std::string & destdir) {
 
 void Repo::load() {
     // Each repository can only be loaded once. This one has already loaded, so just return instantly
-    if (p_impl->loaded) {
+    if (is_loaded()) {
         return;
     }
 
@@ -233,8 +232,6 @@ void Repo::load() {
 
     p_impl->solv_repo->set_needs_internalizing();
     p_impl->base->get_rpm_package_sack()->p_impl->invalidate_provides();
-
-    p_impl->loaded = true;
 }
 
 void Repo::add_libsolv_testcase(const std::string & path) {
@@ -617,7 +614,7 @@ RepoDownloader & Repo::get_downloader() const {
 }
 
 bool Repo::is_loaded() const {
-    return p_impl->loaded;
+    return p_impl->solv_repo.get();
 }
 
 SolvRepo & Repo::get_solv_repo() const {


libdnf5/repo/repo_sack.cpp
```diff
@@ -201,6 +201,7 @@ std::map<std::string, libdnf5::rpm::Package> RepoSack::add_cmdline_packages(
     }
 
     auto cmdline_repo = get_cmdline_repo();
+    cmdline_repo->load();
```
j-mracek, line 204: The patch is OK, but it calls the load() only for one reason - it sets `p_impl->loaded = true;`. Everything else is redundant or skipped.

j-mracek: What about to add `p_impl->loaded = true;` to `rpm::Package Repo::add_rpm_package(const std::string & path, bool with_hdrid)` or create a private method that will set repo as loaded. Or at least add comment above the line and to load() method. The description of the `load()` method does not cover the usage here => it is confusing why it is used here because it does not load libsolv repo, because it is already created.

jan-kolarik: Yep, that's correct. The other things inside the current `load()` implementation are eventually called via the subsequent `add_rpm_package` calls. I considered simply setting the `loaded` flag to true, but the `load()` function exists there to load the repository. And that's at least strange that the `load()` method isn't called at all for the command-line repository. This could be problematic because, in the future, more functionality might be added to the `load()` method. But of course, if you have a better idea on how to address this issue, I'm open to hearing it.

jan-kolarik: "Oh, I started writing before I saw your last message :slightly_smiling_face:

jan-kolarik: > The description of the load() method does not cover the usage here => it is confusing why it is used here because it does not load libsolv repo, because it is already created.

Hmm, so I was debugging the existing code and after the `auto cmdline_repo = get_cmdline_repo();` statement is executed, the `solv_repo` for the `cmdline_repo` is actually nullptr yet.

And also the current docstring for the `load()` method starts with `/// Loads the repository objects into sacks.` and that's actually pretty matching.

jan-kolarik: But when looking into the former implementation, like the checking for `solv_repo != nullptr`, it means that the `loaded` flag should have now the same semantics. So wouldn't it be better in the end to set the `loaded` flag just in the `make_solv_repo` method?

jan-kolarik:
@j-mracek Thanks for pointing me to a different direction. I've simplifed the solution, hopefully it's better now.

jan-kolarik:
I've modified the commit message to match the current implementation of the patch.

j-mracek:
LGTM. This solution also fixes the same problem for repository loaded from a test case.

PR #1436, opened 2024-04-22
API: no inline methods, add p_impl

jrohel:
Converts inline methods to non-inline in several classes.
Adds p_impl to some classes.
Moves public class variables to the private/p_impl section.
Deletes some unwanted constructors and assigns operators.

Related https://github.com/rpm-software-management/dnf5/issues/1025


jan-kolarik (approved):
Great, thanks for also improving the stability of libdnf5-cli and dnf5 headers.

Issue #1353, opened 2024-03-28
`dnf5 -x=dnf install  dnf` does not work as expected


j-mracek:
`-x=dnf` adds `exclude = =dnf` to configuration. Try `--dump-main-config` to see it.

`-x dnf` works as expected

Proposing the bug as a bug with high priority, because the option is important and other options might be affected as well.

m-blaha:
I think we do not support syntax with `=` for short options. Only for the long ones.

jan-kolarik:
Also this behavior should be documented in the dnf4 vs dnf5 changes doc.

jrohel:
DNF5 command line options format:

Prefix of long option `--`.
The prefix of a list of short options `-`. The last option in the list may have a value.

**Examples with the same meaning:**
- `--assumeyes --exclude=dnf` (preferred long format, unambiguous format)
- `--assumeyes --exclude dnf` (alternative long format, For the parser to correctly decode the string `dnf` - is it a value of the value of the `--exclude` option or the next argument - it must know whether or not the `--exclude` option expects a value.)
- `-yxdnf`
- `-yx dnf`
- `-y -xdnf`
- `-y -x dnf`
- `--assumeyes -xdnf`
- `--assumeyes -x dnf`
- `-y --exclude=dnf`
- `-y --exclude dnf`

In DNF5, there is already a lot of freedom when entering options on the command line.
Introducing another possibility, a short option with a value separated by `=`, would be really confusing.

m-blaha:
> In DNF5, there is already a lot of freedom when entering options on the command line. Introducing another possibility, a short option with a value separated by `=` (as requested `-x=dnf`), would be really confusing.

I agree. I'd prefer to not implement `-x=package` format in dnf5 and just document the difference (as dnf4 supports it).
The format without `=` is also inline with POSIX Utility conventions (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)

j-mracek:
Be honest `-yxdnf` looks confusing. Anyway it might be a compatibility issue. It looks like that RPM supports short
 options with both ` ` and `=`

j-mracek:
I see as a problem that DNF5 is not unified in using option separators.

`--exlude=dnf` works but `-x=dnf` doesn't.

j-mracek:
> We don't. dnf5(8) manual agrees it's an invalid invocation. I would rather fix the option parser to reject "-x=dnf" as an invalid option.

[...]

PR #1442, opened 2024-04-24
Raise an exception when `=` is used with short option

j-mracek:
It prevents silently skipping  the change in behavior between DNF4 and DNF5 Where `-x=dnf` excluded `dnf` but in DNF5 it excludes `=dnf`. Without the exception it would be hard to detect that something might work differently then expected.

Closes: https://github.com/rpm-software-management/dnf5/issues/1353


j-mracek:
Upstream wants to implement it differently = closing

PR #1443, opened 2024-04-24
spec: Add conflict with the old provider of dnf.conf

jan-kolarik:
Another bit needed for switching dnf5 in Rawhide. We need to ensure there are no multiple providers of the dnf configuration on the system simultaneously.

Related: https://github.com/rpm-software-management/dnf5/pull/1444.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -342,7 +342,7 @@ Requires:       libsolv%{?_isa} >= %{libsolv_version}
 Requires:       librepo%{?_isa} >= %{librepo_version}
 Requires:       sqlite-libs%{?_isa} >= %{sqlite_version}
 %if %{with dnf5_obsoletes_dnf}
-Conflicts:      dnf-data < 4.16.0
+Conflicts:      dnf-data < 4.20.0
 %endif
 
 %description -n libdnf5


dnf5.spec
```diff
@@ -342,7 +342,7 @@ Requires:       libsolv%{?_isa} >= %{libsolv_version}
 Requires:       librepo%{?_isa} >= %{librepo_version}
 Requires:       sqlite-libs%{?_isa} >= %{sqlite_version}
 %if %{with dnf5_obsoletes_dnf}
-Conflicts:      dnf-data < 4.16.0
+Conflicts:      dnf-data < 4.20.0
```
m-blaha, line 345: Wouldn't be better using `dnf-data < 5`? Otherwise we would need to remember bumping this conflict in case we release dnf-4.20.

jan-kolarik: Ehm, yep. That's a great idea. I've created the PR before my morning coffee, so... :slightly_smiling_face:

jan-kolarik: Actually no, we want to allow simultaneously installed `dnf-data >= 4.20.0` and dnf5, so that's why it is like that.

m-blaha: So now it's my turn to grab a coffee, I guess. :-)

jan-kolarik: IOW this should trigger updating dnf-data to 4.20.0 where the dnf5_obsoletes_dnf macro will be applied for Fedora 41 and therefore /etc/dnf/dnf.conf will not be provided by dnf package but dnf5. If we won't change this conflicts condition here, for the users with installed dnf-data < 4.20.0, /etc/dnf/dnf.conf will be owned by the dnf-data package, but if new dnf5 upgrade package would come, then there will be conflict of two providers of this conf file. This can be currently seen in the Rawhide TMT failing tests for dnf5 PRs:
```
Error: Transaction test error:
  file /etc/dnf/dnf.conf from install of libdnf5-5.2.0.0-1.20240424063204861995.pr1442.225.g46c1ae4e.fc41.x86_64 conflicts with file from package dnf-data-4.19.2-1.fc41.noarch
```

jan-kolarik: Of course, if you find anything illogical in my previous statement or spot any gaps in our current approach, please let me know.

m-blaha (approved):
Thank you!

PR #1452, opened 2024-04-26
Add `history undo` command

kontura:
It adds `add_revert_transaction(...)` API to `Goal`. This is used by `undo` and later should also be used for `rollback` but that requires merging of history transactions.


j-mracek:
May I ask you why users wants to use `history undo` command?

What do you think could be improved in comparison to DNF4?

pmatilai:
> May I ask you why users wants to use history undo command?

This almost looks like a trick-question to me :sweat_smile: 

For me personally the most common use is to uninstall a thing after trying it out, ie typically "undo last". Undoing an upgrade to test whether upgrade broke/fixed stuff would be super useful, but of course that depends on more than just dnf features (the repos dont typically carry all possible versions)

j-mracek:
May be I would be good to be verbose here. In DNF4, history commands are important but not useful much because the implementation is very strict. Undo is very important and can be very powerful.

Use case - revert the last transaction, but why
1. The transaction does not provide expected functionality
2. The transaction did not finish successfully and user wants to undo it
The second option might require reinstall of packages that we want to have after undo, but they are already installed.

Both operation of the same transaction have a different starting point - some packages might be not on the system.

The command does not provide only undo of the last transaction but there is no logic that react to any changes.
We can think about following cases:
```
1. install a-1-1.noarch
2. upgrade a-1-1.noarch to a-2-1.noarch
# dnf5 history undo 1
```
This may end up with following results
 - Fail due to package is installed in a different version (requires good message)
 - Pass with `nothing to do`, message - `No packages to remove for argument: zsh-html-5.9-5.fc38.noarch` (current behavior). It is may be a good behavior for installonly packages
 - Remove `a-2-1.noarch` because package `a` wasn't installed prior the transaction (only for packages that are not installonly)

Another use case
```
1. install a-1-1.noarch
2. remove a-1-1.noarch
# dnf5 history undo 1
```
This may end up with following results
 - Pass - nothing to do
 - Fail - due to the version is not  present on the system

Another example:
```
1. install a-1-1.noarch that requires b, install dependency b-1-1.noarch
2.  install c-1-1.noarch that requires b
# dnf5 history undo 1
```
This may end up with following results
 - Remove only `a`
 - remove `a`, `b`, `c`
 - Fail due to transaction differs from expectations (removal of dependent package).

Additional cases with installonly packages.

Please don't take it as this implementation is wrong. I would like to brainstorm expectation for various use cases. Also I am sure that I've not covered all possibilities.

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2222,115 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        package_replay.nevra = libdnf5::rpm::to_nevra_string(pkg);
+        auto reverted_action = REVERT_ACTION.find(pkg.get_action());
+        libdnf_assert(
+            reverted_action != REVERT_ACTION.end(),
+            "Cannot revert action: \"{}\"",
+            transaction_item_action_to_string(pkg.get_action()));
+        package_replay.action = reverted_action->second;
+
+        if (package_replay.action == Action::INSTALL && pkg.get_reason() == Reason::CLEAN) {
+            package_replay.reason = Reason::DEPENDENCY;
```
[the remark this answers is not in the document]
j-mracek: Would be possible to use the same logic like is it used for `Action::REASON_CHANGE`?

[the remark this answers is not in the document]
kontura: I see, since we cannot tell whether the previous reason was `DEPENDENCY` or `WEAK DEPENDENCY` we have to look into the history.

kontura:
> Use case - revert the last transaction, but why
> 
>     1. The transaction does not provide expected functionality
> 
>     2. The transaction did not finish successfully and user wants to undo it
>        The second option might require reinstall of packages that we want to have after undo, but they are already installed.
> 
> 
> Both operation of the same transaction have a different starting point - some packages might be not on the system.

I have also used it for this in the past but now that I think about it `rollback` command seems as a better fit for this use case because it doesn't have to deal with the questions @j-mracek asked below, it always reverts `n` **last** transactions.

> 
> The command does not provide only undo of the last transaction but there is no logic that react to any changes...

These are basically the same problems that the `replay` command has to deal with when replaying a transaction that doesn't apply cleanly. (The `undo` command is `replay` but the source of the transaction isn't a stored file but the history.)

The `replay` command in dnf4 is strict by default but there are options:
```
--ignore-installed    For the replay command, don't check for installed packages matching those in transaction
--ignore-extras       For the replay command, don't check for extra packages pulled into the transaction
--skip-unavailable    For the replay command, skip packages that are not available or have missing dependencies
```
We already have `--skip-unavailable`.
What about to preserve this approach and use it also for the `undo` command?

j-mracek:
I don't think that the option mentioned above answer the question related to expected behavior. They only make no changes if something is unexpected.

Lets return to the first example
```
1. install a-1-1.noarch
2. upgrade a-1-1.noarch to a-2-1.noarch
# dnf5 history undo 1
```
The current behavior reports nothing to do with success and any of those options would help. I don't think that success is a correct result of the transaction, because it did nothing.

For this situation a user might expect
- it will move package `a` to state prior transaction `1` - no package `a` installed on the system
- ensure that `a-1-1.noarch` is not installed but other versions are fine
- ensure that `a-1-1.noarch` is not installed but if something related to `a` is different (version), fail to allow user to
make a decision.

j-mracek:
> > * ensure that a-1-1.noarch is not installed but if something related to a is different (version), fail to allow user to
> >   make a decision.
> 
> This would be the new default behavior and the user could specify `--ignore-installed` to make it pass (essentially like this test case: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/transaction-sr/replay.feature#L1577-L1599).
> 
> What do you think?

I think this is a better option then  the current behavior. It requires good messaging to ensure that  user understand the problem and possible solutions.

kontura:
Ok I have added several commits and I tried to focus on clear messages and logs.

@j-mracek for the cases you mentioned it looks something like this:
```
1. install a-1-1.noarch
2. upgrade a-1-1.noarch to a-2-1.noarch
# dnf5 history undo 1
Repositories loaded.
Cannot perform action 'Remove' for Package 'a-1-1.noarch' becasue it is not installed.
You can try to add to command line:
  --ignore-installed to allow mismatches between installed and stored transaction packages.
# dnf5 history undo 1 --ignore-installed
Repositories loaded.
Cannot perform action 'Remove' for Package 'a-1-1.noarch' becasue it is not installed.

Nothing to do.

```
(With `--ignore-installed` the operation succeeds but the warning remains.)


```
1. install a-1-1.noarch
2. remove a-1-1.noarch
# dnf5 history undo 1
Repositories loaded.
Cannot perform action 'Remove' for Package 'a-1-1.noarch' becasue it is not installed.
You can try to add to command line:
  --ignore-installed to allow mismatches between installed and stored transaction packages.
# dnf5 history undo 1 --ignore-installed
Repositories loaded.
Cannot perform action 'Remove' for Package 'a-1-1.noarch' becasue it is not installed.

Nothing to do.
```
(Same as the previous example.)
```
1. install a-1-1.noarch that requires b, install dependency b-1-1.noarch
2. install c-1-1.noarch that requires b
# dnf5 history undo 1
Repositories loaded.
Extra package 'c-1-1.noarch' (with action 'Remove') which is not present in the stored transaction was pulled into the transaction.

You can try to add to command line:
  --ignore-extras to allow extra packages in the transaction
# dnf5 history undo 1 --ignore-extras
Removes a, b and c.
```

include/libdnf5/transaction/transaction_history.hpp
```diff
@@ -73,10 +73,21 @@ class TransactionHistory {
     /// @since 5.0
     libdnf5::BaseWeakPtr get_base() const;
 
+    /// Get reason for package specified by name and arch at a point in history
+    /// specified by transaction id.
+    ///
+    /// @param name Name of rpm package
+    /// @param arch Arch of rpm package
+    /// @return Reason of the last transaction item before transaction_id_point that
```
j-mracek, line 81: Parameter `transaction_id_point` is not described. May I ask you to add a description with a hint how to obtain it?

kontura: Thank you, added.

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,138 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        } else if (
+            package_replay.action == Action::REMOVE &&
+            (pkg.get_reason() == Reason::DEPENDENCY || pkg.get_reason() == Reason::WEAK_DEPENDENCY)) {
+            package_replay.reason = Reason::CLEAN;
+        } else {
+            package_replay.reason = pkg.get_reason();
+        }
+
+        // Don't fill in @System repo_id because it's not possible to perform forward actions from it
+        if (pkg.get_repoid() != "@System") {
```
j-mracek, line 2472: I think that setting `repoid` is not right for undo operation or it must be performed differently.

Think about case reverting:
upgrade a-2-1.noarch  repository updates
`replaced a-1-1.noarch @system`

It is transformed to operation
remove a-2-1.noarch  repository to_repo_id updates
`install a-1-1.noarch o_repo_id @system`

For remove step it is ignored. For install step it is not not set. It means the code make no differences.

I see two options:
 - store installed from in history DB and use that for undo operation
 - Ignore repo for `undo` operation to allow to use optimizations like local repository. Personally I prefer this option for undo operation. But storing installed from remove steps makes also some volue, but without impact on `undo` command.

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,138 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+            "Cannot revert action: \"{}\"",
+            transaction_item_action_to_string(pkg.get_action()));
+        package_replay.action = reverted_action->second;
+
+        // We cannot tell the previous reason if the action is REASON_CHANGE it could have been anything.
+        // For reverted action INSTALL and reason CLEAN the previous reason could have been either DEPENDENCY or WEAK DEPENDENCY
+        // to pick the right one we have to look into history.
+        if ((package_replay.action == Action::REASON_CHANGE) ||
+            (package_replay.action == Action::INSTALL && pkg.get_reason() == Reason::CLEAN)) {
+            auto nevras = rpm::Nevra::parse(pkg.to_string(), {rpm::Nevra::Form::NEVRA});
```
j-mracek, line 2455: I believe that `nevras` object is not required for following reasons:

1. It was already created above `package_replay.nevra = libdnf5::rpm::to_nevra_string(pkg);`
2. It is not required, because it can be used already parsed form
` package_replay.reason = history->transaction_item_reason_at(pkg.get_name(), pkg.get_arch(), reverting_transaction.get_id() - 1);`

j-mracek: Or i there any other reason why you want to construct and parse NEVRA from transaction pkg object?

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,138 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+                nevras.size() > 0, "Cannot parse rpm nevra \"{}\" while reverting transaction.", pkg.to_string());
+            // We look up the reason based on only name and arch, this means we could find a different
+            // version of installonly package however we store only one reason for ALL versions of
+            // installonly packages so it doesn't matter.
+            package_replay.reason = history->transaction_item_reason_at(
+                nevras[0].get_name(), nevras[0].get_arch(), reverting_transaction.get_id() - 1);
+        } else if (
+            package_replay.action == Action::REMOVE &&
+            (pkg.get_reason() == Reason::DEPENDENCY || pkg.get_reason() == Reason::WEAK_DEPENDENCY)) {
+            package_replay.reason = Reason::CLEAN;
```
j-mracek, line 2466: May I ask you why the reason is set here to `CLEAN`? What about to not set no reason for any remove step?

libdnf5/base/goal.cpp
```diff
@@ -2414,6 +2754,7 @@ base::Transaction Goal::resolve() {
 
     ret |= p_impl->add_reason_change_specs_to_goal(transaction);
 
+
```
j-mracek, line 2757: It is not a problem but additional line here is not required.

libdnf5/base/goal.cpp
kontura: I see, ok, I will drop the condition and ignore the repo.

libdnf5/base/goal.cpp
kontura: There is no valid reason here I will remove it. Thank you.

libdnf5/base/goal.cpp
kontura: My thinking was that this makes the transaction table more consistent.
It basically ensures that the removed dependencies are reported as `Removing unused dependencies` (like they would be for normal [not-reverted] transaction) not as `Removing dependent packages`.
If we keep the `DEPENDENCY` or `WEAK_DEPENDENCY` this reason is saved to the history because after resolving the replay code overrides the reasons.

However if there is some problem with this I am not against dropping it.

j-mracek: Thank you, this is a good idea. I am just curious whether the reason might be something else then `user`. Please keep it like it is.

j-mracek:
I have tested several scenario and it looks good. Only one thing is not perfect and it is related to messaging when a different version is installed and undo is supposed to remove. The behavior is correct.

```
1. install a-1-1.noarch
2. upgrade a-1-1.noarch to a-2-1.noarch
# dnf5 history undo 1
Repositories loaded.
Cannot perform action 'Remove' for Package 'a-1-1.noarch' becasue it is not installed.
You can try to add to command line:
  --ignore-installed to allow mismatches between installed and stored transaction packages.
```

In this case I would prefer to inform that  'a-1-1.noarch' is installed in a different version. And which one. The reason is that `--ignore-installed` will simply skip the action and undo will do nothing - no reason to perform undo at all.

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,130 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        // actions. We only have to keep the UPGRADE action to have a record of the operation.
+        if (group.get_action() != transaction::TransactionItemAction::UPGRADE) {
+            auto reverted_action = REVERT_ACTION.find(group.get_action());
+            if (reverted_action == REVERT_ACTION.end()) {
+                libdnf_throw_assertion(
+                    "Cannot revert action: \"{}\"", transaction_item_action_to_string(group.get_action()));
+            }
+            group_replay.action = reverted_action->second;
+        } else {
+            group_replay.action = transaction::TransactionItemAction::UPGRADE;
```
j-mracek, line 2487: I've tested it and I think that the revert of upgrade by upgrade is confusing. I am suggesting to add a warning, that group upgrade operation cannot be reverted, only RPMs will be modify. No error, only a warning. Or not a warning but event log would be better (new type I guess).

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,130 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        // actions. We only have to keep the UPGRADE action to have a record of the operation.
+        if (env.get_action() != transaction::TransactionItemAction::UPGRADE) {
+            auto reverted_action = REVERT_ACTION.find(env.get_action());
+            if (reverted_action == REVERT_ACTION.end()) {
+                libdnf_throw_assertion(
+                    "Cannot revert action: \"{}\"", transaction_item_action_to_string(env.get_action()));
+            }
+            env_replay.action = reverted_action->second;
+        } else {
+            env_replay.action = transaction::TransactionItemAction::UPGRADE;
```
j-mracek, line 2522: I am recommending to not present the operation as an upgrade, because in background it does nothing. But if new group appeared then it might modify groups.

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,130 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        }
+
+        // Don't fill in @System repo_id because it's not possible to perform forward actions from it
+        if (env.get_repoid() != "@System") {
+            env_replay.repo_id = env.get_repoid();
```
j-mracek, line 2527: Ibelieve that this is the same type of problem like with RPMs. What about to not use repo_id for undo?

libdnf5/base/goal.cpp
```diff
@@ -2206,6 +2412,130 @@ GoalProblem Goal::Impl::add_reason_change_to_goal(
...
+        if (group_replay.action == Action::INSTALL && group.get_reason() == Reason::CLEAN) {
+            group_replay.reason = Reason::DEPENDENCY;
+        } else if (group_replay.action == Action::REMOVE && group.get_reason() == Reason::DEPENDENCY) {
+            group_replay.reason = Reason::CLEAN;
+        } else {
+            group_replay.reason = group.get_reason();
+        }
+
+        // Don't fill in @System repo_id because it's not possible to perform forward actions from it
+        if (group.get_repoid() != "@System") {
```
j-mracek, line 2499: I believe that this is the same type of problem like with RPMs. What about to not use repo_id for undo?

libdnf5/base/goal.cpp
j-mracek: The reason can be group and the remove operation works nicely.

libdnf5/base/goal.cpp
kontura: Ok, I have turned into it a log event. Now a revert of a revert of group upgrade won't mention the group at all because the group action will disappear on the first revert but I think that is acceptable.

kontura:
I have added the additional reporting of other installed versions, to do that I had to rework it a bit. I have added several new replay goal actions but I think overall the code is now clearer.
Also the `--ignore-installed` hint message was enhanced.

kontura:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1497

dnf5/commands/history/arguments.hpp
```diff
@@ -41,6 +44,30 @@ class ReverseOption : public libdnf5::cli::session::BoolOption {
...
 };
 
+class IgnoreInstalledOption : public libdnf5::cli::session::BoolOption {
+public:
+    explicit IgnoreInstalledOption(libdnf5::cli::session::Command & command)
+        : BoolOption(
+              command,
+              "ignore-installed",
+              '\0',
+              _("Don't consider mismatches between installed and stored transaction packages as errors."),
```
j-mracek, line 54: I am thinking whether the description is clear enough. The first part is OK, but `stored transaction packages` is somehow unclear. It does not explain what in stored transaction is related and how the mismatch will be handled. For upgrade it will replace currently installed packages, but for remove step it will skip remove completely. I think that detailed description can be in man pages.

j-mracek:
I have discovered that PR does not contain an update of documentation. Or did I overlooked it somehow?

include/libdnf5/base/goal_elements.hpp
```diff
@@ -137,12 +140,22 @@ enum class GoalAction {
...
     DISABLE,
-    RESET
+    RESET,
+    REPLAY_INSTALL,
+    REPLAY_REMOVE,
+    REPLAY_UPGRADE,
+    REPLAY_REINSTALL,
+    REPLAY_REASON_CHANGE,
+    REPLAY_REASON_OVERRIDE,
+    REVERT_COMPS_UPGRADE
```
j-mracek, line 150: Just a question - why the last item does not use replay prefix?

include/libdnf5/base/goal_elements.hpp
```diff
@@ -137,12 +140,22 @@ enum class GoalAction {
...
 };
 
 /// Convert GoalAction enum to user-readable string
 std::string goal_action_to_string(GoalAction action);
 
+/// Check whether the action is a replay action
+bool goal_action_is_replay(GoalAction action);
```
j-mracek, line 157: Are you sure that you want to have this method in API? It is very specific and it is only used in `dnf5/main.cpp`.

j-mracek: I am so sorry, I've overlooked multiple usage somewhere else. Please ignore the comment above.

include/libdnf5/transaction/transaction_history.hpp
```diff
@@ -73,10 +73,23 @@ class TransactionHistory {
...
+    ///         has an rpm with matching name and arch.
+    TransactionItemReason transaction_item_reason_at(
+        const std::string & name, const std::string & arch, int64_t transaction_id_point);
+
 private:
     /// Create a new Transaction object.
     libdnf5::transaction::Transaction new_transaction();
 
+
```
j-mracek, line 92: Minor - Are you sure that you want to use 2 new lines in private section?

include/libdnf5/base/goal_elements.hpp
kontura: My thinking was that since it happens during the reverting the `REVERT` prefix is more appropriate. Technically it has nothing to do with replay even though the result of the revert is used by replay.

kontura:
> I have discovered that PR does not contain an update of documentation. Or did I overlooked it somehow?

Yes the docs update is missing, can we leave that for a separate PR please? This one is quite big enough.

libdnf5/transaction/transaction_sr.hpp
```diff
@@ -49,6 +49,7 @@ struct PackageReplay {
     TransactionItemAction action;
     TransactionItemReason reason;
     std::string group_id;
+    // This nevra doesn't contain epoch if it is 0
```
j-mracek, line 52: Perfect comment, thank you.

libdnf5/base/goal.cpp
```diff
@@ -170,6 +172,12 @@ class Goal::Impl {
     void set_exclude_from_weak(const std::vector<std::string> & exclude_from_weak);
     void autodetect_unsatisfied_installed_weak_dependencies();
 
+    GoalProblem add_replay_to_goal(
+        base::Transaction & transaction,
+        const transaction::TransactionReplay & replay,
+        GoalJobSettings settings,
+        std::filesystem::path replay_location = "");
```
j-mracek, line 179: May I ask you to somehow explain the relation between `replay_location` and `package_replay.package_path`.

dnf5/commands/history/arguments.hpp
j-mracek: The hint is perfect, I love it, and can be used as an inspiration. 

```
Cannot perform Remove action because 'zsh-html-5.9-8.fc39.noarch' is installed in a different version: 'zsh-html-5.9-10.fc39.noarch'.
You can try to add to command line:
  --ignore-installed to allow mismatches between installed and stored transaction packages. This can result in an empty transaction because among other things the option can ignore failing Remove actions
```

libdnf5/base/goal.cpp
```diff
@@ -633,11 +663,49 @@ GoalProblem Goal::Impl::add_transaction_replay_specs_to_goal(base::Transaction &
...
+
     for (const auto & package_replay : replay.packages) {
+        rpm_nevra_cache.insert(package_replay.nevra);
         libdnf5::GoalJobSettings settings_per_package = settings;
         settings_per_package.set_clean_requirements_on_remove(libdnf5::GoalSetting::SET_FALSE);
+
+        std::optional<libdnf5::rpm::Package> local_pkg;
+        if (!package_replay.package_path.empty()) {
+            local_pkg =
+                base->get_repo_sack()->add_stored_transaction_package(replay_location / package_replay.package_path);
```
j-mracek, line 696: This path merge is not clear why one path is extended by a prefix in `replay_location`. May I ask you for a comment?

libdnf5/base/goal.cpp
```diff
@@ -633,11 +663,49 @@ GoalProblem Goal::Impl::add_transaction_replay_specs_to_goal(base::Transaction &
...
+
+        const auto nevras = rpm::Nevra::parse(package_replay.nevra, {rpm::Nevra::Form::NEVRA});
+        libdnf_assert(
+            nevras.size() > 0, "Cannot parse rpm nevra \"{}\" while replaying transaction.", package_replay.nevra);
+
+        rpm::PackageQuery query_na(base);
+        query_na.filter_name(nevras[0].get_name());
+        query_na.filter_arch(nevras[0].get_arch());
+        auto query_nevra = query_na;
+        query_nevra.filter_nevra(package_replay.nevra);
```
j-mracek, line 707: In theory it would be possible to use  NEVRA object to avoud double parsing `query_nevra.filter_nevra(nevras[0]);`.

libdnf5/base/goal.cpp
```diff
@@ -633,11 +663,49 @@ GoalProblem Goal::Impl::add_transaction_replay_specs_to_goal(base::Transaction &
...
+
+        std::optional<libdnf5::rpm::Package> local_pkg;
+        if (!package_replay.package_path.empty()) {
+            local_pkg =
+                base->get_repo_sack()->add_stored_transaction_package(replay_location / package_replay.package_path);
+        }
+
+        const auto nevras = rpm::Nevra::parse(package_replay.nevra, {rpm::Nevra::Form::NEVRA});
+        libdnf_assert(
+            nevras.size() > 0, "Cannot parse rpm nevra \"{}\" while replaying transaction.", package_replay.nevra);
```
j-mracek, line 701: What about to change a condition - `nevras.size() == 1`. I believe that NEVRA type has only one solution.

libdnf5/base/log_event.cpp
```diff
@@ -153,23 +153,52 @@ std::string LogEvent::to_string(
                 } else {
                     return ret.append(utils::sformat(_("No match for group package: {}"), *spec));
                 }
+            } else if (goal_action_is_replay(action)) {
+                return ret.append(
+                    utils::sformat(_("Cannot perform {}, no match for: {}."), goal_action_to_string(action), *spec));
```
j-mracek, line 158: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators.

libdnf5/base/log_event.cpp
```diff
@@ -153,23 +153,52 @@ std::string LogEvent::to_string(
...
         case GoalProblem::NOT_FOUND_IN_REPOSITORIES:
             return ret.append(utils::sformat(
                 _("No match for argument '{0}' in repositories '{1}'"),
                 *spec,
                 utils::string::join(settings->get_to_repo_ids(), ", ")));
-        case GoalProblem::NOT_INSTALLED:
+        case GoalProblem::NOT_INSTALLED: {
+            if (goal_action_is_replay(action)) {
+                return ret.append(utils::sformat(
+                    _("Cannot perform {} for {} '{}' becasue it is not installed."),
```
j-mracek, line 169: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators?

libdnf5/base/log_event.cpp
```diff
@@ -153,23 +153,52 @@ std::string LogEvent::to_string(
...
+            }
             return ret.append(utils::sformat(_("Packages for argument '{}' available, but not installed."), *spec));
+        }
         case GoalProblem::NOT_INSTALLED_FOR_ARCHITECTURE:
             return ret.append(utils::sformat(
                 _("Packages for argument '{}' available, but installed for a different architecture."), *spec));
         case GoalProblem::ONLY_SRC:
+            if (goal_action_is_replay(action)) {
+                return ret.append(utils::sformat(
+                    _("Cannot perform {} because '{}' matches only source packages."),
```
j-mracek, line 182: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators? In this case the context message is questionable.

libdnf5/base/log_event.cpp
```diff
@@ -153,23 +153,52 @@ std::string LogEvent::to_string(
...
+                    goal_action_to_string(action),
+                    *spec));
+            }
             return ret.append(utils::sformat(_("Argument '{}' matches only source packages."), *spec));
         case GoalProblem::EXCLUDED:
+            if (goal_action_is_replay(action)) {
+                return ret.append(utils::sformat(
+                    _("Cannot perform {} because '{}' matches only excluded packages."),
```
j-mracek, line 190: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators?

libdnf5/base/log_event.cpp
```diff
@@ -153,23 +153,52 @@ std::string LogEvent::to_string(
...
+                    goal_action_to_string(action),
+                    *spec));
+            }
             return ret.append(utils::sformat(_("Argument '{}' matches only excluded packages."), *spec));
         case GoalProblem::EXCLUDED_VERSIONLOCK:
+            if (goal_action_is_replay(action)) {
+                return ret.append(utils::sformat(
+                    _("Cannot perform {} because '{}' matches only packages excluded by versionlock."),
```
j-mracek, line 198: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators?

libdnf5/base/log_event.cpp
```diff
@@ -191,9 +220,16 @@ std::string LogEvent::to_string(
         case GoalProblem::INSTALLED_IN_DIFFERENT_VERSION:
             if (action == GoalAction::REINSTALL) {
                 return ret.append(utils::sformat(
-                    _("Packages for argument '{}' installed and available, but in a different version => cannot "
-                      "reinstall"),
-                    *spec));
+                    _("Installed packages for argument '{}' are not available in repositories in the same version, "
```
j-mracek, line 223: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators.

libdnf5/base/log_event.cpp
```diff
@@ -191,9 +220,16 @@ std::string LogEvent::to_string(
...
+                      "available versions: {}, cannot reinstall."),
+                    *spec,
+                    libdnf5::utils::string::join(additional_data, ",")));
+            } else if (goal_action_is_replay(action)) {
+                return ret.append(utils::sformat(
+                    _("Cannot perform {} because '{}' is installed in a different version: '{}'."),
```
j-mracek, line 229: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators? Subjects are not explicitly mentioned, therefore the message without a context might be not clear.

libdnf5/base/log_event.cpp
```diff
@@ -220,6 +256,13 @@ std::string LogEvent::to_string(
         case GoalProblem::WRITE_DEBUG:
             return ret.append(utils::sformat(_("Debug data written to \"{}\""), *additional_data.begin()));
         case GoalProblem::UNSUPPORTED_ACTION:
+            if (action == GoalAction::REVERT_COMPS_UPGRADE) {
+                return ret.append(utils::sformat(
+                    _("{} upgrade cannot be reverted, however associated package actions will be. ({} id: '{}') ."),
```
j-mracek, line 261: May I ask you to mark an order of arguments `{1}`. It will allow to change order for translators. In this case what about to provide a context message for translators?

libdnf5/base/log_event.cpp
```diff
@@ -287,6 +330,13 @@ std::string LogEvent::to_string(
                 _("Error: It is not possible to switch enabled streams of a module unless explicitly enabled via "
                   "configuration option module_stream_switch."));
         }
+        case GoalProblem::EXTRA: {
+            return ret.append(utils::sformat(
+                _("Extra package '{}' (with action '{}') which is not present in the stored transaction was pulled "
```
j-mracek, line 335: May I ask you to mark an order of arguments `{1}`.

libdnf5/transaction/transaction_sr.hpp
```diff
@@ -49,6 +49,7 @@ struct PackageReplay {
...
     std::string nevra;
     std::filesystem::path package_path;
```
j-mracek, line 54: May I ask you for a doc string for a package path?

j-mracek (changes requested):
Thank you very much for the PR update. This version is is good, but I have some suggestions  that are not critical, let say they are more cosmetic. It means, it works according to expectations.

libdnf5/transaction/transaction_sr.hpp
kontura: Added.

include/libdnf5/base/goal_elements.hpp
j-mracek: Thank you

libdnf5/base/log_event.cpp
kontura: Added.

libdnf5/base/log_event.cpp
kontura: Added the marks but I don't know what are context messages? Can we perhaps add them separately?

libdnf5/base/log_event.cpp
kontura: Added the marks but I don't know what are context messages? Can we perhaps add them separately?

libdnf5/base/log_event.cpp
kontura: Added the marks but I don't know what are context messages? Can we perhaps add them separately?

libdnf5/base/log_event.cpp
kontura: Added the marks but I don't know what are context messages? Can we perhaps add them separately?

libdnf5/base/log_event.cpp
kontura: Added the marks but I don't know what are context messages? Can we perhaps add them separately?

libdnf5/base/log_event.cpp
kontura: Added.

libdnf5/base/log_event.cpp
kontura: Added.

libdnf5/base/goal.cpp
kontura: Updated.

libdnf5/base/goal.cpp
kontura: Added.

libdnf5/base/goal.cpp
kontura: Added a comment.

dnf5/commands/history/arguments.hpp
kontura: I have updated the description.

libdnf5/base/log_event.cpp
j-mracek: It is used to explain what message means. The method is `C_(<message for translaters>, <string to translate>)` with two string arguments and in DNF5 it is used only here - https://github.com/rpm-software-management/dnf5/blob/3fa8d63db9c060e1744cfd268cbba918941d2fd0/libdnf5/base/transaction.cpp#L1373.

j-mracek:
LGTM, documentation will be delivered in the next PR.

PR #1454, opened 2024-04-26
`history list`: count also groups and envs in total `Altered`

kontura:
dnf4 also counts them.

(It is required for enabling `undo` tests for dnf5.)


m-blaha (approved):
Thank you!

PR #1456, opened 2024-04-26
docs: Add a page about public API changes in dnf 5.2.0.0

jan-kolarik:
To better communicate the changes introduced with the new major version.

This will be then also referenced on the 5.2.0.0 GitHub release page.

### DIFF
--- a/doc/api/changes_dnf5.2.rst
+++ b/doc/api/changes_dnf5.2.rst
@@ -4,6 +4,8 @@
 
 This page lists the differences in the public API of `DNF5 5.2 <https://github.com/rpm-software-management/dnf5/releases/tag/5.2.0.0>`_ compared to the previous major version, DNF5 5.1.
 
+`Here <https://raw.githubusercontent.com/rpm-software-management/dnf5/main/doc/api/changes_dnf5.2.rst>`_'s the full diff showing the changes in the libdnf5 API.
+
 
 Wrapping struct attributes
 ==========================
@@ -51,7 +53,7 @@ Other changes
 
 * ``libdnf5::advisory::AdvisoryQuery``
 
-  * ``filter_*`` methods have the default ``cmp_type`` ``IEXACT`` (was ``EQ`` before)
+  * ``filter_type`` and ``filter_severity`` methods have the default ``cmp_type`` ``IEXACT`` (was ``EQ`` before)
 
 * ``libdnf5::repo::RepoSack``
 
--- a/doc/api/changes_dnf5.2.rst
+++ b/doc/api/changes_dnf5.2.rst
@@ -1,0 +1,64 @@
+###############################################
+ Modifications to the public API in dnf 5.2.0.0
+###############################################
+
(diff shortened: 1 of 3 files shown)


doc/api/changes_dnf5.2.rst
```diff
@@ -0,0 +1,64 @@
...
+
+
+Wrapping struct attributes
+==========================
+
+Public attributes in struct objects have been wrapped in getter and setter methods.
+
+Before: ``bool ignore_case;`` Now: ``void set_ignore_case(bool); bool get_ignore_case() const;``
+
+* ``libdnf5::GoalJobSettings``
```
bcl, line 15: It isn't immediately obvious what all this encompasses. For example, when `group_with_name` stops working I search the changelog for mention of it, not of GoalJobSetting.

Detailing all the changes would help when searching.

doc/api/changes_dnf5.2.rst
```diff
@@ -0,0 +1,64 @@
...
+* ``libdnf5::comps::GroupSack``
+* ``libdnf5::repo::Repo::fresh()`` (use ``is_expired()`` instead)
+
+
+Other changes
+=============
+
+* ``libdnf5::advisory::AdvisoryQuery``
+
+  * ``filter_*`` methods have the default ``cmp_type`` ``IEXACT`` (was ``EQ`` before)
```
gotmax23, line 54: What's the difference between `EQ` and `IEXACT`? Is it just that `IEXACT` is case-insensitive? What was the rationale for this change?

gotmax23: Hmm, I still see `EQ` as the default in

https://github.com/rpm-software-management/dnf5/blob/1fe03c783b4bba7288eb4aa795718199102caa31/include/libdnf5/rpm/package_query.hpp#L83

and elsewhere. Am I missing something?

pkratoch: Yes, `IEXACT` is just case insensitive. It's not all filter methods, just `filter_severity` and `filter_type`. The reason is that these two accept only specified values and it shouldn't matter if e.g. the type is "Bugfix" or "bugfix". See the commit: https://github.com/rpm-software-management/dnf5/commit/355aef18061bbee5599d70bd07da71d8db07c318

jan-kolarik: Also, the `IEXACT` default is related to the mentioned `libdnf5::advisory::AdvisoryQuery` object. You are referencing the `libdnf5::rpm::PackageQuery` object which should be unchanged.

doc/api/changes_dnf5.2.rst
jan-kolarik: I've added a file and linked it to the docs with diff between dnf5-5.1.17 and dnf5-5.2.0.0 API changes. I guess that would be better for the text search.

doc/api/changes_dnf5.2.rst
jan-kolarik: But yeah, I've used `filter_*`, although just some methods were affected by this. Fixed now.

jan-kolarik:
@bcl @gotmax23 Please check again. I've fixed the mentioned things and added an explicit diff with the API changes for better text search.

bcl (approved):
Thanks for the diff.

Issue #849, opened 2023-08-30
dnf5 not raising error until the end when running with no privileges


mattiaverga:
Running `dnf upgrade foo` immediately stops with an error warning the user that the command needs to be executed in privileged mode (sudo or root).

Running `dnf5 upgrade foo` instead performs all the tasks, downloads the update and then fails with a `permission denied`.

ppisar:
Already reported in Fedora Bugzilla <https://bugzilla.redhat.com/show_bug.cgi?id=2217842>.

ppisar:
I believe that DNF4 lies. Checking for EUID is not a proper test because in the world of SELinux a nonsuperuser can have plenty of other ways to obtain required privileges. Especially when multi-level security is employed.

DNF5 could temporarily acquire its lock before a transaction test. That would prevent users without a write access to a DNF database to continue in the operations which are supposed to modify the system (install, upgrade, remove, etc.). 

However, I believe that DNF5 should not deny users from performing a dependency resolution which happens before the transaction check. Querying DNF5 for action it would perform, in a sense of "dnf5 --assumeno ..." is a legitimate operation.

j-mracek:
> However, I believe that DNF5 should not deny users from performing a dependency resolution which happens before the transaction check. Querying DNF5 for action it would perform, in a sense of "dnf5 --assumeno ..." is a legitimate operation.

We can only allow that if we will be able to allow cache sharing between users. If `dnf5 --assumeno upgrade` and `sudo dnf5 -y upgrade` results in double cache download and processing to solv file then user might  get upset.

jan-kolarik:
Related existing issue: https://github.com/rpm-software-management/dnf5/issues/337.

jan-kolarik:
I would like to discuss various ways to accomplish this task. 

Here are some of my thoughts on the topic:
- **How to check** 🔒
  - Try to acquire a transaction write lock (instead of the current euid approach)
  - Allow user to pass a specific parameter explicitly ("I know I am not root, I just want to resolve deps")
- **When to check** 🕑
  - After the goal is resolved
  - Before the command is executed (after parsing)
  - At the beginning of the transaction run method
- **How to notify the user** 🔔
  - For each command requiring elevated privileges
  - After the transaction table is printed

As we seem to have reached a consensus for the default DNF5 behavior to be **strict**, I prefer the check to be performed before command processing (and before metadata is downloaded). I also support @ppisar's suggestion of switching to the approach of acquiring the lock.

[...]

PR #1458, opened 2024-04-27
Run transaction.test() before downloading and installing packages

krydos:
Hi, 
I recently faced the issue described in https://github.com/rpm-software-management/dnf5/issues/849 and [here on bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2217842) about "too late" permissions check while installing/upgrading packages. 

I've solved it by calling `Transaction::test()` in the beginning of `Context::Impl::download_and_run()`.
It is clearly [stated](https://github.com/rpm-software-management/dnf5/blob/4a05cfb857ee4a6c732051833dab1065cf4ddedb/include/libdnf5/base/transaction.hpp#L118) in the `test()` definition that it is redundant to call `test()` before `run()` but without calling it we should wait for download process to finish. 

Another possible inconvenience is when `Transaction::test()` returns non `TransactionRunResult::SUCCESS` I do not customize the error message. Let me know if you think I should check for error types and maybe customize the error message based on that. 

Please let me know if you think it can/should be implemented in a different way and share your recommendations.
Thanks!

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -395,6 +395,11 @@ void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
 }
 
 void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
+    if (transaction.test() != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
+      std::cerr << "Transaction has failed. Try to run it with superuser privileges (under the root user on most systems)"
+                << std::endl;
+      throw libdnf5::cli::SilentCommandExitError(1);
+    }
     if (!transaction_store_path.empty()) {
         auto transaction_location = transaction_store_path / "transaction.json";
         constexpr const char * packages_in_trans_dir{"./packages"};


dnf5/context.cpp
```diff
@@ -363,6 +363,12 @@ void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
         base.get_config().get_tsflags_option().set(libdnf5::Option::Priority::RUNTIME, "test");
     }
 
+    if (transaction.test() != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
```
j-mracek, line 366: I think it would be good to perform the test before download, but I don't think it is a good idea to call `transaction.test()` before RPMs are downloaded, because they are required for the test. The code also  does not consider use case with downloadonly option.

krydos: Could you please help me to figure out where exactly the downloaded RPMs are required for the test to run? 
It isn't really clear for me unfortunately. I don't see anything related to it in `libdnf5::base::Transaction::Impl::_run()`

Also the downloadonly option is apparently unaltered because `Context::Impl::download_and_run()` isn't called for `dnf5 download package_name`. I'm not that fluent in DNF though... feel free to correct me here. 

It probably would make way more sense to check for the particular error `TransactionRunResult::ERROR_LOCK` and show the error message only in case of this error... 

Should we rather have something like `transaction.lock_test()` that only check for permissions and don't rely on the whole `transaction.test()` functionality (which as you said requires RPMs to be downloaded). 

P.S.
Rebased on top of main as `download_and_run` just got some updates.

jan-kolarik:
Hi, thanks for your PR. I've reviewed it as I've just started working on issue #849. Based on the discussion there, I'm considering a more robust solution that would take place after parsing the command line but before downloading any metadata. I plan to check if the command and options provided by the user would cause any system state changes (including `downloadonly`, `assumeno`, etc.). If so, we'll verify write access to the RPM database before proceeding. I should have a prototype PR ready this week. Let me know if you have any concerns with this approach :slightly_smiling_face:

krydos:
@jan-kolarik, It definitely sounds way better than what I have here. Thanks!
If it's possible please let me know when you have your PR available and I'll close this one then.

jan-kolarik:
> @jan-kolarik, It definitely sounds way better than what I have here. Thanks! If it's possible please let me know when you have your PR available and I'll close this one then.

Here it is: https://github.com/rpm-software-management/dnf5/pull/1504.

jan-kolarik:
@krydos Anyway, thanks for your effort. It is greatly appreciated! Looking forward to any future contributions. :slightly_smiling_face:

krydos:
@jan-kolarik thank you for taking it further! I really miss this dnf5's behavior

Issue #1311, opened 2024-03-14
Skipped PGP checks - but not sure which package or repo


praiskup:
```
$ sudo dnf5 update --refresh
..
Total size of inbound packages is 408 MiB. Need to download 408 MiB.
After this operation 2 MiB will be used (install 963 MiB, remove 961 MiB).
Is this ok [y/N]: y
...
Warning: skipped PGP checks for 1 package(s).
```
It would be nice if DNF5 admitted what packages are guilty.

j-mracek:
What about to only extend message and include IDs of repositories, because as it was mentioned skipping is not per package property, but  per repository including commandline repository.

What about something like `Warning: skipped PGP checks for 1 package(s) from 'fedora' repository.`

praiskup:
> I believe that it simply pertains repositories which have signature verification disabled. (Either specifically in the repository configuration, or globally for all of them.) Otherwise, RPM would reject installing that package.

Don't we want to drop the warning, then? I mean, if I (intentionally) use `gpgcheck=0`, the warning seems to raise eyebrows unnecessarily.

PR #1459, opened 2024-04-29
Enhance warning about RPMs that were not validate by RPM

j-mracek:
DNF5 informs about number of packages that signature was not verified, but without any additional detail. The ID of repository provides a good hint for user why the check was skipped.

Closes: https://github.com/rpm-software-management/dnf5/issues/1311

Requires CI modification: https://github.com/rpm-software-management/ci-dnf-stack/pull/1498

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -1264,7 +1264,8 @@ bool Transaction::Impl::check_gpg_signatures() {
     // TODO(mblaha): DNSsec key verification
     libdnf5::rpm::RpmSignature rpm_signature(base);
     std::set<std::string> processed_repos{};
-    int num_checks_skipped = 0;
+    unsigned long num_checks_skipped = 0;
+    std::set<std::string> repos_with_skipped_checks;
     for (const auto & trans_pkg : packages) {
         if (transaction_item_action_is_inbound(trans_pkg.get_action())) {
             auto const & pkg = trans_pkg.get_package();
@@ -1277,6 +1278,7 @@ bool Transaction::Impl::check_gpg_signatures() {
             auto check_result = rpm_signature.check_package_signature(pkg);
             if (check_result == libdnf5::rpm::RpmSignature::CheckResult::SKIPPED) {
                 num_checks_skipped += 1;
+                repos_with_skipped_checks.insert(pkg.get_repo_id());
             } else if (check_result != libdnf5::rpm::RpmSignature::CheckResult::OK) {
                 // these two errors are possibly recoverable by importing the correct public key
                 auto is_error_recoverable =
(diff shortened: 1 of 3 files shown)


libdnf5/base/transaction.cpp
```diff
@@ -1098,7 +1100,11 @@ bool Transaction::Impl::check_gpg_signatures() {
         }
     }
     if (num_checks_skipped > 0) {
-        auto warning_msg = utils::sformat(_("Warning: skipped PGP checks for {} package(s)."), num_checks_skipped);
+        auto repo_string = libdnf5::utils::string::join(repo_of_checks_skipped, ", ");
+        auto warning_msg = utils::sformat(
+            _("Warning: skipped PGP checks for {0} package(s) from {1} repository(s)."),
+            num_checks_skipped,
+            repo_string);
```
pkratoch, line 1107: The "(s)" for "repository" doesn't work as well as for "package". Firstly, it doesn't match the plural form, secondly, it's less clear when there is a list of repositories and not just a number. For example, this reads a bit weird: "skipped PGP checks for 82 package(s) from fedora, updates repository(s)." I would change it, so that the list of repos is at the end: "skipped PGP checks for 82 package(s) from repositories: fedora, updates". What do you think?

libdnf5/base/transaction.cpp
```diff
@@ -1048,6 +1048,7 @@ bool Transaction::Impl::check_gpg_signatures() {
     libdnf5::rpm::RpmSignature rpm_signature(base);
     std::set<std::string> processed_repos{};
     int num_checks_skipped = 0;
+    std::set<std::string> repo_of_checks_skipped;
```
pkratoch, line 1051: I don't get this variable name. Wouldn't something like `repos_with_skipped_checks` be more intuitive?

pkratoch (commented):
Thanks for the patch. I have just a few minor notes that might be a matter of preference.

Also, the commit message says that "the ID of repository provides a good hint", but it doesn't say why. Can you please mention it's because the check is skipped per repo?

libdnf5/base/transaction.cpp
m-blaha: We can even leverage plural forms here:
```P_("skipped PGP checks for {} package from repository {}", "skipped PGP check for {} packages from repositories: {}", num_checks_skipped)```
Or something similar, plurals translations are currently not used in dnf5 so this would be a pioneer :)

Comment (j-mracek) [reply libdnf5/base/transaction.cpp:1107]: I was thinking about `P_`, but there are 3 possibility
"skipped PGP checks for {} package from repository {}"
"skipped PGP checks for {} packages from repository {}"
"skipped PGP checks for {} packages from repositories {}"

Comment (m-blaha) [reply libdnf5/base/transaction.cpp:1107]: Yeah, this is not exactly a straightforward example for `P_` as it contains two plurals. The only way would be splitting the message into two parts - one for package(s) and the other for repository(ies).
I'm not sure if this complication for translators is worth it, as in some languages it might be difficult to join such split sentences back together.

File: libdnf5/base/transaction.cpp [hunk 3]
```diff
@@ -1098,7 +1100,11 @@ bool Transaction::Impl::check_gpg_signatures() {
         }
     }
     if (num_checks_skipped > 0) {
-        auto warning_msg = utils::sformat(_("Warning: skipped PGP checks for {} package(s)."), num_checks_skipped);
+        auto repo_string = libdnf5::utils::string::join(repo_of_checks_skipped, ", ");
```
ppisar, line 1103: Please mark ", " for translation. The joiner is language-specific.

libdnf5/base/transaction.cpp
ppisar:
~~~~
utils::sformat(
  P_(
    "skipped PGP checks for {} package from repositories: {}",
    "skipped PGP check for {} packages from repositories: {}",
     num_checks_skipped),
  num_checks_skipped
  repo_string);
~~~~

j-mracek:
The PR and tests were updated according to suggestions.

Issue #1450, opened 2024-04-25
python: cannot set `ResolveSpecSettings`'s `nevra_forms` field


gotmax23:
Formally

``` python
import libdnf5

settings = libdnf5.base.ResolveSpecSettings()
settings.nevra_forms.append(libdnf5.rpm.Nevra.Form_NA)
```

allowed changing `nevra_forms` on a `ResolveSpecSettings` object. Now, it seems there's no way to do it.

``` python
import libdnf5

settings = libdnf5.base.ResolveSpecSettings()
settings.set_nevra_forms([libdnf5.rpm.Nevra.Form_NA])
```

fails

``` console
$ python repro.py
Traceback (most recent call last):
  File "repro.py", line 4, in <module>
    settings.set_nevra_forms([libdnf5.rpm.Nevra.Form_NA])
  File "/usr/lib64/python3.12/site-packages/libdnf5/base.py", line 2032, in set_nevra_forms
    return _base.ResolveSpecSettings_set_nevra_forms(self, nevra_forms)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: in method 'ResolveSpecSettings_set_nevra_forms', argument 2 of type 'std::vector< libdnf5::rpm::Nevra::Form,std::allocator< libdnf5::rpm::Nevra::Form > >'
```

kontura:
Thanks for the report, currently it has to be set like:
```python
settings.set_nevra_forms(libdnf5.rpm.VectorNevraForm(1, libdnf5.rpm.Nevra.Form_NA))
```
We also have it in a test: https://github.com/rpm-software-management/dnf5/blob/main/test/python3/libdnf5/rpm/test_package_query.py#L127-L128

I would like to make it work with the expected: `settings.set_nevra_forms([libdnf5.rpm.Nevra.Form_NA])` but so far I wasn't able to configure SWIG in such a way.

gotmax23:
How do I instantiate `VectorNevraForm` with more than one item?

kontura:
Currently this seems to be possible:
```python
forms = libdnf5.rpm.VectorNevraForm()
forms.push_back(libdnf5.rpm.Nevra.Form_NA)
forms.push_back(libdnf5.rpm.Nevra.Form_NAME)
forms.push_back(libdnf5.rpm.Nevra.Form_NEVRA)
[...]

PR #1466, opened 2024-05-02
bindings: Add Python attributes for structs

jan-kolarik:
This aligns with existing attributes for options in the configuration classes.

It should also resolve the issues breaking our libdnf5 clients, as the struct wrappers change was the only thing removed from the Python public API without prior deprecation.

### DIFF
--- a/bindings/libdnf5/base.i
+++ b/bindings/libdnf5/base.i
@@ -75,3 +75,12 @@
 
 %include "libdnf5/base/goal.hpp"
 %include "libdnf5/base/goal_elements.hpp"
+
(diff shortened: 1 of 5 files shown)


m-blaha:
@jan-kolarik Does this PR resolve https://github.com/rpm-software-management/dnf5/issues/1450 ?

m-blaha:
> @jan-kolarik Does this PR resolve #1450 ?

Oh, I see, it's only loosely related, but the actual problem there is a different one.

m-blaha (approved):
Thank you!

PR #1467, opened 2024-05-02
Filter advisory types and severity using AND operator and not OR

j-mracek:
It allows to better specify user request when user wants to update only advisory with certain type and severity.

Closes: https://github.com/rpm-software-management/dnf5/issues/1401


m-blaha:
The DNF team has decided not to change the command behavior at this stage and will only clarify the documentation instead.

PR #1474, opened 2024-05-09
Improve docs regarding the keepcache option and download command

### LINKED ISSUE #1278: Document download command behavior with relation to the keepcache option
We should explicitly mention in the documentation, particularly in `caching.7.rst`, that packages downloaded using the `download` command are exempt from the configured `keepcache` option.

jan-kolarik:
Closes #1278.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -387,6 +387,9 @@ Changes to individual options
 ``disable_excludes``
   * To disable all configuration file excludes, the ``*`` glob character is used now instead of the ``all`` to unify the behavior with query objects on the API.
 
+``keepcache``
+  * The behavior has been slightly modified, see the :ref:`Caching <caching_packages_ref-label>` man page for more information.
+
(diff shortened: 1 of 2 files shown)


doc/changes.rst
```diff
@@ -387,6 +387,9 @@ Changes to individual options
 ``disable_excludes``
   * To disable all configuration file excludes, the ``*`` glob character is used now instead of the ``all`` to unify the behavior with query objects on the API.
 
+``keepcache``
+  * The behavior has been slightly modified, see the :ref:`Caching <caching_packages_ref-label>` man page for more information.
```
kontura, line 391: I would find it more helpful if we directly said what is the change.

jan-kolarik: Sure, me too. Though I am not sure I'd come up with a concise and short description. Also, I am not entirely sure about the behavior of all use cases in dnf4, but I can try to examine...

Issue #1477, opened 2024-05-09
"dnf5 needs-restarting -s" segfaults


ppisar:
~~~~
# dnf5 needs-restarting -s
Updating and loading repositories:
Repositories loaded.
Segmentation fault (core dumped)
~~~~

Reproducible with dnf5 compiled from the current main HEAD (4b62aef58cb04b134e08c9bde2384ef0c0c1d55c) in Fedora 41:

~~~~
$ DNF5_PLUGINS_DIR=$PWD/redhat-linux-build/dnf5-plugins/needs_restarting_plugin gdb --args ./redhat-linux-build/dnf5/dnf5 -C needs-restarting -s
[...]
[New Thread 0x7ffff4c006c0 (LWP 5832)]
[Thread 0x7ffff4c006c0 (LWP 5832) exited]

Thread 1 "dnf5" received signal SIGSEGV, Segmentation fault.
0x00007ffff76c9a9e in std::__uniq_ptr_impl<libdnf5::solv::RpmPool, std::default_delete<libdnf5::solv::RpmPool> >::_M_ptr (this=0x8) at /usr/include/c++/14/bits/unique_ptr.h:193
193           pointer    _M_ptr() const noexcept { return std::get<0>(_M_t); }
(gdb) bt
#0  0x00007ffff76c9a9e in std::__uniq_ptr_impl<libdnf5::solv::RpmPool, std::default_delete<libdnf5::solv::RpmPool> >::_M_ptr (this=0x8) at /usr/include/c++/14/bits/unique_ptr.h:193
#1  0x00007ffff76c894a in std::unique_ptr<libdnf5::solv::RpmPool, std::default_delete<libdnf5::solv::RpmPool> >::get (this=0x8) at /usr/include/c++/14/bits/unique_ptr.h:464
#2  0x00007ffff76c78c4 in std::unique_ptr<libdnf5::solv::RpmPool, std::default_delete<libdnf5::solv::RpmPool> >::operator bool (this=0x8) at /usr/include/c++/14/bits/unique_ptr.h:481
#3  0x00007ffff76c6a47 in libdnf5::Base::Impl::get_rpm_pool (this=0x0)
    at /home/test/dnf5/libdnf5/./base/base_impl.hpp:48
#4  0x00007ffff76c6b88 in libdnf5::InternalBaseUser::get_rpm_pool (base=...)
    at /home/test/dnf5/libdnf5/./base/base_impl.hpp:116
[...]

PR #1478, opened 2024-05-09
needs_restarting: Fix invalid reference usage

m-blaha:
`current` is a reference to the last item of the `stack` vector, which
is being used after being invalidated by the `stack.pop_back()` call.

Resolves: https://github.com/rpm-software-management/dnf5/pull/1477


evan-goode (approved):
Thanks!

Issue #1480, opened 2024-05-09
Missing space before error message in string


HarmfulBreeze:
There should probably be a space between the colon and the error message: https://github.com/rpm-software-management/dnf5/blob/4b62aef58cb04b134e08c9bde2384ef0c0c1d55c/libdnf5/base/log_event.cpp#L249

ppisar:
The missing space is good because the second argument starts with a new-line as you can see 8 lines above. 

Though I agree that it's not obvious from from the message catalog. We should either document, or change the code. Provided the code suffers from other localization problem (missing plural forms, nonlilized item separators) , I'd rather change the code.

PR #1482, opened 2024-05-10
i18n: Improve formatting an error message for multiple streams

ppisar:
The leading message glued paragraphs and a missing space in _("Unable to resolve argument '{}':{}") was confusing for the translators.

A continuations of the message also did not suport plural forms.

Both issues are fixed with this patch. The singular form will never be used, but the many plural ones will do.

Fixes: #1480


m-blaha (approved):
Thank you!

PR #1485, opened 2024-05-11
download: add `--source` alias for `--srpm`

jlebon:
For compatibility with the download plugin in dnf4:

https://github.com/rpm-software-management/dnf-plugins-core/blob/a054dcbd64db7743b0af78ac8e06104c4bd6be6e/plugins/download.py#L52-L53


jan-kolarik:
Thanks! The two CI tests failures are unrelated and are currently being fixed in our `ci-dnf-stack` repo.

PR #1499, opened 2024-05-16
dnf5daemon: The buildtime attribute has been added to the package_attrs option

hollowiette:
Hello. I'm trying to develop a desktop application and using the capabilities of dnfdaemon, I came across the lack of a really useful buildtime attribute. Without it, I will not be able to correctly and quickly separate packages by date and recommend recent updates. I have added small changes that work on my stand. Now I can get this attribute by specifying it in package_attrs when accessing dbus methods, but I'm not completely sure that I won't need to change it somewhere else.


m-blaha (approved):
Thank you!

Issue #1502, opened 2024-05-17
Duplicated environment names in `dnf5 environment list`


bc-lee:
I encountered duplicate environment names when running the `dnf5 environment list` command. This might be due to upgrading Fedora 40 from an older version, as indicated by the following output:
```
$ dnf5 environment list
Updating and loading repositories:
Repositories loaded.
ID                                Name                                Installed
basic-desktop-environment         Basic Desktop                              no
budgie-desktop-environment        Budgie Desktop                             no
cinnamon-desktop-environment      Cinnamon Desktop                           no
cloud-server-environment          Fedora Cloud Server                        no
custom-environment                Fedora Custom Operating System             no
deepin-desktop-environment        Deepin Desktop                             no
developer-workstation-environment Development and Creative Workstatio        no
i3-desktop-environment            i3 desktop                                 no
infrastructure-server-environment Infrastructure Server                      no
kde-desktop-environment           KDE Plasma Workspaces                      no
lxde-desktop-environment          LXDE Desktop                               no
lxqt-desktop-environment          LXQt Desktop                               no
mate-desktop-environment          MATE Desktop                               no
minimal-environment               Minimal Install                            no
phosh-desktop-environment         Phosh Desktop                              no
server-product-environment        Fedora Server Edition                      no
sugar-desktop-environment         Sugar Desktop Environment                  no
sway-desktop-environment          Sway Desktop                               no
web-server-environment            Web Server                                 no
workstation-product-environment   Fedora Workstation                        yes
workstation-product-environment   Fedora Workstation                         no
xfce-desktop-environment          Xfce Desktop                               no
$ dnf5 environment info workstation-product-environment
Updating and loading repositories:
Repositories loaded.
Id                   : workstation-product-environment
Name                 : Fedora Workstation
Description          : Fedora Workstation is a user friendly desktop system for laptops and PCs.
Order                : 2
Installed            : True
Repositories         : @System
Required groups      : container-management
                     : core
                     : desktop-accessibility
                     : firefox
                     : fonts
                     : gnome-desktop
                     : guest-desktop-agents
                     : hardware-support
                     : libreoffice
                     : multimedia
                     : networkmanager-submodules
                     : printing
                     : workstation-product
Optional groups      : base-x
                     : domain-client

Id                   : workstation-product-environment
Name                 : Fedora Workstation
Description          : Fedora Workstation is a user friendly desktop system for laptops and PCs.
Order                : 2
Installed            : False
[...]

PR #1503, opened 2024-05-18
Added de-duplication for environment list print

bc-lee:
Previously, `dnf5 environment list` would print duplicate environment with the same id and name. It would be confusing for the user to see duplicate environment names. This patch introduces de-duplication for the environment list print feature to merge duplicate environment entries. The original order of the environments is preserved by assigning an index value to each environment entry.

Fixes: #1502

### DIFF
--- a/libdnf5-cli/output/environmentlist.cpp
+++ b/libdnf5-cli/output/environmentlist.cpp
@@ -23,6 +23,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include <libsmartcols/libsmartcols.h>
 
+#include <map>
+
 namespace libdnf5::cli::output {
 
 namespace {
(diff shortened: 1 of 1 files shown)


pkratoch:
Hi, thanks for the patch! I agree it should be de-duplicated, but I'd rather have the code same as the one for the groups where it's in the dnf5/commands part: https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/group/group_list.cpp#L73-L84 . It will be easier to maintain.

I also wanted to say that another advantage is that the de-duplication will work for both list and info, but unfortunately, the environment command doesn't reuse the list subcommand in the same way as group command does. But that's something we should change as well, though, it's not necessary to resolve in this PR, I can create a follow-up issue.

pkratoch:
> Thanks for the feedback. I appreciate it. I recognize that my solution in this PR isn't perfect, even from my own perspective. I'll take another look at it when I get a chance and see if I can improve it.

Hi, it's been some time. Do you still plan to work on this?

bc-lee:
Sorry, mate. I generally prefer to fix FOSS issues and submit PRs, but at the moment, I don't have enough time to work on this DNF5 PR, as it's a minor polish rather than a new feature that's immediately relevant to me.

pkratoch:
No problem. I'll close this, so that it's clear that the issue can be worked on by someone else. But if you ever want to come back to it, feel free to reopen it.

PR #1504, opened 2024-05-21
Implementation of unified privileges checking for CLI commands

jan-kolarik:
Using the existing RPM lock file functionality.

There are a hard-coded list of commands that are not allowed to run without superuser privileges. Then `assumeno` and `downloadonly` options are taken into account. Lastly, we check if provided command has defined the `store` option which should be defined for all transactional commands and therefore it should mean a need for elevated privileges.

For: #849.


jan-kolarik:
Actually, I think adding some suffix to the lock file path will be needed in case of testing the write privileges. That would still test the write access in the same directory, but not touching the existing transaction file in the potential concurrent dnf instance ...

jan-kolarik:
I am creating a new PR to address this as the Packit Copr repo is broken due to previous additional fixes.

PR #1505, opened 2024-05-22
Implementation of unified privileges checking for CLI commands

jan-kolarik:
Using the existing RPM lock file functionality.

There are a hard-coded list of commands that are not allowed to run without superuser privileges. Then `assumeno` and `downloadonly` options are taken into account. Lastly, we check if provided command has defined the `store` option which should be defined for all transactional commands and therefore it should mean a need for elevated privileges.

For: #849.
Successor of: #1504.

### DIFF
--- a/dnf5-plugins/builddep_plugin/builddep.cpp
+++ b/dnf5-plugins/builddep_plugin/builddep.cpp
@@ -88,6 +88,7 @@ void BuildDepCommand::set_argument_parser() {
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     create_allow_downgrade_options(*this);
+    create_store_option(*this);
 }
 
 void BuildDepCommand::configure() {
--- a/dnf5/commands/clean/clean.cpp
+++ b/dnf5/commands/clean/clean.cpp
@@ -127,7 +127,6 @@ void CleanCommand::set_argument_parser() {
     });
 
     cmd.register_positional_arg(cache_types);
-    create_store_option(*this);
 }
 
 void CleanCommand::run() {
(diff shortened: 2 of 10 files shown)


m-blaha:
I have several concerns regarding this pull request:

Firstly, I'm not sure that the centralized `cmd_requires_privileges()` function is the best choice. I'd prefer to delegate the decision of whether elevated privileges are required to the commands themselves, perhaps by adding a `bool Command::privileges_required()` method. For example, imagine a third-party dnf5 plugin implementing a new command that does not run a transaction but requires administrator privileges for some reason. How can this be handled with the current design? The plugin could be completely private to a customer, and dnf5 would not know anything about it, not even its name, to handle it in the centralized function.

Secondly, I don't think that having read/write access to `<install root>/run/dnf/rpmtransaction.lock` is sufficient to determine if a user has the necessary privileges to run the transaction. Consider the example mentioned in this comment https://github.com/rpm-software-management/dnf5/issues/849#issuecomment-1991314680. Although the user has full control over the specified install root, they still cannot run the RPM transaction successfully due to a failed `chown()` call.

jan-kolarik:
Hi Marek,

>  I'd prefer to delegate the decision of whether elevated privileges are required to the commands themselves, perhaps by adding a bool Command::privileges_required() method

Yeah, I was also thinking about this approach. Just this would require changing each command and making this a prerequisite when implementing a new one. This is a simple approach to the things.

> For example, imagine a third-party dnf5 plugin implementing a new command that does not run a transaction but requires administrator privileges for some reason

I was focused just on the core commands and plugins here. The 3rd party plugins could handle the situation in their way. Anyway, the current PR is really just trying to setup a single place where the initial check of privileges happens. If user doesn't have privileges for the operation, it would be denied anyway even without this PR.

> Secondly, I don't think that having read/write access to <install root>/run/dnf/rpmtransaction.lock is sufficient to determine if a user has the necessary privileges to run the transaction

Yes, it is not sufficient for this use case. This specific scenario could be handled later e.g. by the `install` command itself. The idea here is to handle the majority of common use cases by giving the user a unified message. In other cases, operation would fail as it was before...

m-blaha:
Thanks for clarification! Let's go with this simple approach than.

PR #1506, opened 2024-05-23
Fix Packit releases

jan-kolarik:
- Use new [v1.2](https://github.com/inknos/prepare-release/tree/v1.2) prepare-release action with 4-digits version format
- Create dist-git PR only for Rawhide


m-blaha (approved):
Thanks!

PR #1507, opened 2024-05-23
Build libdnf5 static library, re-enable unit tests that use hidden (private) libdnf5 symbols

jrohel:
Some unit tests use private symbols that are not exported from libdnf5.so shared library.
Therefore, a static version of the libdnf5 library is built in which all symbols are available. Unit tests that need private symbols use the static library.

Please merge after https://github.com/rpm-software-management/dnf5/pull/1307 .

### DIFF
--- a/include/libdnf5/base/goal_elements.hpp
+++ b/include/libdnf5/base/goal_elements.hpp
@@ -157,7 +157,7 @@ enum class GoalAction {
 LIBDNF_API std::string goal_action_to_string(GoalAction action);
 
 /// Check whether the action is a replay action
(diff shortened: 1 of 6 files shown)


jan-kolarik (approved):

jan-kolarik:
I am merging this, although CI is broken now. It seems to be due to previous issues in Packit buildroot in combination with race during merging history undo PR and [this](https://github.com/rpm-software-management/dnf5/pull/1307) one. Let's see and eventually fix the issues right in the subsequent PR.

PR #1509, opened 2024-05-26
builddep: Support the --with/--without options to toggle bconds

dm0-:
This adds the same options used by `rpmbuild`, `mock`, etc. for setting macros to flip build-time conditional switches when parsing spec files.  It provides a consistent interface across related tools.

They are aliases in `rpmbuild`, I'm not sure if there is a better way to accomplish something like that here:
```
rpmbuild alias --with           --define "_with_!#:+     --with-!#:+" \
        --POPTdesc=$"enable configure <option> for build" \
        --POPTargs=$"<option>"
rpmbuild alias --without        --define "_without_!#:+  --without-!#:+" \
        --POPTdesc=$"disable configure <option> for build" \
        --POPTargs=$"<option>"
```

### DIFF
--- a/dnf5-plugins/builddep_plugin/builddep.cpp
+++ b/dnf5-plugins/builddep_plugin/builddep.cpp
@@ -85,6 +85,36 @@ void BuildDepCommand::set_argument_parser() {
         });
     cmd.register_named_arg(defs);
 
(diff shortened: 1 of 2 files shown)


dm0-:
Okay, thanks, I updated and rebased it.

Issue #1493, opened 2024-05-15
dnfdaemon: Generate "transfer_id" in the daemon for "list_fd" operations


mcrha:
Follow up from https://github.com/rpm-software-management/dnf5/pull/1284#issuecomment-2068839230

It would be good to make the `transfer_id` an `out` argument of the `list_fd` operation, thus it can be generated on the server side, making sure it's a unique ID, instead of having complicated code in each client to try to achieve the uniqueness.

CC @m-blaha @anaselli

PR #1517, opened 2024-05-29
daemon: Generate transfer_id on server side

m-blaha:
The list_fd() method of the Rpm interface uses a unique transfer_id to
identify which particular file descriptor transfer has finished.
Previously, it was the client's responsibility to pass this unique
identifier to the server in the list_fd() method call. With this patch,
the identifier is generated on the server side, simplifying the method's
usage for clients.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1493


jan-kolarik (approved):
LGTM

PR #1519, opened 2024-05-29
Add --minimal option for check-ugrade command

j-mracek:
It is required for a feature parity with upgrade command where the option is already available.

Related: https://issues.redhat.com/browse/RHEL-38804

CI: https://github.com/rpm-software-management/ci-dnf-stack/pull/1511

### DIFF
--- a/dnf5/commands/check-upgrade/check-upgrade.cpp
+++ b/dnf5/commands/check-upgrade/check-upgrade.cpp
@@ -46,6 +46,18 @@ void CheckUpgradeCommand::set_argument_parser() {
     auto & cmd = *get_argument_parser_command();
     cmd.set_description("Check for available package upgrades");
 
+    minimal = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+    auto minimal_opt = parser.add_new_named_arg("minimal");
+    minimal_opt->set_long_name("minimal");
+    minimal_opt->set_description(
+        _("Reports the lowest versions of packages that fix advisories of type bugfix, enhancement, security, or "
+          "newpackage. In case that any option limiting advisories is used it reports the lowest versions of packages "
+          "that fix advisories matching selected advisory properties"));
+    minimal_opt->set_const_value("true");
+    minimal_opt->link_value(minimal);
+    cmd.register_named_arg(minimal_opt);
+
     changelogs = dynamic_cast<libdnf5::OptionBool *>(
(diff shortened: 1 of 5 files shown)


dnf5/commands/check-upgrade/check-upgrade.cpp
```diff
@@ -153,8 +176,13 @@ void CheckUpgradeCommand::run() {
             std::move(advisories.value()), installed_query, libdnf5::sack::QueryCmp::GTE);
     }
 
-    // Last, only include latest upgrades
-    upgrades_query.filter_latest_evr();
+    if (minimal) {
```
kontura, line 179: I believe this should be `minimal->get_value()`.
It is also breaking a couple of tests.

dnf5/commands/check-upgrade/check-upgrade.cpp
```diff
@@ -46,6 +46,17 @@ void CheckUpgradeCommand::set_argument_parser() {
     auto & cmd = *get_argument_parser_command();
     cmd.set_description("Check for available package upgrades");
 
+    minimal = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+    auto minimal_opt = parser.add_new_named_arg("minimal");
+    minimal_opt->set_long_name("minimal");
+    minimal_opt->set_description(
+        "Reports the lowest versions of packages that fix a bugfix, enhancement or a fix for a security issue "
+        "(security) on the system.");
```
kontura, line 55: Both this description and the man page don't mention the `newpackage` advisory type which is also used.

What about to change it a bit to: `Reports the lowest versions of packages that fix advisories of type bugfix, enhancement, security, or newpackage.`

Anyway I do like your description much more than the one in `upgrade` command which is quite ambiguous. Could they be unified? [There is even a TODO for it.](https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/upgrade/upgrade.cpp#L47)

kontura (changes requested):

dnf5/commands/check-upgrade/check-upgrade.cpp
j-mracek: May I ask you to explain your comment in more details?

j-mracek: There is an update for tests - https://github.com/rpm-software-management/ci-dnf-stack/pull/1511

j-mracek: I've got it, the test was really incorrect. The test was for pointer and not for bool value in the pointer of BoolOption. Fixed

kontura:
Great, thank you.

PR #1520, opened 2024-05-30
Move offline from dnf5 to libdnf5

m-blaha:
We will need to share the code also with dnf5daemon-server to support
offline upgrades using D-Bus API.

For: https://bugzilla.redhat.com/show_bug.cgi?id=2124503

### DIFF
--- a/include/libdnf5/transaction/offline.hpp
+++ b/include/libdnf5/transaction/offline.hpp
@@ -51,27 +51,47 @@ const std::string STATE_HEADER{"offline-transaction-state"};
 const std::filesystem::path DEFAULT_DATADIR{std::filesystem::path(libdnf5::SYSTEM_STATE_DIR) / "offline"};
 const std::filesystem::path TRANSACTION_STATE_FILENAME{"offline-transaction-state.toml"};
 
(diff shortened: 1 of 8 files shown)


include/libdnf5/transaction/offline.hpp
```diff
@@ -0,0 +1,115 @@
...
+    const std::exception_ptr & get_read_exception() const;
+    /// Returns path to the state file.
+    std::filesystem::path get_path() const;
+
+private:
+    /// Read offline transaction state data from the file
+    void read();
+    std::exception_ptr read_exception;
+    std::filesystem::path path;
+    OfflineTransactionStateData data;
```
kontura, line 98: Could we hide this into a `p_impl`?
Perhaps also the `OfflineTransactionStateData` could store the data in `p_impl` and have a bunch of getters/setters for the values?

This would allow us to do modifications later without breaking ABI.

m-blaha: Sounds reasonable, thanks!

m-blaha: I've added a commit with pImpls for both classes.

kontura:
Thanks!

PR #1523, opened 2024-05-31
RFC: Start documenting updateinfo.xml

stewartsmith:
The `yum`/`dnf` repository format isn't as well documented as one would necessarily like.

For `updateinfo.xml` specifically, it is currently parsed by many third party tooling for the purpose of gathering information about updates applicable to Linux distributions using `yum`/`dnf` repositories. Since the format of `updateinfo.xml` has not been well defined over the years, this presents challenges to authors of these tools.

With clear documentation on what exists in the wild, all authors of tools producing or parsing `updateinfo.xml` can refer to one canonical place for what exists in the wild and how to process it.

---

In order to get to this schema, I've gathered as many samples of `updateinfo.xml` as I can find. That's about 189 files and 1.8GB - which likely should not sit in the `dnf5` repository, but should probably sit somewhere.

Something that should come from that data set is various smaller snippets of `updateinfo.xml` that can be used by various tools for testing.

This is the list of the files I've used for validating the schema:
[updateinfo-xml-urls.txt](https://github.com/user-attachments/files/15517553/updateinfo-xml-urls.txt)

If anyone can think of any more, especially ones that may be different in some interesting way, please let me know!
---

I have the intent to create a version of the schema that is a Strict variant - with the view towards using that for code that *produces* `updateinfo.xml`. I'm putting this PR up first for initial feedback.

---

One question I have is if the `dnf5` repository is the right location for this. @Conan-Kudo assembled https://pagure.io/rpm-metadata/ from existing sources, but I did not find anywhere with complete documentation on the `updateinfo.xml` schema and what variations existed in the wild.

Issue #867, opened 2023-09-05
Implement machine-readable output for commands


jan-kolarik:
Based on the https://github.com/rpm-software-management/dnf5/issues/512, the target format for machine-readable output should be `json`, which is also supported by the `libsmartcols` library, commonly used for command output in DNF.

For the supported commands, machine-readable output will be requested with the `--json` option.

Commands to be implemented:
- [x] https://github.com/rpm-software-management/dnf5/pull/1522
- [x] https://github.com/rpm-software-management/dnf5/pull/1529
- [x] https://github.com/rpm-software-management/dnf5/pull/1531

This issue is targeted for Fedora 41. Other commands to be implemented are scheduled in the follow-up https://github.com/rpm-software-management/dnf5/issues/1906.

softwaresale:
It looks like there is a [JSON wrapper in copr already](https://github.com/rpm-software-management/dnf5/blob/dccea32e5a12e8fe6a4afa96139959d4c76556e6/dnf5-plugins/copr_plugin/json.hpp#L27). Would it be worth while to either:
1. Move that wrapper into libdnf5 so that it can be used across the different commands
2. Create a new wrapper in libdnf5 instead of reusing copr's
3. Not create a wrapper at all and just use raw json-c

jan-kolarik:
> It looks like there is a [JSON wrapper in copr already](https://github.com/rpm-software-management/dnf5/blob/dccea32e5a12e8fe6a4afa96139959d4c76556e6/dnf5-plugins/copr_plugin/json.hpp#L27). Would it be worth while to either:
> 
> 1. Move that wrapper into libdnf5 so that it can be used across the different commands
> 2. Create a new wrapper in libdnf5 instead of reusing copr's
> 3. Not create a wrapper at all and just use raw json-c

Hi, thanks for the suggestion. I was looking into copr plugin and there it is really used for processing the JSON input. Here for the machine-readable output we would probably just utilize the `libsmartcols` functionality to print the output in JSON format and probably do some transformation of the existing scols table into the json scols one. So it looks to me so far that no JSON library would be needed for this use case.

jan-kolarik:
Following-up on https://github.com/rpm-software-management/dnf5/issues/1072#issuecomment-1887165817, I was trying to utilize the `libsmartcols` API for JSON outputs. Unfortunately, I couldn't find a way to produce nested JSON objects. Even without that, the code became cumbersome when trying to transpose the existing tables for a more natural JSON output.

[...]

Issue #512, opened 2023-05-04
Decide on the format of machine readable outputs


m-blaha:
We should select and consistently use one output format. The obvious options are

- json (copr plugin needs json parsing)
- toml (already used by system state)
- yaml
- xml (used in comps group's `serialize` method, but the parsing itself is done by libsolv for us, so no external library used)
- something different?

Here is a (incomplete) list of dnf5 commands that would profit from machine readable outputs

- repoquery
- list
- info
- repoclosure
- updateinfo

praiskup:
Other option is YAML.  I prefer using it for most use-cases nowadays (for good human-readability, too).

praiskup:
> json (copr plugin needs json parsing)

Also [libdnf/transaction/Transformer.cpp](https://github.com/rpm-software-management/dnf5/blob/92624a6830ab764983068f5e6c1a495162b70702/libdnf/transaction/Transformer.cpp#L459).

praiskup:
The Copr plugin would benefit from (opt-in) machine-readable output as well.

evan-goode:
My vote would be for JSON, it is very simple and probably the most popular choice for this type of use case. It's plenty human-readable if pretty-printed. There's `jq` for parsing it in shell scripts.

TOML is nice for configuration, but it's not good IMO for arbitrarily-structured data with potentially many layers of nesting.

I have no strong opinion on YAML, I think that would be fine too.

j-mracek:
If smartcols has an option to export to json I would prefer json.

inknos:
> If smartcols has an option to export to json I would prefer json.
it does
https://cdn.kernel.org/pub/linux/utils/util-linux/v2.31/libsmartcols-docs/libsmartcols-Table.html#scols-table-enable-json

evan-goode:
Have we decided on JSON? Or maybe this is worth a post to fedora-devel to gather some more input?

jan-kolarik:
I think we "sort of" decided the JSON is the winner. Also, if we want to use `libsmartcols` for generating the output, it seems this is the only supported format there after I did some brief browsing through the docs...

jan-kolarik:
Agreed on `json`, using `--json` parameter to produce machine-readable output from dnf commands.

PR #1524, opened 2024-05-31
RFC: Add XML output option to 'dnf5 advisories list'

stewartsmith:
Add an option to output the list of applicable advisories in the same XML format as the updateinfo.xml file in the repository.

This allows tooling to have a consistent machine readable format for getting what advisories apply to a system. Resorting to parsing text output of commands is (naturally) error prone, and designed for humans to look at rather than machines.

Alternatives for these tools include using the APIs of dnf/yum/other-rpm-package-manager, and keeping up with their evolution. For tooling that supports a wide range of distributions and versions, this can become rather onerous, especially across the many different security/patch management products.

With a documented format, we can help these tools get the correct information both on available updates, and what ones are applicable to particular systems. Using the same format these tools need to be able to parse already avoids the need to get multiple parsers correct.

---
One issue I have with this code is that the user needs to pass `-q` to `dnf5` in order to get valid XML on `stdout`. Even with that, sometimes you get ` Fedora 40 openh264 (From Cisco) - aarc 100% |   3.1 KiB/s | 990.0   B |  00m00s` and similar. Hints on how to avoid this are welcome - I'd prefer to not have to go via a file on disk.

I dislike the `--xml-output` option name, perhaps a `--format=[xml|text|morse-code]`? Suggestions welcome.

I'd like feedback on if this should be part of core `dnf5` or a plugin. Being part of the core command does mean that third party tools can start *relying* on it being there rather than needing to also parse text output forever as well.

https://github.com/rpm-software-management/dnf5/pull/1523 is related, as it starts documenting the `updateinfo.xml` format.

I also intend to backport to https://github.com/rpm-software-management/dnf so this functionality can be added to existing distributions, but wanted feedback here first.

---
Sample output of `./dnf5/dnf5 -q advisory list --xml-output` on Fedora 40:
[advisory-list.xml.txt](https://github.com/user-attachments/files/15517832/advisory-list.xml.txt)


kontura:
We do have a plan for machine readable output format but we plan to use `json`: https://github.com/rpm-software-management/dnf5/issues/512 and https://github.com/rpm-software-management/dnf5/issues/867.
Is `xml` something you require or would you be fine with `json`?

> One issue I have with this code is that the user needs to pass -q to dnf5 in order to get valid XML on stdout. Even with that, sometimes you get  Fedora 40 openh264 (From Cisco) - aarc 100% |   3.1 KiB/s | 990.0   B |  00m00s and similar. Hints on how to avoid this are welcome

This appears to be a bug. The -q option should suppress that output. It was discussed here: https://github.com/rpm-software-management/dnf5/pull/1522#discussion_r1627348360

stewartsmith:
The main motivation for XML was to match a format that the tools have to parse anyway, avoiding them needing to correctly parse two formats.

The output of "dnf advisories list" may be the first time these tools are learning of a particular advisory. Parsing updateinfo from the repo is useful for a UI to browse and look up advisories, but for what applies to an instance, it needs to come from the instance to ensure all enabled yum repositories on that instance are covered.

Thus, if these are the same formats, it saves some extra code in these tools.

So, while JSON would functionally work, it would mean there's now two formats  that must be parsed correctly rather than one.

What are your thoughts on just having --output=text|updateinfoxml|json ?

stewartsmith:
One of the tools in question is SSM Patch Manager, a widely used AWS service. One benefit of a format with a well defined schema is that any output deviating from that schema can be alarmed on.

stewartsmith:
With https://github.com/rpm-software-management/dnf5/pull/1531 merged, I'm going to close this PR as agreement has been reached on JSON, and the code ends up being a bunch simpler.

PR #1529, opened 2024-06-05
repoinfo: Implement JSON output

jan-kolarik:
Turn on `quiet` mode when `--json` is requested. Also fixed suppressing repository download progress when `quiet` mode is turned on.

For: #867.

### DIFF
--- a/libdnf5-cli/output/repo_info.cpp
+++ b/libdnf5-cli/output/repo_info.cpp
@@ -24,7 +24,23 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5-cli/utils/units.hpp"
 
+#include <json-c/json_object.h>
+
+#include <iostream>
+
 namespace libdnf5::cli::output {
+
+namespace {
+
+std::vector<std::string> flatten_distro_tags(const std::vector<std::pair<std::string, std::string>> & distro_tags) {
+    std::vector<std::string> distro_tags_flat;
+    for (auto & key_value : distro_tags) {
+        distro_tags_flat.push_back(key_value.second + " (" + key_value.first + ")");
+    }
+    return distro_tags_flat;
+}
+
+}  // namespace
 
 class RepoInfo::Impl : public KeyValueTable {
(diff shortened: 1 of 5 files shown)


libdnf5-cli/output/repo_info.cpp
```diff
@@ -22,10 +22,25 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "key_value_table.hpp"
 #include "utils/string.hpp"
 
+#include "libdnf5-cli/tty.hpp"
 #include "libdnf5-cli/utils/units.hpp"
 
+#include <json.h>
```
kontura, line 28: This is a nitpick but I would prefer to include as little as possible, only: `json-c/json_object.h` should suffice but feel free to leave it as it is.

libdnf5-cli/output/repo_info.cpp
```diff
@@ -22,10 +22,25 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "key_value_table.hpp"
 #include "utils/string.hpp"
 
+#include "libdnf5-cli/tty.hpp"
```
kontura, line 25: Can I ask why not to include directly `<iostream>`?

libdnf5-cli/output/repo_info.cpp
jan-kolarik: Yep, agree.

libdnf5-cli/output/repo_info.cpp
jan-kolarik: Hmm, just a leftover, I've initially used something from that `tty` header, will change that...

kontura:
Thank you.

kontura:
Oh I also forgot to mention, I think it would be good to have some tests for the `--json` option.
Plus it would be nice to have a test for the `--refresh --quiet` combo.

jan-kolarik:
Yep, sure. Thanks for the notice. I will create a follow-up ticket for that.

PR #1531, opened 2024-06-05
Add JSON output to advisory list

stewartsmith:
Similar to the repolist --json output, do it for listing advisories that apply to the system ("dnf advisory list") as well as for the references currently supported Bugzilla and CVE types of references.

(examples edited as of 417cb70e248cfe938de60a7bacc13ba3b5257ca6)
Example of advisory list --json:
```json
[
  {
    "name":"FEDORA-2024-143f0443e1",
    "type":"unspecified",
    "severity":"None",
    "nevra":"yum-4.19.2-1.fc40.noarch",
    "buildtime":"2024-04-19 21:20:20"
  },
  {
    "name":"FEDORA-2024-1e8a70c30e",
    "type":"enhancement",
    "severity":"None",
    "nevra":"zstd-1.5.6-1.fc40.aarch64",
    "buildtime":"2024-04-19 21:20:20"
  }
]
```

Example of advisory list --with-bz --json:
```json
[
  {
    "advisory_id":"FEDORA-2024-b05fbc1127",
    "advisory_type":"enhancement",
    "advisory_severity":"enhancement",
    "nevra":"expat-2.6.2-1.fc40.aarch64",
    "references":[
      {
        "id":"2269400",
        "type":"bugzilla",
        "buildtime":"2024-04-19 21:20:20"
      }
    ]
  },
  {
    "advisory_id":"FEDORA-2024-b05fbc1127",
    "advisory_type":"enhancement",
    "advisory_severity":"enhancement",
    "nevra":"expat-devel-2.6.2-1.fc40.aarch64",
    "references":[
      {
        "id":"2269400",
        "type":"bugzilla",
        "buildtime":"2024-04-19 21:20:20"
      }
    ]
  }
]
```
---

Alternative implementation of https://github.com/rpm-software-management/dnf5/pull/1524 using the same style as the `repolist` JSON output added in 4ffa9d14a9ff5894635f71f7a4b004eb5f56e324.

### DIFF
--- a/include/libdnf5-cli/output/advisorylist.hpp
+++ b/include/libdnf5-cli/output/advisorylist.hpp
@@ -34,7 +34,16 @@ LIBDNF_CLI_API void print_advisorylist_table(
     std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_not_installed,
     std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_installed);
 
+LIBDNF_CLI_API void print_advisorylist_json(
+    std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_not_installed,
+    std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_installed);
+
 LIBDNF_CLI_API void print_advisorylist_references_table(
+    std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_not_installed,
+    std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_installed,
+    std::string reference_type);
+
+LIBDNF_CLI_API void print_advisorylist_references_json(
     std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_not_installed,
     std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_installed,
     std::string reference_type);
--- a/libdnf5-cli/output/advisorylist.cpp
+++ b/libdnf5-cli/output/advisorylist.cpp
@@ -21,8 +21,11 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "utils/string.hpp"
 
+#include <json.h>
 #include <libsmartcols/libsmartcols.h>
 #include <unistd.h>
+
+#include <iostream>
 
 namespace libdnf5::cli::output {
 
@@ -72,6 +75,14 @@ void add_line_into_advisorylist_table(
 
 }  // namespace
 
+static std::string get_reference_type_pretty_name(std::string type) {
+    if (type == "bugzilla") {
+        return "Bugzilla";
+    } else if (type == "cve") {
+        return "CVE";
+    }
+    return type;
+}
 
 void print_advisorylist_table(
     std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_not_installed,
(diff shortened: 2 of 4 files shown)


stewartsmith:
Mainly as a draft because I don't have any kind of schema / end user documentation on it (let alone tests), and I want to ensure that exists before anyone may start consuming it (or I backport to... 😢 )

libdnf5-cli/output/advisorylist.cpp
```diff
@@ -104,6 +106,70 @@ void print_advisorylist_table(
...
+        json_object * json_reference = json_object_new_object();
+	json_object_object_add(json_reference,
+			       "id", json_object_new_string(reference.get_id().c_str()));
+	json_object_object_add(json_reference,
+			       "type", json_object_new_string(reference.get_type().c_str()));
+	json_object_object_add(json_reference,
+			       "title", json_object_new_string(reference.get_title().c_str()));
+	json_object_object_add(json_reference,
+			       "url", json_object_new_string(reference.get_url().c_str()));
+        json_object_array_add(json_adv_references, json_reference);
```
kontura, line 152: List of references like this will be part of the `info --json` command output, here as well I would keep the information and its layout identical to the plain text command.

kontura (commented):
There appear to be some formatting problems (we have a style defined in `.clang-format`) check out the `Pre Commit Hooks` action.

libdnf5-cli/output/advisorylist.cpp
```diff
@@ -104,6 +106,70 @@ void print_advisorylist_table(
...
+    json_object_object_add(json_advisory, "type",
+			   json_object_new_string(advisory->get_type().c_str()));
+    json_object_object_add(json_advisory, "title",
+			   json_object_new_string(advisory->get_title().c_str()));
+    json_object_object_add(json_advisory, "description",
+			   json_object_new_string(advisory->get_description().c_str()));
+    std::string message = advisory->get_message();
+    if (message.size() > 0)
+        json_object_object_add(json_advisory, "message",
+			       json_object_new_string(message.c_str()));
```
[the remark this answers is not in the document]
stewartsmith: I'll poke at it to ensure it can be used for SSM Patch Manager (and likely others)

libdnf5-cli/output/advisorylist.cpp
stewartsmith: Updated to be much closer to the text output. I added the advisory id / name to the reference output (`--with-bz`/`--with-cve`) to make it easier to then pump these into `advisory info`, or to only pump the ones you don't already know about into it.

stewartsmith:
> There appear to be some formatting problems (we have a style defined in `.clang-format`) check out the `Pre Commit Hooks` action.

I should have fixed this up now...

libdnf5-cli/output/advisorylist.cpp
```diff
@@ -104,18 +107,68 @@ void print_advisorylist_table(
...
+    unsigned long long buildtime = advisory->get_buildtime();
+    json_object_object_add(
+        json_advisory, "buildtime", json_object_new_string(libdnf5::utils::string::format_epoch(buildtime).c_str()));
+    return json_advisory;
+}
+
+static json_object * adv_refs_as_json(
+    auto & pretty_reference_type, auto & reference_type, auto & adv_pkg, auto & advisory) {
+    json_object * json_advisory = json_object_new_object();
+    json_object_object_add(json_advisory, "advisory_id", json_object_new_string(advisory.get_name().c_str()));
```
kontura, line 126: What do you think about calling this `name` instead of `id`?
Just to be consistent with the output without references and with the API.

libdnf5-cli/output/advisorylist.cpp
```diff
@@ -104,18 +107,68 @@ void print_advisorylist_table(
...
+    json_object_object_add(json_advisory, "advisory_type", json_object_new_string(advisory.get_type().c_str()));
+    json_object_object_add(json_advisory, "advisory_severity", json_object_new_string(advisory.get_type().c_str()));
```
kontura, line 128: I am thinking about the `advisory_` attribute name prefix, while it makes clearer where the attribute belongs I don't like the change between the output without and with (`--with-bz/cve`) references.

What about to completely remove it or add `reference_` prefix to the references?

include/libdnf5-cli/output/advisorylist.hpp
```diff
@@ -34,9 +34,20 @@ LIBDNF_CLI_API void print_advisorylist_table(
...
     std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_installed);
 
+LIBDNF_CLI_API void print_advisorylist_json(
+    std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_not_installed,
+    std::vector<std::unique_ptr<IAdvisoryPackage>> & advisory_package_list_installed);
+
 LIBDNF_CLI_API void print_advisorylist_references_table(
     std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_not_installed,
     std::vector<libdnf5::advisory::AdvisoryPackage> & advisory_package_list_installed,
+    std::string pretty_reference_type,
```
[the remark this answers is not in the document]
stewartsmith: okay, I'll keep the `reference_type` and move the pretty name over to a separate function it can call.

Since `updateinfo.xml` in the wild lists a lot more than `bugzilla` and `cve`, see https://github.com/rpm-software-management/dnf5/blob/65076dc83ecccf3d75d18271c8922a7ab3c0cef1/doc/repository_format/updateinfo-permissive.xsd#L506-L531 

Thus, for a future patch, I want to support these too.... just not sure how we would want to deal with the pretty names for them.

include/libdnf5-cli/output/advisorylist.hpp
```diff
@@ -34,9 +34,20 @@ LIBDNF_CLI_API void print_advisorylist_table(
...
+    std::string reference_type);
+
+LIBDNF_CLI_API void print_advisorylist_references_json(
```
stewartsmith, line 47: I'll convert these to be similar to the table signature above, if only for consistency.

libdnf5-cli/output/advisorylist.cpp
stewartsmith: In the `updateinfo.xml` it is referred to as `id`:
https://github.com/rpm-software-management/dnf5/blob/65076dc83ecccf3d75d18271c8922a7ab3c0cef1/doc/repository_format/updateinfo-permissive.xsd#L86-L105

So I guess it's a bit of a preference here on if we continue with the `name` as is in the table output, or use `id` that is in the source data.

I don't have a strong opinion here, happy to use `name`

libdnf5-cli/output/advisorylist.cpp
stewartsmith: I'll add the `reference_` prefix, it seems the clarity would be useful here considering the interesting array of bugs I've seen around `updateinfo.xml` and friends.

libdnf5-cli/output/advisorylist.cpp
```diff
@@ -104,18 +107,68 @@ void print_advisorylist_table(
...
+    for (auto reference : references) {
+        json_object * json_reference = json_object_new_object();
+        json_object_object_add(json_reference, "id", json_object_new_string(reference.get_id().c_str()));
+        json_object_object_add(json_reference, "type", json_object_new_string(reference_type.c_str()));
+        json_object_object_add(json_reference, "pretty_type", json_object_new_string(pretty_reference_type.c_str()));
+        unsigned long long buildtime = advisory.get_buildtime();
+        json_object_object_add(
+            json_reference,
+            "buildtime",
+            json_object_new_string(libdnf5::utils::string::format_epoch(buildtime).c_str()));
```
stewartsmith, line 142: I just noticed this should be on the advisory rather than on the reference. Fixing.

include/libdnf5-cli/output/advisorylist.hpp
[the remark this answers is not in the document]
kontura: I was thinking about it and personally I would just drop the pretty name from the json output. For the plain text output it is just a column header to let the user know what it means but in json output the type information is already there. Plus we have to guess how it should look like.

I like the fallback to the type string (even for the plain text). :+1: 

Otherwise this looks quite good to me. I have [made an issue](https://github.com/rpm-software-management/ci-dnf-stack/issues/1513) to add tests to our CI and I think we can merge this.

kontura:
Oh and there is a couple of formatting issues again.

stewartsmith:
> Oh and there is a couple of formatting issues again.

I'm going to judge the `emacs` `clang-format`, but will fix.

kontura:
Great, thank you!

include/libdnf5-cli/output/advisorylist.hpp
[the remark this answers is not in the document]
kontura: The CI is a somewhat complicated. You can read the [README](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/README.md) for general overview and how to run it in a container.

Its also possible to run directly (which is fine in this case since these commands don't modify the host system in any way) described [here](https://github.com/rpm-software-management/ci-dnf-stack/tree/main/dnf-behave-tests). Though its a bit outdated, for dnf5 the behave command should look something like:
```
behave --tags @dnf5 -Ddnf_command=dnf5 ./ci-dnf-stack/dnf-behave-tests/dnf/updateinfo.feature
```
Dependencies from `requirements.spec` and python `requirements.txt` are needed.

For the actual tests I would say the features are quite self explanatory, look at: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/updateinfo.feature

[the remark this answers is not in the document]
kontura: I should add that the build with the `advisory list --json` option will be available in https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/builds/ tomorrow so if you build the container today by default it won't be there.

PR #1539, opened 2024-06-10
[libdnf, actions plugin] Support get/set repositories options, ver 1.1.0

### LINKED ISSUE #1345: libdnf actions plugin: possibility to modify per-repository configuration
It is possible to output `conf.<option_name>=<value>` from .actions to alter DNF configuration, but it'd also be nice to be able to override repository configuration that way, similarly to how it was possible to, say,

    for name, repo in self.base.repos.items():
        repo.<option_name> = <value>

Would probably also be useful to provide a `${repository_names}` to iterate over.

jrohel:
Argument `${conf.<repoid_pattern>.<option_name>[=<value_pattern>]}` is substituted by a list of "repoid.option=value" pairs for the matching repositories. Pairs are separated by ',' character. The ',' character in the value is replaced by escape sequence "\x2C".
The input `repoid_pattern` - is the repository id and can contain globs.
If `value_pattern` is used, only pairs with the matching value are listed. The `value_pattern` can contain globs.

Output line format for setting repository option:
`conf.<repoid_pattern>.<option_name>=<value>` - sets the value of option `option_name` for all matching repositories.
The `repoid_pattern` - is repository id and can contain globs.

Closes #1345

### DIFF
--- a/doc/libdnf5_plugins/actions.8.rst
+++ b/doc/libdnf5_plugins/actions.8.rst
@@ -94,6 +94,7 @@ Each non-comment line defines an action and consists of five items separated by 
    * ``${pid}`` - process ID
    * ``${plugin.version}`` - version of the actions plugin (added in version 0.3.0)
    * ``${conf.<option_name>}`` - option from base configuration
+   * ``${conf.<repoid_pattern>.<option_name>[=<value_pattern>]}`` - list of "repoid.option=value" pairs (added in version 1.1.0)
    * ``${var.<variable_name>}`` - variable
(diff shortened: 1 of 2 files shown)


jan-kolarik (approved):
LGTM, please add also CI tests for the newly introduced functionality.

jan-kolarik:
F39 tests are OK. Build in Rawhide is also OK. The other tests failures is very likely due to creating the PR before merging the https://github.com/rpm-software-management/dnf5/pull/1546 where the issues with `builddep` plugin was fixed. Merging this ...

PR #1543, opened 2024-06-10
dnfdaemon: Support to run transactions offline

m-blaha:
Adds support for "offline" option to Goal::do_transaction() dbus API method.

### DIFF
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
@@ -85,6 +85,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         Following @options are supported:
             - comment: string
                 Adds a comment to a transaction.
+            - offline: boolean, default false
+                If true, the transaction will be prepared to run during the next reboot. Otherwise, it will run immediately.
 
         Unknown options are ignored.
     -->
--- a/dnf5daemon-server/services/goal/goal.cpp
+++ b/dnf5daemon-server/services/goal/goal.cpp
@@ -28,7 +28,6 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "utils.hpp"
 
 #include <fmt/format.h>
-#include <libdnf5/transaction/offline.hpp>
 #include <libdnf5/transaction/transaction_item.hpp>
 #include <libdnf5/transaction/transaction_item_action.hpp>
 #include <sdbus-c++/sdbus-c++.h>
--- a/dnf5daemon-server/services/goal/goal.cpp
+++ b/dnf5daemon-server/services/goal/goal.cpp
@@ -28,7 +28,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "utils.hpp"
 
 #include <fmt/format.h>
-#include <libdnf5/repo/package_downloader.hpp>
+#include <libdnf5/transaction/offline.hpp>
 #include <libdnf5/transaction/transaction_item.hpp>
 #include <libdnf5/transaction/transaction_item_action.hpp>
 #include <sdbus-c++/sdbus-c++.h>
@@ -265,24 +265,6 @@ sdbus::MethodReply Goal::get_transaction_problems(sdbus::MethodCall & call) {
 }
 
 
-// TODO (mblaha) callbacks to report the status
-void download_packages(Session & session, libdnf5::base::Transaction & transaction) {
-    libdnf5::repo::PackageDownloader downloader(session.get_base()->get_weak_ptr());
-
-    // container is owner of package callbacks user_data
-    std::vector<std::unique_ptr<dnf5daemon::DownloadUserData>> user_data;
-    for (auto & tspkg : transaction.get_transaction_packages()) {
-        if (transaction_item_action_is_inbound(tspkg.get_action()) &&
-            tspkg.get_package().get_repo()->get_type() != libdnf5::repo::Repo::Type::COMMANDLINE) {
-            auto & data = user_data.emplace_back(std::make_unique<dnf5daemon::DownloadUserData>());
-            data->download_id = "package:" + std::to_string(tspkg.get_package().get_id().id);
-            downloader.add(tspkg.get_package(), data.get());
-        }
-    }
-
-    downloader.download();
-}
-
 sdbus::MethodReply Goal::do_transaction(sdbus::MethodCall & call) {
     transaction_resolved_assert();
     if (!session.check_authorization(dnfdaemon::POLKIT_EXECUTE_RPM_TRANSACTION, call.getSender())) {
@@ -293,29 +275,34 @@ sdbus::MethodReply Goal::do_transaction(sdbus::MethodCall & call) {
     dnfdaemon::KeyValueMap options;
     call >> options;
 
-    auto * transaction = session.get_transaction();
+    bool offline = dnfdaemon::key_value_map_get<bool>(options, "offline", false);
 
-    download_packages(session, *transaction);
+    if (offline) {
+        session.store_transaction_offline();
+        // TODO(mblaha): signalize reboot?
+    } else {
+        session.download_transaction_packages();
 
-    std::string comment;
-    if (options.find("comment") != options.end()) {
-        comment = dnfdaemon::key_value_map_get<std::string>(options, "comment");
+        std::string comment;
+        if (options.find("comment") != options.end()) {
+            comment = dnfdaemon::key_value_map_get<std::string>(options, "comment");
+        }
+
+        auto * transaction = session.get_transaction();
+        transaction->set_callbacks(std::make_unique<dnf5daemon::DbusTransactionCB>(session));
+        transaction->set_description("dnf5daemon-server");
+        transaction->set_comment(comment);
+
+        auto rpm_result = transaction->run();
+        if (rpm_result != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
+            throw sdbus::Error(
+                dnfdaemon::ERROR_TRANSACTION,
+                fmt::format(
+                    "rpm transaction failed with code {}.",
+                    static_cast<std::underlying_type_t<libdnf5::base::Transaction::TransactionRunResult>>(rpm_result)));
+        }
+        // TODO(mblaha): clean up downloaded packages after successful transaction
     }
-
-    transaction->set_callbacks(std::make_unique<dnf5daemon::DbusTransactionCB>(session));
-    transaction->set_description("dnf5daemon-server");
-    transaction->set_comment(comment);
-
-    auto rpm_result = transaction->run();
-    if (rpm_result != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
-        throw sdbus::Error(
-            dnfdaemon::ERROR_TRANSACTION,
-            fmt::format(
-                "rpm transaction failed with code {}.",
-                static_cast<std::underlying_type_t<libdnf5::base::Transaction::TransactionRunResult>>(rpm_result)));
-    }
-
-    // TODO(mblaha): clean up downloaded packages after successful transaction
 
     auto reply = call.createReply();
     return reply;
--- a/dnf5daemon-server/services/offline/offline.cpp
+++ b/dnf5daemon-server/services/offline/offline.cpp
@@ -1,0 +1,190 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "offline.hpp"
+
+#include "dbus.hpp"
+#include "utils/string.hpp"
+
+#include <libdnf5/transaction/offline.hpp>
+#include <sdbus-c++/sdbus-c++.h>
+
+#include <exception>
+#include <filesystem>
+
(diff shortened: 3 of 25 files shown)


dnf5daemon-server/services/goal/goal.cpp
```diff
@@ -293,30 +275,35 @@ sdbus::MethodReply Goal::do_transaction(sdbus::MethodCall & call) {
     dnfdaemon::KeyValueMap options;
     call >> options;
 
-    auto * transaction = session.get_transaction();
+    bool offline = dnfdaemon::key_value_map_get<bool>(options, "offline", false);
 
-    download_packages(session, *transaction);
+    if (offline) {
+        session.store_transaction_offline();
+        // TODO(mblaha): signalize reboot?
```
mcrha, line 282: Could you add API to check whether there's a pending offline update, please? It can be used by gnome-shell, to check whether the offline update is ready, then it asks the user `[x] Install pending update` in the power off/restart dialog and, when the user unchecks the option, the prepared update is "cancelled" (another new API) and the machine powers off/reboots without applying the offline update.

m-blaha: Thanks for suggestions! I'll add them. We actually have similar funcionality on CLI already - `dnf offline status` and `dnf offline clean`.

m-blaha: Added.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
```diff
@@ -94,6 +94,36 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+
+    <!--
+        offline_transaction_clean:
+        @options: an array of key/value pairs
+        @transaction_cleaned: boolean, returns `false` if there was an error during the transaction cleanup. Returns `true` if the offline transaction was successfully cleaned or if no offline transaction was configured.
+
+        Clean the offline transaction configured for the next reboot by removing the /system-update symlink (see systemd.offline-updates(7)).
+
+        Following @options are supported:
+            - clean_datadir: boolean, default true
```
mcrha, line 117: Would using a `false` as the default be a bit safer variant? You do not want to download 3GB of packages again just because you need to turn off the laptop and the app calling this does not pass the option, relying on the default value (or not knowing it exists, or when you split the option into clear packages, clear the other thing,...).

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
```diff
@@ -94,6 +94,36 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+                If true, all stored offline transaction data, including downloaded packages, will be removed along with the system-update symlink.
+
+        Unknown options are ignored.
+    -->
+    <method name="do_transaction">
```
mcrha, line 122: correct the `name`

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
```diff
@@ -94,6 +94,36 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        <arg name="options" type="a{sv}" direction="in" />
+        <arg name="transaction_cleaned" type="b" direction="out" />
+    </method>
+
 </interface>
 
 </node>
```
mcrha, line 129: I'm not sure whether adding this to the `Goal` is the best place. For me, the Goal manages actual transactions, something ready to be started. These two look like a method not related to actual transactions.

dnf5daemon-server/services/goal/goal.cpp
mcrha: I tested the `do_transaction` part and it works as expected. Thank you.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
```diff
@@ -94,6 +94,36 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
     </method>
 
+
+    <!--
+        offline_transaction_pending:
+        @pending: boolean, true if there is a pending offline transaction
+
+        Check whether there is an offline transaction configured for the next reboot (i.e. checks the /system-update symlink existence).
+    -->
+    <method name="offline_transaction_pending">
```
mcrha, line 104: maybe call it `get_offline_transaction_pending`?

mcrha:
Also, looking in how gnome-shell uses the PackageKit side of thing, there is a "trigger" method in the PackageKit's D-Bus "offline-update" interface and depending on what the user chose in the gnome-shell they run:
* trigger("reboot") - when the user chose to reboot the system and left "install updates" checked in the dialog
* trigger("power-off") - when the user chose to power off the system and left "install updates" checked in the dialog
* cancel offline update - when the user unchecked the "install updates" option in any of these dialogs.

In other words, the "trigger" function can set what to do after the offline update is finished.

I see you have there a `poweroff_after = false` in the `/usr/lib/sysimage/libdnf5/offline/offline-transaction-state.toml`, thus it might be doable (if that thing does what I think it does).

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
m-blaha: I agree it's not the best possible place. It's kind of loosely connected - the offline transactions are scheduled using `Goal::do_transaction(offline=true)`. But maybe introducing completely new `Offline` interface would be cleaner way.

mcrha: I did not think of a completely new interface, I thought some could be extended, but it seems none matches cleanly. The `Base` does only the `read_all_repos` and some signals, the `SessionManager` is not ideal either. I mean, you are right, adding the `Offline` interface looks like the best option.

m-blaha:
The `poweroff_after` option says that after the updates were applied (during the reboot), the system should power off instead of rebooting to the updated environment - see https://github.com/rpm-software-management/dnf-plugins-core/pull/467 .
It's not completely clear to me when this `trigger` function is executed - it seems that it somehow handles the `Reboot` and `Shut down` buttons to allow cancellation of the scheduled offline transaction (which can be done by calling the `clean` method of the `Offline` interface).

mcrha:
> The `poweroff_after` option says that after the updates were applied (during the reboot), the system should power off instead of rebooting to the updated environment - see [rpm-software-management/dnf-plugins-core#467](https://github.com/rpm-software-management/dnf-plugins-core/pull/467) .

Nice, in that case it behaves as I expected and it can be (re)used.

> It's not completely clear to me when this `trigger` function is executed

I'm not sure whether the `trigger` method only sets the desired action to be done after the offline update is finished, or whether it also does the reboot itself. From reading the gnome-shell code they do both, call PackageKit method and then initiate the reboot. From that the naming `trigger` is inaccurate, because it does not trigger the update at all, it also tells the PackageKit what to do after reboot; at least from my point of view.

What dnf5 may do is to add something like `set_finish_action(action)` method with `reboot` and `power-off` values values, beside the existing `clean()` method.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
m-blaha: Given we are moving to a new Offline interface, I'd call it simply `check_pending()`. What do you think?

mcrha: Sure, sounds good to me. The `get_status` would imply it can be more than just "is/is-not ready".

m-blaha: Exactly. Actually we can later provide a call returning all the details about the scheduled transaction.

m-blaha:
I pushed a new version that

- moves offline transaction methods to a separate `Offline` interface
- sets default of `clean_datadir` option for `Offline::clean()` method to false
- fixes the interface xml file
- adds new `set_finish_action()` method

evan-goode:
Looks good to me overall! A few comments:

- Does there need to be a `check_pending` method at all if there are plans to create a more general `get_status` method? 
- If we do have a `check_pending` method, it should probably also check offline-transaction-state.toml to check that the symlink points to a valid transaction.
- I would prefer to split `clean(bool clean_datadir)` into `clean()` and `cancel()`, since it's a bit confusing that the default behavior of `clean` is to not delete the contents of the datadir, which is different from `dnf5 offline clean`.

mcrha:
> Does there need to be a `check_pending` method at all if there are plans to create a more general `get_status` method?

You are right, having a single method sounds better than having two.

> I would prefer to split clean(bool clean_datadir) into clean() and cancel(),

The suggestion makes sense to me too.

I'm not the change owner, my comment is from the "client/API consumer" point of view.

m-blaha:
>     * Does there need to be a `check_pending` method at all if there are plans to create a more general `get_status` method?
> 
>     * If we do have a `check_pending` method, it should probably also check offline-transaction-state.toml to check that the symlink points to a valid transaction.

Currently `check_pending` is a simple method that only checks whether any service (not only dnf5) has scheduled an offline upgrade that will be performed during the next reboot (i.e. checks the presence of /system-update magic symlink). This can be used by gnome software to offer the user option to reboot the system without performing the update.

It differs from `dnf5 offline status` command which prints info from offline-transaction-state.toml regardless presence of the magic symlink.

>     * I would prefer to split `clean(bool clean_datadir)` into `clean()` and `cancel()`, since it's a bit confusing that the default behavior of `clean` is to not delete the contents of the datadir, which is different from `dnf5 offline clean`.

I like this idea!

mcrha:
> Currently `check_pending` is a simple method that only checks whether any service (not only dnf5) has scheduled an offline upgrade

I think it's not great, neither the `dnf offline status` you described (or the later at least if it won't tell me that the upgrade is ready, while it's not scheduled).

When I ask dnf5 whether there's an offline update/upgrade ready, I mean with it that the dnf5 has scheduled update/upgrade for the next boot, not that dnf5 or any other software happened to have scheduled an offline update. Aka the symlink presence is not a good indicator that the offline update is ready. I understand it still means that something will do the update/upgrade", but as long as it's not the dnf5, I'd just return "no" when asking dnf5 about the state.

For me, the symlink is just an implementation detail. Maybe it'll work differently in the future, or without systemd, or whatever, thus I use the library API to check the status instead of using the symlink directly. I'd not even mention it in the API documentation.

Imagine a case where something else will schedule the offline update - the symlink is there. The user will pick in the Power Off dialog of the GNOME Shell to not update, then the gnome-shell will ask dnf5 to discard/postpone the update. Will dnf5 unconditionally remove the symlink, even when it's not its, or it'll give up? What if the cancellation means for the other app something else than just removing the symlink, maybe some cleanup, state change written somewhere, then the dnf5 would make the other app in an inconsistent state.

I'm only thinking aloud. Feel free to ignore me.

m-blaha:
> I'm only thinking aloud. Feel free to ignore me.

Actually your comment makes a lot of sense. You are right that dnf5 should not touch upgrades not scheduled by it.

m-blaha:
I did another change suggested by reviewers:

- check_pending returns true only if the offline transaction was scheduled by dnf5
- split clean(bool) into
    - cancel() which only removes the magic symlink (if created by dnf5) so the offline transaction is not executed during reboot
    - clean() which cancels the transaction and removes all dnf5 stored offline data including downloaded packages
- I'm still not sure whether we should replace check_pending with something more powerful like get_status()

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -0,0 +1,190 @@
...
+    // try load the offline transaction state
+    const std::filesystem::path state_path{
+        libdnf5::offline::MAGIC_SYMLINK / libdnf5::offline::TRANSACTION_STATE_FILENAME};
+    libdnf5::offline::OfflineTransactionState state{state_path};
+    const auto & read_exception = state.get_read_exception();
+    if (read_exception == nullptr) {
+        // set the poweroff_after item accordingly
+        std::string finish_action;
+        call >> finish_action;
+        state.get_data().set_poweroff_after(finish_action == "poweroff");
```
mcrha, line 175: I'm fine to have it like this, even case sensitive, but it might be good to at least test whether the other action is `"reboot"` and claim a warning on the console about "unknown finish action" if it's neither `"poweroff"` nor `"reboot"`. That way one can check for typos and the older library will not crash when a new value is proposed in the future.

m-blaha: I can fill the `error_msg` output parameter with such warning. Returning an error reply seems too me.

mcrha: Returning an error, aka failing the call, would make the parameter strict. It's fine by me, it only has certain consequences.

evan-goode:
> It differs from `dnf5 offline status` command which prints info from offline-transaction-state.toml regardless presence of the magic symlink.

Yeah, `dnf5 offline status` doesn't check for the symlink since it's assumed that the system will be immediately rebooted after `dnf5 offline reboot` sets the status to `STATUS_READY` and creates the symlink. But now with this dnf5daemon functionality, `dnf5 offline status` should probably check and report the status of the symlink (in user-friendly terms, not necessarily mentioning the word "symlink"). The user should be able to tell whether the transaction will occur when they reboot via any means or whether it will only occur when they run `dnf5 offline reboot`.

> I'm still not sure whether we should replace check_pending with something more powerful like get_status()

I could go either way, but I still prefer one `get_status` function that returns two outputs (maybe `is_pending` and `status`). A caller interested in the `status` field (`download-incomplete`, `download-complete`, `ready`, etc.) is probably also interested in whether the transaction is scheduled to be performed next boot.

m-blaha:
> > I'm still not sure whether we should replace check_pending with something more powerful like get_status()
> 
> I could go either way, but I still prefer one `get_status` function that returns two outputs (maybe `is_pending` and `status`). A caller interested in the `status` field (`download-incomplete`, `download-complete`, `ready`, etc.) is probably also interested in whether the transaction is scheduled to be performed next boot.

OK, so let's replace check_pending() with get_status() which will have two out parameters:
- boolean is_pending (true if **dnf5** offline transaction is scheduled for the next reboot, i.e. magic symlink exists and points to /usr/lib/sysimage/libdnf5/offline/)
- dictionary string->variant with status of dnf5 offline transaction. It will contain all fields from offline-transaction-state.toml file (if it exists).

@evan-goode are you OK with this?

m-blaha:
Thanks! I don't plan any other changes right now, feel free to merge it.

mcrha:
I can confirm these changes do what I'd expect to be done from the gnome-software side.

Issue #1544, opened 2024-06-10
dnf5 `undo` - documentation and context message for translators


kontura:
The method is `C_(<message for translaters>, <string to translate>)` with two string arguments and in DNF5 it is used only here - https://github.com/rpm-software-management/dnf5/blob/3fa8d63db9c060e1744cfd268cbba918941d2fd0/libdnf5/base/transaction.cpp#L1373

kontura:
Resolved by: https://github.com/rpm-software-management/dnf5/pull/1549

PR #1549, opened 2024-06-11
`undo` command man page and translations

kontura:
I chose not to add the context messages for translators for now (https://github.com/rpm-software-management/dnf5/issues/1544) but I can add them if required.

### DIFF
--- a/doc/commands/history.8.rst
+++ b/doc/commands/history.8.rst
@@ -65,15 +65,55 @@ Subcommands
     | Replay the transaction that was previously stored into the file.
 
 
-Options
-=======
+Options for ``list`` and ``info``
+=================================
 
 ``--reverse``
     | Reverse the order of transactions in the output.
 
 
+Options for ``undo``
+====================
+
+``--skip-unavailable``
+    | Allow skipping packages actions that are not possible perform.
+
+``--ignore-extras``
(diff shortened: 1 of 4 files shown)


jan-kolarik:
Could you please add a reference to the history into `dnf5.8.rst` and also to `doc/commands/index.rst`?

doc/commands/history.8.rst
```diff
@@ -65,16 +65,56 @@ Subcommands
...
+    | Don't consider extra packages pulled into the transaction as errors.
+    | They will still be reported as warnings.
+
+``--ignore-installed``
+    | Don't consider mismatches between installed and stored transaction packages as errors.
+    | For install actions skip already installed packages.
+    | For upgrade actions skip groups or environments that are not installed.
+    | For remove actions skip not installed packages/groups/environments.
+    | They will still be reported as warnings.
+    | Using this option result in an empty transaction.
```
jan-kolarik, line 91: Maybe move the last two lines right after the first `Don't ...` line? Otherwise it's not really clear what `They will still be reported as warnings.` is referring to.

kontura: Sure, moved.

jan-kolarik (approved):

jan-kolarik:
Thanks!

Issue #613, opened 2023-06-14
Add message at the end of process?


hrw:
I run "dnf5 distro-sync" on my Fedora 38 system. It fetches packages, installs/upgrades, runs postinstall scripts, erases.

Then runs triggers. Many of them. And ends.

Can we get something like "Operation completed." message at the end?

```
[503/504] Erasing kf5-kquickcharts-0:5.106.0-1.fc38.x86_64                                                                                           100% |   9.8 KiB/s |  30.0   B |  00m00s
[504/504] Erasing kf5-kholidays-1:5.106.0-1.fc38.x86_64                                                                         100% [==================] |   0.0   B/s |  91.0   B | -00m00s
[504/504] Erasing kf5-kholidays-1:5.106.0-1.fc38.x86_64                                                                         100% [==================] |   0.0   B/s |  91.0   B | -00m00s
>>> Running post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
[504/504] Erasing kf5-kholidays-1:5.106.0-1.fc38.x86_64                                                                         100% [==================] |   0.0   B/s |  91.0   B | -00m00s
>>> Running post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
>>> Stop post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
[504/504] Erasing kf5-kholidays-1:5.106.0-1.fc38.x86_64                                                                         100% [==================] |   0.0   B/s |  91.0   B | -00m00s
>>> Running post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
>>> Stop post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
>>> Running post-transaction scriptlet: kernel-core-0:6.3.7-200.fc38.x86_64
[504/504] Erasing kf5-kholidays-1:5.106.0-1.fc38.x86_64                                                                         100% [==================] |   0.0   B/s |  91.0   B | -00m00s
>>> Running post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
>>> Stop post-transaction scriptlet: kernel-modules-core-0:6.3.7-200.fc38.x86_64
>>> Running post-transaction scriptlet: kernel-core-0:6.3.7-200.fc38.x86_64
[...]

PR #1553, opened 2024-06-20
Add "Complete!" message after succesfull transaction

j-mracek:
Keeping the last message something abou scriptlets is not clear message that everything passed correctly and program was terminated correctly. The commit uses the same message like DNF4. Contrary to DNF4, this commit does not add Complete! message when there is anything to perform - no transaction table.

Closes: https://github.com/rpm-software-management/dnf5/issues/613

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -1444,6 +1444,11 @@ int main(int argc, char * argv[]) try {
 
     log_router.info("DNF5 finished");
 
+    // Print Complete! message only when transaction is created to prevent poluting output from repoquery or other command
+    if (auto * transaction = context.get_transaction(); transaction && !transaction->empty()) {
+        context.print_info(_("Complete!"));
+    }
+
     return static_cast<int>(libdnf5::cli::ExitCode::SUCCESS);
 } catch (const libdnf5::Error & e) {
     std::cerr << libdnf5::format(e, libdnf5::FormatDetailLevel::WithName);


dnf5/main.cpp
```diff
@@ -1444,6 +1444,11 @@ int main(int argc, char * argv[]) try {
 
     log_router.info("DNF5 finished");
 
+    // Print Complete! message only when transaction is created to prevent poluting output from repoquery or other command
+    if (auto * transaction = context.get_transaction(); transaction && !transaction->empty()) {
+        std::cout << "Complete!\n";
```
jrohel, line 1449: Do we want to print the message also in quiet (`-q`) mode? If not, use `context.print_info("Complete");` instead of printing directly to `std::cout`.

In dnf5 quiet mode only partially works. For example, transaction table and transaction progress is visible even in quiet mode. We probably want to fix that as well. But that's out of the scope of this PR.

ppisar:
I hope the message does not pollute stdout of "dnf5 repoquery" output.

jan-kolarik:
> I hope the message does not pollute stdout of "dnf5 repoquery" output.

`// Print Complete! message only when transaction is created to prevent poluting output from repoquery or other command`

It seems, it should not :slightly_smiling_face:

dnf5/main.cpp
j-mracek: Good point, I am changing it to react to `-q`.

jan-kolarik, line 1449: I have just one point about the message, can we make it translatable?

ppisar, line 1449: I find the exclamation mark for an information message inappropriate. Could we use "Complete.\n" instead?

j-mracek: Anyway scriptlets are still visible in `-q` mode and even worse,  the new line is not added after the last entry.

j-mracek: I have no preferences. `"Complete!"` was used by DNF4.

jrohel:
Thanks.

Issue #1559, opened 2024-06-24
dnf5 has different groups then dnf-4


dschwoerer:
I tried to install gnome with `dnf5` - but that fails:
```bash
fakeroot dnf5 group install "Fedora Workstation"
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
No match for argument: Fedora Workstation
```

It does however succeed with `dnf-4`:
```bashfakeroot dnf group install "Fedora Workstation" 
Last metadata expiration check: 0:43:32 ago on Mon Jun 24 11:34:01 2024.
Dependencies resolved.
================================================================================
 Package                        Arch       Version            Repository   Size
================================================================================
Installing group/module packages:
[...]
```

I did not find anything related to why the group list is so much smaller in `dnf-4` compared to `dnf5`.

This was done on `fedora:40`

jan-kolarik:
Hi, the mentioned `Fedora Workstation` group is actually the Environment Group. These could be managed with `dnf5 environment` command or installed also with the common install command with `dnf5 install @<group-id>`.

jan-kolarik:
Though I agree it's confusing as the subcommands of `environment` are not consistent with the `group` command. This is already reported and tracked here: https://github.com/rpm-software-management/dnf5/issues/724.

dschwoerer:
Sorry, I did read the other issue, but failed to realise that dnf-4 list environments with groups.

This is indeed a duplicate of #724

Issue #724, opened 2023-07-14
[dnf5] Confusion when working with 'group' and 'environment' commands


zdohnal:
Hi,

usually if I need to upgrade my virtual machine to a desktop, I install environment group for Fedora Workstation. My DNF4 workflow was:

```
# dnf group list
# # find the correct string
# dnf -y group install <group name>
```

With DNF5, there is a confusion (at least for me :) ) - environment groups are not listed by ```dnf group list```, but by ```dnf environment list``` - so a common user would think: "Ok, 'environment' is your command for this now", however there is no 'install' subcommand for 'environment'... and user has to use 'group install' again, which is IMHO confusing, especially the fact the group is not listed by group command, but can be installed by it.

I thought about this and I would see the following solutions:

1. list environment groups under 'dnf groups list' again, which will provide backwards compatibility in behavior,
2. 'environment' command will have subcommand 'install'
3. mention in 'environment' command documentation+man page that installation is handled by 'dnf group' command.

WDYT?

jan-kolarik:
I remember we've already discussed in team that we should implement the `env install` and keep `group install` working with both groups and envs.

PR #1562, opened 2024-06-25
Add some docs about environments

dschwoerer:
This is not meant to stop implementing a nicer solution, but at least add some documentation on how it currently is.
I would be very happy to get the dnf-4 behaviour back, where `group` meant both env and group.

Related to #724 and #1559

### DIFF
--- a/doc/commands/group.8.rst
+++ b/doc/commands/group.8.rst
@@ -34,8 +34,9 @@ Description
 ===========
 
 The ``group`` command in ``DNF5`` offers several queries for getting information about groups, packages
-related to them and it is also used for groups installation. Note that some
-things dnf-4 listed as groups are now listed with ``environment list``.
+related to them and it is also used for groups installation.
+To query environments use separate ``environment`` command.
+Note: ``dnf-4`` listed both environments and groups with the ``group`` command.
 
 Optional ``group-spec`` arguments could be passed to filter only groups with given names.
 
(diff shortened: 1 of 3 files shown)


doc/commands/group.8.rst
```diff
@@ -34,7 +34,8 @@ Description
 ===========
 
 The ``group`` command in ``DNF5`` offers several queries for getting information about groups, packages
-related to them and it is also used for groups installation.
+related to them and it is also used for groups installation. Note that some
+things dnf-4 listed as groups are now listed with ``environment list``.
```
kontura, line 38: I find this new sentence little bit cryptic.

What about to move it to a separate line and change it to something like:
```
To query environments use separate ``environment`` command.
```

kontura (changes requested):
This is definitely a step in the right direction, thank you.

In general the docs for handling comps (groups and environments) need updating.
I have made an issue that should include this: https://github.com/rpm-software-management/dnf5/issues/1602

doc/commands/group.8.rst
dschwoerer, line 38: I think it would be useful to include that those environments used to be `groups`.

I read something about environments, but ignored that because I used to look at groups (not knowing that some groups where actual environments)

How about:
```suggestion
related to them and it is also used for groups installation.
To query environments use separate ``environment`` command.
Note: ``dnf-4`` listed both environments and groups with the ``group`` command.
```

kontura: ok, that makes sense.

PR #1569, opened 2024-07-01
TransactionReplay: handle group package types

kontura:
This information is stored in system state for each group and is used during group upgrade.

### DIFF
--- a/libdnf5/transaction/transaction_sr.cpp
+++ b/libdnf5/transaction/transaction_sr.cpp
@@ -130,6 +130,8 @@ TransactionReplay parse_transaction_replay(const std::string & json_serialized_t
         std::string reason;
         std::string group_path;
         std::string repo_id;
+        std::string joined_package_types;
+        comps::PackageType package_types;
 
         if (json_object_is_type(json_groups, json_type_array) == 0) {
             throw TransactionReplayError(M_("Unexpected type of \"groups\", array expected."));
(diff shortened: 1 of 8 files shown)


pkratoch:
Hi, can you please rebase? Not that there are merge requests, but in the commit https://github.com/rpm-software-management/dnf5/commit/1090963ff0488e107308557a306c2ffc6fddb336 , `libdnf5::transaction::GroupReplay` objects are created without the new `libdnf5::comps::PackageType` attribute.

kontura:
> Hi, can you please rebase? Not that there are merge requests, but in the commit [1090963](https://github.com/rpm-software-management/dnf5/commit/1090963ff0488e107308557a306c2ffc6fddb336) , `libdnf5::transaction::GroupReplay` objects are created without the new `libdnf5::comps::PackageType` attribute.

Thank you for catching that. :+1: 
Updated.

libdnf5/transaction/transaction_sr.cpp
```diff
@@ -158,13 +160,20 @@ TransactionReplay parse_transaction_replay(const std::string & json_serialized_t
             if (json_object_object_get_ex(group, "repo_id", &value) != 0) {
                 repo_id = json_object_get_string(value);
             }
+            if (json_object_object_get_ex(group, "package_types", &value) != 0) {
+                joined_package_types = json_object_get_string(value);
+                auto package_types_vec = libdnf5::utils::string::split(joined_package_types, ",");
+                std::for_each(package_types_vec.begin(), package_types_vec.end(), libdnf5::utils::string::trim);
+                package_types = comps::package_type_from_string(package_types_vec);
+            }
```
pkratoch, line 168: If the `package_types` is not in the json, it will remain uninitialized, and I think that would result in undefined behavior. So either the value should be required by the json format or it should default perhaps to the `group_package_types` configuration option.

pkratoch (changes requested):
This is probably just a matter of personal preference, but I would find it more logical if the "." were in the original `TransactionReplayError`s instead of adding it in the `LogEvent::to_string`, especially when I imagine that there could be a `TransactionReplayError` with multiple sentences as a message. That way, it would be well formatted on it's own. But I don't insist on this.

libdnf5/transaction/transaction_sr.cpp
kontura: I have added an `else` branch to initialize it to zero/no package types.
This matches the dnf4 behavior.

I believe this is the expected behavior, when no `package_types` are specified the value is empty.
This is the same like when `"package_types":""` is in the json.

I think this should happen when installing a group with the `--no-packages` option but currently its using the default `package_types` and I consider that a bug.

kontura:
> This is probably just a matter of personal preference, but I would find it more logical if the "." were in the original `TransactionReplayError`s instead of adding it in the `LogEvent::to_string`, especially when I imagine that there could be a `TransactionReplayError` with multiple sentences as a message. That way, it would be well formatted on it's own. But I don't insist on this.

I don't have a strong preference here, I would be fine either way but it seems most exceptions in libdnf5 don't end with the `.` so I wanted to make it consistent.

kontura:
I have added one more commit that fixes a related problem with group upgrade and package types.

pkratoch:
Thank you!

pkratoch (approved):

Issue #1518, opened 2024-05-29
Verbose mode for transaction table


jan-kolarik:
This is a follow-up issue to a prototype implementation introduced in https://github.com/rpm-software-management/dnf5/pull/1191.

For now, we don't want to make significant changes to the default transaction table layout. However, we can enable more detailed output via a CLI or configuration option. The output should preferably have a defined format and not be dependent on console width.

Here are the main points to discuss:

- What should be present in the table?
- When upgrading or replacing, should the original and new package values be on the same row or separate rows?
- What should be the standard console width for testing (80 / 100 / 120)?
- Should we introduce a hard limit for repository names? (e.g., Copr packages can be quite lengthy)
- Should we have different layouts for different operations (e.g., install vs. upgrade)?

Once the main points are settled, we can proceed with either finalizing the originally proposed PR or creating a new one, whichever makes more sense.

Related: https://github.com/rpm-software-management/dnf5/issues/326.

jan-kolarik:
We should also think about including scriptlet outputs (as mentioned [here](https://github.com/rpm-software-management/dnf5/issues/522#issuecomment-2132404834)) only with verbose mode. Otherwise keep it in log or even nowhere based on setup verbosity.

carlwgeorge:
> What should be present in the table?

I think the most important fields are:

- package name
- old version
- new version
- repository

Everything else is of secondary importance.  For the sake of minimalism and supporting more narrow terminals, I think limited ourselves to those four fields by default would be the best choice.  Additional fields could be enabled by configuration settings or CLI flags.

> When upgrading or replacing, should the original and new package values be on the same row or separate rows?

Definitely same row.  Yum and DNF4 only had one row per package, albeit with only the new values.  Aside from DNF5, I don't know of any other package manager with tabular output that uses separate rows to display old and new values of the same package.  Right now I think this is the biggest visual disruption for users switching from DNF4 to DNF5, and has resulted in at least three issues (#326, #407, and #1510) and one pull request (#1191) being filed so far.

> What should be the standard console width for testing (80 / 100 / 120)?

I don't have strong feelings on this one.  If we limit the default output to the four fields I mentioned above, 80 seems reasonable.  If we are going to include additional default fields like the size and/or architecture, it may be unavoidable to set the minimum at 100 or 120.

> Should we introduce a hard limit for repository names? (e.g., Copr packages can be quite lengthy)

[...]

Issue #785, opened 2023-08-01
Upgrade summary output too wide (> 80 cols)


rwmjones:
```
# dnf install wine --enablerepo=rawhide
Updating and loading repositories:
Repositories loaded.
Package                                      Arch   Version                 Repository      Size
Downgrading:                                                                                    
 openssl                                     x86_64 1:3.0.8-4.fc39          rawhide      1.6 MiB
  replacing openssl                          x86_64 1:3.0.9-1.fc38          updates      1.6 MiB
 openssl-devel                               x86_64 1:3.0.8-4.fc39          rawhide      4.1 MiB
  replacing openssl-devel                    x86_64 1:3.0.9-1.fc38          updates      4.1 MiB
 openssl-libs                                x86_64 1:3.0.8-4.fc39          rawhide      6.4 MiB
   replacing openssl-libs                    x86_64 1:3.0.9-1.fc38          updates      6.4 MiB
Upgrading:                                                                                      
 SDL2                                        x86_64 2.26.5-2.fc39           rawhide      1.8 MiB
  replacing SDL2                             x86_64 2.26.5-1.fc39           <unknown>    1.8 MiB
 SDL2-devel                                  x86_64 2.26.5-2.fc39           rawhide      2.7 MiB
  replacing SDL2-devel                       x86_64 2.26.5-1.fc39           <unknown>    2.8 MiB
 alsa-lib                                    x86_64 1.2.9-4.fc39            rawhide      1.4 MiB
  replacing alsa-lib                         x86_64 1.2.9-3.fc39            <unknown>    1.4 MiB
[etc]
```

This is wider than 80 columns.

rwmjones:
(Apparently you cannot put a screenshot here)

It looks like this: http://oirase.annexia.org/tmp/Screenshot_2023-08-01_11-34-49.png

kontura:
If there is an interactive terminal `dnf5` is trying to use the full width it has available so yes it can be wider than 80. I believe this is consistent with old `dnf` as well.

But it shouldn't be stretching the outputs over multiple lines like in your screenshot (unless you first print the transaction table and then resize the window).
In my testing if there is not enough space it cuts off the too long elements rather than spilling over the lines.

rwmjones:
I didn't resize the terminal.  Everything would fit into 80 columns fine if there wasn't so much whitespace.

kontura:
Interesting, do you have a reproducer for this?
Does it happen on every dnf5 command?

[...]

PR #1574, opened 2024-07-04
Drop replacing lines from the table when they are exact replacements

jan-kolarik:
Exclude `replacing` lines from the transaction table output when replacing a single package with the same name.

For https://github.com/rpm-software-management/dnf5/issues/326.


jrohel:
I understand the motivation for this change - fewer rows in the table.
The problem is that useful information is being removed, not just duplicated as stated in the linked issue.

For example, I check from which version to which version the change is happening before approving the upgrade.
Occasionally it happens that a package is replaced by a package with a different architecture, an arch package for noarch and vice versa. Or a package from a different repository.

I understand that not everyone checks what is happening in the system during an upgrade. But I do use this information and it is one of the reasons for me why dnf5 is better than dnf. 

I guess we'll have to figure out the "verbosity" of the output somehow, since everyone has different preferences. This PR will satisfy users who want a shorter output at the expense of information (or who have super memory and know exactly what versions and from where they have installed and thus don't need to display it). But it will annoy those who are interested in details. Especially when I manage several different computers I am interested in the exact state of the upgraded computer when upgrading.

jan-kolarik:
There is an intent to have the verbosity configurable with a default level "not detailed". The verbose mode should be implemented by the https://github.com/rpm-software-management/dnf5/issues/1518 after the discussion in https://github.com/rpm-software-management/dnf5/pull/1191.

jan-kolarik:
There isn't a consensus within the team on this approach. Let's wait until things settle around the verbose mode, and then we can decide. We can eventually merge it together when the verbose option is ready.

jan-kolarik:
> @jan-kolarik can you please shine some light on the status? more and more tickets on this topic are open, it would be great to point users to one place with up-to-date info to manage expectations. Thanks :)

As of now, I can't provide specific details about this. It's on our radar, and we plan to implement verbose mode and make some changes to the default output—though preferably minimal compared to the current state (mainly wrt "replacing" things). I see this as more of a Q1 2025 or later task, as we're currently overwhelmed with higher-priority work.

ferdnyc:
As an in-betweening of this, in https://github.com/rpm-software-management/dnf5/issues/785#issuecomment-2585599434 I suggested changing the "replacing" line format for same-package replacements from:

```text
<package_name>             <arch> <new_version>   <repo> <size>
  replacing <package_name> <arch> <old_version>   <repo> <size>
```

to:

```text
<package_name>             <arch> <new_version>   <repo> <size>
  replacing                       <old_version>
```

Which would help with the line-wrapping issues frequently caused by the overly-long "`␣␣replacing <package_name>`" lines in the first column.

mpx:
Please consider ensuring that output always fits on a single line. It's easier (for me) to look at a specific column for the data I want, rather than interleaving the data across lines.

From a user experience POV: previously I was able to tell at a glance what was happening (which packages/versions were being upgraded/installed) - now it is easily several orders of magnitude more difficult. I need to carefully look at each line (often hundreds) and it is difficult to absorb - this is made worse since the vertical space I need to scan has been doubled.

Most of the old/new version numbers are basically the same due to small updates so the signal to noise is extremely low for me which is disproportionate to the time cost.

Usually the only time I care about the previous version is when investigating a fault, and I'm going to look at the past logging for that.

Issue #1577, opened 2024-07-05
DNF5 Bug: Absolute path to local RPM package activates download of filelists


jrohel:
If the path to the local RPM package is specified as relative, the filelists download will not be activated. If it is specified absolutely (starts with the '/' character), the filelists download is triggered.
This is because DNF5 thinks that user wants to install a missing file and not a local RPM package. However, it later determines correctly that it is a RPM package based on the ".rpm" extension.

In all places, DNF5 should behave the same way and recognize that it is a RPM package.

A bug was discovered in https://github.com/rpm-software-management/dnf5/issues/1570 issue.

PR #1578, opened 2024-07-05
Fix DNF5: Don't trigger filelists download if abs path to local RPM

jrohel:
If the path to the local package was specified absolutely (starting with '/'), filelists download was triggered.
Now, as in `Goal`, DNF5 checks if the path ends with the ".rpm" suffix. If it does, it is treated as the path to the local RPM package and filelists download is not triggered.

Closes: https://github.com/rpm-software-management/dnf5/issues/1577


m-blaha (approved):
Thank you!

PR #1582, opened 2024-07-15
Use `SOLVER_FLAG_FOCUS_NEW` to install latests versions of deps

kontura:
This is a new flag in the recent libsolv 0.7.30 release (not yet present in rawhide).

We previously tried to use `SOLVER_FLAG_FOCUS_BEST` but it turned out to be too strong, this new flag should install latest versions of dependencies and update packages only if it is required.

We had several reports where users expect this behavior.

This should properly resolve issue such as:
https://bugzilla.redhat.com/show_bug.cgi?id=2150300
https://issues.redhat.com/browse/RHEL-1374


kontura:
Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1526

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ option(WITH_MODULEMD "Build with modulemd modules support" ON)
 option(WITH_ZCHUNK "Build with zchunk delta compression support" ON)
 option(WITH_SYSTEMD "Build with systemd and D-Bus features" ON)
 option(ENABLE_SOLV_URPMREORDER "Build with support for URPM-like solution reordering?" OFF)
+option(ENABLE_SOLV_FOCUSNEW "Build with SOLVER_FLAG_FOCUS_NEW libsolv flag enabled to ensure new dependencies are installed in latests versions?" OFF)
```
Conan-Kudo, line 37: We should just set this to be enabled by default and turn it off for incompatible releases.

dnf5.spec
```diff
@@ -798,7 +809,9 @@ automatically and regularly from systemd timers, cron jobs or similar.
     -DVERSION_PRIME=%{project_version_prime} \
     -DVERSION_MAJOR=%{project_version_major} \
     -DVERSION_MINOR=%{project_version_minor} \
-    -DVERSION_MICRO=%{project_version_micro}
+    -DVERSION_MICRO=%{project_version_micro} \
+    \
+    -DENABLE_SOLV_FOCUSNEW=%{?focus_new:ON}%{!?focus_new:OFF}
```
Conan-Kudo, line 814: Can you move this up above the `-DWITH_DNF5DAEMON_CLIENT` line?

dnf5.spec
```diff
@@ -108,6 +108,13 @@ Provides:       dnf5-command(versionlock)
 %bcond_with    performance_tests
 %bcond_with    dnf5daemon_tests
 
+# Enable SOLVER_FLAG_FOCUS_NEW only for current rawhide
+%if 0%{?rhel} || 0%{?fedora} < 41
+%bcond_with    focus_new
+%else
+%bcond_without focus_new
+%endif
```
Conan-Kudo, line 116: Let's flip this, since we only care about old releases failing here.

```suggestion
# Disable SOLVER_FLAG_FOCUS_NEW only for old RHEL
%if 0%{?rhel} && 0%{?rhel} < 10
%bcond_with    focus_new
%else
%bcond_without focus_new
%endif
```

Conan-Kudo (changes requested):

dnf5.spec
kontura: We did wanted to enable this only for rawhide to test it out but since dnf5 is default only in rawhide maybe we could do this to simplify? @jan-kolarik

jan-kolarik: Yes I think we should take the approach as with other features, to have upstream matching Rawhide and the older releases to use a patch turning this off. Just beware that for now we still want to keep this turned on just for Rawhide, so we also need to switch off for the upcoming Fedora 41 release.

Conan-Kudo: Why do we want to switch it off for Fedora at all? It's not used by default in older releases, and I'm not sure what's problematic about this flag. If you think there's potentially issues with it, we shouldn't do it at all. My understanding here is that the problem is related to having a new enough libsolv, but that's already in progress to land across all Fedora releases currently supported.

jan-kolarik: > I'm not sure what's problematic about this flag

This is basically the reason. We are also not sure :slightly_smiling_face: So, we want to be more careful about it. We'll offer it first to Rawhide users and see if any issues emerge.

Conan-Kudo: Maybe we should think about making solver flags runtime-configurable like Zypper does...

Conan-Kudo:
The libsolv update for F40 is queued for stable this evening, the update for F39 still needs positive karma to land today.

PR #1585, opened 2024-07-16
Improvements and fixes for storing transactions

kontura:
For: https://github.com/rpm-software-management/ci-dnf-stack/pull/1527


kontura:
Just to clarify the `Store to the db only requested package types for group` commit, when a history transaction is stored (`history store` command) is takes the used `package_types` from the db. The problem was that the db always contained all types present in the selected group, now it contains only the selected types.

For example if someone preformed a `group install` with just mandatory packages and then `history store`d the transaction in the json it would wrongly say the transaction contained all present types (not just mandatory).

Issue #1583, opened 2024-07-15
offline: Running update with local package fails


mcrha:
I downloaded a package (`patch`) from koji, with a newer version than I've installed, and I wanted to update it, but in the offline (to save some bandwidth) and when I do that, I get:

```
# dnf update --offline ./patch-2.7.6-24.fc40.x86_64.rpm
Updating and loading repositories:
Repositories loaded.
Package                 Arch   Version                  Repository          Size
Upgrading:
 patch                  x86_64 2.7.6-24.fc40            @commandline   262.8 KiB
   replacing patch      x86_64 2.7.6-23.fc40            @commandline   262.8 KiB

Transaction Summary:
 Upgrading:         1 packages
 Replacing:         1 packages

Total size of inbound packages is 131 KiB. Need to download 0 B.
After this operation 0 B will be used (install 263 KiB, remove 263 KiB).
Is this ok [y/N]: y

Failed to access RPM "/usr/lib/sysimage/libdnf5/offline/./packages/patch-2.7.6-24.fc40.x86_64.rpm": No such file or directory
```

Not using pre-downloaded rpm file works.

mcrha:
I forgot to add, this is with dnf5-5.2.4.0-20240715005730.21.g9c2e2150.fc41.x86_64

ppisar:
I confirm this issue. It behaves like that for relative path, absolute paths, and for file:// schema. Also "dnf install --offline" is affected.

I guess this caused by an optimization for not copying local packages into a package cache.

PR #1586, opened 2024-07-16
Make offline transactions work with local rpm files

m-blaha:
Offline transactions now copy also local (command line repo) packages to the
cache.

Changes:

f30fd653 (Marek Blaha, 2 hours ago)
   dnf5: Offline transactions work with local rpm files

9ab4e873 (Marek Blaha, 2 hours ago)
   package_downloader: Handle local files

   Local packages are not passed to librepo for download, but directly copied
   to the destination directory.

74e01778 (Marek Blaha, 7 hours ago)
   transaction: Flag whether download local packages

   By default the transaction during the `download()` call does not copy local
   packages to the destination directory. These packages are used directly
   from their original location. For some use-cases (like offline transaction
   involving the command line repo packages) we need to copy them to the
   package cache.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1583

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -436,6 +436,7 @@ void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
         std::filesystem::create_directories(offline_datadir);
 
         base.get_config().get_destdir_option().set(offline_datadir / "packages");
+        transaction.set_download_local_pkgs(true);
     }
 
     transaction.download();
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -187,6 +187,12 @@ public:
     /// specified path.
     void store_comps(const std::filesystem::path & comps_path) const;
 
+    /// Set whether local packages should be copied to the destination directory during the download().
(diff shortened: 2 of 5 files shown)


kontura:
I have created a couple of tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1529

kontura:
There are some failing CI tests.

I think the problem is with normal repo packages that are already downloaded in cache (using `--downloadonly` option).

They don't have the repo type `COMMANDLINE` so they are always set to be downloaded, in the `PackageDownloader::download` they are available locally so they are selected for copy but their source (`pkg_target.package.get_location()`) comes from the repo data - for example: `Packages/h/hwloc-libs-2.10.0-3.fc40.x86_64.rpm` and this fails to copy because the packages is not present at the path.

m-blaha:
Thanks! The `get_location()` was wrong, `get_package_path()` is needed to cover also the cached packages.
I added a commit that fixes it and also executes download callbacks to inform the user about the package being copied.

kontura:
Thanks :+1:

PR #1588, opened 2024-07-17
dnfdaemon: system-upgrade API and command

m-blaha:
Changes:

c5831738 (Marek Blaha, 72 minutes ago)
   doc: Add system-upgrade example using D-Bus API

960882be (Marek Blaha, 6 days ago)
   dnf5daemon-client: system-upgrade command

fe4b380f (Marek Blaha, 7 weeks ago)
   dnfdaemon: Add system_upgrade() method to Rpm interface

   The method fills the goal to perform upgrade of the distribution to the new
   release.

### DIFF
--- a/dnf5daemon-client/commands/system-upgrade/system-upgrade.cpp
+++ b/dnf5daemon-client/commands/system-upgrade/system-upgrade.cpp
@@ -1,0 +1,81 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "system-upgrade.hpp"
+
+#include "commands/shared_options.hpp"
+#include "context.hpp"
+#include "exception.hpp"
+#include "utils/auth.hpp"
+
+#include <dnf5daemon-server/dbus.hpp>
+#include <libdnf5/conf/option_string.hpp>
+
+namespace dnfdaemon::client {
+
+using namespace libdnf5::cli;
+
+void SystemUpgradeCommand::set_parent_command() {
+    auto * arg_parser_parent_cmd = get_session().get_argument_parser().get_root_command();
+    auto * arg_parser_this_cmd = get_argument_parser_command();
+    arg_parser_parent_cmd->register_command(arg_parser_this_cmd);
+    arg_parser_parent_cmd->get_group("software_management_commands").register_argument(arg_parser_this_cmd);
+}
+
+void SystemUpgradeCommand::set_argument_parser() {
+    auto & parser = get_context().get_argument_parser();
+    auto & cmd = *get_argument_parser_command();
+
+    cmd.set_description("prepare system for upgrade to a new release");
+
+    auto no_downgrade = parser.add_new_named_arg("no_downgrade");
+    no_downgrade->set_long_name("no-downgrade");
+    no_downgrade->set_description(
+        "Do not install packages from the new release if they are older than what is currently installed");
+    no_downgrade->set_has_value(true);
+    no_downgrade->set_arg_value_help("<yes|no>");
+    no_downgrade->link_value(&no_downgrade_option);
+    cmd.register_named_arg(no_downgrade);
+
+    // TODO(mblaha): set the releasever named arg as required (currently no API for this)
+}
+
+void SystemUpgradeCommand::run() {
+    auto & ctx = get_context();
+
+    if (!libdnf5::utils::am_i_root()) {
+        throw UnprivilegedUserError();
+    }
+
+    // TODO(mblaha): check the target releasever is set and different from the detected one
+
+    dnfdaemon::KeyValueMap options = {};
+    if (no_downgrade_option.get_value()) {
+        options["mode"] = "upgrade";
+    }
+
+    ctx.session_proxy->callMethod("system_upgrade")
+        .onInterface(dnfdaemon::INTERFACE_RPM)
+        .withTimeout(static_cast<uint64_t>(-1))
+        .withArguments(options);
+
+    run_transaction(true);
+}
+
+}  // namespace dnfdaemon::client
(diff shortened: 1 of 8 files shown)


dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -212,6 +212,23 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        @options: an array of key/value pairs to modify system_upgrade behavior
+
+        Prepare a transaction for upgrade to the new distribution release. The prepared transaction should be executed during reboot (see `offline` option of the `do_transaction` method of the `Goal` interface).
+
+        Following @options are supported:
+
+            - mode: string (one of "distrosync", "upgrade", default is "distrosync")
+                By default the system_upgrade behaves like `dnf distro-sync`, always installing packages from the new release, even if they are older then the currently installed version. If set to "upgrade", packages from the new release are not installed if they are older then what is currently installed (behave like `dnf upgrade`).
+
+        Unknown options are ignored.
```
mcrha, line 226: how to I set the release version to be upgraded to? It seems to be "detected" somehow, but one can upgrade both to N+1 and N+2. An explicit "upgrade to `releasever` X" would be better.

The python example suggests the `releasever` is set within the `open_session`. Could there be added an argument, which will override that value if set, here as well, please?

dnf5daemon-server/services/rpm/rpm.cpp
```diff
@@ -623,3 +633,46 @@ sdbus::MethodReply Rpm::remove(sdbus::MethodCall & call) {
...
+    libdnf5::GoalJobSettings settings;
+    if (upgrade_mode == "upgrade") {
+        goal.add_rpm_upgrade(settings);
+    } else if (upgrade_mode == "distrosync") {
+        goal.add_rpm_distro_sync(settings);
+    } else {
+        throw sdbus::Error(
+            dnfdaemon::ERROR,
+            fmt::format(
+                "Unsupported system-ugrade mode \"{}\". Only \"distrosync\" and \"upgrade\" modes are supported.",
```
mcrha, line 672: `s/system-ugrade/system-upgrade/` (missing `p` in the second word)

doc/dnf_daemon/examples/system_upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+# Libdnf is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+This is an example how to perform a system-upgrade with  dnf5daemon-server.
```
mcrha, line 20: are the two spaces near the end of the line intentional?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
```diff
@@ -212,6 +212,23 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+    <!--
+        system_upgrade:
+        @options: an array of key/value pairs to modify system_upgrade behavior
+
+        Prepare a transaction for upgrade to the new distribution release. The prepared transaction should be executed during reboot (see `offline` option of the `do_transaction` method of the `Goal` interface).
+
+        Following @options are supported:
+
+            - mode: string (one of "distrosync", "upgrade", default is "distrosync")
+                By default the system_upgrade behaves like `dnf distro-sync`, always installing packages from the new release, even if they are older then the currently installed version. If set to "upgrade", packages from the new release are not installed if they are older then what is currently installed (behave like `dnf upgrade`).
```
mcrha, line 224: Are the similarities with those modes the `dnf5 distro-sync` and `dnf5 system-upgrade`? I see some differences between the two, like the first can run "online" or "offline" while the other is always "offline". The first has `--download only`, the second is always split into two phases, `download` and `reboot`.

For the gnome-software it'll be always `download` and then `reboot-to-install`. It should be possible to change my mind and not install previously prepared system upgrade.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
m-blaha: Releasever should be set using the `open_session()` option. It needs to be set before the repositories metadata are loaded and setting it here can be too late. I'll add this information to the method description.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
m-blaha: It's more about what method `system-upgrade` uses in background. By default the `dnf5 system-upgrade download` is roughly equivalent to `dnf5 distro-sync --offline`. But with `dnf5 system-upgrade download --no-downgrade` option the system upgrade makes sure, that packages from the new release that are of older version then currently installed are not used. In this case its roughly equivalent to `dnf5 upgrade --offline`.

I will add code to enforce "offline" behavior to "Rpm.system_upgrade()" call, because currently it relies on `offline` option being passed to `Goal.do_transaction()` method.

kontura (commented):
This looks good I am only thinking how we could CI test it.
Perhaps some special step?

dnf5daemon-client/commands/system-upgrade/system-upgrade.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "commands/shared_options.hpp"
+#include "context.hpp"
+#include "exception.hpp"
+#include "utils/auth.hpp"
+
+#include <dnf5daemon-server/dbus.hpp>
+#include <libdnf5/conf/option_string.hpp>
+
+#include <memory>
```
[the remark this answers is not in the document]
m-blaha: Thanks, removed.

doc/dnf_daemon/examples/system_upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+This is an example how to perform a system-upgrade with dnf5daemon-server.
+"""
+
+import dbus
+import os
```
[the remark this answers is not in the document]
m-blaha: Thanks, removed.

mcrha:
I'm trying it here, and I noticed one thing:
```
$ cat /system-update/offline-transaction-state.toml 

[offline-transaction-state]
module_platform_id = ""
poweroff_after = false
verb = ""
system_releasever = "39"
target_releasever = "40"
cachedir = "/var/cache/libdnf5"
status = "download-incomplete"
cmd_line = "dnf5daemon-server"
state_version = 1

```

The `status = "download-incomplete"` is scary, because this is the "magic directory", thus restarting the machine will cause an upgrade attempt, with incomplete download. I'd say, if anything like this fails (or just really anything), then this should fail, report the error to the caller and do not prepare the upgrade for the next reboot.

This is with the default options, thus the `distrosync` mode.

PR #1590, opened 2024-07-22
Implement a default user_agent string

lleyton:
This implements a default user_agent string as described in https://github.com/rpm-software-management/dnf5/blob/main/doc/dnf5.conf.5.rst.

I also moved the `OSRelease` class from the COPR plugin into a library util for reuse, as well as a few other functions in `conf/vars`.


kontura (changes requested):
This is great, thank you!

I have added just one note + can you fix the formatting issues detected by the Pre Commit Hooks? You can see the specific issues here in the [Details](https://github.com/rpm-software-management/dnf5/actions/runs/10038130164/job/27808486363?pr=1590).

libdnf5/conf/config_main.cpp
```diff
@@ -84,6 +87,15 @@ static int str_to_bytes(const std::string & str) {
...
 }
 
+static std::string get_user_agent() {
+    utils::OSRelease os_release;
+    return std::format("libdnf ({} {}; {}; {}.{})",
+        os_release.get_value("NAME"),
+        os_release.get_value("VERSION_ID"),
+        os_release.get_value("VARIANT_ID"),
+        utils::get_os(),
+        rpm::get_base_arch(utils::detect_arch()));
```
kontura, line 97: Old libdnf had a fallback if either `NAME`, `VERSION_ID`, `os` or `arch` were missing it set just `'libdnf'` user agent. Can we do that here as well to be compatible?

libdnf5/utils/os_release.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (initialized_)
+        return;
+
+    initialized_ = true;
+    std::ifstream infile(path);
+    if (!std::filesystem::exists(path))
+        return;
+
+    const std::regex r_no_quotes("^([A-Z_]+)=(\\w+)");
+    const std::regex r_quotes("^([A-Z_]+)=\"([\\w\\s]+)\"");
```
kontura, line 46: I noticed some of our (currently disabled) CI tests would fail because of this regex.
For example if there is a value such as: `VERSION_ID="8.1"` in the `/ets/os-release` this regex doesn't match it.

Looking at the os-release.5 man page it seems that in general it can contain anything apart from non-printable characters. Some keys are further restricted to: no spaces or other characters outside of 0–9, a–z, ".", "_" and "-".
In any case this regex is way too strict.

kontura (changes requested):
While working on some related functionality I noticed couple more issue.

kontura:
@lleyton any update here?

We require these changes also for https://github.com/rpm-software-management/ci-dnf-stack/pull/1501.

lleyton:
oh sorry! missed your replies, gonna push changes to address the issues in a bit

lleyton:
Should be good now!

kontura:
Great, thanks.

There is still couple formatting issues.
You can use the `pre-commit` to do the formatting automatically with `pre-commit run --all-files`.

kontura:
The failing CI for 41 and rawhide is not related.

lleyton:
formatting issues should be fixed

kontura (approved):

Issue #1192, opened 2024-01-22
Refresh PGP keys, e.g. when prolonging an expiration time


xsuchy:
When the package is signed by the GPG key, and the key is prolonged, then GPG does not recognize that the content of the GPG key was changed and refuses to update the package.

This is affecting Copr users and originally reported as https://github.com/fedora-copr/copr/issues/2894

Steps to reproduce:

1. Create repository, sign the package with GPG key expiring in 5 years (this is what we have in Copr, for testing purpose you can lower it to one day :) )
2. wait 5 years
3. Other projects and user usually generate and use new GPG key. In Copr, we instead prolong the key. E.g. see https://filipe.kiss.ink/renew-expired-gpg-key/
4. This DOES NOT change the gpg  id.
5. after the old expire date try to install the rpm

You will get:
```
  1. Certificiate 454724A7D1C452B2 invalid: certificate is not alive
      because: The primary key is not live
      because: Expired on 2022-09-02T17:42:01Z
  2. Key 454724A7D1C452B2 invalid: key is not alive
      because: The primary key is not live
      because: Expired on 2022-09-02T17:42:01Z
error: Verifying a signature using certificate 519B71E71D5251A03A517DF8454724A7D1C452B2 (praiskup_myvpn (None) <praiskup#myvpn@copr.fedorahosted.org>):
  1. Certificiate 454724A7D1C452B2 invalid: certificate is not alive
      because: The primary key is not live
      because: Expired on 2022-09-02T17:42:01Z
  2. Key 454724A7D1C452B2 invalid: key is not alive
      because: The primary key is not live
      because: Expired on 2022-09-02T17:42:01Z
Copr repo for myvpn owned by praiskup                                                   194  B/s | 998  B     00:05    
GPG key at https://download.copr.fedorainfracloud.org/results/praiskup/myvpn/pubkey.gpg (0xD1C452B2) is already installed
The GPG keys listed for the "Copr repo for myvpn owned by praiskup" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: myvpn-1.3-6.fc38.x86_64
 GPG Keys are configured as: https://download.copr.fedorainfracloud.org/results/praiskup/myvpn/pubkey.gpg
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: GPG check FAILED
```

The workaround is to:
```
[...]

PR #1592, opened 2024-07-24
expired-pgp-keys: New plugin for detecting expired PGP keys

jan-kolarik:
Based on the implementation from https://github.com/rpm-software-management/dnf-plugins-core/pull/533.

Workaround for: https://github.com/rpm-software-management/dnf5/issues/1192.

CI fix: https://github.com/rpm-software-management/ci-dnf-stack/pull/1625.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -59,7 +59,7 @@ namespace dnf5 {
 namespace {
 
 // The `KeyImportRepoCB` class implements callback only for importing repository key.
-class KeyImportRepoCB : public libdnf5::repo::RepoCallbacks {
+class KeyImportRepoCB : public libdnf5::repo::RepoCallbacks2_1 {
 public:
     explicit KeyImportRepoCB(libdnf5::ConfigMain & config) : config(&config) {}
 
@@ -85,6 +85,34 @@ public:
 
     void repokey_imported([[maybe_unused]] const libdnf5::rpm::KeyInfo & key_info) override {
         std::cerr << _("The key was successfully imported.") << std::endl;
+    }
+
+    bool repokey_remove(const libdnf5::rpm::KeyInfo & key_info, const libdnf5::Message & removal_info) override {
+        if (config->get_assumeno_option().get_value()) {
+            return false;
+        }
+
+        std::cerr << libdnf5::utils::sformat(
+                         _("The following PGP key (0x{}) is about to be removed:"), key_info.get_short_key_id())
+                  << std::endl;
+        std::cerr << libdnf5::utils::sformat(_(" Reason     : {}\n"), removal_info.format(true));
+        for (auto & user_id : key_info.get_user_ids()) {
+            std::cerr << libdnf5::utils::sformat(_(" UserID     : \"{}\"\n"), user_id);
+        }
+        std::cerr << libdnf5::utils::sformat(_(" From       : {}\n"), key_info.get_url());
+
+        std::cerr << std::endl << _("As a result, installing packages signed with this key will fail.") << std::endl;
+        std::cerr << _("It is recommended to remove the expired key to allow importing") << std::endl;
+        std::cerr << _("an updated key. This might leave already installed packages unverifiable.") << std::endl
+                  << std::endl;
+
+        std::cerr << _("The system will now proceed with removing the key.") << std::endl;
+
+        return libdnf5::cli::utils::userconfirm::userconfirm(*config);
+    }
+
+    void repokey_removed([[maybe_unused]] const libdnf5::rpm::KeyInfo & key_info) override {
+        std::cerr << _("The key was successfully removed.") << std::endl;
     }
 
 private:
--- a/include/libdnf5/plugin/iplugin.hpp
+++ b/include/libdnf5/plugin/iplugin.hpp
@@ -125,6 +125,18 @@ private:
     ImplPtr<Impl> p_impl;
 };
 
+/// @brief Extended plugin interface with additional hooks introduced in version 2.1 of the plugin API.
+class LIBDNF_PLUGIN_API IPlugin2_1 : public IPlugin {
+public:
+    explicit IPlugin2_1(IPluginData & data);
+    ~IPlugin2_1();
+
+    /// The goal resolved hook.
+    /// It is called right after the goal is resolved.
+    /// @param transaction Contains the transaction that was resolved.
+    virtual void goal_resolved(const libdnf5::base::Transaction & transaction);
+};
+
 }  // namespace libdnf5::plugin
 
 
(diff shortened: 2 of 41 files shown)


jrohel:
I looked into the source code. It's a rewrite of the new "expired-gpg-keys" plugin from DNF4 to libdnf5.

And here's the problem. The original plugin is in Python. It is not part of the libdnf library -> this makes it ignored by users using libdnf directly - `microdnf`, `PackageKit`, ... And it behaves more like an application plugin - it interacts with the user via the console.

This PR implements the `libdnf5` plugin. So it is a library plugin. The advantage is that it will be active for all users of the `libdnf5` library (`dnf5` application, `dnf5daemon-server`, ...). However, the library must not interfere with the console. For example, `dnf5daemon-server` does not have a console to communicate with the user - it is a background daemon. 

The `dnf5` application only supports "command" plugins. And generally making this plugin as an application plugin is not a good way to go. The same problem would occur as with DNF4. It would not work for other applications (users of the libdnf5 library).

How to get out of this?
We need an interface defined in the `libdnf5` library for this case. 
Queries whether to do key import are done via callbacks. Similarly, queries to remove them (with possible information why) should be handled via callbacks.

jan-kolarik:
> I looked into the source code. It's a rewrite of the new "expired-gpg-keys" plugin from DNF4 to libdnf5.
> 
> And here's the problem. The original plugin is in Python. It is not part of the libdnf library -> this makes it ignored by users using libdnf directly - `microdnf`, `PackageKit`, ... And it behaves more like an application plugin - it interacts with the user via the console.
> 
> This PR implements the `libdnf5` plugin. So it is a library plugin. The advantage is that it will be active for all users of the `libdnf5` library (`dnf5` application, `dnf5daemon-server`, ...). However, the library must not interfere with the console. For example, `dnf5daemon-server` does not have a console to communicate with the user - it is a background daemon.
> 
> The `dnf5` application only supports "command" plugins. And generally making this plugin as an application plugin is not a good way to go. The same problem would occur as with DNF4. It would not work for other applications (users of the libdnf5 library).
> 
> How to get out of this? We need an interface defined in the `libdnf5` library for this case. Queries whether to do key import are done via callbacks. Similarly, queries to remove them (with possible information why) should be handled via callbacks.

That's a good point Jarek, I wasn't thinking about non-CLI users here. I will rework the plugin as you've suggested, thanks for that!

pmatilai:
Looking at the expiriation date will miss what I think is the more common issue just now: all the obsolete SHA1 based signatures that will fail in various other ways. Sequoia generally defers this checking to verify rather than import time, but you might want to additionally call pgpPubKeyLint() on the key material to catch at least some of those issues too.

jan-kolarik:
I performed some manual testing, and everything appears to be working as expected. One potential minor issue is that when GPG keys are available remotely (e.g., Copr keys), users always see output about downloading the keys, like this:
```
[kolage@ecstatic_dewdney dnf5]# dnf upgrade libdnf
Repositories loaded.
 https://download.copr.fedorainfracloud. 100% |  13.9 KiB/s |   1.0 KiB |  00m00s
 https://download.copr.fedorainfracloud. 100% |  12.2 KiB/s |   1.0 KiB |  00m00s
...
```
This happens because the `FileDownloader` in the `RpmSignature` class uses the standard download callbacks from the `Base` class, and there’s currently no proper way to customize these callbacks for one-time use.

jan-kolarik:
This PR adds a new libdnf5 workflow hook at the `Resolved` stage, enabling custom logic to be added immediately after the goal is resolved. Related updates were also made to the `Actions` plugin to support this new behavior.

jan-kolarik:
**How to test:**

If the plugin is enabled and any COPR repository is enabled on your system, you can test it by faketiming the system date and disabling `sslverify`, then running a transaction. This should trigger a prompt to remove the PGP key:

```
faketime '+5years' dnf5 install nudoku --setopt=sslverify=False
```

CI tests will be delivered in a follow-up PR.

jan-kolarik:
The only failing CI test seems to be about updating the Actions CI test text expectation.

libdnf5-plugins/expired-pgp-keys/expired-pgp-keys.cpp
```diff
@@ -0,0 +1,236 @@
...
+    auto key_id = key.get_short_key_id();
+    while ((h = rpmdbNextIterator(mi)) != nullptr) {
+        if (headerGetAsString(h, RPMTAG_VERSION) == libdnf5::utils::string::tolower(key_id)) {
+            rpmtsAddEraseElement(ts, h, -1);
+            retval = rpmtsRun(ts, nullptr, RPMPROB_FILTER_NONE) == 0;
+            break;
+        }
+    }
+    rpmtsFree(ts);
+    rpmdbFreeIterator(mi);
```
jrohel, line 139: I'm not sure if this is mandatory, but at least in general it's a good idea to release the iterators first and then the transaction.

dnf5/context.cpp
```diff
@@ -87,6 +87,35 @@ class KeyImportRepoCB : public libdnf5::repo::RepoCallbacks {
         std::cerr << _("The key was successfully imported.") << std::endl;
     }
 
+    bool repokey_remove(const libdnf5::rpm::KeyInfo & key_info) override {
+        if (config->get_assumeno_option().get_value()) {
+            return false;
+        }
+
+        std::cerr << libdnf5::utils::sformat(
+                         _("The following PGP key (0x{}) has expired on {}:"),
```
jrohel, line 96: The `repokey_remove` callback is called before the key is removed and allows you to confirm or deny its removal.

The message "The following PGP key (0x{}) expired on {}:" may be wrong. We do not know who called the callback and why they want to remove the key. It could was the expired-pgp-keys plugin. But also some other plugin, which for example will check if the key has been revoked. Or whether the key is still on the company's list of allowed keys. The reason for removing the key could be different.

jrohel (changes requested):
Thank you. Nice work. I wrote comments what still needs to be resolved.

include/libdnf5/plugin/iplugin.hpp
```diff
@@ -125,6 +125,17 @@ class LIBDNF_PLUGIN_API IPlugin {
...
+/// @brief Extended plugin interface with additional hooks introduced in version 2.1 of the plugin API.
+class LIBDNF_PLUGIN_API IPlugin2_1 : public IPlugin {
+public:
+    explicit IPlugin2_1(IPluginData & data);
+
+    /// The goal resolved hook.
+    /// It is called right after the goal is resolved.
+    /// @param transaction Contains the transaction that was resolved.
+    virtual void goal_resolved(const libdnf5::base::Transaction & transaction);
+};
```
jrohel, line 137: Please add destructor with `= default` definition in cpp file.

jrohel (approved):

jrohel:
Thank you.

Issue #1598, opened 2024-07-27
Improve English: After this operation 4 MiB will be used


hboetes:
I'd like to propose the following improvement to the English used by dnf5:

Original: "After this operation 4 MiB will be used"
Improved: "After this operation, 4 MiB extra will be used" or "After this operation, 4 MiB less will be used"

ppisar:
For the shrink case we already have "4 MiB will be freed". Is it good enough? I will add the comma for both cases and change a wording for the enlargement case.

hboetes:
Yes, an even better solution.

PR #1600, opened 2024-07-29
Improve "After this operation" disk usage messages

ppisar:
Resolves: #1598


jan-kolarik (approved):

jan-kolarik:
Thanks for the text improvements!

Issue #1213, opened 2024-01-29
Document how to enable and disable repositories


lnicola:
E.g. `dnf5 repo disable foo` or `dnf5 config-manager --set-disabled foo`

lnicola:
Oh, is that `dnf5 config-manager setopt enabled=0 foo`?

lnicola:
Oh, is that `dnf5 config-manager setopt foo.enabled=0`?

ppisar:
The latter: `dnf5 config-manager setopt foo.enabled=false`.
We need to write a documentation for `dnf5 repolist` and `dnf5 config-manager` commands (part of issue #1011.

lnicola:
Wouldn't it be nice if `dnf5 repo disable foo` worked though?

ppisar:
I agree that `dnf5 repo disable foo` would be handy. Though I'm not sure whether the move of all configuration changes to `dnf5 config-manager` was intentional. I will check it up with other DNF5 developers.

ppisar:
According to more experienced DNF developers, the missing way for enabling/disabling repositories with `dnf5 repo` was intentional. They prefer a single way over multiple ways. Mainly to simplify documentation and supported interfaces.

lnicola:
Looks like this was added for `dnf5daemon` in https://github.com/rpm-software-management/dnf5/pull/1266. Maybe it would still be nice to have it in `dnf5` for symmetry? AFAICT, `dnf5daemon-client` generally exposes less functionality than `dnf5` itself.

jan-kolarik:
> Looks like this was added for `dnf5daemon` in #1266. It would be good to have it in `dnf5` for symmetry. AFAICT, `dnf5daemon-client` generally exposes less functionality than `dnf5` itself.

The way it is exposed in dnf5daemon is not aligned with dnf5, as the config-manager functionality is not exposed in dnf5daemon.

jrohel:
> Oh, is that dnf5 config-manager setopt foo.enabled=0?

Yes, this is mentioned in the examples at https://dnf5.readthedocs.io/en/latest/dnf5_plugins/config-manager.8.html :
```
dnf5 config-manager setopt repoid1.enabled=1 repoid2.enabled=0
    Sets override to enable repository identified by repoid1 and disable repository identified by repoid2.
```

kontura:
Since `config-manager` already is documented I have added only an example to the `repo` command man page that links to it.
It could help users looking for that functionality.
[...]

PR #1601, opened 2024-07-30
Add an example how to disable repo to `repo` command man page

kontura:
While this example isn't for the `repo` command it could be useful to users looking into the man page.

For: https://github.com/rpm-software-management/dnf5/issues/1213


jan-kolarik (approved):
Thanks!

Issue #1533, opened 2024-06-06
Excessive scriptlet output during the transaction


jan-kolarik:
Follow-up on https://github.com/rpm-software-management/dnf5/issues/522#issuecomment-2132404834.

I remember this was also mentioned in another (BZ?) ticket, but cannot find it now, therefore creating a new one.

jan-kolarik:
If it's easily manageable, I think it would be reasonable to show the output only during the transaction and collapse it once the scriptlet is done, possibly leaving the output in the logs.

PR #1606, opened 2024-08-02
Reduce the noise around running scriptlets

m-blaha:
Changes:

d1ffe89f (Marek Blaha, 9 minutes ago)
   dnf5: Reduce the noise around running scriptlets

   If the scriptlet finishes successfully, do not print the
   "Running scriptlet..." / "Stop scriptlet..." pair of messages in the 
   transaction progress bar.

   The "Running scriptlet" message is displayed during the scriptlet execution
   to inform the user that a time-consuming operation is in progress. If the
   scriptlet completes successfully, the message is removed. If it fails, the
   message remains, and an error/warning message and the "Stop scriptlet"
   message is printed to enable debugging.

b3163df9 (Marek Blaha, 12 minutes ago)
   cli: Method to drop the last progress bar message

   Useful to create "temporary" messages, will be used to reduce the
   "Running scriptlet" / "Stop scriptlet" rpm transaction noise.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1533

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -41,6 +41,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 #include <libdnf5/utils/fs/file.hpp>
 #include <libdnf5/utils/patterns.hpp>
+#include <rpm/rpmtypes.h>
 
 #include <algorithm>
 #include <cctype>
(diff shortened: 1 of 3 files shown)


ppisar:
I haven't run code yet, but it looks reasonable.

The only concern I have is flickering the line by periodical printing and erasing the message. Maybe we should think about rate limiting of the progress bar. But that's an idea for another pull request.

ppisar (approved):
I tested the code and it works for me.

m-blaha:
It should not flicker - the starting message for the scriptlet is printed and then erased only once. So it should be visible during execution of the scriptlet, and eventually erased after the scriptlet successfully finishes.

dnf5/context.cpp
```diff
@@ -871,10 +872,24 @@ void RpmTransCB::script_stop(
...
     libdnf5::rpm::TransactionCallbacks::ScriptType type,
-    [[maybe_unused]] uint64_t return_code) {
-    active_progress_bar->add_message(
-        libdnf5::cli::progressbar::MessageType::INFO,
-        fmt::format("Stop {} scriptlet: {}", script_type_to_string(type), to_full_nevra_string(nevra)));
+    uint64_t return_code) {
+    switch (return_code) {
+        case RPMRC_OK:
+            // remove the script_start message in case the script finished successfully
+            active_progress_bar->pop_message();
```
jan-kolarik, line 879: As @ppisar mentioned, I've also noticed the flickering. There is always a flash of an empty line after the messages is popped and before the next one is printed. I've tried to modify the implementation locally to always just switch on the flag that we have to rewrite the last message next time when adding a new one. It seems the flickering is gone then, but still some edge cases seem need to be covered, like when the message is the last one in the given bar. Let's defer it to a separate issue.

jan-kolarik (approved):

Conan-Kudo (approved):

PR #1615, opened 2024-08-07
spec: fix cmake focus_new arg

kontura:
Follow up for: https://github.com/rpm-software-management/dnf5/pull/1582


ppisar (approved):
Good.

Issue #1602, opened 2024-07-31
Fix docs for commands accepting specs


kontura:
`install`, `upgrade` and `remove` commands are documented to accept only packages (`<package-spec>`) but they work also with comps.

ppisar:
It's not only documentation. "dnf install --help" also reads:

~~~~
Arguments:                                     
  specs                                        List of package specs to install
~~~~

I think `package-spec` identifier should be kept reserved for packages and would should change documentation to `package-spec|group-spec` where appropriate. Otherwise, the grammar would accept `@group > 2` which is obviously unwanted.

Issue #1498, opened 2024-05-16
install does not satisfy missing dependencies of an installed package


t184256:
The `install` command does not do what it claims to do, specifically it doesn't bother satisfying missing dependencies when the package is already installed, but the dependencies are missing.

> The install command in DNF5 is used for installing packages. It makes sure that all given packages defined in package-spec arguments and their dependencies are installed on the system.

How to reproduce:
1. upgrade to rawhide
2. `dnf download glances`
3. `rpm -i --nodeps glances*.rpm`
4. `dnf install glances`

Observed behaviour:
```
Updating and loading repositories:
 Fedora rawhide - x86_64                100% | 239.6 KiB/s |   4.3 KiB |  00m00s
Repositories loaded.
Package "glances-3.4.0-5.fc40.noarch" is already installed.
```

dnf did not, in fact, make sure that "all given packages ... and their dependencies are installed on the system."

Expected behaviour:

dnf downloads and installs all missing direct and transitive dependencies of glances

t184256:
`dnf reinstall glances` reinstalls glances and satisfies the dependencies. But I don't want to reinstall (in my actual case, it's about a package that dnf rejects, but RPM can be forced to install), just satisfy the dependencies.

`dnf5 check` detects the problem, but it doesn't offer a way to satisfy the dependencies.

IMO, there are, broadly, at least two and a half sensible ways forward:

1. make install do what it claims to do and satisfy the dependencies, turning it into a dependency problem fixing tool it's advertised as
2. change install description to reflect the behaviour, e.g., by adding an "unless the package is installed", or moving the claim about satisfying dependencies to reinstall, which actually does that
2a. if it's not doing anything anyway on the grounds of having the package already installed, maybe also make it skip "Updating and loading repositories" phase while you're at it

j-mracek:
If I am not mistaken in past yum-3 did a check for system integrity in background, therefore it allowed to detect such a situation. The question whether we should implement such a mechanism in DNF5? The problem is about price (CPU) where this will be not for free and current implementation of check command is quite slow.

jan-kolarik:
The man pages were fixed, I am closing this.

PR #1621, opened 2024-08-12
doc: fix arguments for install, upgrade and remove

kontura:
For: https://github.com/rpm-software-management/dnf5/issues/1602
For: https://github.com/rpm-software-management/dnf5/issues/1498

Also adds some clarification to `group_package_types` docs.

### DIFF
--- a/doc/commands/install.8.rst
+++ b/doc/commands/install.8.rst
@@ -25,15 +25,25 @@
 Synopsis
 ========
 
-``dnf5 install [options] <package-spec>...``
+``dnf5 install [options] <package-spec>|@<group-spec>|@<environment-spec>...``
 
 
 Description
 ===========
 
-The ``install`` command in ``DNF5`` is used for installing packages. It makes sure that
-all given packages defined in ``package-spec`` arguments and their dependencies are installed
-on the system. Environments can be installed with ``@environment-id`` as ``<package-spec>``.
+The ``install`` command in ``DNF5`` is used for installing packages, groups or environments.
+
+When installing packages defined in ``package-spec`` arguments, ``DNF5`` ensures that the packages
+and their dependencies are installed on the system.
+If the specified packages are already installed, DNF5 does not check their dependencies again and
+simply verifies that the packages themselves are present.
+
+When installing groups defined in ``group-spec`` arguments, ``DNF5`` ensures that the groups and
+their packages are installed on the system. Installs only group packages matching configured package
+type. See :manpage:`dnf5-conf(5)`, :ref:`group_package_types <group_package_types_options-label>`.
+
+When installing environments defined in ``environment-spec`` arguments, ``DNF5`` ensures that the
+environments and their groups are installed on the system.
 
 
 Options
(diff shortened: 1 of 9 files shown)


doc/commands/install.8.rst
```diff
@@ -25,15 +25,23 @@
...
 Description
 ===========
 
-The ``install`` command in ``DNF5`` is used for installing packages. It makes sure that
-all given packages defined in ``package-spec`` arguments and their dependencies are installed
-on the system. Environments can be installed with ``@environment-id`` as ``<package-spec>``.
+The ``install`` command in ``DNF5`` is used for installing packages, groups or environments.
+
+When installing packages defined in ``package-spec`` arguments, ``DNF5`` ensures that the packages
+and their dependencies are installed on the system.
```
jan-kolarik, line 37: > ensures that the packages and their dependencies are installed on the system.

Hmm, it seems we are still promising to do something that we aren't actually doing, e.g. what's described in https://github.com/rpm-software-management/dnf5/issues/1498 ...

jan-kolarik: Consider adding something like: "If a specified package is already installed, DNF5 does not check its dependencies again and simply verifies that the package itself is present."

jan-kolarik: If the described situation also occurs for group and env specs, I’d suggest adding it as a general note under the first sentence of the description.

doc/misc/specs.7.rst
```diff
@@ -161,12 +161,12 @@ Specifying File Provides
...
 
 
-Groups
+Comps
 ======
 
-``<group-spec>`` allows one to select (environment) groups a particular operation should work
-on. It is a case insensitive string (supporting globbing characters) that is
-matched against a group's ID, canonical name and name translated into the
+``<group-spec>``|``<environment-spec>`` allows one to select groups|environments a particular
```
jan-kolarik, line 167: Perhaps mention that commands accepting multiple types of specs uses the `@` prefix?

doc/commands/install.8.rst
kontura: I tried to phrase it in such a way that specifies we check the dependencies only when actually installing the package so I think it should cover the mentioned issue.

However I can add the sentence to make it explicit.

kontura: > If the described situation also occurs for group and env specs, I’d suggest adding it as a general note under the first sentence of the description.

It doesn't, both groups and environments can actually be installed (synchronized) over and over again and each time the dependencies/content is processed.

I have added the note specific for packages.

doc/misc/specs.7.rst
kontura: Added.

doc/commands/install.8.rst
jan-kolarik: > However I can add the sentence to make it explicit.

Thanks, it's now more understandable and explicit.

jan-kolarik (approved):
Thanks a lot for improving the docs!

PR #1625, opened 2024-08-13
Update toml11-devel usage for 4.0.0

kontura:
This should fix the rawhide build.

### DIFF
--- a/dnf5/CMakeLists.txt
+++ b/dnf5/CMakeLists.txt
@@ -61,6 +61,11 @@ else()
   message (STATUS "Using fallback bash completion dir ${BASH_COMPLETION_COMPLETIONSDIR}")
 endif()
 
+find_package(toml11 REQUIRED)
+if (toml11_VERSION VERSION_LESS 4.0.0)
+    add_definitions(-DTOML11_COMPAT)
+endif()
+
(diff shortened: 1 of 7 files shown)


kontura:
I believe the failed `Pre Commit` formatting is wrong, I prefer the current format.

evan-goode:
Thanks. I will wait for this before releasing DNF 5.2.6.0.

> I believe the failed `Pre Commit` formatting is wrong, I prefer the current format.

Maybe the current format is better, but I think it's still better just to keep the formatter happy.

kontura:
To make the `f41` and `rawhide` builds work we need: https://src.fedoraproject.org/rpms/toml11/pull-request/5 and https://src.fedoraproject.org/rpms/toml11/pull-request/4.

Since the builds on those versions are failing anyway I guess we don't need to block this PR on that.

jan-kolarik:
> To make the `f41` and `rawhide` builds work we need: https://src.fedoraproject.org/rpms/toml11/pull-request/5 and https://src.fedoraproject.org/rpms/toml11/pull-request/4.
> 
> Since the builds on those versions are failing anyway I guess we don't need to block this PR on that.

Thanks for taking care of it!

Issue #1593, opened 2024-07-24
dnf5daemon: Get dependencies to be downloaded (sizes are good enough)


mcrha:
The gnome-software can show how many bytes the app will take to download, but it shows only the main package, not its dependencies. Consider app like `hedgewars`, which also requires `hedgewars-data` and possibly other.

On the command line:
```
# dnf install hedgewars
Dependencies resolved.
==============================================================================================================================================================
 Package                                    Architecture                    Version                                    Repository                        Size
==============================================================================================================================================================
Installing:
 hedgewars                                  x86_64                          1.0.2-7.fc39                               updates                          2.7 M
Installing dependencies:
 SDL2_mixer                                 x86_64                          2.6.3-2.fc39                               fedora                           159 k
 SDL2_net                                   x86_64                          2.2.0-3.fc39                               fedora                            19 k
 SDL2_ttf                                   x86_64                          2.20.2-2.fc39                              fedora                            43 k
 compat-lua-libs                            x86_64                          5.1.5-23.fc39                              fedora                           167 k
 dejavu-sans-fonts                          noarch                          2.37-20.fc39                               updates                          1.3 M
 hedgewars-data                             noarch                          1.0.2-7.fc39                               updates                          139 M
 physfs                                     x86_64                          3.0.2-12.fc39                              fedora                            84 k
 wqy-zenhei-fonts                           noarch                          0.9.46-30.fc39                             fedora                           7.6 M

Transaction Summary
==============================================================================================================================================================
Install  9 Packages

Total download size: 151 M
Installed size: 183 M
```

while gnome-software shows only 2.8MB for the Hedgewars app.

A naive solution could be to get the package dependencies and check what it'll return, at least for not-yet-installed packages (it won't cover updates, I guess), but I did not find a way to it. The `rpm.Rpm` interface allows "whatrequires", but not "requires", as I could otherwise:

```
   $ dnf repoquery --requires hedgewars
```

Nonetheless, I'm not interested in precise dependencies, the sizes (downloaded and installed) are good enough for my use case (and I do not think I could catch all the downloaded cases with the simple (and single) call to get the `requires`, because those requires can have their own requires to be installed and so on). A new package property `install_size_with_deps` and `download_size_with_deps`, both calculated on demand, only if needed (because it's an expensive operation), would do the trick too, but I understand it's too specific and single-purpose thing, thus not so great in the generic API the daemon (tries to/) does have.

[...]

PR #1626, opened 2024-08-13
daemon: New API to get download size of pkg including deps

m-blaha:
The new API simulates install operation on given pkg returning all dependencies that are going to be installed with their sizes.

Changes:

be3c8b6a (Marek Blaha, 49 seconds ago)
   dnfdaemon: Example of Rpm.check_install() usage

be9411ba (Marek Blaha, 4 minutes ago)
   dnfdaemon: API to get download size including dependencies

   This can be used to get the "cost" of installation of a package - how much
   data needs to be downloaded and how much disk space will be used.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1593

### DIFF
--- a/dnf5daemon-server/services/rpm/rpm.cpp
+++ b/dnf5daemon-server/services/rpm/rpm.cpp
@@ -22,6 +22,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "dbus.hpp"
 #include "package.hpp"
 
+#include <libdnf5/base/goal_elements.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/rpm/package_set.hpp>
 #include <sdbus-c++/sdbus-c++.h>
@@ -122,6 +123,16 @@ void Rpm::dbus_register() {
         {},
         [this](sdbus::MethodCall call) -> void {
             session.get_threads_manager().handle_method(*this, &Rpm::system_upgrade, call, session.session_locale);
+        });
+    dbus_object->registerMethod(
+        dnfdaemon::INTERFACE_RPM,
+        "check_install",
+        "sa{sv}",
+        {"pkg_spec", "options"},
+        "aa{sv}",
+        {"packages"},  // inbound packages with {reason, name, epoch, version, release, evr, arch, install size, download size}
+        [this](sdbus::MethodCall call) -> void {
+            session.get_threads_manager().handle_method(*this, &Rpm::check_install, call, session.session_locale);
         });
 
     dbus_object->registerSignal(
@@ -674,5 +685,46 @@ sdbus::MethodReply Rpm::system_upgrade(sdbus::MethodCall & call) {
     }
 
     auto reply = call.createReply();
+    return reply;
+}
+
+sdbus::MethodReply Rpm::check_install(sdbus::MethodCall & call) {
+    auto base = session.get_base();
+    session.fill_sack();
+
+    // read options from dbus call
+    std::string pkg_spec;
+    call >> pkg_spec;
+    dnfdaemon::KeyValueMap options;
+    call >> options;
+
+    dnfdaemon::KeyValueMapList out_packages;
+    libdnf5::Goal goal(*base);
+    libdnf5::GoalJobSettings settings;
+    goal.add_install(pkg_spec, settings);
+    auto transaction = goal.resolve();
+    if (transaction.get_problems() == libdnf5::GoalProblem::NO_PROBLEM) {
+        std::vector<std::string> pkg_attrs{
+            "name",
+            "epoch",
+            "version",
+            "release",
+            "arch",
+            "repo_id",
+            "download_size",
+            "install_size",
+            "evr",
+        };
+        for (auto & tspkg : transaction.get_transaction_packages()) {
+            if (transaction_item_action_is_inbound(tspkg.get_action())) {
+                auto pkgmap = package_to_map(tspkg.get_package(), pkg_attrs);
+                pkgmap["install_reason"] = transaction_item_reason_to_string(tspkg.get_reason());
+                out_packages.emplace_back(std::move(pkgmap));
+            }
+        }
+    }
+
+    auto reply = call.createReply();
+    reply << out_packages;
     return reply;
 }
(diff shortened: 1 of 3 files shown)


m-blaha:
This is only a mere POC. @mcrha, @jan-kolarik, do you consider this worth having?
The other solution would be to enhance `Goal.resolve()` return values to also contain packages sizes. Then the same result as `Rpm.check_install(pkg_spec)` can be achieved using two calls - `Rpm.install(pkg_spec)` and `Goal.resolve()`.

mcrha:
Having something like this would work for me. Being able to verify whether the install will work and how much it'll "cost" is a good idea.

Having there `pkg_attrs` as another option would be nice to have, thus one can ask what it wants to see.

The name `check_install` is fine, I'm only thinking, what about `simulate_install`? I thought there was some sort of `--simulate` or `--dry-run` or some such argument for the command line tool, but it looks like there is no such thing, thus I might see it with another software/tool.

By the way, it can sometimes fail with the GPG key not being installed/updated yet. Will that break the check/simulation completely, like the call returning failure/nothing?

The call can take a long time, due to dependency resolution, slow connection to update local caches and such. That can easily cause timeout on the D-Bus call. I set indefinite timeout in some cases in the gnome-software, thus not a big deal, I'm only mentioning it as a possible bottleneck for any archive reader.

mcrha:
Aha, it's what Marek meant, there is already a way to do this, it's only hidden with two D-Bus calls. I do not like code duplication, thus ideally share the code. While a dedicated function makes it clearer how to achieve that, that not-so-obvious-two-calls works for me too.

Looking on what Jan pointed to, that part does a lot more things.

My comment about customizable `pkg_attrs` still applies.

m-blaha:
I considered extending the `Rpm.list()` method with new pkg_attrs (like `download_size_with_deps`), but ultimately decided against it. These attributes are quite time-consuming to calculate, and if a user were to run `Rpm.list('*')` with them, it could take an exceptionally long time to complete.
But I will consider extending the `resolve` call.

mcrha:
I tried the `resolve` call on a `hedgewars` package and the `transaction` output variant does contain the sizes for the packages, as Honza pointed to above.

```
resolved with result:0 and transaction:'[
   ('Package', 'Install', 'User', @a{sv} {}, {'arch': <'x86_64'>, 'download_size': <uint64 2844164>, 'epoch': <'0'>,
    'evr': <'1.0.2-7.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'hedgewars-0:1.0.2-7.fc40.x86_64'>, 'id': <8073>,
    'install_size': <uint64 5784790>, 'name': <'hedgewars'>, 'reason': <'None'>, 'release': <'7.fc40'>, 'repo_id': <'updates'>,
    'version': <'1.0.2'>}),
    ('Package', 'Install', 'Dependency', {}, {'arch': <'noarch'>, 'download_size': <uint64 145588917>, 'epoch': <'0'>,
     'evr': <'1.0.2-7.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'hedgewars-data-0:1.0.2-7.fc40.noarch'>, 'id': <8074>,
     'install_size': <uint64 161683968>, 'name': <'hedgewars-data'>, 'reason': <'None'>, 'release': <'7.fc40'>,
     'repo_id': <'updates'>, 'version': <'1.0.2'>}),
    ('Package', 'Install', 'Dependency', {}, {'arch': <'x86_64'>, 'download_size': <uint64 20076>, 'epoch': <'0'>,
      'evr': <'2.2.0-5.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'SDL2_net-0:2.2.0-5.fc40.x86_64'>, 'id': <29954>,
      'install_size': <uint64 27818>, 'name': <'SDL2_net'>, 'reason': <'None'>, 'release': <'5.fc40'>, 'repo_id': <'fedora'>,
      'version': <'2.2.0'>}),
    ('Package', 'Install', 'Dependency', {}, {'arch': <'x86_64'>, 'download_size': <uint64 86028>, 'epoch': <'0'>,
      'evr': <'3.0.2-14.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'physfs-0:3.0.2-14.fc40.x86_64'>, 'id': <68521>,
      'install_size': <uint64 171340>, 'name': <'physfs'>, 'reason': <'None'>, 'release': <'14.fc40'>, 'repo_id': <'fedora'>,
      'version': <'3.0.2'>}),
    ('Package', 'Install', 'Dependency', {}, {'arch': <'noarch'>, 'download_size': <uint64 7946036>, 'epoch': <'0'>,
      'evr': <'0.9.46-31.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'wqy-zenhei-fonts-0:0.9.46-31.fc40.noarch'>, 'id': <102476>,
      'install_size': <uint64 17157417>, 'name': <'wqy-zenhei-fonts'>, 'reason': <'None'>, 'release': <'31.fc40'>,
      'repo_id': <'fedora'>, 'version': <'0.9.46'>}),
    ('Package', 'Install', 'Dependency', {}, {'arch': <'x86_64'>, 'download_size': <uint64 166484>, 'epoch': <'0'>,
      'evr': <'2.8.0-1.fc40'>, 'from_repo_id': <''>, 'full_nevra': <'SDL2_mixer-0:2.8.0-1.fc40.x86_64'>, 'id': <3859>,
      'install_size': <uint64 353616>, 'name': <'SDL2_mixer'>, 'reason': <'None'>, 'release': <'1.fc40'>, 'repo_id': <'updates'>,
      'version': <'2.8.0'>})]'
```

Thus, from my point of view, this change is not needed.

What would be good to have are two things:
1. extend the API to provide the `pkg_attrs` myself, thus I can read for the transaction only those bits I'm interested in; the default value can the attributes used right now, thus any current API users will not need changing anything;
2. be able to cancel the transaction, or better run it as "simulate", thus it's not saved for later when I only want to know how much the install would cost. The `resolve` function does have an `options` argument, thus adding there a boolean `simulate` or `dry-run` or whichever name any other common tools use, to indicate to the dnf5daemon to discard the result as soon as it's resolved, would be ideal, both because the daemon can do it straight away and for the client not needing to do two D-Bus calls.

mcrha:
From my previous comment, the `1.` is a feature request, not belonging here. The `2.` is covered by the https://github.com/rpm-software-management/dnf5/pull/1678 . I propose to close this, because the extra API is not needed (proved by being able to get to the values without this change too).

Issue #1355, opened 2024-03-28
docs: Describe the logic behind matching source and debug repository names


jan-kolarik:
This is a dnf5 equivalent of https://github.com/rpm-software-management/dnf/issues/2071.

PR #1627, opened 2024-08-14
doc: Naming of source and debug repos

pkratoch:
Resolves: https://github.com/rpm-software-management/dnf5/issues/1355

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -587,6 +587,15 @@ Repo Options
 
     Default: ``True``.
 
+
+Source and debuginfo repository names
+=====================================
+
+For a given repository with an identifier in the form "<ID>-rpms", its corresponding source repository is expected to have an identifier in the form "<ID>-source-rpms" and debuginfo repository an identifier in the form "<ID>-debug-rpms". Otherwise (if the repository identifier doesn't have the "-rpms" suffix), the source repository is expected to have an identifier in the form "<ID>-source" and debuginfo repository an identifier in the form "<ID>-debuginfo".
+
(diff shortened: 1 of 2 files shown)


doc/dnf5.conf.5.rst
```diff
@@ -1358,6 +1358,15 @@ Variables Directories
 ``Distribution Variables Directory``
     /usr/share/dnf5/vars.d/
 
+
+Source and debuginfo repository names
```
ppisar, line 1362: I'd like to see this piece of information to be close to where repository file format is documented. We don't have a dedicated page for it. But we have "Repo Options" and "Repo Variables" sections in this page. Could you move this section between the two sections?

doc/dnf5.conf.5.rst
```diff
@@ -1358,6 +1358,15 @@ Variables Directories
...
+=====================================
+
+For a given repository with an ID in the form "<ID>-rpms", its corresponding source repository is expected to have an ID in the form "<ID>-source-rpms" and debuginfo repository an ID in the form "<ID>-debug-rpms". Otherwise (if the repository doesn't have the suffix "-rpms"), the source repository is expected to have an ID in the form "<ID>-source" and debuginfo repository an ID in the form "<ID>-debuginfo".
```
ppisar, line 1365: I find it difficult to follow when an ID means a prefix and when it means the whole identifier. Could you spell all "ID"s which are not meant to be a replaceable text as "identifier"? Like "For a given repository with an identifier in the form "\<ID\>-rpms"".

ppisar, line 1365: Please change parenthesized note `if the repository doesn't have the suffix "-rpms"` into `if the repository identifier doesn't have the "-rpms" suffix`. I.e. add "identifier and move "-rpms" ahead because of "the" article.

ppisar (commented):
Otherwise, the text is good.

pkratoch:
Thanks for the review, should be fixed now.

pkratoch:
Also, there is the same documentation for dnf4: https://github.com/rpm-software-management/dnf/pull/2119

Issue #1620, opened 2024-08-08
dnf fails to install packages with `destdir` option set


pkratoch:
I belive the packages are downloaded to the given destdir, but then the installation fails, because it searches for the packages in the repo cache. See `libdnf5::rpm::package::get_package_path()`, but maybe the fix will be needed on more places.

Reproducer:
```
# dnf5 -y --setopt=destdir=newdestdir install nano
Updating and loading repositories:
Repositories loaded.
Package                          Arch      Version                          Repository             Size
Installing:
 nano                            x86_64    8.1-1.fc41                       rawhide             2.9 MiB

Transaction Summary:
 Installing:        1 packages

Total size of inbound packages is 729 KiB. Need to download 729 KiB.
After this operation 3 MiB will be used (install 3 MiB, remove 0 B).
[1/1] nano-0:8.1-1.fc41.x86_64                                 100% |   8.9 MiB/s | 728.9 KiB |  00m00s
-------------------------------------------------------------------------------------------------------
[1/1] Total                                                    100% |   4.9 MiB/s | 728.9 KiB |  00m00s
Running transaction

Transaction failed: Signature verification failed.
PGP check for package "nano-8.1-1.fc41.x86_64" (/var/cache/libdnf5/rawhide-2d95c80a1fa0a67d/packages/nano-8.1-1.fc41.x86_64.rpm) from repo "rawhide" has failed: Problem occurred when opening the package.
# echo $?
1
# ls newdestdir/
nano-8.1-1.fc41.x86_64.rpm
```

PR #1628, opened 2024-08-14
Use correct path when destdir option is set

m-blaha:
Changes:

1e83b073 (Marek Blaha, 9 minutes ago)
   repo: New Repo.get_packages_download_dir() method

   Currently, the path where a package should be downloaded is computed in 
   several places. This patch introduces a single location for this 
   computation to reduce code duplication. The patch also fixes usage of
   incorrect package path in Package::get_package_path() call - the destdir
   config option was not taken into account.

07105d0b (Marek Blaha, 2 hours ago)
   doc: Document destdir main config option

Resolves: https://github.com/rpm-software-management/dnf5/issues/1620
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1536

### DIFF
--- a/doc/dnf5.conf-todo.5.rst
+++ b/doc/dnf5.conf-todo.5.rst
@@ -189,10 +189,6 @@ This section does not track any deprecated option. For such options see :ref:`De
 
 ``history_list_view``
 
-.. _destdir_options-label:
-
-``destdir``
-
 .. _comment_options-label:
 
 ``comment``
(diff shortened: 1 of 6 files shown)


pkratoch:
Thanks for the patch! I noticed one test that looks related to the change broke ("skip downloading of already downloaded rpm file" in the dnf/download.feature).

m-blaha:
Yeah, I was lazy to run the tests locally and that's the result... I'll check it out, meanwhile switching to draft state.

m-blaha:
So the issue is not with this PR itself, but was introduced in commit 59538f3e which did not properly call download callbacks in case the source and destination of a downloaded file were the same. The problem did not surface before this PR, because the package was not correctly identified as locally available due to an incorrect path being used.

pkratoch:
Thank you!

Issue #1629, opened 2024-08-14
Document that "dnf5 install FILE_NAME" searches /usr/{s,}bin/FILE_NAME in file provides


ppisar:
"dnf5 install FOO" has a feature that if FOO is not a package name, it also searches for /usr/bin/FOO and /usr/sbin/FOO.

Try "dnf5 install cat", resulting to coreutils, or read PackageQuery::resolve_pkg_spec() in  libdnf5/rpm/package_query.cpp:

~~~~
    // If spec is file path than it is not a binary
    if (settings.get_with_binaries() and !is_file_pattern) {
        ReldepList reldep_list(p_impl->base);
        std::array<std::string, 2> binary_paths{"/usr/bin/", "/usr/sbin/"};
~~~~

I cannot find this feature nowhere documented. dnf5-install(8) only links to dnf5-specs(7). dnf5-specs(7) only documents that the search is also performed in file provides. But it does not mention that file name-like arguments undergo the /usr/{s,}bin/ expansion.

Contrary, "dnf5 repoquery --whatprovides" does not perform the path expansion. It searches for exact file provides and for glob-expanded file provides. But not for path-expanded values. (I and [someone else](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/DYREMWS66P4726VFSGHADGJ4ZT3F36NS/) find this asymmetry  counterintuitive.)

I'm not sure where to document it:

* I would prefer in dnf5-install(8) rather than in dnf5-specs(8) as it only works with "dnf5 install".
* Another option is in dnf5-specs(8) as a special exception for package installation.
* Another option is change dnf5-install(8) synopsis from `dnf5 install [options] <package-spec>...` to `dnf5 install [options] <install-spec>...` and then add a new section about install-spec to dnf5-specs(8) page.

(A tangent problem is that we should use $PATH environment variable instead of the two hard-coded directories "/usr/bin/", "/usr/sbin/".)

PR #1630, opened 2024-08-15
doc: specs.7: document looking for file provides in /usr/[s]bin

marcusmueller:
As discussed on devel@lists.fedoraproject.org, https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/Z2NFCBGNT2VLIDTVKS5HL5QJZ72UGHEG/.

In short, I was wondering how `dnf repoquery --whatprovides gnuradio-companion` turned up empty, and still, `dnf install gnuradio-companion` installed the `gnuradio` package. Petr Pisar was nice enough to explain that to me, so here's a PR :)


ppisar:
Thanks for the patch. But this issue is not so simple because the path lookup is applied only for package installation. I've already opened #1629 to track the problem and the DNF team will discuss the best place for the this piece of information soon.

PR #1631, opened 2024-08-15
feat: set up function to check for duplicate executables in PATH

solomoncyj:


m-blaha:
Thanks for a contribution!
Would you mind elaborating on the use-case you are addressing? Once the duplicities are identified, it would be helpful to understand how you propose dnf5 (as a package manager) should handle them.

dnf5/commands/check/check.cpp
```diff
@@ -110,6 +112,62 @@ cli::ArgumentParser::NamedArg * create_option(
...
+    // parse PATH variable
+    for (const auto path : PATH | std::views::split(':')) {
+        PATH_dir.insert({path.data(), {}});
+    }
+
+    // populate file list
+    for (const auto& [dir, _] : PATH_dir) {
+        for (const auto& entry : std::filesystem::directory_iterator(dir)) {
+            std::string path = entry.path().string();
+            std::size_t found = path.find_last_of("/\\");
```
ppisar, line 129: The backslash is wrong here as POSIX requires a forward slash. That you assume POSIX is visible few lines above when segmenting PATH environment variable by colons.

ppisar:
I remember somebody asking for a check that after uninstalling a package, DNF should check that none of the uninstalled command files is present on other PATH directories. The use case was that a user installs a program to /usr/local/bin or ~/bin not with an RPM manager, then uninstalls the same-named command file with an RPM manager from /usr/bin and is surprised that the command is still available. At that time I explained to the user that PATH which DNF process sees can differ from PATH which sees a user and thus we concluded that it does not make sense to add this check to DNF.

Reading the code, the check is similar, though not bound to a package removal.

While I understand that system administrators would like to have a tool like that, the problem is completely independent of DNF (this code does not work RPM at all), and thus in my opinion it should be a separate tool, not maintained by DNF. I wouldn't be surprises if various security and policy compliance scanners, like OpenSCAP, already implemented this check.

solomoncyj:
> Thanks for a contribution! Would you mind elaborating on the use-case you are addressing? Once the duplicities are identified, it would be helpful to understand how you propose dnf5 (as a package manager) should handle them.

when a dup is found, you can use dnf to remove the offending package, or tru to remember which manual way you used to uninstall it

solomoncyj:
ok.. now stuck on generating the pkgid and nevera... how do i generate them?

m-blaha:
> ok.. now stuck on grabbing the pkgid and nevera... how do i generate them?

These are attributes of the package object.

While I appreciate the effort you've put into this, I have to agree with @ppisar's assessment. It seems that this task may not be the best fit for dnf. As Petr mentioned, the varying `$PATH` content for different users and the fact that duplicates are usually not caused by another packages (such as a user creating a symlink in `~/bin/`) make it challenging to address this issue effectively within the package manager.

PR #1639, opened 2024-08-20
spec: fix modularity man page

kontura:
This should fix the currently failing builds.

For f41 and rawhide they will still be failing because of the toml11 break.


m-blaha (approved):
Thank you!

Issue #1538, opened 2024-06-07
Bidirectional communication of libdnf5 actions plugin with running processes


jrohel:
Why?
It will provide more flexibility for sub-plugins created as processes running with the actions plugin. These sub-plugins (processes) can be written in any programming language. They are more isolated from libdnf5. They can run with their own permissions.

Proposal:
- Running processes will be able to communicate with the actions plugin using standard output and input.
- Request-Response style communication.
- Process sends query/command to standard output, actions plugin returns response/acknowledgement to standard input of process.
- Running processes will be able to read/modify main configuration, repository configuration, libdnf5 variables, plugin support variables. They will also be able to query the packages in the transaction and write to the log.

A communication protocol needs to be designed and implemented.

PR #1642, opened 2024-08-21
Birectional communication of libdnf5 actions plugin with running processes - "json" mode

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1538

Introduces a "mode" option in the hook handler definition.
Option "mode" defines communication model between actions plugin and process.
Introduces 2 modes of communication: "plain" and "json".

### "mode=plain" - Original and default communication model.

Parameters are passed to the process only by arguments when it starts. Arguments may contain variables that are substituted before the process is started. A process can modify libdnf5 settings (main configuration, repository configuration, variables) and plugin actions variables by writing to standard
output.

**Output line format:**
- `tmp.<actions_plugin_variable_name>=<value>` - sets the value of action plugins variable `<actions_plugin_variable_name>`
-` tmp.<actions_plugin_variable_name>` - removes the action plugins variable if it exists
- `conf.<option_name>=<value>` - sets the value of option <option_name> in the base configuration
- `conf.<repoid_pattern>.<option_name>=<value>` - sets the value of option <option_name> in the matching repositories
- `var.<variable_name>=<value>` - sets value of the variable <variable_name>

### "mode=json" - Newly introduced communication model.

Parameters can be passed to the process before it starts using arguments just like in "plain" mode. This mode adds bidirectional communication between the actions plugin and the process. This is request/response communication.
The process sends requests and the actions plugin sends responses. The process writes requests to the standard output and reads responses from the standard input.

**Processes send requests and the actions plugin responds:**

- Request format:
    `{"op":"<operation>", "domain":"<domain>", "args":{<op_domain_spec_args>}}`

"domain" is ignored in the case of the "log" operation.

- Response format:
    `{"op": "reply", "requested_op":"<op_from_request>", "domain": "<domain_from_request>", "status": "OK", "return":{<data>}}`

    `{"op": "reply", "requested_op":"<op_from_request>", "domain": "<domain_from_request>", "status": "ERROR", "message":"<error message>"`}

"return" is not included in the response to a request for a "log"
operation.

If the JSON request is unparsable - a syntax error or incomplete message, or the response cannot be written, the action plugin closes both communication pipes. The process must terminate if the communication pipes are closed.

### DIFF
--- a/libdnf5-plugins/actions/actions.cpp
+++ b/libdnf5-plugins/actions/actions.cpp
@@ -18,11 +18,13 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 */
 
 #include <fmt/format.h>
+#include <json.h>
 #include <libdnf5/base/base.hpp>
 #include <libdnf5/base/transaction.hpp>
 #include <libdnf5/common/exception.hpp>
 #include <libdnf5/common/sack/match_string.hpp>
 #include <libdnf5/plugin/iplugin.hpp>
+#include <libdnf5/repo/repo_errors.hpp>
 #include <libdnf5/repo/repo_query.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
@@ -45,7 +47,7 @@ using namespace libdnf5;
 namespace {
 
 constexpr const char * PLUGIN_NAME = "actions";
-constexpr plugin::Version PLUGIN_VERSION{1, 1, 1};
+constexpr plugin::Version PLUGIN_VERSION{1, 2, 0};
 
 constexpr const char * attrs[]{"author.name", "author.email", "description", nullptr};
 constexpr const char * attrs_value[]{"Jaroslav Rohel", "jrohel@redhat.com", "Actions Plugin."};
@@ -103,11 +105,15 @@ public:
 
     void repos_loaded() override { on_hook(repos_loaded_actions); }
 
-    void pre_add_cmdline_packages(const std::vector<std::string> &) override {
+    void pre_add_cmdline_packages(const std::vector<std::string> & paths) override {
+        this->cmdline_packages_paths = &paths;
         on_hook(pre_add_cmdline_packages_actions);
     }
 
-    void post_add_cmdline_packages() override { on_hook(post_add_cmdline_packages_actions); }
+    void post_add_cmdline_packages() override {
+        this->cmdline_packages_paths = nullptr;
+        on_hook(post_add_cmdline_packages_actions);
+    }
 
     void pre_transaction(const libdnf5::base::Transaction & transaction) override {
         on_transaction(transaction, pre_trans_actions);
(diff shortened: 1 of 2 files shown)


libdnf5-plugins/actions/actions.cpp
```diff
@@ -844,7 +854,777 @@ void Actions::process_command_output_line(std::string_view line) {
...
+
+                libdnf5::sack::ExcludeFlags query_flags{0};
+                auto * jparams = get_any_object_or_null(jargs, "params");
+                if (jparams) {
+                    get_array(jargs, "params");  // check, must be array
+                    for (size_t idx = 0; idx < json_object_array_length(jparams); ++idx) {
+                        auto * jparam = json_object_array_get_idx(jparams, idx);
+                        const auto param_key = get_string_view(jparam, "key");
+                        //auto * jvalue = get_any_object_or_null(jparam, "value");
+                        //const auto value = jvalue ? std::string(get_string_view(jvalue)) : "";
```
kontura, line 1319: These can be removed right?

libdnf5-plugins/actions/actions.cpp
```diff
@@ -844,7 +854,777 @@ void Actions::process_command_output_line(std::string_view line) {
...
+                        }
+                    }
+                }
+
+                auto * jfilters = get_any_object_or_null(jargs, "filters");
+                if (jfilters) {
+                    get_array(jargs, "filters");  // check, must be array
+                }
+
+                enum class Direction { IN, OUT, ALL } edirection = Direction::ALL;
```
kontura, line 1341: Why isn't the `Action::Direction` used here?

libdnf5-plugins/actions/actions.cpp
```diff
@@ -844,7 +854,777 @@ void Actions::process_command_output_line(std::string_view line) {
...
+                    for (size_t idx = 0; idx < json_object_array_length(jfilters); ++idx) {
+                        auto * jfilter = json_object_array_get_idx(jfilters, idx);
+                        const auto filter_key = get_string_view(jfilter, "key");
+                        auto * jvalue = get_any_object_or_null(jfilter, "value");
+                        const auto value = jvalue ? std::string(get_string_view(jvalue)) : "";
+                        auto * joperator = get_any_object_or_null(jfilter, "operator");
+                        const auto oper = joperator ? cmp_operator_from_string(get_string_view(joperator))
+                                                    : libdnf5::sack::QueryCmp::EQ;
+                        if (filter_key == "direction") {
+                        } else if (filter_key == "name") {
```
kontura, line 1381: Minor suggestion but perhaps we could add a comment here saying that the `direction` filtering was already done above.

kontura:
I believe a docs update is needed and I assume it will be done in a follow up PR.

libdnf5-plugins/actions/actions.cpp
jrohel: Yes. I removed it.

libdnf5-plugins/actions/actions.cpp
jrohel: Thank you. Good point. This is a good candidate for using the `Action::Direction` type. I somehow missed it when I reorganized the code.

libdnf5-plugins/actions/actions.cpp
jrohel: I added comments.

PR #1645, opened 2024-08-21
Show the output of failed scriptlets to the user

m-blaha:
Currently, the output of RPM scriptlets is kind of buried in the log. This patch introduces a new `script_output` callback, which is executed just before `script_end` or `script_error`. While it would be more intuitive to include the scriptlet output directly as a new parameter in `script_end` and `script_error`, I opted to add a new callback to maintain compatibility.

The scriptlet output is only printed in the event of a failure; it is not shown when the execution is successful.

Resolves: https://github.com/rpm-software-management/dnf5/issues/522

Changes:

eaf5dd98 (Marek Blaha, 39 minutes ago)
   dnf5: Handle script_output callback for transaction

0d6e2cff (Marek Blaha, 41 minutes ago)
   transaction: New script_output callback

   Used to pass the outputs (stdout and stderr) of an executed scriptlet to 
   the user.

b915955e (Marek Blaha, 26 hours ago)
   transaction: Move scriptlet outputs to rpm::Transaction

   The code works on rpm level anyway and I'd like to use this mechanism in 
   transaction callback to present the outputs to the user. Also adds
   rpm::Transaction class member to store the output of the last scriptlet.

### DIFF
--- a/include/libdnf5/rpm/transaction_callbacks.hpp
+++ b/include/libdnf5/rpm/transaction_callbacks.hpp
@@ -92,6 +92,14 @@ public:
     virtual void script_error(const TransactionItem * item, Nevra nevra, ScriptType type, uint64_t return_code);
     virtual void script_start(const TransactionItem * item, Nevra nevra, ScriptType type);
     virtual void script_stop(const TransactionItem * item, Nevra nevra, ScriptType type, uint64_t return_code);
+    /// Called after the scriptlet finished
+    /// @param item Transaction item
+    /// @param nevra NEVRA of the package the scriptlet belongs to
+    /// @param type Type of the scriptlet
+    /// @param return_code Return code of the scriptlet: OK, non-critical error, error
+    /// @param output Output produced by the scriptlet
+    virtual void script_output(
+        const TransactionItem * item, Nevra nevra, ScriptType type, uint64_t return_code, const std::string output);
     virtual void elem_progress(const TransactionItem & item, uint64_t amount, uint64_t total);
     virtual void verify_progress(uint64_t amount, uint64_t total);
     virtual void verify_start(uint64_t total);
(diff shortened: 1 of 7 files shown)


m-blaha:
@jrohel @jan-kolarik may I ask you for your opinion?

include/libdnf5/rpm/transaction_callbacks.hpp
```diff
@@ -92,6 +92,14 @@ class LIBDNF_API TransactionCallbacks {
     virtual void script_error(const TransactionItem * item, Nevra nevra, ScriptType type, uint64_t return_code);
     virtual void script_start(const TransactionItem * item, Nevra nevra, ScriptType type);
     virtual void script_stop(const TransactionItem * item, Nevra nevra, ScriptType type, uint64_t return_code);
+    /// Called after the scriptlet finished
+    /// @param item Transaction item
+    /// @param nevra NEVRA of the package the scriptlet belongs to
+    /// @param type Type of the scriptlet
+    /// @param return_code Return code of the scriptlet: OK, non-critical error, error
+    /// @param output Output produced by the scriptlet
+    virtual void script_output(
```
jrohel, line 101: PR inserts a new virtual method into the public callback class. This does not break the API.

However, adding a virtual function will definitely change the `vtable` and how it changes depends on the compiler.
**The result will be a broken ABI. And in a very bad way.** An application using a library with an added or missing virtual callback method will run and then crash during the callback function call.

If we want to keep ABI backward compatibility, we need to do it differently.

**Example solution:**
Instead of adding a new callback `script_output` (a virtual method), add a non-virtual API method `libdnf5::base::Transaction::get_last_scriptlet_output` that the library user will be able to call while servicing the existing callback `script_stop` .
This solution came to my mind first. Maybe there is a better place for a new non-virtual method. This is an example.

jrohel (changes requested):
PR breaks ABI. And in a very bad way.
I described it in the review with an example solution.

Issue #727, opened 2023-07-17
Introduce `gpgcheck_policy` to control `gpgcheck` defaults and add support for `pkg_gpgcheck`


Conan-Kudo:
### Description of request

Currently, DNF supports two flags: `gpgcheck` and `repo_gpgcheck`. However, it is not clear what `gpgcheck` actually covers. In DNF, `gpgcheck` only covers package signature checking, but in Zypper, it covers both package and repository metadata signature checking.

Zypper has three options:
1. `gpgcheck`, covering repo and packages
2. `pkg_gpgcheck`, covering packages only (equivalent to current gpgcheck in Yum/DNF)
3. `repo_gpgcheck`, covering repo only (DNF supports this the same way Zypper does)

Each of these can be set globally or per-repository. By default, Zypper does both repository and package checks, and the other two options can be used to override bits of that behavior in Zypper.

To control the behavior of `gpgcheck`, we should have a global config option called `gpgcheck_policy`, with the following options: "legacy", "full", and "all".

* `gpgcheck_policy=legacy` makes "gpgcheck=1" equivalent to setting `pkg_gpgcheck=1`, `repo_gpgcheck=0`, and `localpkg_gpgcheck=0`.

* `gpgcheck_policy=full` makes `gpgcheck=1` equivalent to setting `pkg_gpgcheck=1`, `repo_gpgcheck=1`, and `localpkg_gpgcheck=0`.

* `gpgcheck_policy=all` makes `gpgcheck=1` equivalent to setting `pkg_gpgcheck=1`, `repo_gpgcheck=1`, and `localpkg_gpgcheck=1`.

Regardless of modes, each configuration section (`[main]` or repo sections in `dnf.conf` or in repo files) can override the policy behavior by setting `pkg_gpgcheck` and `repo_gpgcheck` accordingly. Obviously `localpkg_gpgcheck` has no bearing in repo sections, but could be switched on separately in the main section in `dnf.conf`.

The upstream default in the code (that is, when the `gpgcheck_policy` option is unset) would be the equivalent of setting `gpgcheck_policy=full`. However, for the `dnf.conf` that will ship in Fedora and RHEL, it would be set to `gpgcheck_policy=legacy` to conform with the existing behavior and not break anything.

This aligns our options and behaviors with other RPM package managers using rpm-md and makes the GPG checking policy coherent.

### Benefit to distributions

[...]

PR #1648, opened 2024-08-22
doc: configuration options update

kontura:
Regarding the `gpgcheck` option, we might want to match the dnf4 behavior described in https://bugzilla.redhat.com/show_bug.cgi?id=1614351, like https://github.com/rpm-software-management/dnf/pull/1293.
However for now  lets remove the incorrect info.


Conan-Kudo:
Could we add `pkg_gpgcheck` as the supported option and make `gpgcheck` an alias to that (for now)? That would let us make a step toward #727...

m-blaha (approved):
Thank you!

Issue #1622, opened 2024-08-12
--config short option


alcir:
In dnf4 we had
`-c, --config [config file]`
in dnf5 we only have
`--config=CONFIG_FILE_PATH              Configuration file location`

Adding the short version `-c` could prevent breakage of some scripts [^1]

[^1]: : i.e. I've spotted BitDefender unable to upgrade its agent on Fedora due to the use of `-c` option

PR #1650, opened 2024-08-23
[dnf5] Add argument "-c" - alias to "--config" (dnf4 compatibility)

jrohel:
We prefer the "--config" argument, for which we support bash completion. The short version "-c" is added to prevent problems with old scripts using dnf4 that may use "-c".

Closes: https://github.com/rpm-software-management/dnf5/issues/1622


jan-kolarik (approved):

jan-kolarik:
Thanks!

PR #1652, opened 2024-08-26
Show the output of failed scriptlets to the user

m-blaha:
Currently, the output of RPM scriptlets is somewhat hidden in the log. This patch introduces a new `libdnf5::base::Transaction::get_last_script_output()` method, which returns the outputs of the most recently executed RPM scriptlet. This method is then utilized in the transaction callbacks in DNF5 to display the output to the user.

Although it would be more intuitive to add the scriptlet output as a new parameter in script_end and script_error, doing so would break the existing API.

The scriptlet output is only printed in the event of a failure; it is not displayed when the execution is successful.

Resolves: https://github.com/rpm-software-management/dnf5/issues/522
Replaces: https://github.com/rpm-software-management/dnf5/pull/1645

Changes:

57715716 (Marek Blaha, 5 minutes ago)
   dnf5: Print rpm scriptlets outputs to the user

   If a scriptlet fails, its output is displayed to the user in addition to 
   being written to the log. For scriptlets that complete successfully, the 
   output is only logged.

27d322a8 (Marek Blaha, 9 minutes ago)
   transaction: Store outputs of the last rpm scriptlet

   The outputs of the most recently executed RPM scriptlet in the transaction
   are stored and can be accessed by the user through the 
   Base::Transaction::get_last_script_output() method.

55cf66fe (Marek Blaha, 21 minutes ago)
   rpm::transaction: Base::Transaction to callback holder

   Handling scriptlet outputs requires rpm transaction callbacks to have 
   access to the `libdnf5::Base::Transaction` instance the rpm transaction 
   belongs to.

### DIFF
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -193,12 +193,17 @@ public:
     void set_download_local_pkgs(bool value);
     bool get_download_local_pkgs() const noexcept;
 
+    /// Retrieve output of the last executed rpm scriptlet.
+    std::string get_last_script_output();
+
 private:
     friend class TransactionEnvironment;
     friend class TransactionGroup;
     friend class TransactionModule;
     friend class TransactionPackage;
     friend class libdnf5::Goal;
+    friend class libdnf5::rpm::
+        Transaction;  // to access clear_last_script_output() from the rpm transaction SCRIPT_START callback
 
     LIBDNF_LOCAL Transaction(const libdnf5::BaseWeakPtr & base);
 
@@ -209,6 +214,9 @@ private:
     std::optional<uint32_t> user_id;
     std::string comment;
     std::string description;
+
+    /// Clear the recorded last scriptlet output
+    LIBDNF_LOCAL void clear_last_script_output();
 };
 
 }  // namespace libdnf5::base
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -843,6 +843,17 @@ void RpmTransCB::cpio_error(const libdnf5::rpm::TransactionItem & item) {
     multi_progress_bar.print();
 }
 
+void RpmTransCB::script_output_to_progress(const libdnf5::cli::progressbar::MessageType message_type) {
+    auto transaction = context.get_transaction();
+    auto output = transaction->get_last_script_output();
+    if (!output.empty()) {
+        active_progress_bar->add_message(message_type, "Scriptlet output:");
+        for (auto & line : libdnf5::utils::string::split(output, "\n")) {
+            active_progress_bar->add_message(message_type, line);
+        }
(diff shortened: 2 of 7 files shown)


include/libdnf5/base/transaction.hpp
```diff
@@ -30,6 +30,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/rpm/transaction_callbacks.hpp"
 
 #include <optional>
+#include <string_view>
```
jrohel, line 33: Why is the `string_view` file included to the public header?
It is not used in this header. Please move the line to the `transaction_impl.hpp` file.

include/libdnf5/base/transaction.hpp
```diff
@@ -193,12 +194,17 @@ class LIBDNF_API Transaction {
     void set_download_local_pkgs(bool value);
     bool get_download_local_pkgs() const noexcept;
 
+    /// Retrieve output of the last executed rpm scriptlet.
+    std::string get_last_script_output();
```
jrohel, line 198: I'm wondering if we need to make a copy in the future or if it would be better to just return a constant reference.
`const std::string & get_last_script_output();`
If you change here to a reference, change in `transaction_impl.hpp` as well

jrohel (changes requested):
Thank you for the PR. Just a minor change, please. See notes in review.

include/libdnf5/base/transaction.hpp
m-blaha: Yes, I intended to make a copy here. The thing is that this string is regularly changed from another thread and I wanted to ensure that the string the client received will remain the same.

include/libdnf5/base/transaction.hpp
m-blaha: Thanks, I will fix it.

include/libdnf5/base/transaction.hpp
jrohel: If the user `calls get_last_script_output` in a `script_error` or `script_stop` callback, the script has already finished and a new script will be started after the next `script_start` callback. The rpm executes scripts serially (there is no parallelization). So the contents of the string should not change while the `script_error` and `script_stop` callbacks are being serviced. And if the user needs to keep it for longer, he will make his own copy. 

Your solution allows the user to call the `get_last_script_output` method from another thread. We don't generally support multithreading - it is not supported by the libraries used, for example libsolv. So I still don't know if this copy will help us in the future. But I will merge it.

m-blaha: I was specifically thinking about situations where the client keeps the output longer then during script_start/script_end execution. You are right that it can be handled on the client side. Thank you.

PR #1661, opened 2024-08-28
doc: add typical dnf5 workflow

kontura:
It also enables doc generation for couple classes.


jan-kolarik (approved):
Nice, thanks for that—it looks great! I hope this helps our API users, CLI users, and those using dnf5 in scripts better understand the entire dnf5 machinery.

jan-kolarik:
The build is passing and the failing tests are unrelated as these are just docs changes ...

Issue #1603, opened 2024-07-31
`config-manager addrepo` fails to add docker repository config


liemeldert:
Following https://docs.docker.com/engine/install/fedora/ and updating commands to DNF5 causes this error when adding their repository:
```
❯ sudo dnf config-manager addrepo --from-repofile=https://download.docker.com/linux/fedora/docker-ce.repo

https://download.docker.com/linux/fedora/docker-ce.repo                                                 100% |  20.6 KiB/s |   1.9 KiB |  00m00s
Error in added repository configuration file. Cannot set repository option "#1= ": Option "#1" not found
```

Importing the repo from dnf4 and then installing with dnf5 works without issues.

I'm facing this issue on Ultramarine (Fedora) 40, dnf5 version 5.1.17, config-manager 0.1.0, kernel 6.9.11-200.fc40.x86_64.

jan-kolarik:
Hi, thanks for the report and simple reproducer. I confirm the issue persist with the latest upstream dnf5 version. With dnf4 config-manager it works correctly.

PR #1641, opened 2024-08-20
Print diagnostic messages on stderr, not stdout

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/1361

This patchset changes many messages to be printed on stderr instead of stdout.

Opinions vary, but by general convention, the "useful output" of a program (that might be e.g. piped to a pager or consumed by a script), should be on stdout. All other messages, including warnings, progress bars, "Are you sure you want to continue?"-style prompts should be on stderr.

There are gray areas here, especially when you consider the same piece of output in different contexts. You could argue that the "Updating and loading repositories" message and multiprogressbar are "useful output" of the `makecache` command, but the same messages are not "useful output" of the `repoquery` command. And DNF 5 commands such as `install` and `upgrade` are primarily "imperative" and don't have "useful output" in the same way that commands like `repolist` and `repoquery` do. So here, I've tried to use my best judgment and err on the side of keeping the status quo.

The following messages are moved from stdout to stderr:
- All messages printed via `Context::print_info` go to stderr, which includes:
    - "Updating and loading repositories:"
    - "Repositories loaded."
    - "Running transaction"
    - "Testing offline transaction"
    - "After this operation, x extra will be used (install x, remove x)."
    - "After this operation, x extra will be freed (install x, remove x)."
    - "Total size of inbound packages is x. Need to download x."
    - "Complete!"
- Warnings, errors (most were already on stderr), and userconfirm prompts
- Progress bars
- Output related to key import

To observe these changes, it's useful to redirect both streams to files:

```
sudo dnf5 <args> >/tmp/stdout 2>/tmp/stderr
```

and watch them from other terminals:

```
tail -Fqf /tmp/stdout /tmp/stderr   # merged stdout stderr
tail -Ff /tmp/stdout                # just stdout
tail -Ff /tmp/stdout                # just stderr
```

<details>

<summary>Output of <code>dnf5 --refresh rq rawhide xz</code> after this patch</summary>

(before this patch, all output was on stdout.)

<details>

<summary>merged stdout and stderr</summary>

```
Updating and loading repositories:
 Fedora 40 openh264 (From Cisco) - x86_64    100% |   6.7 KiB/s | 989.0   B |  00m00s
 Fedora 40 - x86_64 - Updates                100% | 223.7 KiB/s |  26.2 KiB |  00m00s
 Fedora 40 - x86_64                          100% | 220.3 KiB/s |  28.0 KiB |  00m00s
[...]

PR #1671, opened 2024-09-03
config-manager: Fix addrepo from-repofile with empty/comment lines

jrohel:
Comments and blank lines were interpreted as options in the configuration file validity test. Now they are ignored.

Closes: https://github.com/rpm-software-management/dnf5/issues/1603

### DIFF
--- a/dnf5-plugins/config-manager_plugin/addrepo.cpp
+++ b/dnf5-plugins/config-manager_plugin/addrepo.cpp
@@ -346,6 +346,12 @@ void ConfigManagerAddRepoCommand::add_repos_from_repofile(
         // Test if the repository options can be set.
         for (const auto & [repo_id, repo_opts] : parser.get_data()) {
             for (const auto & [key, key_val] : repo_opts) {
+                // Skip empty lines and comment lines (ConfigParser stores the empty line and comment line
+                // in an automatically generated key whose name starts with the '#' character).
+                if (key.starts_with('#')) {
+                    continue;
+                }
+
                 try {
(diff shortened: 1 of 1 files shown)


dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -346,6 +346,9 @@ void ConfigManagerAddRepoCommand::add_repos_from_repofile(
         // Test if the repository options can be set.
         for (const auto & [repo_id, repo_opts] : parser.get_data()) {
             for (const auto & [key, key_val] : repo_opts) {
+                if (key.starts_with("#")) {
```
jan-kolarik, line 349: I understand this fixes the comment lines, but how it fixes the empty ones?

jan-kolarik: Or at least at the comment here why it is skipping empty and comment lines, because just looking at this, it's a bit mysterious to me :)

jrohel: OK. I added a comment.

jan-kolarik: Thanks!

jan-kolarik (approved):

jan-kolarik:
The 3 failing tests are not related to this PR, it seems they were introduced after merging https://github.com/rpm-software-management/dnf5/pull/1641 and related tests. I will look into that...

Issue #1668, opened 2024-09-02
Skipping PGP checks for packages from offline transaction


jan-kolarik:
Using dnf5-5.2.5.0-20240828005842.55.g4b2fb499.fc40.x86_64.

I've upgraded the machine from f39 to f40.

When I run `dnf5 update` in the f40 it asks to import the f40 PGP key, as expected.

When I run `dnf5 update --offline`, it does _not_ ask to import the key. It claims on the terminal:

> Warning: skipped PGP checks for 2 packages from repository: @stored_transaction

Similarly when running an offline update using the D-Bus API of the dnf5daemon-server; it does not ask to import the key too.

I think it makes sense to have the PGP key imported and checked for (when the repo has it enabled) regardless whether the update is done immediately or offline.

_If you think this bug is different from that mine, then just say so and I can create a new bug instead._

_Originally posted by @mcrha in https://github.com/rpm-software-management/dnf5/issues/991#issuecomment-2314891892_

m-blaha:
Looks like it's a regression of caused by 5c00c1a
We also need to add a test for this scenario.

Issue #1667, opened 2024-09-02
system-upgrade download provides confusing output and stuck at some point


j-mracek:
`dnf5 system-upgrade download --releasever 40` ended with following outputs

```
[2687/2687] Total                                                                                                                                     100% |  20.2 MiB/s |   3.2 GiB |  02m40s
Testing offline transaction
Warning: skipped PGP checks for 2687 packages from repository: @stored_transaction
Transaction stored to be performed offline. Run `dnf5 offline reboot` to reboot and run the transaction. To cancel the transaction and delete the downloaded files, use `dnf5 offline clean`.
Complete!

```

After line `[2687/2687] Total ` it requires 3m 40s probably to generate file with description of transaction. Then it reports `Testing offline transaction`. We could discuss whether 10 or 20 seconds is a limit for user where to start to think what is going on, but after nearly 4 minute many user will interrupt. I recommend to improve performance and create a bar with progress.

The next line is problematic: `Warning: skipped PGP checks for 2687 packages from repository: @stored_transaction`
How I can trust the transaction? It mentions repository that is not in transaction and it anyway does not check PGP keys.  It scares me a lot.

jan-kolarik:
Hi Jarda, thanks for the info!

> We could discuss whether 10 or 20 seconds is a limit for user where to start to think what is going on, but nearly 4 minute many user will interrupt. I recommend to improve performance and create a bar with progress.

I completely agree. We should regularly update users on the progress to ensure they know that something is happening.

> The next line is problematic: Warning: skipped PGP checks for 2687 packages from repository: @stored_transaction
How I can trust the transaction? It mentions repository that is not in transaction and it anyway does not check PGP keys. It scares me a lot.

We've already received a similar report and created an issue for it: https://github.com/rpm-software-management/dnf5/issues/1668. Let's focus just on the progress report in this issue.

jan-kolarik:
> We should regularly update users on the progress to ensure they know that something is happening.

The key issue raised in this ticket remains unresolved, even after the changes in #1672. I tested upgrading from Fedora 40 to Fedora 41 using dnf5 system-upgrade and encountered insufficient feedback in two scenarios:

- **Post-reboot, before the transaction starts:** A spinning wheel is displayed for over a minute with no output for the user.

- **End of the transaction**: There’s still a long delay with no updates provided, even if the duration may have slightly improved after the PR.

These gaps make the upgrade process unclear and might be frustrating for users.

PR #1672, opened 2024-09-03
dnf5: Run transaction test for offline transactions

m-blaha:
A serialized offline transaction consists only of local packages
downloaded to the `<system state dir>/offline` location. Unfortunately,
PGP checks for these packages are disabled by default due to the
`localpkg_gpgcheck` option default value.
Instead of testing the serialized transaction, this patch performs an
RPM transaction test on the originally resolved transaction, which still
retains information about the repositories from which the packages were
downloaded.
This approach also saves time required for deserialization and resolving
the testing transaction.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1668
Resolves: https://github.com/rpm-software-management/dnf5/issues/1667

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1549

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -349,7 +349,6 @@ void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
     const auto & installroot = base.get_config().get_installroot_option().get_value();
     const auto & offline_datadir = installroot / libdnf5::offline::DEFAULT_DATADIR.relative_path();
     std::filesystem::create_directories(offline_datadir);
-
     constexpr const char * packages_in_trans_dir{"./packages"};
     constexpr const char * comps_in_trans_dir{"./comps"};
     const auto & comps_location = offline_datadir / comps_in_trans_dir;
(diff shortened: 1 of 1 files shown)


m-blaha:
The original approach had one advantage that the exactly same transaction that would be performed after `dnf offline reboot` was tested. Do you think it makes sense to at least try to resolve the serialized transaction (although it takes a long time so we should properly inform the user on what's going on). @evan-goode @jan-kolarik

evan-goode (approved):
I'm fine with this, I tested upgrading F39->F40 and it works.

> Do you think it makes sense to at least try to resolve the serialized transaction (although it takes a long time so we should properly inform the user on what's going on).

I think we can skip it. The main reason I implemented it this way was that it seemed to not be possible to test the (unserialized) offline transaction because the RPMs are all stored in the same directory. But that turned out to be due to a different bug. 

Assuming the transaction serialization logic is reliable, testing the serialized transaction is unnecessary and costs time.

dnf5/context.cpp
```diff
@@ -321,6 +321,27 @@ void Context::Impl::load_repos(bool load_system) {
 }
 
 void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
+    // Test the transaction
```
jrohel, line 324: Scenario:
1. a new transaction is tested
2. `get_status()` of the offline transaction status is tested - if there is already an offline transaction in the queue and an `AbortedbyUserError` exception is thrown, the tested transaction is discarded
3. serialize the transaction under test

Can we change the order? Test new transaction only if `get_status()` test passes?
1. `get_status()` of offline transaction status is tested
2. a new transaction is tested
3. serialize the transaction under test

m-blaha: That's a good point. Maybe we should do the status test even before downloading transaction packages.

m-blaha: I've added a commit with such optimization. Now the offline transaction status is checked before the download phase.

PR #1677, opened 2024-09-05
Fix regressions from stderr/stdout changes

evan-goode:
Follow-up to https://github.com/rpm-software-management/dnf5/pull/1641, I forgot to push these two commits.

More details in the commit messages.

Fixes the following ci-dnf-stack tests:
- `dnf/install-file-conflicts.feature:8  An error is reported when a package with a file conflict is tried to be installed`
- `dnf/dnf-automatic/reboot.feature:95  dnf-automatic shows error message when reboot command failed`


jan-kolarik (approved):
Thanks for the fix!

Issue #1653, opened 2024-08-26
dnf5daemon: Session caching can be harmful


mcrha:
When using only a single session a package install followed by the same package uninstall results into: `rpm transaction failed with code 5` and the package is not uninstalled, while using a different session for the uninstall works fine.

The package flags are also affected by this, because starting a new session and removing the package and immediately after that using the same session to get the package attributes keeps the `is_installed` flag of the package on `true`, while using two different sessions provides accurate information to the dnf5daemon client.

I believe both things are closely related, thus I filled them as a single issue.

This is with dnf5-5.2.5.0-20240826005752.50.gb1f839cd

m-blaha:
This issue prevents gnome software from re-using the same dnf5daemon session. Bumping priority.

m-blaha:
Proposed PR https://github.com/rpm-software-management/dnf5/pull/1678 which re-sets the goal automatically after the do_transaction() execution. Also adds a new D-Bus API call to reset the goal manually.

m-blaha:
Now I'm thinking about the second part (the is_installed flag), this would require re-reading the system repo. Let me explore what options there are.

mcrha:
I tried the changes from the #1757 in action and it seems to do the proper thing (I've just done some quick/smoke testing). Thank you.

PR #1678, opened 2024-09-06
daemon: Reset the goal

m-blaha:
New D-Bus API to reset the goal.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1653

Changes:

3cd0a6e3 (Marek Blaha, 5 minutes ago)
   dnfdaemon: D-Bus API to reset the goal

   With this patch the user can re-use the session to resolve multiple 
   transactions without running them. The use case might be to check what 
   dependencies would be installed with given package.

d77231b8 (Marek Blaha, 17 hours ago)
   dnfdaemon: Add reset_goal() method

   The method resets the current goal and transaction, which is necessary for
   scenarios where the user wants to reuse the D-Bus session to resolve or
   execute multiple transactions.

5dbe16a0 (Marek Blaha, 17 hours ago)
   goal: Add missing members to reset() method

   The reset() method currently does not reset all Goal members. This patch 
   adds those that are missing.

### DIFF
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -3296,12 +3296,17 @@ void Goal::add_redo_transaction(
 void Goal::reset() {
     p_impl->module_specs.clear();
     p_impl->rpm_specs.clear();
+    p_impl->rpm_reason_change_specs.clear();
     p_impl->rpm_ids.clear();
     p_impl->group_specs.clear();
     p_impl->rpm_filepaths.clear();
     p_impl->resolved_group_specs.clear();
     p_impl->resolved_environment_specs.clear();
+    p_impl->group_specs.clear();
     p_impl->rpm_goal = rpm::solv::GoalPrivate(p_impl->base);
+    p_impl->serialized_transaction.reset();
+    p_impl->revert_transactions.reset();
+    p_impl->redo_transaction.reset();
 }
 
 BaseWeakPtr Goal::get_base() const {
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.Goal.xml
@@ -106,6 +106,13 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         <arg name="error_msg" type="s" direction="out" />
     </method>
 
+    <!--
+        reset:
+        Reset the prepared rpm transaction. After this call the session is ready to perform another rpm transaction.
+    -->
+    <method name="reset">
+    </method>
+
 </interface>
 
 </node>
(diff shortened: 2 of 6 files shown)


m-blaha:
I'm not totally sure about automatically resetting after the transaction is done. The transaction might still be needed for follow-up calls, like collecting RPM errors that happened during execution.
@mcrha Would it be OK for you to manually call the reset() method at some point after the transaction is done?

mcrha:
Aha, you are right. If the errors reported by `Goal::get_transaction_problems_string()` or `Goal::get_transaction_problems()` still require the failed transaction, then it's perfectly fine to have the `reset()` called manually.

In fact, I planned to call the `reset()` always before I start a transaction (not exactly before `Goal::do_transaction()` call), to make sure nothing is left from the previous calls/transactions.

mcrha:
I gave a quick test to the changes in the current form and it works as expected in my use case. I tried to install/uninstall packages with not shared and with shared session and both cases work fine, including when "simulating" a transaction to figure out needed size to download/install.

libdnf5/base/goal.cpp
```diff
@@ -3296,12 +3296,17 @@ void Goal::add_redo_transaction(
...
     p_impl->rpm_ids.clear();
     p_impl->group_specs.clear();
     p_impl->rpm_filepaths.clear();
     p_impl->resolved_group_specs.clear();
     p_impl->resolved_environment_specs.clear();
+    p_impl->group_specs.clear();
     p_impl->rpm_goal = rpm::solv::GoalPrivate(p_impl->base);
+    p_impl->serialized_transaction.reset();
+    p_impl->revert_transactions.reset();
+    p_impl->redo_transaction.reset();
```
jan-kolarik, line 3309: That's fine, but I was thinking why not we recreate the goal member in the daemon session. The goal object's destructor should handle cleanup automatically, right? That way, we wouldn't need to explicitly clear individual items here or worry about adding new ones in the future.

m-blaha: We can do that in the daemon, no problem.
But since the `libdnf5::Goal::reset()` public API has already existed, I wanted to fix it. We can of course deprecate it and remove in the future.

jan-kolarik: Let's keep it as it is now, I will create a new issue for that.

jan-kolarik: https://github.com/rpm-software-management/dnf5/issues/1718

jan-kolarik (approved):

jan-kolarik:
@m-blaha The tests should be passing after merging the https://github.com/rpm-software-management/dnf5/pull/1711. Could you please rebase to check it?

m-blaha:
> @m-blaha The tests should be passing after merging the #1711. Could you please rebase to check it?

Sorry for the delay, I missed the message. Rebased.

PR #1689, opened 2024-09-11
transaction_callbacks: Deprecate confusing alias

m-blaha:
Given that there is another TransactionItem class in
libdnf5::transaction namespace, the alias libnf5::rpm::TransactionItem ->
libdnf5::base::TransactionPackage is confusing and does not add any
value to the user.
This patch deprecates it and uses directly TransactionPackage in all
places.

### DIFF
--- a/dnf5-plugins/automatic_plugin/transaction_callbacks_simple.cpp
+++ b/dnf5-plugins/automatic_plugin/transaction_callbacks_simple.cpp
@@ -33,7 +33,7 @@ void TransactionCallbacksSimple::transaction_start([[maybe_unused]] uint64_t tot
 
 
 void TransactionCallbacksSimple::install_start(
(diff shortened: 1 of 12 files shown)


ppisar (commented):
Looks good. It preserves API and ABI.
The test failures are probably caused by toml11 (#1693).

I have only one request: Could you make the commit title more concrete? E.g.  "transaction_callbacks: Deprecate libdnf5::rpm::TransactionItem alias". Those titles go into release notes.

m-blaha:
> Looks good. It preserves API and ABI. The test failures are probably caused by toml11 (#1693).
> 
> I have only one request: Could you make the commit title more concrete? E.g. "transaction_callbacks: Deprecate libdnf5::rpm::TransactionItem alias". Those titles go into release notes.

Thanks, changed.

ppisar (approved):

Issue #1624, opened 2024-08-13
automatic: Consider reading the dnf4 configuration file for backward compatibility


jan-kolarik:
The configuration for the `automatic` functionality has been moved to a different location compared to dnf4, as this functionality is now provided as a plugin (see the [changes docs](https://dnf5.readthedocs.io/en/latest/changes.html#changes-to-individual-commands) and [automatic command docs](https://dnf5.readthedocs.io/en/latest/dnf5_plugins/automatic.8.html#automatic-plugin-ref-label)). The content is backwards compatible, and since we are already reading multiple configuration file locations, we might consider applying the values from the original location for users who have migrated to dnf5.

keszybz:
I don't think the new location makes much sense. As a user, I really don't want to care or think about the fact this functionality is provided by a plugin. I'd vote for just using the old location (which is backwards compatible and makes the most sense for the user).

jan-kolarik:
> I don't think the new location makes much sense. As a user, I really don't want to care or think about the fact this functionality is provided by a plugin. I'd vote for just using the old location (which is backwards compatible and makes the most sense for the user).

Yeah, good point, thanks for the feedback! If the plugin still needs to access the configuration at the standard plugin location, we can just add a symlink there.

PR #1690, opened 2024-09-11
automatic: Use original dnf4 config file location

m-blaha:
Since the automatic configuration file remains compatible, this change
makes the upgrade dnf4->dnf5 easier for the user.
Having so many automatic config files seems overly complicated so the
change also deprecates /etc/dnf/dnf5-plugins/automatic.conf location and
warns the user if they use it.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1624

### DIFF
--- a/dnf5-plugins/automatic_plugin/automatic.cpp
+++ b/dnf5-plugins/automatic_plugin/automatic.cpp
@@ -281,22 +281,33 @@ void AutomaticCommand::pre_configure() {
 
     // read the config files in following order (/etc overrides /usr):
     //      - [installroot]/usr/share/dnf5/dnf5-plugins/automatic.conf
-    //      - [installroot]/etc/dnf/dnf5-plugins/automatic.conf
+    //      - [installroot]/etc/dnf/dnf5-plugins/automatic.conf [deprecated]
+    //      - [installroot]/etc/dnf/automatic.conf
     auto & main_config = base.get_config();
     bool use_host_config{main_config.get_use_host_config_option().get_value()};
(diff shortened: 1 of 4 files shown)


dnf5-plugins/automatic_plugin/automatic.cpp
```diff
@@ -281,17 +281,27 @@ void AutomaticCommand::pre_configure() {
...
         if (!use_host_config) {
             conf_file_path = installroot_path / conf_file_path.relative_path();
         }
         if (std::filesystem::exists(conf_file_path)) {
+            if (deprecated) {
+                auto msg = libdnf5::utils::sformat(
+                    _("Warning: Configuration file location \"{}\" for dnf automatic is deprecated. See "
+                      "dnf5-automatic(8) for details."),
+                    conf_file_path.string());
+                logger.warning(msg);
```
ppisar, line 302: The build fails:

~~~~
dnf5-plugins/automatic_plugin/automatic.cpp:302:17: error: ‘logger’ was not declared in this scope
~~~~

You might want to replace the call with `(*base.get_logger()).warning(msg)`.

ppisar: `base.get_logger()->warning(msg)`, actually.

m-blaha: Sorry. I made this change some time ago and forgot to test it. I'll fix it.

ppisar (approved):
Good. It works for me. The build failures in CI are caused by toml11 which not relevant here.

PR #1691, opened 2024-09-11
spec: Recommend dnf5-plugins if dnf-plugins-core installed

evan-goode:
Tested F40 -> F41 upgrade using `dnf4 system-upgrade --releasever=40` and `dnf4 distro-sync --releasever=40`. If `dnf-plugins-core` is installed, then the system-upgrade (or distrosync) will install `dnf5-plugins` as a weak dependency. If `dnf-plugins-core` is missing, then `dnf5-plugins` will not be installed.

For https://bugzilla.redhat.com/show_bug.cgi?id=2309697


m-blaha (approved):
Thank you!

Issue #1693, opened 2024-09-12
offline and system-upgrade tests die: Error reading state: [error] toml::parse_file: invalid table key


ppisar:
These ci-dfn-stack tests fail in Fedora ≥ 41:

Failing scenarios:
  dnf/offline.feature:28  Test offline when an offline transaction is already queued
  dnf/offline.feature:73  Test offline clean
  dnf/offline.feature:91  Test offline install and offline status
  dnf/offline.feature:128  Test offline upgrade
  dnf/offline.feature:151  Test offline group install
  dnf/offline.feature:174  Test offline-distrosync
  dnf/offline.feature:202  Test offline-upgrade
  dnf/offline.feature:230  Test offline install with local rpm (absolute path)
  dnf/offline.feature:246  Test offline install with local rpm (relative path)

Failing scenarios:
  dnf/system-upgrade.feature:45  Test system-upgrade basic functionality
  dnf/system-upgrade.feature:70  Test system-upgrade with --cachedir
  dnf/system-upgrade.feature:95  Test system-upgrade with --no-downgrade
  dnf/system-upgrade.feature:118  Test system-upgrade transaction file not found
  dnf/system-upgrade.feature:141  Test system-upgrade downloading a package from a different repo
  dnf/system-upgrade.feature:222  Test system-upgrade with --setopt=cachedir doesn't delete user files -- @1.1 
  dnf/system-upgrade.feature:223  Test system-upgrade with --setopt=system_cachedir doesn't delete user files -- @1.2 

An example:

~~~~
    Given I successfully execute dnf with args "offline reboot"                        # dnf/steps/cmd.py:211
      Assertion Failed: Command has returned exit code 1: dnf5 -y --installroot=/tmp/dnf_ci_installroot_lrs2l8nv --releasever=29 --setopt=module_platform_id=platform:f29 --setopt=cachedir=/tmp/dnf_ci_installroot_lrs2l8nv/var/cache/dnf offline reboot
      Captured stdout:
      [31m
      Last Command: [31m[1mdnf5 -y --installroot=/tmp/dnf_ci_installroot_lrs2l8nv --releasever=29 --setopt=module_platform_id=platform:f29 --setopt=cachedir=/tmp/dnf_ci_installroot_lrs2l8nv/var/cache/dnf offline reboot
      [36m[1m
      Last Command stderr:[90m
      Error reading state: [error] toml::parse_file: invalid table key
[...]

PR #1694, opened 2024-09-12
offline: Update usage of toml11-devel

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/1693.


ppisar (approved):
This indeed fixes the reported test failures.

Remaining test failure "dnf/plugins-core/debuginfo-install.feature:76  installs debuginfo when a -debuginfo package is provided" is not related.

Build on Fedora 41 fails because there is still in use unpatched toml11-4.1.0-1.fc41.

ppisar:
I guess this fix should be applied to Fedora 41 before releasing Fedora. Otherwise, people won't be able to do an offline upgrade after a fresh installation.

Issue #1596, opened 2024-07-25
Document how to do forcearch as an API user


jan-kolarik:
Currently we have documentation for CLI users [here](https://dnf5.readthedocs.io/en/latest/misc/forcearch.7.html), but it's not clear how to do the same thing through API. Also, the process involves setting up the `arch` variable which is different compared with dnf4 where `arch` was also a configuration option. We should also improve our "dnf4 vs dnf5 changes" docs in this manner.

PR #1695, opened 2024-09-13
doc: Document arch override for API users

m-blaha:
We have documentation for CLI users (dnf5-forcearch(7)), but it's not
clear how to do the same thing through API. This patch fills the gap.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1596


doc/tutorial/index.rst
```diff
@@ -11,3 +11,4 @@ Tutorial
     repos
     queries
     transaction
+    force-arch
```
kontura, line 14: It looks good, I am only thinking if it would be a bit clearer if it was added right after the `session`?
I like how the tutorials are ordered according to the workflow progression and adding this at the end disrupts that.

I had also one more idea that the page could be a child of the `session` page but I am not sure about that.

m-blaha: Originally I was thinking about just adding the commented section to the `session` example - but given that this really is kind of a corner case, I didn't want to over-complicate the tutorial example.
But let me experiment with a session sub-page, it might work.

m-blaha: I've changed the PR to make the forcearch tutorial a sub-section of session and I like it. What do you think?

kontura: I like it as well. Thank you!

Issue #1688, opened 2024-09-10
Many texts in "dnf install" output are not internationalized


bittin:
dnf5 in Fedora 41 installed with https://kojipkgs.fedoraproject.org/compose/branched/Fedora-41-20240910.n.0/compose/Workstation/x86_64/iso/Fedora-Workstation-Live-x86_64-41-20240910.n.0.iso is still in English: https://github.com/rpm-software-management/dnf5/blob/main/dnf5/po/sv.po even if there is a Swedish translation

https://i.imgur.com/FjgHw5d.png

ppisar:
Could you tell us what locale do you use (`locale` command output) and what DNF5 version you have installed (`rpm -q dnf5`)?

For me it works. The first message in the linked sv.po is:

~~~~
#: commands/advisory/advisory_info.hpp:33
msgid "Print details about advisories"
msgstr "Skriv ut detaljer om rekommendationer"
~~~~

It's a description of "info" command in `dnf advisory --help` output.

When I install a Swedish locale (`dnf install glibc-langpack-sv`) and call `dnf5 advisory --help` in sv_SE.utf8 locale, I get:

~~~~
# LC_ALL=sv_SE.utf8 dnf5 advisory --help
Användning:
  dnf5 [GLOBAL OPTIONS] advisory <COMMAND> ...
                                
Query Commands:                 
  list                          Lista rekommendationer
  info                          Skriv ut detaljer om rekommendationer
  summary                       Skriv ut en sammanfattning av rekommendationer
~~~~

That matches the translation.

bittin:
> Could you tell us what locale do you use (`locale` command output) and what DNF5 version you have installed (`rpm -q dnf5`)?
> 
> For me it works. The first message in the linked sv.po is:
> 
> ```
> #: commands/advisory/advisory_info.hpp:33
> msgid "Print details about advisories"
> msgstr "Skriv ut detaljer om rekommendationer"
> ```
> 
> It's a description of "info" command in `dnf advisory --help` output.
> 
> When I install a Swedish locale (`dnf install glibc-langpack-sv`) and call `dnf5 advisory --help` in sv_SE.utf8 locale, I get:
> 
> ```
> # LC_ALL=sv_SE.utf8 dnf5 advisory --help
> Användning:
>   dnf5 [GLOBAL OPTIONS] advisory <COMMAND> ...
>                                 
> Query Commands:                 
>   list                          Lista rekommendationer
[...]

PR #1696, opened 2024-09-13
I18N: Mark messages in "dnf install" output for a translation

ppisar:
Many texts were always in English because the code did not internatialinze them. This patch fixes it.

fmt::format() calls were replaced with libdnf5::utils::sformat() calls to be able to pass a localized string as a formatting string.

Sentences glued from words, e.g. "Installing dependencies:" in a transaction table, were changed to a single-string sentences in the code. Having a full sentece is important for the translators.

I fixed spelling at few places, especially around importing PGP keys.

I used internatinoalized plural forms in the transaction summary ("Installing: 1 package" vs "Installing: 2 packages").

I also internationalized output to DNF log. At the end, it's a human-oriented text.

I intentionally left out libdnf_throw_assertion() arguments from the internationalization because that function needs changing its implmentation to accept nonconst-literal formatting string. That will be implemented in a separate patch.

I manully tested significant parts of the output to be sure the messages propetly undergo localization.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1688


m-blaha (approved):
Thank you!

PR #1697, opened 2024-09-13
swig: Add wrappers for TransactionEnvironment and TransactionGroup

pkratoch:
Resolves: https://issues.redhat.com/browse/SWM-2993


ppisar (approved):
It looks sane. I guess including the header files with %include directive instead of #include in %{%} block is for processing the declared C++ templates by Swig.

Issue #1687, opened 2024-09-10
dnf5 does not use localized package summaries and descriptions


musicinmybrain:
As reported in https://github.com/rpm-software-management/dnf5/issues/1685#issuecomment-2340920728:

There are also differences in localization, which kind of worked in `rpm`, kind of worked *differently* in `dnf` 4, and doesn’t work at all in `dnf5`. You can compare these in Fedora 40:

- `LC_ALL=zh_CN.utf8 rpm -qi python3-fastapi`: uses localized summary and description text and build date; does not localize its own fixed strings (field labels)
- `LC_ALL=zh_CN.utf8 dnf info python3-fastapi`: localizes all of its own strings (dates, field labels), but does not use the localized summary and description text
- `LC_ALL=zh_CN.utf8 dnf5 info python3-fastapi`: ignores locale completely, hex-escapes `•` in the default English description

(Don’t try other locales for now; I just fixed an issue with the other localized descriptions in the `python-fastapi` spec file, and the updates for this are not yet in stable repositories.)

Displaying localized text will require proper Unicode support, https://github.com/rpm-software-management/dnf5/issues/1685, for most locales.

jrohel:
In fedora 40 there is now dnf version 5.1.17. This version does not use locale.
Support was added to the upstream some time ago. I assume the problem was solved by this commit https://github.com/rpm-software-management/dnf5/commit/3d302e0b32174eab40176befddb26628f283757a .
That is, since dnf5 version 5.2.0.

musicinmybrain:
What I see when I try this in a Fedora 41 mock chroot with `--enable-network` and `dnf5 version 5.2.5.0` is *weird*.

```
# dnf info python3-fastapi
[…]
               :   • Fast: Very high performance, on par with NodeJS and Go (thanks to Starlette
[…]
# dnf info uv
[…]
               :   • ⚖️ Drop-in replacement for common pip, pip-tools, and virtualenv commands.
[…]
```

…so Unicode output is working in my default locale. But now, if I set it explicitly:

```
# LC_ALL=en_US.UTF-8 dnf info python3-fastapi
[…]
               :   \xe2\x80\xa2 Fast: Very high performance, on par with NodeJS and Go (thanks to Starlette
[…]
# dnf info uv
[…]
               :   \xe2\x80\xa2 \xe2\x9a\x96\xef\xb8\x8f Drop-in replacement for common pip, pip-tools, and virtualenv commands.
[…]
```

[...]

PR #1698, opened 2024-09-13
I18N: Mark messages in "dnf info" output for a translation

ppisar:
Related: https://github.com/rpm-software-management/dnf5/issues/1687

### DIFF
--- a/libdnf5-cli/output/packageinfo.cpp
+++ b/libdnf5-cli/output/packageinfo.cpp
@@ -52,25 +52,27 @@ void print_package_info(
     scols_table_enable_noheadings(table, 1);
     scols_table_new_column(table, "key", 1, 0);
     scols_table_new_column(table, "value", 1, SCOLS_FL_WRAP);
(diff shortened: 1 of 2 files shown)


libdnf5-cli/output/packageinfo.cpp
```diff
@@ -52,25 +52,25 @@ void print_package_info(
     scols_table_enable_noheadings(table, 1);
     scols_table_new_column(table, "key", 1, 0);
     scols_table_new_column(table, "value", 1, SCOLS_FL_WRAP);
-    scols_table_set_column_separator(table, " : ");
+    scols_table_set_column_separator(table, _(" : "));
```
pkratoch, line 55: Is it expected that the translators know what purpose this string serve or should a note, such as https://github.com/rpm-software-management/dnf5/blob/main/dnf5daemon-client/commands/advisory/arguments.hpp#L153-L154 , be added?

pkratoch: Thank you!

PR #1699, opened 2024-09-13
I18N: Set C++ locale and format sizes according to the locale

ppisar:
Displaying a package size, an installation size, a download speed etc. formats decimal numbers to 1-digit precision after the decimal point (49.5 KiB). However, users expect the number to be formatted according their locale. E.g. in cs_CZ.UTF-8, it is "49,5 KiB".

DNF5 formats these values with fmt::format() which utilizes C++ locale if "L" formatting option is used.

C++ locale (std::locale::global()) and C locale (setlocale(), C++-wrapped as std::setlocale()) are two different things and DNF5 only has set the C locale up to now.

This patch starts setting C++ locale, which also implicitly sets C locale. This patch also modifies
libdnf5::cli::utils::units::format_size_aligned() to use the locale-dependent decimal seperator (available since fmt-8.0.0).

I manually tested dnf5 and dnf5daemon-client and they work for me.

Though there is a risk that the new C++ locale will affect some code uknown to me, like regular expression matching, or thread-specific locales. If the affected code was unfixable, we can resort to saving the desired C++ locale into a dedicated object accessible to format_size_aligned() and pass it explicitly for fmt::format. Thorough testing is welcome.

Related: https://github.com/rpm-software-management/dnf5/issues/1687


jrohel:
I wonder how many inputs, outputs and regexes this PR will break.
We already have broken parsing of some configuration values for some locales. Probably from this https://github.com/rpm-software-management/dnf5/pull/1408. And it's because parsing uses functions that depend on the C locale (at least `std::stod` is used for `log_size`, `bandwidth`, `minrate`, all options using class `OptionSeconds` and `std::stof` for `throttle`).
This PR breaks the parsing of other numeric configuration values for some locales (`OptionNumber` class uses `operator >>` for parsing).

I'm considering whether to merge this PR now or first prepare a PR to fix the bugs I mentioned above.

ppisar:
First we need to fix the parser.

PR #1701, opened 2024-09-17
Fix: libdnf5-cli: TransactionSummary counters data type

jrohel:
Fixes `error: implicit conversion changes signedness: 'const int' to 'unsigned long' [-Werror,-Wsign-conversion]`

The bug was caused by commit https://github.com/rpm-software-management/dnf5/pull/1696

### DIFF
--- a/libdnf5-cli/output/transaction_table.cpp
+++ b/libdnf5-cli/output/transaction_table.cpp
@@ -186,14 +186,14 @@ public:
     }
 
 private:
-    int installs = 0;
-    int reinstalls = 0;
-    int upgrades = 0;
-    int downgrades = 0;
-    int removes = 0;
-    int replaced = 0;
-    int reason_changes = 0;
-    int skips = 0;
+    unsigned int installs = 0;
+    unsigned int reinstalls = 0;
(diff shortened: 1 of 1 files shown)


ppisar:
Sorry. I did not notice that ngettext() expects an unsigned long int as a counter. Shouldn't we add -Wsign-conversion to CXXFLAGS so that GCC catches it too?

m-blaha (approved):
Thank you!

m-blaha:
> Sorry. I did not notice that ngettext() expects an unsigned long int as a counter. Shouldn't we add -Wsign-conversion to CXXFLAGS so that GCC catches it too?

Good idea. I tried, but I must have done it incorrectly, because even with `-Wsign-conversion` the gcc did not catch the error.

jrohel:
We have `Package Build / CCache Build (clang)`, which uses `clang` for compilation. Unfortunately wrong. As you can see with this PR, the build fails because it uses source code without this PR. In other words, it doesn't test this PR but the older code without it. So if there's a bug in the PR, it will show up later in some other PRs. The cache is working strangely.

PR #1702, opened 2024-09-17
RepoSack: New API to reload system repo

m-blaha:
Introduce a new API to clear all solvables from the system repo and
reload fresh data from the RPM database.

Re-creating the Base object can be resource-intensive, and there is a
use case in dnfdaemon where the current session (which wraps around the
Base instance) needs to be reused even after executing an RPM
transaction - whether for queries or running another transaction.

### DIFF
--- a/libdnf5/repo/solv_repo.cpp
+++ b/libdnf5/repo/solv_repo.cpp
@@ -792,4 +792,9 @@ void SolvRepo::create_environment_solvable(
     repodata_internalize(data);
 }
 
+void SolvRepo::empty() {
+    // TODO(mblaha): is resuseid=true safe enough?
+    repo_empty(repo, true);
+}
+
 }  // namespace libdnf5::repo
--- a/include/libdnf5/repo/repo.hpp
+++ b/include/libdnf5/repo/repo.hpp
@@ -319,10 +319,15 @@ private:
     friend class PackageDownloader;
     friend class solv::Pool;
 
-    /// Loads the repository objects into sacks.
+    /// Loads the repository objects into sacks. Only the first call actually loads the repo.
     ///
     /// Also writes the libsolv's solv/solvx cache files.
     LIBDNF_LOCAL void load();
+
+    /// Empties currently loaded repo and loads it again.
+    LIBDNF_LOCAL void reload();
+
+    LIBDNF_LOCAL void _load();
 
     /// Downloads repository metadata.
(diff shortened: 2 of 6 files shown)


m-blaha:
@jrohel @kontura do you see any issues with this?
It can also be used by dnfdaemon to handle messages from `rpm-plugin-dbus-announce` rpm plugin.

kontura:
> @jrohel @kontura do you see any issues with this? It can also be used by dnfdaemon to handle messages from `rpm-plugin-dbus-announce` rpm plugin.

It does make me feel uneasy. It basically invalidates the pool right? Libsolv `ids` might have changed. So any `Package` object, `Goal` or `PackageQuery` in progress should be invalidated as well.

If I understand correctly for example existing `Package` could simply transform into a different one after the reload is used. I could see this causing quite hard to find bugs.

m-blaha:
Yes, that's right. Any place that keeps an id of an installed package is invalid after this call.

libdnf5/repo/solv_repo.cpp
```diff
@@ -792,4 +792,9 @@ void SolvRepo::create_environment_solvable(
     repodata_internalize(data);
 }
 
+void SolvRepo::empty() {
+    // TODO(mblaha): is resuseid=true safe enough?
```
jrohel, line 796: Code from libsolv:
```
  if (reuseids && repo->end == pool->nsolvables)
    {
      /* it's ok to reuse the ids. As this is the last repo, we can
         just shrink the solvable array */
```
It looks like `reuseids` works only when the repo is last. DNF places the system repo first. Behind it are available repos. The system repo is last only if it is the only one (we don't have available repos loaded).

jrohel:
You are introducing a new API method `void reload_system_repo()`. DNF saves the system repository to the pool first, followed by the others. Thus, installed solvables have a lower `id` than available. What do you expect to happen after calling the libsolv function `repo_empty(repo, true)` when there are other repos after the repo? Will space and ID for the packages be able to be reused? I'm not sure about that. And what if the new system repo is larger (a package has been installed)? Surely the repo won't fit in the original space. Will the entire repo be saved to the end of the pool? Or will it be scattered in the pool - part in the original location and part at the end of the pool?

When we sort packages and they have the same nevra, the solvable `id` comes into play. Thus, moving the system repo to the end affects the sorting of the packages. The question is how much does it matter.

In general we are getting into the untested area of using libsolv.

PR #1703, opened 2024-09-17
Warn on sign conversion

ppisar:
This patch set adds -Wsign-conversion to compiler flags and fixes the code be compilable with it. It should prevent from surprises like in #1701.

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -85,6 +85,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <cstring>
 #include <filesystem>
 #include <iostream>
+#include <utility>
 
 constexpr const char * DNF5_LOGGER_FILENAME = "dnf5.log";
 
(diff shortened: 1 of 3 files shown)


dnf5/main.cpp
```diff
@@ -773,29 +774,38 @@ static void print_versions(Context & context) {
...
         } else if (transaction_item_action_is_outbound(trans_pkg.get_action())) {
             remove_size += pkg.get_install_size();
+            if (remove_size < pkg.get_install_size())
+                return;
         }
     }
 
-    if (in_pkgs_size != 0) {
-        const auto [in_pkgs_size_value, in_pkgs_size_unit] = libdnf5::cli::utils::units::to_size(in_pkgs_size);
+    if (in_pkgs_size != 0 && std::in_range<int64_t>(in_pkgs_size) && std::in_range<int64_t>(download_pkgs_size)) {
```
jrohel, line 804: `std::in_range<int64_t>(in_pkgs_size)` - Maybe a bit unnecessary test. The transaction will not exceed 8 exabytes in the near future. And if it does, I suspect there will be more serious problems on other places. :-)

jrohel:
Thank you for the PR.

PR #1712, opened 2024-09-19
Do not install /var/cache/libdnf5 directory

ppisar:
Bootc systems tend to prune /var directory
<https://containers.github.io/bootc/filesystem.html#var>. That breaks verifying integrity with "rpm --verify libdnf5".

This patch stops installing /var/cache/libdnf5 directory, but keeps tracking it by libdn5 RPM package. The directory is created by libdnf5 whenever it is needed.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2313032

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -381,7 +381,7 @@ Package management library.
 %dir %{_libdir}/libdnf5
 %{_libdir}/libdnf5.so.2*
 %license lgpl-2.1.txt
-%{_var}/cache/libdnf5/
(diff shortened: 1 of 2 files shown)


ppisar:
Is this change suitable for upstream (make install), or should we restrict it to dnf5.spec ("rm %{buildroot}/%{_var}/cache/libdfn5" in %install section)?

m-blaha (approved):
I like it. There is no need to create cache directory during package build.

Issue #1685, opened 2024-09-09
dnf5 CLI does not handle emojis (and potentially other unicode symbols?) in %description


gotmax23:
The uv package's description in Fedora has emojis in it, which dnf5 does not handle correctly. See the difference between `dnf5 repoquery` and `dnf5 info`'s output:

![sudo dnf5 repoquery --qf='%{description}' uv properly renders emojis](https://github.com/user-attachments/assets/14f4cce1-3730-4065-a38d-27e9104eeab8)

![dnf5 info does not](https://github.com/user-attachments/assets/a1f6783c-4005-4371-be05-7f4de7da4b22)

``` console
$ dnf5 --version
dnf5 version 5.1.17
dnf5 plugin API version 1.0
libdnf5 version 5.1.17
libdnf5 plugin API version 1.0

Loaded dnf5 plugins:
  name: builddep
  version: 1.0.0
  API version: 1.0

  name: changelog
  version: 1.0.0
  API version: 1.0

  name: config-manager
  version: 0.1.0
  API version: 1.0

  name: copr
  version: 0.1.0
  API version: 1.0

  name: needs_restarting
  version: 1.0.0
  API version: 1.0

  name: repoclosure
  version: 1.0.0
  API version: 1.0
```

hroncok:
@musicinmybrain said in the Fedora Python matrix channel:

> I am not deeply committed to the emoji in the description of uv in particular, although all else equal, I default to preserving them since they are a very intentional part of upstream’s writing.
music
> I tend to fall on the side of thinking that emoji are part of Unicode support, and (subject to font availability) failure to display them is a bug. Emoji are not the only things that use codepoints outside the BMP or combine several codepoints into one glyph, but they are the most visible to English speakers. So I tend to assume that if emoji don’t work, something in CJK languages doesn’t work either.

musicinmybrain:
I just tried this on Fedora 40:

```
dnf5 info python3-fastapi
```

and got (excerpted):

```
               : The key features are:
               : 
               :   \xe2\x80\xa2 Fast: Very high performance, on par with NodeJS and Go (thanks to Starlette
               :     and Pydantic). One of the fastest Python frameworks available.
```

Now, these are merely Unicode bullets, `•`! So it looks like `dnf5` is treating everything as ASCII and backslash-escaping everything that isn’t ASCII.

The above works fine with `dnf` 4.21.0 and with `rpm -qi`.

musicinmybrain:
[...]

PR #1713, opened 2024-09-19
setlocale: If locale setting fails, try using C.UTF-8 as fallback

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1685


m-blaha (approved):
Thank you!

PR #1715, opened 2024-09-20
chore: static_cast to fix sign conversion warning

evan-goode:
Without this patch, the i686 build fails due to `-Werror` and a sign conversion warning, introduced in https://github.com/rpm-software-management/dnf5/pull/1703.

I only noticed the error after making an upstream release and trying to build DNF 5.2.6.1 in Koji. It might be good to have builds for non-x86_64 architectures tested in our upstream CI so we can catch these issues sooner.

### DIFF
--- a/libdnf5/rpm/package_query.cpp
+++ b/libdnf5/rpm/package_query.cpp
@@ -388,7 +388,7 @@ std::vector<std::vector<unsigned int>> kosaraju(const std::vector<std::vector<un
         }
 
         if (sccredges.empty()) {
-            std::vector scc(stack.begin() + s, stack.end());
+            std::vector scc(stack.begin() + static_cast<std::vector<unsigned int>::difference_type>(s), stack.end());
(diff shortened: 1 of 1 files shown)


jrohel:
> Without this patch, the i686 build fails due to -Werror and a sign conversion warning

On x86_64, `std::vector<some_type>::difference_type` is a 64 bit signed type. Thus it can hold a larger number than `unsigned int`, a 32 bit unsigned type.
But, on i686, `std::vector<some_type>::difference_type` is a 32 bit signed type. So the compiler correctly warned when adding `unsigned int` - a 32 bit unsigned type.

> It might be good to have builds for non-x86_64 architectures tested in our upstream CI.

Yes, good idea. Especially testing a 32 bit build.

We could also test builds for big-endian architecture. To catch endian related bugs. Fedora supports s390x - big-endian architecture.

jrohel:
Thanks for the fix.

PR #1722, opened 2024-09-23
comps: add get_base() to {Group,Environment}{,Query}

gotmax23:
This mirrors the similar classes in libdnf5::rpm that provide the same methods. This will help fedrq implement group querying functionality.

### DIFF
--- a/test/python3/libdnf5/comps/test_comps_iterators.py
+++ b/test/python3/libdnf5/comps/test_comps_iterators.py
@@ -24,8 +24,10 @@ import base_test_case
 class TestCompsIterators(base_test_case.BaseTestCase):
     def test_group_query_iterable(self):
         query = libdnf5.comps.GroupQuery(self.base)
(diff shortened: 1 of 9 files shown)


test/python3/libdnf5/comps/test_comps_iterators.py
```diff
@@ -24,8 +24,10 @@
 class TestCompsIterators(base_test_case.BaseTestCase):
     def test_group_query_iterable(self):
         query = libdnf5.comps.GroupQuery(self.base)
+        self.assertIsInstance(query.get_base().get(), libdnf5.base.Base)
         _ = [grp.get_groupid() for grp in query]
```
gotmax23, line 28: I wanted to test `get_base()` on the `Group`/`Environment` classes, as well, but this test is broken and doesn't actually contain anything in the query.

kontura:
Thank you!

PR #1725, opened 2024-09-24
copr: use pubkey URL returned by Copr API

FrostyX:
See https://github.com/fedora-copr/copr/issues/3375

When an alternative storage (such as Pulp) for Copr is used, the pubkey.gpg is still being stored on Copr backend. Therefore we can't always construct the correct pubkey URL here in the DNF plugin.


FrostyX:
Pinging @praiskup for a review.

praiskup (approved):

praiskup:
Thank you!

evan-goode (approved):
LGTM

Issue #1674, opened 2024-09-03
Comparison to DNF4, doc/changes.rst, is not packaged


ppisar:
A Fedora users searches for a documentation describing changes from DNF4 <https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/GHQWNVP3OGOPNKCYQRFFJFWELGXBOH5N/>.

We have doc/changes.rst.  But I cannot find that document packaged in RPM, preferably as a manual page.
Also please use a better name than "changes". E.g. dnf5-dnf4delta(7).

PR #1729, opened 2024-09-25
Package file documenting dnf4/dnf5 changes as man page

m-blaha:
Resolves: https://github.com/rpm-software-management/dnf5/issues/1674


evan-goode:
I would rather the document be called `changes-from-dnf4`. What do you think?

m-blaha:
> I would rather the document be called `changes-from-dnf4`. What do you think?

Yeah, that might be even clearer.

m-blaha:
The document is renamed to `changes-from-dnf4`.

ppisar (approved):
Good.

Issue #1734, opened 2024-09-27
Option to retreive download url of a pacakge from all available mirrors


alimirjamali:
With dnf4, our `downloadurl` plugin ([fork](https://github.com/QubesOS/qubes-core-agent-linux/blob/main/package-managers/dnf-plugin-downloadurl.py) of download plugin) allowed us to get urls for specific package from all mirrors (our own infrastructure). So we could shuffle them and use curl/wget/... to download the package via our own interactive CLI/GUI tools from multiple mirrors (They are multi-gigabyte VM templates). The command was `dnf downloadurl --url --all-mirrrors <pkg>`.

It appears that `downloadurl` plugin is not compatible with dnf5 anymore. Neither `dnf repoquery <pkg> --location` nor `dnf download <pkg> --url ` could do the same. Is there any workaround for dnf5?

alimirjamali:
Never mind. Some confusion here. It is a forked version of `download` plugin. Not an official one.

marmarek:
But maybe it still makes sense? What about adding `--all-mirrors` option to the `download` command? The way we have it in our `ownloadurl` plugin makes sense only for a single package (it prints all URLs but doesn't mark which one is for which package), but maybe still something that would be useful for others too?

alimirjamali:
[This line](https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/download/download.cpp#L276) in dnf5 download plugin is what that has to be changed/improved to allow `--all-mirrors` (to be used together with `--url` option). I am ready to work on a patch if DNF5 maintainers approve it.

alimirjamali:
Related: #497

alimirjamali:
> Would you print every URL on a separate line, or would you group URLs pertaining the same package on a single line, separating them with a white space?

For our current use case, we have been using the old DNF4 patched download plugin only for one package. So the grouping was not necessary. But grouping is doable and is very useful. I will amend the PR and apply the recommended change. Force-push it after testing.

PR #1735, opened 2024-09-28
Add --allmirros option for `dnf download --url`

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/1734

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -69,6 +69,9 @@ void DownloadCommand::set_argument_parser() {
     url_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
+    allmirrors_option = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+
     srpm_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
@@ -96,6 +99,12 @@ void DownloadCommand::set_argument_parser() {
     url->set_description("Print a URL where the rpms can be downloaded instead of downloading");
     url->set_const_value("true");
     url->link_value(url_option);
+
+    auto allmirrors = parser.add_new_named_arg("allmirrors");
+    allmirrors->set_long_name("allmirrors");
+    allmirrors->set_description(_("When running with --url, prints URLs from all available mirrors"));
+    allmirrors->set_const_value("true");
+    allmirrors->link_value(allmirrors_option);
 
     urlprotocol_valid_options = {"http", "https", "ftp", "file"};
     urlprotocol_option = {};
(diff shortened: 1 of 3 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -115,6 +118,12 @@ void DownloadCommand::set_argument_parser() {
             return true;
         });
 
+    auto urlallmirrors = parser.add_new_named_arg("all-mirrors");
```
ppisar, line 121: Could you please move this block just after `auto url = parser.add_new_named_arg("url");` block? So that it is handled in the same order as urlallmirrors_option is initialized just after url_option?

dnf5/commands/download/download.cpp
```diff
@@ -115,6 +118,12 @@ void DownloadCommand::set_argument_parser() {
...
+    urlallmirrors->set_long_name("all-mirrors");
+    urlallmirrors->set_description("When running with --url, prints URLs from all available mirrors");
```
ppisar, line 123: Could you mark this string for a translation with `_("When...")`?

dnf5/commands/download/download.cpp
```diff
@@ -273,7 +283,13 @@ void DownloadCommand::run() {
                 ctx.get_base().get_logger()->warning("Failed to get mirror for package: \"{}\"", pkg.get_name());
                 continue;
             }
-            std::cout << urls[0] << std::endl;
+            std::cout << urls[0];
+            if (urlallmirrors_option->get_value()) {
+                for (long unsigned index = 1; index < urls.size(); ++index) {
```
ppisar, line 288: Plese use "size_t" or "auto" instead of "long unsigned". The type should match a type returned by urls.size().

dnf5/commands/download/download.cpp
ppisar, line 121: I would use "allmirros" witout a hypen. `--urlprotocol` is also spelled without a hyphen.

dnf5/commands/download/download.cpp
```diff
@@ -97,6 +100,12 @@ void DownloadCommand::set_argument_parser() {
     url->set_const_value("true");
     url->link_value(url_option);
 
+    auto urlallmirrors = parser.add_new_named_arg("allmirrors");
+    urlallmirrors->set_long_name("allmirrors");
+    urlallmirrors->set_description _("When running with --url, prints URLs from all available mirrors");
```
ppisar, line 105: I think you forgot parentheses at set_description call: `urlallmirrors->set_description (_("When..."));`

ppisar:
Now, when you renamed the option, also please update a title of this commit.

alimirjamali:
> Now, when you renamed the option, also please update a title of this commit.

This is also done

dnf5/commands/download/download.hpp
```diff
@@ -47,6 +47,7 @@ class DownloadCommand : public Command {
     libdnf5::OptionBool * resolve_option{nullptr};
     libdnf5::OptionBool * alldeps_option{nullptr};
     libdnf5::OptionBool * url_option{nullptr};
+    libdnf5::OptionBool * urlallmirrors_option{nullptr};
```
ppisar, line 50: Other option variables are named after the options. Please rename `urlallmirrors_option' to `allmirrors_option'

alimirjamali:
@ppisar Fedora 39 unittests failed because of missing `libfmt.so`. Is this normal?

ppisar (approved):
Good. The F39 test failure is unrelated.

ppisar:
> The error looks more like a corrupted container.

I  reproduced it. The cause is condor-23.9.6-1.fc39.x86_64 package that RPM-provides libfmt.so.10()(64bit), but does not actually install that library. I reported a bug against condor <https://bugzilla.redhat.com/show_bug.cgi?id=2315859>.

PR #1741, opened 2024-10-01
Set `POOL_FLAG_ADDFILEPROVIDESFILTERED` only when not loading filelists

kontura:
When filelists are loaded and file provides are used it is much faster not to set `POOL_FLAG_ADDFILEPROVIDESFILTERED` because libsolv goes through all the files just once during `pool_addfileprovides_queue`.

This is based on: https://github.com/rpm-software-management/libdnf/issues/1674

Specifically this is an improvement for users that use repos with file provides outside of primary.xml and download filelsits.xml.

### DIFF
--- a/libdnf5/repo/repo_sack.cpp
+++ b/libdnf5/repo/repo_sack.cpp
@@ -27,6 +27,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "conf/config.h"
 #include "repo_cache_private.hpp"
 #include "repo_downloader.hpp"
+#include "solv/pool.hpp"
 #include "solv/solver.hpp"
 #include "solv_repo.hpp"
 #include "utils/auth.hpp"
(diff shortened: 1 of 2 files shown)


kontura:
While working on the backport to dnf4 I noticed some differences in performance, it turns out dnf5 is probably storing the filelists solv cache twice.

When metadata are downloaded with filelists the primary solv file also contains filelists. This can be visible when comparing dnf4 and dnf5 solv files from the same metadata. For example:
```
$ ls -lah /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora.solv /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora-filelists.solvx
-rw-r--r--. 1 root root 61M Oct  2 03:19 /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora-filelists.solvx
-rw-r--r--. 1 root root 87M Oct  2 03:19 /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora.solv

```
vs
```
$ ls -lah /var/cache/dnf/fedora.solv /var/cache/dnf/fedora-filenames.solvx
-rw-r--r--. 1 root root 61M Oct  2 02:30 /var/cache/dnf/fedora-filenames.solvx
-rw-r--r--. 1 root root 27M Oct  2 02:30 /var/cache/dnf/fedora.solv
```

I am looking into it.

m-blaha:
I did some tests and with this patch, the performance issue with `POOL_FLAG_ADDFILEPROVIDESFILTERED` is gone. Good work!

I suggest to track the primary solv file problem as a separate issue, since it's not caused by this patch, and the fix may turn out to be difficult. What do you think?

kontura:
> I did some tests and with this patch, the performance issue with `POOL_FLAG_ADDFILEPROVIDESFILTERED` is gone. Good work!
> 
> I suggest to track the primary solv file problem as a separate issue, since it's not caused by this patch, and the fix may turn out to be difficult. What do you think?

Ok, that is a good suggestion.

m-blaha (approved):
Thank you!

PR #1743, opened 2024-10-01
comps: Fix memory issues in group serialization

m-blaha:
This commit addresses two memory issues that occur if saving of the
group XML file fails:

Memory leak: The `doc` variable is not properly freed before an error is
thrown, leading to a memory leak.

Use after free: The libxml2 error handler continues to reference the
`error_to_strings()` function, which uses the `xml_errors` vector that
is local to the `Group::serialize()` method. If the handler is invoked
later (e.g., during the serialization of an Environment), it may cause a
crash due to accessing freed memory.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2315789

### DIFF
--- a/libdnf5/comps/group/group.cpp
+++ b/libdnf5/comps/group/group.cpp
@@ -323,7 +323,14 @@ void Group::serialize(const std::string & path) {
     }
 
     // Save the document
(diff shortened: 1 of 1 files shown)


jrohel:
Thanks.

Issue #1744, opened 2024-10-01
POT files are never updated causing dropping newly translated messages


ppisar:
As pointed in <https://github.com/rpm-software-management/dnf5/issues/1736#issuecomment-2383660596> comment, installed DNF5 is missing newly added messages for a translation. The cause is that we never update POT files here in DNF5 repository. As a result msgmerge run later during a build will merge the updated po files against the out-dated pot files, resulting in removing the messages from the translation catalogs.

We either need to amend .github/actions/weblate-pull-translations/action.yml script to copy the POT files from dnf5-l10n repository, or stop calling msgmerge at built time.

The latter option would simplify the process. But because we deliver the POT in source archive and code repository, I'd adopt the former option.

PR #1745, opened 2024-10-01
i18n: Update translation templates from Weblate

ppisar:
Translation template (*.pot) files were updated last time in year
2023. As a result msgmerge run during building DNF5 removed all newly added translated messages from message cataloges because the new messages were missing from the template files.

This patch fixes it by copying the templates together with messages catalogs from Weblate on every DNF5 release.

Resolves: #1744


ppisar:
@evan-goode, you do releases recently. Could you review this pull request?

evan-goode (approved):
Looks good, thank you!

Issue #839, opened 2023-08-25
Implement `--color` flag and respect `color` option


evan-goode:
Noticed by a participant in the 2023-08-11 test week: https://fedoraproject.org/wiki/Test_Day:2023-08-11_Fedora_39_DNF_5#cite_note-74.

1. For commands that produce colorful output, such as `info`, `search`, and `list`, colors are currently shown iff stdout is a terminal, which is the behavior of `--setopt=color=auto`. `color=always` and `color=never` are not respected.
2. The `--color=auto|never|always` argument to set the `color` configuration option is missing.

These items are probably low priority; `color=auto` is desired in the vast majority of cases. But some scripts may invoke `dnf --color=never ...` and it would be nice to not break them.

alimirjamali:
PR Submitted.
We specifically require this option for QubesOS

review priority: low

PR #1750, opened 2024-10-02
Reimplement `--color` flag

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/839
related: https://github.com/QubesOS/qubes-issues/issues/9482

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -63,6 +63,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5-cli/output/adapters/transaction.hpp>
 #include <libdnf5-cli/output/transaction_table.hpp>
 #include <libdnf5-cli/session.hpp>
+#include <libdnf5-cli/tty.hpp>
 #include <libdnf5-cli/utils/units.hpp>
 #include <libdnf5-cli/utils/userconfirm.hpp>
 #include <libdnf5/base/base.hpp>
@@ -171,6 +172,14 @@ void RootCommand::set_argument_parser() {
     cacheonly->set_const_value("all");
     cacheonly->link_value(&config.get_cacheonly_option());
     global_options_group->register_argument(cacheonly);
+
+    auto color = parser.add_new_named_arg("color");
+    color->set_long_name("color");
+    color->set_has_value(true);
+    color->set_arg_value_help("COLOR");
+    color->set_description(_("Control whether color is used."));
+    color->link_value(&config.get_color_option());
+    global_options_group->register_argument(color);
 
     auto refresh = parser.add_new_named_arg("refresh");
     refresh->set_long_name("refresh");
@@ -1394,6 +1403,17 @@ int main(int argc, char * argv[]) try {
             }
 
             base.setup();
+
+            auto & coloring = base.get_config().get_color_option().get_value();
+            if (coloring == "always") {
+                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::ALWAYS);
+            } else if (coloring == "never") {
+                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::NEVER);
+            } else if (libdnf5::cli::tty::is_interactive()) {
+                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::ALWAYS);
+            } else {
+                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::NEVER);
+            }
 
             print_no_match_libdnf_plugin_patterns(context);
 
--- a/doc/dnf5.8.rst
+++ b/doc/dnf5.8.rst
@@ -276,6 +276,9 @@ Following options are applicable in the general context for any ``dnf5`` command
 ``--refresh``
     | Force refreshing metadata before running the command.
 
+``--color=<color>``
+    | Control whether color is used in terminal output. Valid values are ``always``, ``never`` and ``auto`` (default).
+
 ``--repo=REPO_ID,...``
     | Enable just specified repositories.
     | This is a list option which can be specified multiple times.
--- a/include/libdnf5-cli/tty.hpp
+++ b/include/libdnf5-cli/tty.hpp
@@ -55,6 +55,11 @@ LIBDNF_CLI_API std::ostream & cursor_show(std::ostream & stream);
 LIBDNF_CLI_API int get_width();
 LIBDNF_CLI_API bool is_interactive();
 
+enum class ColoringEnabled { AUTO, ALWAYS, NEVER };
+
+LIBDNF_CLI_API void coloring_enable(ColoringEnabled);
+LIBDNF_CLI_API bool is_coloring_enabled();
+
 
 }  // namespace libdnf5::cli::tty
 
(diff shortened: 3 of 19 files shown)


alimirjamali:
The `fedora-39-*` unit-tests failure were discussed here:
https://github.com/rpm-software-management/dnf5/pull/1735#issuecomment-2385138020

I am not sure about other test failures with.

alimirjamali:
@jrohel Hi. I did not see that you have self-assigned yourself to the issue. Please let me know if I should rebase the PR and continue working on it, or possibly you are working on a better solution.

doc/dnf5.8.rst
```diff
@@ -276,6 +276,9 @@ Following options are applicable in the general context for any ``dnf5`` command
 ``--refresh``
     | Force refreshing metadata before running the command.
 
+``--color=<color>``
+    | Control whether color is used in terminal output. Valid values are ``always``, ``never`` and ``auto`` (default).
```
jrohel, line 280: Valid values are:
- `always`, `on`, `yes`, `1`, `true`
- `never`, `off`, `no`, `0`, `false`
- `auto`, `tty`, `if-tty`

So we are fully compatible with DNF4. It also accepts all the values listed. But the truth is that the DNF4 man page only lists `always`, `never`, and `auto`. I am considering whether we want to ignore the other possible values in the documentation again.

It may be practical to know about the possibility to write `--color=1` instead of `--color=always` and `--color=0` instead of `--color=never`.

libdnf5-cli/tty.cpp
```diff
@@ -67,6 +67,17 @@ bool is_interactive() {
     return isatty(fileno(stdout)) == 1;
 }
 
+bool colorized = true;  // Review notes: Maybe encapsulate this in a class
```
jrohel, line 70: The `colorized` variable is a private variable. It is not in the header file and is only accessed within that file. Please make it `static`.

include/libdnf5-cli/tty.hpp
```diff
@@ -54,6 +54,8 @@ LIBDNF_CLI_API std::ostream & cursor_show(std::ostream & stream);
 
 LIBDNF_CLI_API int get_width();
 LIBDNF_CLI_API bool is_interactive();
+LIBDNF_CLI_API void set_colorized(bool);
+LIBDNF_CLI_API bool is_colorized();
```
jrohel, line 58: I'm thinking of names. Actually the function does not return whether the tty is colored. It returns the user's preference of whether or not they want colors based on what they set with `set_colorized`. So on a colored `tty` it may return `false` and vice versa.

But I can't think of a better name... `set_use_colors`, `get_use_colors` ... I don't know

jrohel (changes requested):
Hi. Thanks for the PR. You can keep working. I did a review.

dnf5/main.cpp
```diff
@@ -171,6 +172,16 @@ void RootCommand::set_argument_parser() {
     cacheonly->link_value(&config.get_cacheonly_option());
     global_options_group->register_argument(cacheonly);
 
+    auto color = parser.add_new_named_arg("color");  // Review notes: --color is after --refresh in DNF4
+    color->set_long_name("color");
+    color->set_has_value(true);
+    color->set_arg_value_help("COLOR");
+    color->set_description(_(
+        "Control whether color is used."));  // Review notes: Some descriptions end in full-stop and some are not. Please advise
+    color->set_const_value("auto");
```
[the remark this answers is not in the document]
alimirjamali: Done

dnf5/main.cpp
```diff
@@ -1357,6 +1368,17 @@ int main(int argc, char * argv[]) try {
             }
         }
 
+        auto & colorized = base.get_config().get_color_option().get_value();
```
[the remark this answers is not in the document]
alimirjamali: Done

doc/dnf5.8.rst
alimirjamali: > So we are fully compatible with DNF4. It also accepts all the values listed. But the truth is that the DNF4 man page only lists `always`, `never`, and `auto`. I am considering whether we want to ignore the other possible values in the documentation again.

Yes. That was exactly what I did (copy/pasting from DNF4 man page). Please let me know how this should read and I will amend the PR accordingly

libdnf5-cli/tty.cpp
alimirjamali: done

alimirjamali:
> Hi. Thanks for the PR. You can keep working. I did a review.

Hi. Thanks for the review. I rebased the PR (to resolve a merge conflict and for testing purpose) and applied the advised changes. Awaiting CI/CD checks and further instructions for the remaining issues.

include/libdnf5-cli/tty.hpp
jrohel: And if we keep the convention with `is` we get:
`void coloring_enable(bool)` and `bool is_coloring_enabled()`

doc/dnf5.8.rst
jrohel: Frankly, I can't think of an appropriate description. Therefore, leave the documentation as you have written it.

jrohel:
Thank you for the changes. I realized we had a problem. PR changes the default coloring behavior from `auto` to `always`. This breaks other applications using libdnf5-cli that don't call `coloring_enable`. For example, dnf5daemon. 

I am looking for a simple solution. For example:
- Add to "include/libdnf5-cli/tty.hpp":
 `enum class ColoringEnabled { AUTO, ALWAYS, NEVER };`

- Change "libdnf5-cli/tty.cpp" to:
```
static ColoringEnabled coloring = ColoringEnabled::AUTO;

void coloring_enable(ColoringEnabled value) {
    coloring = value;
}

bool is_coloring_enabled() {
    return coloring == ColoringEnabled::AUTO ? is_interactive() : coloring == ColoringEnabled::ALWAYS;
}
```
- Change "dnf5-main.cpp" to:
```
            auto & coloring = base.get_config().get_color_option().get_value();
            if (coloring == "always") {
                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::ALWAYS);
            } else if (coloring == "never") {
                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::NEVER);
            } else if (libdnf5::cli::tty::is_interactive()) {
                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::ALWAYS);
            } else {
                libdnf5::cli::tty::coloring_enable(libdnf5::cli::tty::ColoringEnabled::NEVER);
            }
```

alimirjamali:
> I am looking for a simple solution. For example:

The suggested solution (with some minor changes) passed most of the CI/CD tests. Local tests also pass (two failed before applying the suggestions). Thanks for guidance.

BTW, I rebased before applying the suggestions. For the remaining pending issues:

1. Clang [error](https://github.com/rpm-software-management/dnf5/actions/runs/13969632312/job/39107947794?pr=1750#step:9:271) is unrated to this specific PR. But I can fix the bug in this PR if necessary.
2. To the best of my understanding, the failed test scenario (`dnf/plugins-core/expired-pgp-keys.feature`) should be solved after by either of PR 2145 or 2147 (hopefully).

jrohel:
Thank you for the PR. The failed tests are not due to this PR. I'm merging it.

jrohel (approved):

Issue #1736, opened 2024-09-30
Update POT files for dnf5?


bc-lee:
It seems that the POT file at https://github.com/rpm-software-management/dnf5/blob/5.2.6.2/dnf5/po/dnf5.pot was last updated on 2023-11-21, which is more than 10 months ago. When I run `$ ninja dnf5-pot` in the dnf5 source repository, a surprisingly large number of strings are updated. This implies that current translations are referring to outdated strings, meaning that even though translation contributors might think their work is 100% complete, many of the texts are actually not translated.

Here’s the output of `$ git diff --stat dnf5/po` over generated POT file changes:

```
$ git diff --stat dnf5/po
 dnf5/po/dnf5.pot | 1015 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---
 1 file changed, 977 insertions(+), 38 deletions(-)
```

I believe we need to update the POT files for dnf5, and it might be beneficial to update them automatically (e.g., every month) to avoid this situation.

The image below shows the result of `$ dnf --help` in Korean, where many strings are not translated. (dnf5-5.2.6.2-1.fc41.x86_64):

![스크린샷 2024-09-30 13-56-16](https://github.com/user-attachments/assets/d01d213a-3aaa-49e7-80d3-68af8e196f62)

ppisar:
*.pot files, like dnf5/po/dnf5.pot, are not updated here. They are instead updated in <https://github.com/rpm-software-management/dnf5-l10n> repository every day. The translations are then copied back just before every DNF5 release. It's so not to clutter commit history of this repository with updates from translators. Another reason is to prevent Weblate service from running into merge conflicts. But I agree that it's confusing.

Another question is why the texts you showed in the image are not localized. It's because the texts are not marked for a translation in the code. That needs to be fixed.

bc-lee:
Thanks for the clarification. It turns out that dnf5-i10n's dnf5.pot files are updated regularly, so I don't need to worry about them. I should have checked that first.

bc-lee:
I still need to understand why translated entries are not appearing as translated in dnf5.

When I run `ninja`, the following command is executed:

```
$ ninja -v
[...]

PR #1751, opened 2024-10-02
Make most descriptions for `dnf5 --help` translatable.

bc-lee:
Use `_(msgId)` in `bgettext-lib.h` to mark text as eligible for
translation. This will allow translators to translate the short
descriptions of the commands in `dnf5 --help`. Further changes will be
required to make each command description translatable.

Related: #1736

### DIFF
--- a/dnf5-plugins/config-manager_plugin/config-manager.cpp
+++ b/dnf5-plugins/config-manager_plugin/config-manager.cpp
@@ -25,6 +25,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "unsetopt.hpp"
 #include "unsetvar.hpp"
 
+#include <libdnf5/utils/bgettext/bgettext-lib.h>
+
 namespace dnf5 {
 
 using namespace libdnf5::cli;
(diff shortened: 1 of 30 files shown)


ppisar:
The patch looks good. It seems you tried to mark all short description which are visible in `dnf5 --help` output. However, some commands are still missing. E.g.:

builddep
changelog
clean

Lee, are going to correct your current pull request? Or should I simply merge it as it is and fix the missing descriptions later?

bc-lee:
> Lee, are going to correct your current pull request? Or should I simply merge it as it is and fix the missing descriptions later?

I prefer an incremental approach since it would take much longer if we wanted to finish all of the texts related to localization in this PR, and I don't think these changes are necessarily dependent on other PRs.

ppisar:
Then please change the commit title. E.g. replace "each command"  with "most of the commands".

bc-lee:
> Then please change the commit title. E.g. replace "each command" with "most of the commands".

Done. Could reviewers please check the title and the commit message?

ppisar:
You have changed a title of this pull request. I meant changing a title of the git commit message. That's what counts because it will be forever in the git log. Please update the git commit.

bc-lee:
> You have changed a title of this pull request. I meant changing a title of the git commit message. That's what counts because it will be forever in the git log. Please update the git commit.

Before applying the changes to the commit messages, I intentionally updated the title and description of this PR. This is because: 1) Many GitHub-based projects amend multiple commits in a PR, so each commit message is not as important as the PR title and description, and 2) Changing commit messages requires a force push, which isn't recommended for all projects, and I'm not sure if this project (`dnf5`) allows it.

Nonetheless, I have updated the commit messages as well. Please take a look.

Issue #1749, opened 2024-10-02
Filelists are stored twice in solv file cache


kontura:
When metadata are downloaded with filelists the primary solv file also contains filelists. This can be visible when comparing dnf4 and dnf5 solv file sizes from the same metadata. For example:
```
$ ls -lah /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora.solv /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora-filelists.solvx
-rw-r--r--. 1 root root 61M Oct  2 03:19 /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora-filelists.solvx
-rw-r--r--. 1 root root 87M Oct  2 03:19 /var/cache/libdnf5/fedora-845d89688cb28f31/solv/fedora.solv

```
vs
```
$ ls -lah /var/cache/dnf/fedora.solv /var/cache/dnf/fedora-filenames.solvx
-rw-r--r--. 1 root root 61M Oct  2 02:30 /var/cache/dnf/fedora-filenames.solvx
-rw-r--r--. 1 root root 27M Oct  2 02:30 /var/cache/dnf/fedora.solv
```

_Originally posted by @kontura in https://github.com/rpm-software-management/dnf5/issues/1741#issuecomment-2387798588_

PR #1752, opened 2024-10-03
When writing main solv file (primary.xml) don't store filelists

kontura:
In a repo filelists are stored in different `repodata`, limit the written `repodata` when writing main solv file.

Otherwise the filelists are stored twice: in the main solv file and in the filelists specific solvx file. (This happens when downloading both the primary.xml and filelists.xml at the same time.)

It also improves performance because when filelists were part of main solv file they were always loaded (regardless of the `optional_metadata_types` option).

Closes: https://github.com/rpm-software-management/dnf5/issues/1749


m-blaha (approved):
Thank you! The change looks reasonable and according to my testing it works fine. Unfortunately the parts of libsolv this relies on is basically undocumented, so fingers crossed there won't be any unexpected negative side effects.

PR #1757, opened 2024-10-04
daemon: API to reset the session.base instance

m-blaha:
This allows dnf5daemon users to reset their session and re-use it for another transaction.

For: https://github.com/rpm-software-management/dnf5/issues/1653


dnf5daemon-server/services/goal/goal.cpp
```diff
@@ -283,6 +283,14 @@ sdbus::MethodReply Goal::do_transaction(sdbus::MethodCall & call) {
     if (!session.check_authorization(dnfdaemon::POLKIT_EXECUTE_RPM_TRANSACTION, call.getSender())) {
         throw std::runtime_error("Not authorized");
     }
+
+    auto & transaction_mutex = session.get_transaction_mutex();
```
jan-kolarik, line 287: It seems we need to cover more places with the mutex, what about `Session::set_transaction(transaction)` or maybe rather the callers of this method?

jan-kolarik (commented):
Thanks for nicely separating the work into commits! BTW, it might be helpful to add a test case for resetting the session and running a new transaction afterward. What do you think?

jan-kolarik (approved):
The unit test implementation is deferred to a separate [issue](https://github.com/rpm-software-management/dnf5/issues/1854) as it's not easy and straightforward with the daemon testing...

PR #1762, opened 2024-10-07
test: Normalize Python code

ppisar:
Broken in 00615105e commit ("test python3: improve comps tests").


kontura:
Thanks, that is my bad.

PR #1765, opened 2024-10-09
chore: Clean up Fedora 37-related conditionals in RPM spec

bc-lee:
Fedora 37 was end-of-life on 2023-12-05.


pkratoch:
Hi, thanks for the patch. I think the correct condition should be `%if 0%{?fedora} || 0%{?rhel} > 10`, though, because otherwise it would get resolved to false on Fedora.

pkratoch:
Thanks! Please squash the commits and I will merge it.

Issue #1649, opened 2024-08-22
Could we add `pkg_gpgcheck` as the supported option and make `gpgcheck` an alias to that (for now)?


kontura:
Could we add `pkg_gpgcheck` as the supported option and make `gpgcheck` an alias to that (for now)? That would let us make a step toward #727...

_Originally posted by @Conan-Kudo in https://github.com/rpm-software-management/dnf5/issues/1648#issuecomment-2304404102_

jan-kolarik:
IIUC, this doesn't change any existing behavior. If that's the case, I have no issue with it.

Issue #865, opened 2023-09-04
Document deprecation for API functions


inknos:
Currently, DNF5 does not have a clear policy on how to deprecate public API functions. We should document the process and be clear on how to intervene in such instances.
Take [this](https://github.com/rpm-software-management/dnf5/pull/813/files#diff-cec2e1948707b6302af4658ebfc54a0050e948f35fd3c2cd86e838f32583c0b4) commit, which generated the discussion, as an example.

- [ ] Firstly, considering that DNF5's API is stable, we should remind in the documentation the deprecation steps that we would follow. As an example, we could follow Fedora release steps, the deprecation process should be documented e.g., 
    1. Announcing deprecation during one fedora release
    2. Removing the function on the next fedora, or something like that.
- [ ] document how deprecation is handled as warnings in the codebase. There are three steps that should be targeted:
    - [ ] build time warning: how should warnings be handled and how noisy should the deprecation warnings be at build time
    - [ ] runtime warning: how should warnings be printed when using the C++ API or the Python API at runtime 
    - [ ] documentation warning: currently, warnings are just _a word_ in the docs, they are not visible enough. See [here](https://dnf5.readthedocs.io/en/latest/api/python/libdnf5_base_base.html#libdnf5.base.Base.load_config_from_file). This should be automated and generally improved to avoid confusion.

jan-kolarik:
The current approach to handling deprecated methods has been reported as insufficient by our API clients. Before implementing any changes, we should review existing standards to increase the likelihood of detecting issues using the existing CI checkers deployed by our dependents.

jan-kolarik:
I think it’s time to decide on an approach here. Following @inknos’s proposal, here are my suggestions:

### :clock10: Deprecation timeline documentation

Align with Fedora release cycles: mark objects or methods as deprecated at the start of a release and remove them in the subsequent release. Document this approach in our documentation.

### :loudspeaker: Deprecation signaling implementation

**Documentation visibility**

Mark deprecated elements in docstrings with `@deprecated` and apply a custom style to make this tag more visible in upstream docs.

Optionally, create a dedicated deprecation list section.

**Build-time warnings for C++ users**

Use `[[deprecated]]`, which has been standard since C++14.

**Runtime user notifications**

Define e.g. a `LIBDNF_DEPRECATED` macro wrapper to log a warning to stderr whenever a deprecated method is called.
**Question:** Should this also apply to C++ API users, even though they already receive a build-time warning?

[...]

PR #1766, opened 2024-10-09
Change `gpgcheck` option to `pkg_gpgcheck` but stay compatible

kontura:
Changes the name to a more apt `pkg_gpgcheck` but `gpgcheck` is still accepted. At this point I am not sure we will ever be able to drop it.

It adds new getter API to both main and repo: `get_pkg_gpgcheck_option` and depracates the old `get_gpgcheck_option`.
It also extends `libdnf5::cli::output::IRepoInfo` with `get_pkg_gpgcheck` and the `dnf5daemon-server` should also accept `pkg_gpgcheck` attribute.

For: https://github.com/rpm-software-management/dnf5/issues/1649

### DIFF
--- a/dnf5daemon-server/services/repo/repo.cpp
+++ b/dnf5daemon-server/services/repo/repo.cpp
@@ -58,7 +58,7 @@ enum class RepoAttribute {
     skip_if_unavailable,
 
     gpgkey,
-    gpgcheck,
+    pkg_gpgcheck,
     repo_gpgcheck,
 
     proxy,
@@ -106,7 +106,8 @@ const static std::map<std::string, RepoAttribute> repo_attributes{
     {"skip_if_unavailable", RepoAttribute::skip_if_unavailable},
 
     {"gpgkey", RepoAttribute::gpgkey},
-    {"gpgcheck", RepoAttribute::gpgcheck},
+    {"gpgcheck", RepoAttribute::pkg_gpgcheck},
+    {"pkg_gpgcheck", RepoAttribute::pkg_gpgcheck},
     {"repo_gpgcheck", RepoAttribute::repo_gpgcheck},
 
     {"proxy", RepoAttribute::proxy},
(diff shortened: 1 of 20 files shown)


libdnf5-cli/output/repo_info.cpp
```diff
@@ -252,7 +252,7 @@ void print_repoinfo_json([[maybe_unused]] const std::vector<std::unique_ptr<IRep
         json_object_object_add(json_repo, "gpg_key", json_gpg_keys);
 
         json_object_object_add(json_repo, "repo_gpgcheck", json_object_new_boolean(repo->get_repo_gpgcheck()));
-        json_object_object_add(json_repo, "gpgcheck", json_object_new_boolean(repo->get_gpgcheck()));
+        json_object_object_add(json_repo, "pkg_gpgcheck", json_object_new_boolean(repo->get_pkg_gpgcheck()));
```
kontura, line 255: This change is breaking CI tests which can be fixed easily [enough](https://github.com/rpm-software-management/ci-dnf-stack/pull/1565) but we should be careful about changing the `json` output.

dnf5daemon-server/services/repo/repo.cpp
```diff
@@ -58,7 +58,7 @@ enum class RepoAttribute {
     skip_if_unavailable,
 
     gpgkey,
-    gpgcheck,
+    pkg_gpgcheck,
```
jan-kolarik, line 61: Nice, thanks for taking care of daemon too!

include/libdnf5/conf/config_main.hpp
```diff
@@ -244,8 +244,12 @@ class LIBDNF_API ConfigMain : public Config {
     const OptionString & get_username_option() const;
     OptionString & get_password_option();
     const OptionString & get_password_option() const;
+    /// @deprecated It is going to be removed, use get_pkg_gpgcheck_option()
```
jan-kolarik, line 247: This reminds me, we should already discuss how to consistently handle the deprecated options in the CLI and API (https://github.com/rpm-software-management/dnf5/issues/865). I’ll add it to the agenda for some of our next dev meetings so we can agree on this together.

libdnf5-cli/output/repo_info.cpp
jan-kolarik: I agree, still I see this machine-readable output functionality new enough to do such changes.

test/libdnf5/base/test_transaction.cpp
```diff
@@ -45,7 +45,7 @@ void BaseTransactionTest::test_check_gpg_signatures_no_gpgcheck() {
 void BaseTransactionTest::test_check_gpg_signatures_fail() {
     add_repo_repomd("repomd-repo1");
 
-    base.get_config().get_gpgcheck_option().set(true);
+    base.get_config().get_pkg_gpgcheck_option().set(true);
```
jan-kolarik, line 48: Do we have at least some simple test checking that both options behave in the same way? Either a CI or a unit test ...

jan-kolarik (approved):

jan-kolarik:
Thanks!

Conan-Kudo:
> At this point I am not sure we will ever be able to drop it.

Eventually `gpgcheck=` meaning will be controlled by some kind of policy setting. For now, it doesn't change anything.

Issue #1717, opened 2024-09-20
dnf5 with --installroot does not log rpm transaction messages


m-blaha:
During the rpm transaction execution, dnf does not write rpm messages to log when the `--installroot` option is used.  The logging stops after packages verification and starts again once the transaction finishes.

This is part of dnf.log for command `dnf install acpi --installroot=/tmp/INSTALLROOT --use-host-config`:
```
2024-09-20T10:30:35+0000 [714356] INFO --- DNF5 launched with arguments: "dnf install acpi --installroot=/tmp/INSTALLROOT" ---
------------- 8<------------------
2024-09-20T10:30:53+0000 [714413] DEBUG [librepo] lr_download: Downloading started
2024-09-20T10:30:53+0000 [714413] DEBUG [librepo] Transfer finished: Packages/a/acpi-1.7-21.fc39.x86_64.rpm (Effective url: https://mirror.karneval.cz/pub/linux/fedora/linux/releases/39/Everything/x86_64/os/Packages/a/acpi-1.7-21.fc39.x86_64.rpm)
2024-09-20T10:30:53+0000 [714413] DEBUG [librepo] check_finished_transfer_checksum: Checksum (sha256) 2a7b32f1b42732dc50529e6005a895602f886221951fbe97b378d4bfb6e58efe is OK
2024-09-20T10:30:53+0000 [714413] DEBUG [librepo] lr_download_packages: Restoring an old SIGINT handler
2024-09-20T10:30:53+0000 [714413] INFO RPM callback verify start, total 1
2024-09-20T10:30:53+0000 [714413] INFO RPM callback open file "/tmp/INSTALLROOT/var/cache/libdnf5/fedora-845d89688cb28f31/packages/acpi-1.7-21.fc39.x86_64.rpm"
2024-09-20T10:30:53+0000 [714413] INFO RPM callback close file
2024-09-20T10:30:53+0000 [714413] INFO RPM callback verify stop, total 1
2024-09-20T10:30:53+0000 [714413] INFO RPM callback open file "/tmp/INSTALLROOT/var/cache/libdnf5/fedora-845d89688cb28f31/packages/acpi-1.7-21.fc39.x86_64.rpm"
2024-09-20T10:30:53+0000 [714413] INFO RPM callback close file
2024-09-20T10:30:53+0000 [714413] INFO RPM callback verify start, total 1
[...]

PR #1772, opened 2024-10-11
log: Preserve log messages during RPM transaction

m-blaha:
There is an issue with RotatingFileLogger where log messages during an
RPM transaction in installroot fail to be recorded. When RPM enters a
chroot environment during the transaction, the path resolution changes,
and RotatingFileLogger attempts to reopen the log file from its disk
path for each message. However, once inside the chroot, the log file
path can no longer be resolved, leading to a failure in logging until
the RPM transaction completes.

This patch modifies the RotatingFileLogger behavior to keep the log file
descriptor open, ensuring continuous logging throughout the RPM
transaction.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1717

### DIFF
--- a/libdnf5/logger/rotating_file_logger.cpp
+++ b/libdnf5/logger/rotating_file_logger.cpp
@@ -25,26 +25,33 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <fmt/format.h>
 #include <string.h>
 #include <sys/file.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


jrohel (approved):
There is still a problem if there is a nested installroot with the same path in installroot.

Example:
`dnf5 --installroot=/containers/test upgrade` 
and there exists `/containers/test/containers/test/var/log/` (nested installroot). Messages from rpm callbacks will be logged to the nested installroot.

But in general PR improves the situation.

PR #1778, opened 2024-10-17
Optimize getting counts of transaction items

kontura:
Previously during `history list` we separately fetched all transaction packages, groups and environments only to count them. This could be quite slow because it resulted in a lot of separate DB connections.

Now we fetch only the counts and use only one DB connection. To maintain compatibility the counts are passed as an optional parameter and if missing we fallback to the old (slow) computation.

### DIFF
--- a/include/libdnf5-cli/output/transactionlist.hpp
+++ b/include/libdnf5-cli/output/transactionlist.hpp
@@ -27,7 +27,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 namespace libdnf5::cli::output {
 
-LIBDNF_CLI_API void print_transaction_list(std::vector<libdnf5::transaction::Transaction> & ts_list);
+LIBDNF_CLI_API void print_transaction_list(
+    std::vector<libdnf5::transaction::Transaction> & ts_list,
+    const std::unordered_map<int64_t, int64_t> id_to_item_count = {});
 
 }  // namespace libdnf5::cli::output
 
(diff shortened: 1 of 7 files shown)


jan-kolarik:
Related BZ ticket: https://bugzilla.redhat.com/show_bug.cgi?id=2315665.

include/libdnf5-cli/output/transactionlist.hpp
```diff
@@ -27,7 +27,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 namespace libdnf5::cli::output {
 
-LIBDNF_CLI_API void print_transaction_list(std::vector<libdnf5::transaction::Transaction> & ts_list);
+LIBDNF_CLI_API void print_transaction_list(
```
jan-kolarik, line 30: I think this will break ABI compatibility. Could we possibly load the items inside this function? So like moving `history.get_transaction_item_counts(transactions);` inside this without modifying the API?

libdnf5-cli/output/transactionlist.cpp
```diff
@@ -49,11 +51,15 @@ void print_transaction_list(std::vector<libdnf5::transaction::Transaction> & ts_
...
-        scols_line_set_data(
-            ln,
-            4,
-            std::to_string(ts.get_packages().size() + ts.get_comps_groups().size() + ts.get_comps_environments().size())
-                .c_str());
+        std::string trans_item_count;
+        if (id_to_item_count.at(ts.get_id())) {
+            trans_item_count = std::to_string(id_to_item_count.at(ts.get_id()));
+        } else {
+            // Fallback to manual calculation (very slow if the packages, groups and environments have to be fetched from DB)
```
jan-kolarik, line 58: Is it even necessary to have a fallback here? Couldn't we just call `TransactionHistory::get_transaction_item_counts` before printing the transaction list here and that's it?

kontura: Updated.

jan-kolarik (approved):
Thank you!

PR #1788, opened 2024-10-21
Drop `errorlevel` config option

kontura:
It isn't doing anything and it was deprecated in dnf < 5 probably since dnf was first created.

### DIFF
--- a/include/libdnf5/conf/config_main.hpp
+++ b/include/libdnf5/conf/config_main.hpp
@@ -44,8 +44,6 @@ public:
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
-    OptionNumber<std::int32_t> & get_errorlevel_option();
-    const OptionNumber<std::int32_t> & get_errorlevel_option() const;
     OptionPath & get_installroot_option();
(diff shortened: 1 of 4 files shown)


include/libdnf5/conf/config_main.hpp
```diff
@@ -44,8 +44,6 @@ class LIBDNF_API ConfigMain : public Config {
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
-    OptionNumber<std::int32_t> & get_errorlevel_option();
```
ppisar, line 47: This breaks API. You should bump a major version of the project and a soname of libdnf.

kontura: Indeed, I would probably do that together with some more significant changes.
If you agree, I made a PR to revert the breakage: https://github.com/rpm-software-management/dnf5/pull/1793.

pkratoch: Oops, that didn't occur to me. Thanks for noticing that, Petr.

ppisar: That's also an option.

PR #1793, opened 2024-10-22
Revert "Drop `errorlevel` config option"

kontura:
Keep the defunct options to preserve API.
 
Revert: https://github.com/rpm-software-management/dnf5/pull/1788

It keeps the docs update.

### DIFF
--- a/include/libdnf5/conf/config_main.hpp
+++ b/include/libdnf5/conf/config_main.hpp
@@ -44,6 +44,12 @@ public:
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
+    /// @deprecated errorlevel option does nothing and is scheduled for removal
+    [[deprecated("errorlevel option does nothing and is scheduled for removal")]] OptionNumber<std::int32_t> &
+    get_errorlevel_option();
+    /// @deprecated errorlevel option does nothing and is scheduled for removal
+    [[deprecated("errorlevel option does nothing and is scheduled for removal")]] const OptionNumber<std::int32_t> &
+    get_errorlevel_option() const;
     OptionPath & get_installroot_option();
     const OptionPath & get_installroot_option() const;
     OptionPath & get_config_file_path_option();
--- a/libdnf5/conf/config_main.cpp
+++ b/libdnf5/conf/config_main.cpp
@@ -113,6 +113,7 @@ class ConfigMain::Impl {
     Config & owner;
 
     OptionNumber<std::int32_t> debuglevel{2, 0, 10};
+    OptionNumber<std::int32_t> errorlevel{3, 0, 10};
     OptionPath installroot{"/", false, true};
     OptionBool use_host_config{false};
     OptionPath config_file_path{CONF_FILENAME};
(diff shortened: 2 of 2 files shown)


include/libdnf5/conf/config_main.hpp
```diff
@@ -44,6 +44,8 @@ class LIBDNF_API ConfigMain : public Config {
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
+    OptionNumber<std::int32_t> & get_errorlevel_option();
```
ppisar, line 47: Could you mark these functions as deprecated? I don't know if we have a dedicated document for the deprecations or a macro, but marking the prototypes with `__attribute((deprecated("errorlevel option does nothing and is scheduled for removal")))` would be great. This makes people using these methods to see a compile-time warning.

libdnf5/conf/config_main.cpp
```diff
@@ -113,6 +113,7 @@ class ConfigMain::Impl {
     Config & owner;
 
     OptionNumber<std::int32_t> debuglevel{2, 0, 10};
+    OptionNumber<std::int32_t> errorlevel{3, 0, 10};
```
ppisar, line 116: This could be removed if get_error_level() accessors were were implemented as a dummy `return 0`.

ppisar (commented):
I recommend renaming this commit to "Return ConfigMain::get_errorlevel_option API". Otherwise it could evoke that you reverts the whole commit.

libdnf5/conf/config_main.cpp
kontura: I don't know how to easily do that.
The getters return a reference and I don't think those have a null value.

I have kept it for now.

include/libdnf5/conf/config_main.hpp
kontura: That sounds good, I have also added a `@deprecated` since this is already used in the file.
Thank you.

kontura:
The required formatting seems a bit odd but I have updated it.

I have also renamed the commit. :+1:

include/libdnf5/conf/config_main.hpp
```diff
@@ -44,6 +44,14 @@ class LIBDNF_API ConfigMain : public Config {
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
+    /// @deprecated errorlevel option does nothing and is scheduled for removal
+    __attribute((deprecated("errorlevel option does nothing and is scheduled for removal")))
```
ppisar, line 48: It should be `__attribute__((...))`. Sorry for providing a wrong example. See GCC documentation [https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-deprecated-function-attribute].

kontura: I had to use different syntax, I couldn't convince SWIG to accept it.

ppisar (approved):
The patch looks good.

I compared list of symbols and there is no change comparing to a commit before cc8a0f95aa3d5cfb8e92dd1e9553e2ecc8ed4aad.

I also verified that using the get_errorlevel_option() method emits a warning.

The rawhide failures (duplicate --ruby option) are unrelated (#1794).

PR #1804, opened 2024-10-28
SWIG bindings and unit tests for `Vars::detect_release()` 

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1803


gotmax23:
fedrq [has](https://git.sr.ht/~gotmax23/fedrq/tree/48c5b412501fb143d91f6b7ffb5b8aafbda4929c/item/src/fedrq/backends/libdnf5/backend/__init__.py#L948-955) the following which uses the original version of this that @j-mracek added as a "raw prototype" after discussion in https://github.com/rpm-software-management/dnf5/issues/281.

``` python
def get_releasever() -> str:
    """
    Return the system releasever
    """
    # libdnf5 >= 5.0.10
    # https://github.com/rpm-software-management/dnf5/pull/448
    base = libdnf5.base.Base()
    return libdnf5.conf.Vars.detect_release(base.get_weak_ptr(), "/").get()
```

Returning a proper string instead of a wrapper type is definitely nicer but does represent a breaking change.

Although, with `python3-libdnf5-5.1.17-2.fc40.x86_64`, it seems that is now broken and returns an empty string when it used to return the proper releasever.

jrohel:
> Returning a proper string instead of a wrapper type is definitely nicer

Returning a proper string is nicer.
And, I tested the wrong version. It really returns a wrapped `std::unique_ptr<std::string>`.  And `get()` can be used to get the string. The problem is that it doesn't distinguish between `None` and an empty string, which was probably the original intent of the unpleasant `std::unique_ptr<string>` construct.

>  but does represent a breaking change.

But as you write. It's breaking change.
I'm considering whether to close this PR... Anyway, it complicates my further work.

PR #1807, opened 2024-10-30
Fix parsing of offline transaction JSON file

m-blaha:
Initialize the action, nevra, reason, repo_id, group_id, and package_path
fields to empty strings for each package, when parsing transaction JSON
file. Without this initialization, these fields may retain values
from the previous loop iteration, leading to an incorrect transaction.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2322527


ppisar:
Isn't the same bug few lines above when parsing "groups" and when parsing "environments"?

m-blaha:
Oh, good catch. I'll add it to the fix.

PR #1814, opened 2024-10-31
DownloadCallbacks: Ensure `end` for every successful `add_new_download`

jrohel:
For every successful (ended without exception) call of `add_new_download` callback, the `end` callback is called once.

The code relied on librepo to call the `end` callback just once for each `lr_target`. Which works under normal conditions. But if there is an interruption in the librepo download (error, signal received) librepo will not call `end` callbacks for targets that have not yet started downloading. However, the `add_new_download` callback has already been called for them. For example, if a higher layer allocated resources on `add_new_download` and deallocated on `end`, resources would be lost. 

The new code tracks this and calls the appropriate `end` callbacks. Also, an assert has been added to monitor for any attempt to double call the `end` callback.

### DIFF
--- a/libdnf5/repo/file_downloader.cpp
+++ b/libdnf5/repo/file_downloader.cpp
@@ -61,12 +61,21 @@ public:
           destination(destination),
           user_data(user_data) {}
 
+    ~FileTarget() {
+        if (need_call_end_callback) {
+            if (auto * download_callbacks = base->get_download_callbacks()) {
+                download_callbacks->end(user_cb_data, DownloadCallbacks::TransferStatus::ERROR, "Interrupted");
+            }
+        }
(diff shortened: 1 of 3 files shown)


kontura:
Since the updated unit tests pass even without the `file`/`package_downloader` changes can I ask when is the problem happening?

jrohel:
> Since the updated unit tests pass even without the `file`/`package_downloader` changes can I ask when is the problem happening?

I have added a description of the resolved problem to the commit and PR.
The added unit tests pass with the original code, because under normal conditions librepo will call the `end` callback. The problem is when the librepo download is interrupted (error, signal).

kontura (approved):
Great, thank you!

PR #1825, opened 2024-11-02
MultiProgressBar now buffers the output text to a single write

### LINKED ISSUE #1263: The download progress of multiple packages is flickering and not smooth
The progress always flickers which is causing eye pain. This is not an issue in dnf4. In dnf4, the update in progress is quite seamless.

Giedriusj1:
This is to avoid multiple small writes to the terminal, which can cause flickering.

On my machine (Fedora 41 GNOME) I can see that the new concurrent download status bar flickers quite a bit. I've tested this on gnome-terminal/xfce4-terminal/uterm and kitty. Only Kitty does not flicker. The flickering happens regardless if Wayland/X11 is used.

The old behaviour would issue multiple clear_line writes, followed by multiple writes to "draw" the new status bars.

The new behaviour buffers all of the screen updates for a single "frame" and then writes all of that out with a single syscall.

I've tested the code changes using gnome-terminal/xfce4-terminal/uterm and kitty. Wayland and X11 sessions. It works flawlessly for me now.

I've also tested resizing the terminal window during download, and that works as before (w/o flickering now)

I believe this also fixes #1263

### DIFF
--- a/libdnf5-cli/progressbar/multi_progress_bar.cpp
+++ b/libdnf5-cli/progressbar/multi_progress_bar.cpp
@@ -98,15 +98,22 @@ std::size_t MultiProgressBar::get_total_num_of_bars() const noexcept {
 std::ostream & operator<<(std::ostream & stream, MultiProgressBar & mbar) {
     const bool is_interactive{tty::is_interactive()};
 
+    // We'll buffer the output text to a single string and print it all at once.
+    // This is to avoid multiple writes to the terminal, which can cause flickering.
+    static std::ostringstream text_buffer;
+    text_buffer.str("");
+    text_buffer.clear();
+
     if (is_interactive && mbar.num_of_lines_to_clear > 0) {
-        stream << tty::clear_line;
+        text_buffer << tty::clear_line;
         for (std::size_t i = 1; i < mbar.num_of_lines_to_clear; i++) {
-            stream << tty::cursor_up << tty::clear_line;
+            text_buffer << tty::cursor_up << tty::clear_line;
         }
-        stream << "\r";
+        text_buffer << "\r";
     } else if (mbar.line_printed) {
-        stream << std::endl;
+        text_buffer << std::endl;
     }
+
     mbar.num_of_lines_to_clear = 0;
     mbar.line_printed = false;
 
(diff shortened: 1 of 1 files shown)


psimonyi:
I was just looking at the same problem and want to suggest also simplifying the command sequences (around line 102).

Currently it repeats the cursor-up-one, clear-line control sequences for as many lines as need to be cleared. But the control sequences can tell the terminal how much to move/clear in a single command. No loop required; just fill in *n*:

- ESC `[` *n* `A` — move cursor up *n* lines
- `\r` — move cursor to column 1
- ESC `[J` — clear screen from cursor to the end

Giedriusj1:
@psimonyi thanks for your feedback!

I've implemented your suggestions. In fact, I was able to remove the clear line tty commands altogether!

I figured that it was enough to move the cursor up, and simply let the next progress update write over the previous terminal contents.

I've tested this with multiple terminals while resizing them during download and it all works beautifully :smile:

libdnf5-cli/progressbar/multi_progress_bar.cpp
```diff
@@ -105,10 +105,10 @@ std::ostream & operator<<(std::ostream & stream, MultiProgressBar & mbar) {
     text_buffer.clear();
 
     if (is_interactive && mbar.num_of_lines_to_clear > 0) {
-        text_buffer << tty::clear_line;
-        for (std::size_t i = 1; i < mbar.num_of_lines_to_clear; i++) {
-            text_buffer << tty::cursor_up << tty::clear_line;
+        if (mbar.num_of_lines_to_clear > 1) {
+            text_buffer << tty::cursor_up_multiple_lines(mbar.num_of_lines_to_clear - 1);
```
kontura, line 109: I like the approach, but please don't do it like this via a new public class.

Perhaps just inline it with a comment what the sequences mean.

kontura (changes requested):

libdnf5-cli/progressbar/multi_progress_bar.cpp
kontura: It is much better and simpler, thank you.

kontura (approved):
Thank you for the contribution!

egmontkob:
Let me link to the discussion at https://gitlab.gnome.org/GNOME/vte/-/issues/2837 with lots of details about how terminal emulators (and many other related components of the ecosystem) work under the hood when it comes to processing their input and updating their screen. Summary:

Contrary to the claim of the opening report here, buffering the output cannot fully reliably eliminate flickers. It might make them magnitudes more unlikely to happen, but can never entirely stop them from happening.

The only really 100% robust solution is never to erase the character cells that you soon afterwards re-populate with the desired new text. Just print the new text over the old one, replacing it, with no clearing in between. Use earsing escape sequences only if you actually wish to erase some previously present content and wish to now have blank there.

I'm happy to see that things went in this direction here, I just wanted to leave an explicit note stating that this is the only truly right approach.

Giedriusj1:
Thanks for your comment @egmontkob. The vte discussion was a very interesting read.

If you look at the final solution that was accepted, It contains no "clear line" commands.

We now move the cursor up, and allow the following writes to write over the old text.

PR #1844, opened 2024-11-08
Support repo's Appstream data download and install

mcrha:
Repositories can provide Appstream data for the packages they contain. This Appstream data is consumed by applications like gnome-software or KDE Discover, thus the users can see the packages (apps) in them.

This is to be in pair with PackageKit, which does download and install the repo's Appstream data.

As this adds a dependency on the `appstream` library, there is also a CMake option WITH_APPSTREAM to be able to turn the support off. The support is enabled by default.

-----------------------------------------

That is, the `/var/cache/libdnf5/$REPO_NAME/repodata/` directory can have downloaded also files with `appstream` in their name and, when installed, the data is stored under `/var/cache/swcatalog/xml/`.

CC @m-blaha @jan-kolarik

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ option(WITH_LIBDNF5_CLI "Build library for working with a terminal in a command-
 option(WITH_DNF5 "Build dnf5 command-line package manager" ON)
 option(WITH_DNF5_PLUGINS "Build plugins for dnf5 command-line package manager" ON)
 option(WITH_PLUGIN_ACTIONS "Build a dnf5 actions plugin" ON)
+option(WITH_PLUGIN_APPSTREAM "Build with plugin to install repo's Appstream metadata" ON)
 option(WITH_PLUGIN_RHSM "Build a libdnf5 rhsm (Red Hat Subscription Manager) plugin" OFF)
 option(WITH_PYTHON_PLUGINS_LOADER "Build a special dnf5 plugin that loads Python plugins. Requires WITH_PYTHON3=ON." ON)
 
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -77,6 +77,7 @@ Provides:       dnf5-command(versionlock)
 %bcond_without dnf5
 %bcond_without dnf5_plugins
 %bcond_without plugin_actions
+%bcond_without plugin_appstream
 %bcond_without plugin_rhsm
 %bcond_without python_plugins_loader
 
@@ -604,6 +605,24 @@ Libdnf5 plugin that allows to run actions (external executables) on hooks.
 %{_mandir}/man8/libdnf5-actions.8.*
 %endif
 
+# ========== libdnf5-plugin-appstream ==========
+
+%if %{with plugin_appstream}
+
+%package -n libdnf5-plugin-appstream
+Summary:        Libdnf5 plugin to install repo Appstream data
+License:        LGPL-2.1-or-later
+Requires:       libdnf5%{?_isa} = %{version}-%{release}
+BuildRequires:  pkgconfig(appstream) >= 0.16
+
+%description -n libdnf5-plugin-appstream
+Libdnf5 plugin that installs repository's Appstream data, for repositories which provide them.
+
+%files -n libdnf5-plugin-appstream
+%{_libdir}/libdnf5/plugins/appstream.so
+%config %{_sysconfdir}/dnf/libdnf5-plugins/appstream.conf
+
+%endif
 
 # ========== libdnf5-plugin-plugin_rhsm ==========
 
@@ -805,6 +824,7 @@ automatically and regularly from systemd timers, cron jobs or similar.
     -DWITH_LIBDNF5_CLI=%{?with_libdnf_cli:ON}%{!?with_libdnf_cli:OFF} \
     -DWITH_DNF5=%{?with_dnf5:ON}%{!?with_dnf5:OFF} \
     -DWITH_PLUGIN_ACTIONS=%{?with_plugin_actions:ON}%{!?with_plugin_actions:OFF} \
+    -DWITH_PLUGIN_APPSTREAM=%{?with_plugin_appstream:ON}%{!?with_plugin_appstream:OFF} \
     -DWITH_PLUGIN_RHSM=%{?with_plugin_rhsm:ON}%{!?with_plugin_rhsm:OFF} \
     -DWITH_PYTHON_PLUGINS_LOADER=%{?with_python_plugins_loader:ON}%{!?with_python_plugins_loader:OFF} \
     \
--- a/libdnf5-plugins/CMakeLists.txt
+++ b/libdnf5-plugins/CMakeLists.txt
@@ -2,5 +2,6 @@ set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 
 add_subdirectory("actions")
+add_subdirectory("appstream")
 add_subdirectory("python_plugins_loader")
 add_subdirectory("rhsm")
--- a/libdnf5-plugins/appstream/appstream.cpp
+++ b/libdnf5-plugins/appstream/appstream.cpp
@@ -1,0 +1,128 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <appstream.h>
+#include <libdnf5/base/base.hpp>
+#include <libdnf5/conf/const.hpp>
+#include <libdnf5/plugin/iplugin.hpp>
+#include <libdnf5/repo/repo.hpp>
+#include <libdnf5/repo/repo_query.hpp>
+
+#include <iostream>
+
+using namespace libdnf5;
+
+namespace {
+
+constexpr const char * PLUGIN_NAME{"appstream"};
+constexpr libdnf5::plugin::Version PLUGIN_VERSION{.major = 1, .minor = 0, .micro = 0};
+
+constexpr const char * attrs[]{"author.name", "author.email", "description", nullptr};
+constexpr const char * attrs_value[]{"Milan Crha", "mcrha@redhat.com", "install repo Appstream data."};
+
+class AppstreamPlugin : public plugin::IPlugin {
+public:
+    AppstreamPlugin(libdnf5::plugin::IPluginData & data, libdnf5::ConfigParser &) : IPlugin(data) {}
+    virtual ~AppstreamPlugin() = default;
+
+    PluginAPIVersion get_api_version() const noexcept override { return PLUGIN_API_VERSION; }
+
+    const char * get_name() const noexcept override { return PLUGIN_NAME; }
+
+    plugin::Version get_version() const noexcept override { return PLUGIN_VERSION; }
+
+    const char * const * get_attributes() const noexcept override { return attrs; }
+
+    const char * get_attribute(const char * attribute) const noexcept override {
+        for (size_t i = 0; attrs[i]; ++i) {
+            if (std::strcmp(attribute, attrs[i]) == 0) {
+                return attrs_value[i];
+            }
+        }
+        return nullptr;
+    }
+
+    void repos_loaded() override {
+        Base & base = get_base();
+        repo::RepoQuery repos(base);
+        repos.filter_enabled(true);
+        for (const auto & repo : repos) {
+            auto type = repo->get_type();
+            if (type == repo::Repo::Type::AVAILABLE || type == repo::Repo::Type::SYSTEM) {
+                install_appstream(repo.get());
+            }
+        }
+    }
+
+private:
+    void install_appstream(libdnf5::repo::Repo * repo);
+};
+
+void AppstreamPlugin::install_appstream(libdnf5::repo::Repo * repo) {
+    libdnf5::Base & base = get_base();
+    if (!repo->get_config().get_main_config().get_optional_metadata_types_option().get_value().contains(
+            libdnf5::METADATA_TYPE_APPSTREAM))
+        return;
+
+    std::string repo_id = repo->get_config().get_id();
+    auto appstream_metadata = repo->get_appstream_metadata();
+    for (auto & item : appstream_metadata) {
+        const std::string path = item.second;
+        GError * local_error = NULL;
+
+        if (!as_utils_install_metadata_file(
+                AS_METADATA_LOCATION_CACHE, path.c_str(), repo_id.c_str(), NULL, &local_error)) {
+            base.get_logger()->debug(
+                "Failed to install Appstream metadata file '{}' for repo '{}': {}",
+                path,
+                repo_id,
+                local_error ? local_error->message : "Unknown error");
+        }
+
+        g_clear_error(&local_error);
+    }
+}
+
+}  // namespace
+
+
+PluginAPIVersion libdnf_plugin_get_api_version(void) {
+    return PLUGIN_API_VERSION;
+}
+
+const char * libdnf_plugin_get_name(void) {
+    return PLUGIN_NAME;
+}
+
+plugin::Version libdnf_plugin_get_version(void) {
+    return PLUGIN_VERSION;
+}
+
+plugin::IPlugin * libdnf_plugin_new_instance(
+    [[maybe_unused]] LibraryVersion library_version,
+    libdnf5::plugin::IPluginData & data,
+    libdnf5::ConfigParser & parser) try {
+    return new AppstreamPlugin(data, parser);
+} catch (...) {
+    return nullptr;
+}
+
+void libdnf_plugin_delete_instance(plugin::IPlugin * plugin_object) {
+    delete plugin_object;
+}
--- a/libdnf5/repo/repo.cpp
+++ b/libdnf5/repo/repo.cpp
@@ -52,7 +52,6 @@ extern "C" {
 #include <filesystem>
 #include <set>
 
-
 namespace libdnf5::repo {
 
 static void is_readable_rpm(const std::string & fn) {
@@ -203,6 +202,9 @@ void Repo::read_metadata_cache() {
     p_impl->downloader->load_local();
 }
 
+std::vector<std::pair<std::string, std::string>> Repo::get_appstream_metadata() const {
+    return get_downloader().get_appstream_metadata();
+}
 
 bool Repo::is_in_sync() {
     if (!p_impl->config.get_metalink_option().empty() && !p_impl->config.get_metalink_option().get_value().empty()) {
@@ -412,6 +414,13 @@ void Repo::load_available_repo() {
 
     auto optional_metadata = p_impl->config.get_main_config().get_optional_metadata_types_option().get_value();
     const bool all_metadata = optional_metadata.contains(libdnf5::METADATA_TYPE_ALL);
+
+    if (all_metadata || optional_metadata.contains(libdnf5::METADATA_TYPE_APPSTREAM)) {
+        auto appstream_metadata = p_impl->downloader->get_appstream_metadata();
+        for (auto & item : appstream_metadata) {
+            p_impl->solv_repo->load_repo_ext(RepodataType::APPSTREAM, item.first, *p_impl->downloader.get());
+        }
+    }
 
     if (all_metadata || optional_metadata.contains(libdnf5::METADATA_TYPE_FILELISTS)) {
         p_impl->solv_repo->load_repo_ext(RepodataType::FILELISTS, *p_impl->downloader.get());
--- a/libdnf5/repo/repo_downloader.cpp
+++ b/libdnf5/repo/repo_downloader.cpp
@@ -388,6 +388,25 @@ const std::string & RepoDownloader::get_metadata_path(const std::string & metada
     return it != metadata_paths.end() ? it->second : empty;
 }
 
+bool RepoDownloader::is_appstream_metadata_type(const std::string & type) const {
+    /* TODO: make the list configurable with this default */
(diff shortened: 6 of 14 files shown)


mcrha:
Hrm, the CI either needs to add `appstream-devel` (or better `pkgconfig(appstream)>=1.0`) dependency or turn off build of these bits with `-DWITH_APPSTREAM=OFF` to the `cmake` command. No idea how to do it, but if you could guide me I'll change it either way.

libdnf5/repo/repo.cpp
```diff
@@ -203,6 +206,31 @@ void Repo::read_metadata_cache() {
     p_impl->downloader->load_local();
 }
 
+void Repo::install_appstream() {
+#ifdef WITH_APPSTREAM
+    std::string repo_id = p_impl->config.get_id();
+    const char * as_basenames[] = {"appstream", "appstream-icons", NULL};
```
mcrha, line 212: this might not necessarily work when the appstream data is generated by the `appstream-generator` (instead of the `appstream-builder`), but I cannot test what the generator does in Fedora, because it's simply [broken there](https://bugzilla.redhat.com/show_bug.cgi?id=2325398).

Ideally, the `get_metadata_path()` might return a list of the repo files, which contain certain string, like in this case I'd call it `appstream` and any `*appstream*` files would be processed here, instead of knowing beforehand what file "types" can be used by whichever Appstream library and/or its version.

For example, the appstream-generator might create multiple `appstream-icons` files, each containing the icon size in its name, which is very bad approach, because the consumers need to be updated when the icon sizes change, to keep the list of the possible icon sizes in sync with the others.

Conan-Kudo: Note that `appdata` and `appdata-icons` are also used (by e.g. SUSE distributions).

mcrha: Do they use different names, thus patch PackageKit? Or how does that work?

Conan-Kudo:
We do not want AppStream data being downloaded by default, as it's quite huge, so only callers that can do something with that data should request it.

libdnf5/repo/repo.cpp
mcrha: I see. I just committed a change, which considers any data type which starts with `appstream` or `appdata` prefix as being related to the Appstream and it's tried to be installed.  It allows to avoid hard-coding all the possibilities for the `appstream-icons-64x64.tar.gz`, .... (or what the `appstream-generator` generates). I asked (through a TODO comment), to move that to the configuration, because I do not know how to do it, hence the comment only.

mcrha:
> We do not want AppStream data being downloaded by default, as it's quite huge, so only callers that can do something with that data should request it.

It's not that _you_ (the caller) only can work with it, the Appstream data is installed system wide (to /var/cache/swcatalog/...), where any Appstream-capable app can use it. Preparing it beforehand makes sense, no?

Or do you mean, in the "we" distro, when there's no gnome-software nor KDE Discover, there's also no PacakgeKit (which downloads and installs the appstream data unconditionally)? Note these are only the GUI apps I know of, which consume the appstream data. The `appstreamcli` reads it too, to name one on the command line.

mcrha:
> Hrm, the CI...

I looked around, and please correct me if I'm wrong, because I can be wrong, it seems to me the CI is located in a separate project, in the https://github.com/rpm-software-management/ci-dnf-stack . I looked briefly into it, but I realized it's a blackbox for me, I even do not see the build parameters to be applied there. I'm sorry.

Conan-Kudo:
> > We do not want AppStream data being downloaded by default, as it's quite huge, so only callers that can do something with that data should request it.
> 
> It's not that _you_ (the caller) only can work with it, the Appstream data is installed system wide (to /var/cache/swcatalog/...), where any Appstream-capable app can use it. Preparing it beforehand makes sense, no?
> 
> Or do you mean, in the "we" distro, when there's no gnome-software nor KDE Discover, there's also no PacakgeKit (which downloads and installs the appstream data unconditionally)? Note these are only the GUI apps I know of, which consume the appstream data. The `appstreamcli` reads it too, to name one on the command line.

Right, If PackageKit-DNF5 or GNOME Software through dnf5daemon want it, they can ask for it to be downloaded. But the regular dnf5 and dnf5daemon CLI can't do anything with it, so it's not useful to download.

mcrha:
While I agree, I also do not think it would be helpful. It's the same as if you would want from the PackageKit to not download and install the appstream data from the repository metadata - it cannot be done. This is better than PackageKit, because with dnf and PackageKit on one machine you've duplicated the repo data in the local cache, which is bad (no PackageKit => half repo data stored on the machine).

Could you point me to a repo, which provides that large appstream data, please? I've been in an impression that the distros provide its appstream data in certain packages, not as the repo metadata. Fedora has `appstream-data` package, RPMFusion has `rpmfusion-*-appstream-data` (`rpmfusion-free-appstream-data` and so on). I do not know what package OpenSUSE uses, but for example [the GNOME repo there](https://download.opensuse.org/repositories/GNOME:/Apps/15.6/repodata/) does not provide appstream data as the repo metadata, while that repository contains GUI apps, which do provide their .metainfo.xml files. For example RPMFusion's Steam repo provides the Appstream data, see it [here](https://download1.rpmfusion.org/nonfree/fedora/steam/41/x86_64/repodata/).

Is it possible we are talking about different things?

kontura:
I agree with @Conan-Kudo, I really don't think we should download it by default.

> If the `download & install` of the repo's appstream metadata is a per-session option, then the refresh of the repository needs to run differently, because when something updates the repo without the appstream data and then another app will need the appstream data, then a simple "is repo up-to-date" check is not simple anymore, one would need two checks, one for the repo and one for the appstream data. Handling of a half-updated repo sounds complicated on its own, the more for the maintenance of the code.

I don't think this would be a problem, by default we already download just `primary`, `comps` and `updateinfo` everything else is downloaded only when needed. Typically in different sessions (different dnf invocations) like changelogs (`other`) or `filelists`.

In my option this could fit well into the `optional_metadata_types` config option, then any client can configure it. It could even be set by a distribution in dnf config file.

kontura:
> > Hrm, the CI...
> 
> I looked around, and please correct me if I'm wrong, because I can be wrong, it seems to me the CI is located in a separate project, in the https://github.com/rpm-software-management/ci-dnf-stack . I looked briefly into it, but I realized it's a blackbox for me, I even do not see the build parameters to be applied there. I'm sorry.

Yes, it is located in `ci-dnf-stack` but the build is done according to the `dnf5.spec` file in this repository. You would need to add the dependency there.

Though I think we should discuss if it should be a dependency of libdnf, that is if the install should be done by libdnf. This is related to the other comments but since libdnf is not using the metadata it would make more sense to me if the install was done by the client that requested it.

mcrha:
> This is related to the other comments but since libdnf is not using the metadata it would make more sense to me if the install was done by the client that requested it.

Okay.

In that case I'd need access to the metadata itself from the client (through the dnf5daemon). Either to the `repomd.xml` filenames for each repository (doing `N` D-Bus calls, one for each of the `N` configured and enabled repos, sounds boring and waste of time and resources to me, thus a single "give me repomd filenames for each configured & enabled repo" would be better), with another API to get URL of the repo metadata (which cannot be done in bulk then), to avoid code duplication and avoid expectations which can change in the future, like about where the files are stored and such, which the libdnf knows, but the client having only the repomd.xml does not. Then the client would be able to download the data from the provided URL, though you do more than just the download, you also verify the content, right?

Alternatively, a new API to list provided metadata by respective repo, then a new API to download chosen types, which would result in a list of full path names for the places where it had been downloaded. Ideally allow download in bulk, a list of the types to be downloaded resulting in a list of the filenames.

Alternatively, anything you think would be better and would work for you.

Let me know which it is, please, and I can try to update the merge request accordingly, or pass it to someone whom knows the internals of the dnf5 better than me (I know basically nothing). Thanks in advance.

kontura:
> Alternatively, anything you think would be better and would work for you.
> 

I was thinking the client would just add something like `"appstream"` to the `optional_metadata_types` config option before metadata sync, libdnf would download all the requested metadata as usual and then the client would call [`libdnf5::repo::Repo::get_metadata_path("appstream")`](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/repo/repo.hpp#L198) or some daemon equivalent to get the actual file path(s).

This would probably needed additional work to handle the fact that its multiple files with unknown type with prefix `appstream`/`appdata`.

This is just my idea, I will bring it up on our team meeting today.

mcrha:
I realized the list of the downloaded appstream data is useless, because the client usually runs in the user space and cannot write to the `/var/cache/...`, that can be done only by a privilege processes. Saving the data per-user would be a waste of disk space.

Nonetheless, I went with the config approach. The client can ask to use the repo's appstream data by the `optional_metadata_types` config key (named `appstream` also for the `appdata*` types) and it'll be downloaded & installed only if requested, not by default. The appstream support can be still disabled in the configure time of the project, of course.

The just added commit is not complete, it doesn't work, I'll be looking on it. It seems like I cannot just pick "everything appstream*/appdata*" easily (the added change is about that, see the commit message), thus I might end up with a list of the possible types to be read, which is sad.

Conan-Kudo:
At least this will make the code simpler for the PackageKit DNF5 backend too, I don't have to manually download the appstream data. 😄

mcrha:
I've another go on this. This time:

* the appstream data is downloaded only if the client has set in the `optional_metadata_types` also `appstream`
* there is a hard-coded list of the expected appstream-related types; I would read it from the `repomd.xml`, but that file content is not available in the `RepoDownloader::init_local_handle()` and when it's created in the `RepoDownloader::load_local()`, it's too late to add it into the `LRO_YUMDLIST` (yeah, I tried to re-run the `result`, but then it fails with an error about missing appstream data or what). It's a pita to have the hard coded list of the types, I hoped this can do better than PackageKit, but... I do not know how to do it.

I tested that this works as expected also for a case when another client reads the repository data without the `appstream` (virtual) type first. The client with this type causes load (and install) of this data even if the repo did not change, which is great.

_P.S.: I do not see in this confusing GitHub interface how to set a Draft flag on this merge request; could anyone unset it, please?_

mcrha:
> There is "Ready for review" button down under the checks. I've pressed it :)

Hmm, I guess there might be another button to let me move it to draft again (which I do _not_ want to), but I do not see it there. Either it does not look like a button, or it's named somehow oddly, or I'm stupid. I'm definitely not blind (yet) ;)

mcrha:
> The move it to draft link is at the top right under the list of Reviewers.

I see, a tiny grey link, on a very different place than the other button. Confusing from the start. Anyway, thanks for the pointers.

libdnf5/repo/repo_downloader.cpp
```diff
@@ -493,6 +513,28 @@ void RepoDownloader::common_handle_setup(LibrepoHandle & h) {
...
+        dlist.push_back("appstream-icons-128x128@2");
+        // consult the prefixes with the is_appstream_metadata_type()
+        dlist.push_back("appdata");
+        dlist.push_back("appdata-icons");
+        dlist.push_back("appdata-icons-48x48");
+        dlist.push_back("appdata-icons-48x48@2");
+        dlist.push_back("appdata-icons-64x64");
+        dlist.push_back("appdata-icons-64x64@2");
+        dlist.push_back("appdata-icons-128x128");
+        dlist.push_back("appdata-icons-128x128@2");
```
jan-kolarik, line 536: So we want to always have first just the repomd present to parse appstream data for that. We have currently these scenarios when downloading metadata:

1. No previous metadata => downloading everything once
2. Previous metadata present but old => first download only repomd to check it's old, then download everything (including repomd again)
3. Previous metadata present and valid => only download repomd to check it's valid

It would mean that we just need to alter the 1. scenario to have the additional download of repomd before downloading everything.

I think it should be manageable. Just it would need some refactoring, the repomd-only downloading is currently done in `RepoDownloader::is_repomd_in_sync()`. In our case we want to reuse the `LrYumRepo` object from there for our appstream parsing.

libdnf5/repo/repo_downloader.cpp
```diff
@@ -493,6 +513,28 @@ void RepoDownloader::common_handle_setup(LibrepoHandle & h) {
     if (optional_metadata.extract(libdnf5::METADATA_TYPE_UPDATEINFO)) {
         dlist.push_back(MD_FILENAME_UPDATEINFO);
     }
+    if (get_optional_metadata().extract(libdnf5::METADATA_TYPE_APPSTREAM)) {
```
jan-kolarik, line 516: This block should be conditional with `#ifdef WITH_APPSTREAM` right?

mcrha: Right, it can be conditional.

libdnf5/repo/repo_downloader.cpp
mcrha: Hmm, that could work, yes.

Is it acceptable to update the "global" `LRO_YUMDLIST` in the `is_repomd_in_sync()`? The function would do more than just checking whether the repomd is up-to-date afterwards.

I already lost my code where I've been playing with these things, but it should not be that hard to write it again.

jan-kolarik: If metadata are correctly downloaded to cache, they should not be downloaded again during the next run. They are loaded from cache then, the paths should be stored in `metadata_paths`. So maybe it could be also a problem with that. Would like to debug it, but unfortunately don't have capacity for that in upcoming days...

jan-kolarik: OK, let's keep things as they are. I will create a follow-up issue to try optimizing those hardcoded names.

ppisar:
These are the dependencies of appstream library:

~~~~
# pkgconf --libs appstream
-lappstream -lgio-2.0 -lgobject-2.0 -lglib-2.0
# scanelf -n /usr/lib64/libappstream.so
 TYPE   NEEDED FILE 
ET_DYN libglib-2.0.so.0,libgobject-2.0.so.0,libcurl.so.4,libxmlb.so.2,libgio-2.0.so.0,libxml2.so.2,libyaml-0.so.2,libsystemd.so.0,libzstd.so.1,libstemmer.so.0,libgcc_s.so.1,libc.so.6 /usr/lib64/libappstream.so
~~~~

mcrha:
Would it be acceptable if I add a new method to the `IPlugin`, something like:
```c
    /// The repo_loaded hook.
    /// It is called when a single repository had been loaded (in Impl).
    virtual void repo_loaded(Repo * repo);
```
which I'd call in the place where `repo->install_appstream()` is called in this merge request?

There is that `IPlugin::repos_loaded()`, but seeing how many things are considered before the repo is added into the `repos_for_processing`, I'd prefer to avoid the loop and code duplication and I'd just move the body of the `Repo::install_appstream()` into the new plugin, using this new IPlugin method.

The alternative of traversing all repos (enabled/disabled/broken/...) and check whether the cache has the appstream data and install it if exists is also possible and I can go by that route as the starter, it only feels performance heavy on the first look.

libdnf5-plugins/CMakeLists.txt
```diff
@@ -4,3 +4,7 @@ set(CMAKE_C_VISIBILITY_PRESET hidden)
 add_subdirectory("actions")
 add_subdirectory("python_plugins_loader")
 add_subdirectory("rhsm")
+
+if(WITH_APPSTREAM)
+    add_subdirectory("appstream")
```
mcrha, line 9: It works, the Appstream data is installed as expected, with "one little caveat": when a session ends (on the dnf5daemon-server side), you also destroy the plugin class (which is fine), but you also unload the module (.so file). It breaks glib internals, because when another session opens and loads the plugin and calls the appstream library function, the glib aborts the dnf5daemon-server with:
```
(process:49835): GLib-GObject-CRITICAL **: 15:47:52.469: cannot register existing type 'AsMetadata'

(process:49835): GLib-GObject-CRITICAL **: 15:47:52.469: cannot add private field to invalid (non-instantiatable) type '<invalid>'

(process:49835): GLib-CRITICAL **: 15:47:52.469: g_once_init_leave_pointer: assertion 'result != 0' failed

(process:49835): GLib-GObject-CRITICAL **: 15:47:52.469: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
Segmentation fault (core dumped)
```

This does not exhibit when there's at least one session always loaded, which keeps the plugin module loaded in the memory, thus the glib type system is left in tact.

Is the plugin module unload really needed? I guess it can break more than just glib, it can break anything what stores some global data.

libdnf5-plugins/appstream/appstream.cpp
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return nullptr;
+    }
+
+    void repos_loaded() override {
+        Base & base = get_base();
+        repo::RepoQuery repos(base);
+        repos.filter_enabled(true);
+        for (const auto & repo : repos) {
+            switch (repo->get_type()) {
```
jan-kolarik, line 66: This is minor, but I would rather do:
```
auto type = repo->get_type();

if (type == repo::Repo::Type::AVAILABLE || type == repo::Repo::Type::SYSTEM) {
    install_appstream(repo.get());
}
```

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ option(WITH_COMPS "Build with comps groups and environments support" ON)
 option(WITH_MODULEMD "Build with modulemd modules support" ON)
 option(WITH_ZCHUNK "Build with zchunk delta compression support" ON)
 option(WITH_SYSTEMD "Build with systemd and D-Bus features" ON)
+option(WITH_APPSTREAM "Build with appstream support" ON)
```
jan-kolarik, line 37: I think now we don't want here the `WITH_APPSTREAM` switch and rather move it inside the `appstream/CMakeLists.txt`.

jan-kolarik: Similarly as it is handled for the `actions` plugin.

dnf5.spec
```diff
@@ -143,6 +144,9 @@ BuildRequires:  bash-completion
 BuildRequires:  cmake
 BuildRequires:  doxygen
 BuildRequires:  gettext
+%if %{with appstream}
+BuildRequires:  pkgconfig(appstream) >= 0.16
```
jan-kolarik, line 148: This should be moved inside the `libdnf5-plugin-appstream` package, right? Otherwise the dnf5 itself would depend on appstream.

CMakeLists.txt
```diff
@@ -129,6 +130,10 @@ if (WITH_MODULEMD)
     add_definitions(-DWITH_MODULEMD)
 endif()
 
+if (WITH_APPSTREAM)
```
jan-kolarik, line 133: Should be dropped now.

libdnf5-plugins/CMakeLists.txt
```diff
@@ -4,3 +4,7 @@ set(CMAKE_C_VISIBILITY_PRESET hidden)
 add_subdirectory("actions")
 add_subdirectory("python_plugins_loader")
 add_subdirectory("rhsm")
+
+if(WITH_APPSTREAM)
```
jan-kolarik, line 8: Please add the subdirectory unconditionally and move the condition within the appstream/CMakeLists.txt like it is done e.g. for actions plugin.

libdnf5/repo/repo_downloader.cpp
```diff
@@ -498,8 +517,35 @@ void RepoDownloader::common_handle_setup(LibrepoHandle & h) {
 
         // download the rest metadata added by 3rd parties
         for (auto & item : optional_metadata) {
-            dlist.push_back(item.c_str());
+            // the appstream metadata is a "virtual" type, the list
+            // of the types is read from the repomd file
+            if (item.compare(libdnf5::METADATA_TYPE_APPSTREAM) != 0)
+                dlist.push_back(item.c_str());
         }
+#ifdef WITH_APPSTREAM
```
jan-kolarik, line 525: Hmm, we can probably drop this conditional in the end as it is now in a separate plugin. There is not anyway anything dependent on the appstream library.

libdnf5-plugins/CMakeLists.txt
jan-kolarik: I see. So this issue is related to the code that loads and unloads shared libraries [here](https://github.com/rpm-software-management/dnf5/blob/5.2.8.1/libdnf5/utils/library.cpp#L28-L38).

Could we simply solve this by passing `RTLD_NODELETE` during the library's `dlopen` to prevent unloading the libraries for the duration of the daemon's lifetime? Alternatively, should this be conditional for specific plugins only? Another approach could be to implement custom logic to unload libraries when the daemon terminates. @m-blaha @jrohel, could you please share your thoughts on this? @m-blaha @jrohel, could you please share your thoughts on this?

libdnf5/repo/repo_downloader.cpp
mcrha: it felt like downloading the data without being used (the plugin not compiled) is useless, but maybe it's obsolete by the plugin. I changed it.

libdnf5-plugins/CMakeLists.txt
mcrha: I know I'm not asked here, but if I can add my opinion, then I believe the `RTLD_NODELETE` is a good thing to add. As the plugins are loaded only inside the deamon, they are freed together with the deamon too. Not within the session object the deamon creates, which is good and intentional.

I gave it a try and it works as expected, no more crash of the dnf5-daemon process and no runtime warnings either. See the added (second) commit. I cleaned up the commits too, the merge request is ready for the final review now.

jan-kolarik (approved):

libdnf5-plugins/CMakeLists.txt
jan-kolarik: We agreed on using the `RTLD_NODELETE` on team meeting.

jan-kolarik:
Hey @mcrha, thanks a lot for iterating this multiple times, it's in a great state now! All reviewed bits are resolved, so we can proceed with merging...

ppisar:
This breaks CI jobs (e.g. "CCache Build (gcc)", https://github.com/rpm-software-management/dnf5/pull/1999):

~~~~
-- Checking for module 'appstream>=0.16'
CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:645 (message):
--   Package 'appstream' not found
  The following required packages were not found:

   - appstream>=0.16

Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPkgConfig.cmake:873 (_pkg_check_modules_internal)
  libdnf5-plugins/appstream/CMakeLists.txt:11 (pkg_check_modules)
~~~~

Have you amended various container images to install new dependency?

mcrha:
I did not do anything with the CI images. Apart of not knowing what to do, I thought they get the dependencies from the .spec file. Maybe not all of them?

kontura:
They do get them from the .spec file, the CI host container just needed to be rebuild ([which happened tonight automatically](https://github.com/rpm-software-management/ci-dnf-stack/actions/runs/12702369376)). 
The dnf5 builds should be working now.

Issue #1441, opened 2024-04-24
Python DownloadProgress.add_new_download() callback user data issue


es-fabricemarie:
Hi,

I'm trying to use DownloadProgress callbacks using the Python bindings and I face an issue with the add_new_download method.

The code to reproduce my issue can be simplified to:

``` Python
#!/usr/bin/python3
import libdnf5

class DownloadProgress(libdnf5.repo.DownloadCallbacks):
    # Notify the client that a new download has been created.
    #
    # @param user_data User data entered together with url/package to download.
    # @param description The message describing new download (url/packagename).
    # @param total_to_download Total number of bytes to download.
    # @return Associated user data for new download.
    def add_new_download(self, user_data, description, total_to_download):
        print("Just added package %s to download list" % description)
        return description


base = libdnf5.base.Base()
base.load_config_from_file()
base.setup()
sack = base.get_repo_sack()
sack.create_repos_from_system_configuration()
sack.update_and_load_enabled_repos(True)
goal = libdnf5.base.Goal(base)
goal.add_rpm_upgrade()
transaction = goal.resolve()
downloader_callbacks = DownloadProgress()
base.set_download_callbacks(
    libdnf5.repo.DownloadCallbacksUniquePtr(downloader_callbacks)
)
transaction.download()
```

I'm trying to set the package name as the `Associated user data for new download`, so that it gets sent back to the other callbacks (`progress` and `end` for instance), so I can report fine grained progress.

I run this code, the script crashes and a core is dumped:

```
Just added package docker-buildx-plugin-0:0.14.0-1.fc39.x86_64 to download list
terminate called after throwing an instance of 'Swig::DirectorTypeMismatchException'
  what():  SWIG director type mismatch in output value of type 'void *'
Aborted (core dumped)
```

What am I doing wrong here? Or is this an issue with the Python binding?

jan-kolarik:
The first thing I would look into is probably implementing all the methods from the `libdnf5.repo.DownloadCallbacks` interface in your `DownloadProgress` class - there is also the `progress`, `end`, `mirror_failure` and `fastest_mirror` methods. That might be the source of an error if these methods are missing and the library tries to call them.

es-fabricemarie:
I did of course. But the code above is the shortest way to reproduce my issue.
In my example, as soon as add_new_download() method returns it dumps core.
All the example I found return None. But when returning none, we lose the ability to pass metdata to the suequent `progress`, `end`, etc.. callbacks.

[...]

PR #1849, opened 2024-11-11
SWIG bindings for user_cb_data in repo::DownloadCallbacks, unit tests

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1441

SWIG bindings for user_cb_data in `repo::DownloadCallbacks` in the C++ API, the user can return a pointer (`void *`) to user data in the `add_new_download` method of the `repo::DownlodCallbacks` class. This pointer is then passed as an argument to `void * user_cb_data` in other methods of the `repo::DownloadCallbacks` class.

There was a problem of how to use this mechanism in SWIG bindings.

I implemented and tested several solutions. In the case of passing a pointer, there was a problem with the ownership of the data the pointer points to. Users of Python and some other languages are used to the gargabe collector and automatically holding ownership of the passed data. While I've solved this for `user_cb_data`, passing another void pointer will need to be solved (I'll do that later). And from there, the ownership will be more complicated. We need the solution to work reliably in Python, Ruby and Perl. And possibly be easy to use for other languages in the future.

In the end, I chose the method of passing an integer instead of a pointer. When passing an integer, there is no need to deal with who owns the number. The integer is passed by value. And if the user needs to pass objects, for example, he can create an array of objects and pass the index (integer) to the array. The array also provides ownership of the objects.

PR also improves and adds unit tests for Perl and extends unit tests for other languages.

### DIFF
--- a/bindings/libdnf5/repo.i
+++ b/bindings/libdnf5/repo.i
@@ -53,7 +53,40 @@
 %include "libdnf5/repo/config_repo.hpp"
 
 %feature("director") DownloadCallbacks;
+
+%typemap(directorin, noblock=1) void * user_cb_data {
+    $input = SWIG_From_int(static_cast<int>(reinterpret_cast<intptr_t>($1)));
+}
+
+%typemap(directorout, noblock=1) void * {
+    int swig_val;
+    int swig_res = SWIG_AsVal_int($1, &swig_val);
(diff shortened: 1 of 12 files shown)


m-blaha (approved):
Thank you very much!

PR #1861, opened 2024-11-14
I18N: Mark messages in "dnf search" output for a translation

sunwire:


kontura (changes requested):
Can you please move the include according to the: https://github.com/rpm-software-management/dnf5/actions/runs/11841388613/job/33119622460?pr=1861?

kontura (approved):
Thank you

PR #1863, opened 2024-11-15
Fix copr chroot specification: replace faulty regex with simpler split

kontura:
The regex `-[^-]*` has multiple matches in a chroot like: `fedora-rawhide-x86_64`, the result of the `regex_replace` is just `fedora`.

Replace it with simpler split.

After https://github.com/rpm-software-management/ci-dnf-stack/pull/1587 is merged I would also like to add couple CI tests.


kontura:
Couple tests added here: https://github.com/rpm-software-management/ci-dnf-stack/pull/1594

jan-kolarik (approved):

Issue #1796, opened 2024-10-23
dnf5 builddep plugin: support remote spec file 


kontura:
This is supported by dnf4 version.

There is a TODO in the code: https://github.com/rpm-software-management/dnf5/blob/main/dnf5-plugins/builddep_plugin/builddep.cpp#L142

A CI test is also present and disabled: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/plugins-core/builddep.feature#L6

Issue #1651, opened 2024-08-26
Implement ABI compatibility check in CI


jan-kolarik:
We need to implement an ABI differences check action in our CI pipeline to detect if a submitted PR breaks ABI compatibility compared to the current upstream build. Although an `abidiff` check is already run as part of the `rpminspect` test suite after pushing a Bodhi update, this occurs too late in the process.

ppisar:
When comparing the pull-requested code against a previous good known state, there were two ideas:

David's idea was offload it to Fedora build system and compare against "last" Fedora build in Koji.

My idea was use abidw to extract ABI from a good know state, commit the abidw output (a single XML file) to dnf5 git tree and then use abidiff against that committed abidw file.

PR #1869, opened 2024-11-18
Add packit job to run ABI check on testing farm

kontura:
Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1595


kontura:
For: https://github.com/rpm-software-management/dnf5/issues/1651

kontura:
And actually after the crash fix is released it will fail on the lack of support of running `builddep` on remote files: https://github.com/rpm-software-management/dnf5/issues/1796

jan-kolarik (approved):
Thanks! Seems fine, I'll just wait for the test to be re-run.

Issue #1864, opened 2024-11-15
dnf5 builddep crashes in Fedora 41 container


abbra:
DNF5 builddep crashes in Fedora 41 container. We use fedora-toolbox container in FreeIPA CI and part of the process we run is a rebuild of FreeIPA package. For that we need to install build dependencies with `dnf builddep`. This works well for Fedora 40 and prior versions but since this morning we've got reliable crashes in `dnf builddep`.

It is reproducible outside of a cloud environment as well. Below is the stacktrace we get:

```
Thread 1 "dnf" received signal SIGSEGV, Segmentation fault.
Downloading 161.53 K source file /usr/include/c++/14/bits/basic_string.h
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string (this=<optimized out>, __s=<optimized out>, __n=<optimized out>, __a=..., this=<optimized out>, __s=<optimized out>,                     
    __n=<optimized out>, __a=...) at /usr/include/c++/14/bits/basic_string.h:626
626	      : _M_dataplus(_M_local_data(), __a)
(gdb) bt full
#0  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string (this=<optimized out>, __s=<optimized out>, __n=<optimized out>, __a=..., this=<optimized out>, __s=<optimized out>, 
    __n=<optimized out>, __a=...) at /usr/include/c++/14/bits/basic_string.h:626
No locals.
#1  0x00007f3838392b9d in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string (this=<optimized out>, __svw=..., __a=..., this=<optimized out>, __svw=..., __a=...)
    at /usr/include/c++/14/bits/basic_string.h:182
No locals.
#2  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<std::basic_string_view<char, std::char_traits<char> >, void> (this=<optimized out>, __t=..., __a=..., this=<optimized out>, 
    __t=..., __a=...) at /usr/include/c++/14/bits/basic_string.h:801
No locals.
#3  std::construct_at<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::basic_string_view<char, std::char_traits<char> >&> (__location=<optimized out>)
    at /usr/include/c++/14/bits/stl_construct.h:97
No locals.
[...]

PR #1870, opened 2024-11-18
rpm: Reset RPM log callback upon RpmLogGuard destruction

m-blaha:
This patch addresses undefined behavior that occurs when the RPM log
callback function is invoked after the RpmLogGuard object has been
destroyed. The issue arises because rpmlogCallbackData retains a
reference to the destroyed object.

The problem manifests during the execution of the builddep plugin command:

1. During Vars::detect_release, the RPM log callback is set via RpmLogGuard.
2. After the release version is detected, the RpmLogGuard object is destroyed.
3. A subsequent call to rpmSpecParse() by the builddep command may
   trigger the previously set log callback.
4. This results in undefined behavior because the callback data still
   points to the destroyed RpmLogGuard.

By resetting the RPM log callback upon RpmLogGuard destruction, this
patch ensures safe and predictable behavior.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2326535
Resolves: https://github.com/rpm-software-management/dnf5/issues/1864

### DIFF
--- a/libdnf5/rpm/rpm_log_guard.cpp
+++ b/libdnf5/rpm/rpm_log_guard.cpp
@@ -65,6 +65,16 @@ static int rpmlog_callback(rpmlogRec rec, rpmlogCallbackData data) {
     return 0;
 }
 
+RpmLogGuardBase::~RpmLogGuardBase() {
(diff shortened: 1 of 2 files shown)


jan-kolarik (approved):
Thanks for a quick fix!

PR #1879, opened 2024-11-20
repo: Add option to download all repository metadata

m-blaha:
Needed for `dnf5 reposync --download-metadata` command.

### DIFF
--- a/include/libdnf5/conf/const.hpp
+++ b/include/libdnf5/conf/const.hpp
@@ -63,6 +63,7 @@ constexpr const char * METADATA_TYPE_FILELISTS = "filelists";
 constexpr const char * METADATA_TYPE_OTHER = "other";
 constexpr const char * METADATA_TYPE_PRESTO = "presto";
 constexpr const char * METADATA_TYPE_UPDATEINFO = "updateinfo";
+constexpr const char * METADATA_TYPE_ALL = "all";
 
 const std::set<std::string> OPTIONAL_METADATA_TYPES{
     METADATA_TYPE_COMPS, METADATA_TYPE_FILELISTS, METADATA_TYPE_OTHER, METADATA_TYPE_PRESTO, METADATA_TYPE_UPDATEINFO};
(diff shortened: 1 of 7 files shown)


include/libdnf5/conf/const.hpp
```diff
@@ -63,6 +63,7 @@ constexpr const char * METADATA_TYPE_FILELISTS = "filelists";
 constexpr const char * METADATA_TYPE_OTHER = "other";
 constexpr const char * METADATA_TYPE_PRESTO = "presto";
 constexpr const char * METADATA_TYPE_UPDATEINFO = "updateinfo";
+constexpr const char * METADATA_TYPE_ALL = "all";
 
 const std::set<std::string> OPTIONAL_METADATA_TYPES{
     METADATA_TYPE_COMPS, METADATA_TYPE_FILELISTS, METADATA_TYPE_OTHER, METADATA_TYPE_PRESTO, METADATA_TYPE_UPDATEINFO};
```
kontura, line 69: It looks like this `set` is meant to represent all the optional metadata types but its just the ones known to libdnf5, so it is different to the new `all`.

(Also I think both of its uses are questionable. At least the `changelog` plugin doesn't need to download/load all of these metadata types, I think it needs just `METADATA_TYPE_OTHER`.)

This is not that related so I don't believe it should block this PR but some comment that describes the difference between `all` and `OPTIONAL_METADATA_TYPES` would be nice.

m-blaha: Hm, I'm thinking about getting rid of OPTIONAL_METADATA_TYPES - changelog definitely doesn't need it and system repo loading should be fine with METADATA_TYPE_ALL.
I can remove it's usage, but since it is on public API I can only mark it deprecated. But the set itself needs to stay for a while.

kontura: Regarding the system repo I am not sure what its supposed to mean. I think the system repo can only load `comps`: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/repo/solv_repo.cpp#L286-L318
I don't think the `RepoDownloader::get_optional_metadata()` is ever used for system repo.

I haven't really investigated it but in my mind `RepoDownloader` doesn't seem applicable to system repo.

m-blaha:
@kontura thanks for the review!
I've added documentation for new "all" metadata type, ensure it's taken into account when testing optional_metadata_types value, and removed usage of OPTIONAL_METADATA_TYPES from the changelog plugin.

kontura (approved):

PR #1880, opened 2024-11-20
repo: While cloning root metadata copy also metalink

m-blaha:
Although metalink.xml is located out of repodata directory, it should be
cloned with repository metadata.
Without correct metalink.xml for example doesn't work package remote
location resolving (because mirrors stored in metalink.xml are unknown).


m-blaha:
Yes, I believe so.

m-blaha:
I need to check the failing CI test.

kontura (commented):
Should the `mirrorlist` be handled in the same way?

I have created a couple CI tests for this https://github.com/rpm-software-management/ci-dnf-stack/pull/1597.

m-blaha:
I'll take a look on the mirror lists.

Thinking about it, this PR is kind of a workaround. We should make sure that remote locations work even without metalink / mirrorlist files. The thing currently is, that repository mirrors are updated during `RepoDownloader::load_local()` execution, which runs (as the name suggests) with `h.set_opt(LRO_LOCAL, 1L);`. Thus no metalinks are downloaded from the remote repo and the code relies on the local copy. But this is out of scope of this simple PR.

m-blaha:
I've added also mirrorlist copying during the clone_root_metadata.

kontura (approved):
Thanks!

PR #1884, opened 2024-11-20
automatic: add a setting to not emit boring messages

AdamWill:
Currently, we always emit messages every time automatic runs, even if it does nothing very interesting (the operation succeeds and there are no changed packages). This is flooding Fedora infra with boring emails that effectively just say "nothing happened".

This adds a setting to only emit if *something* interesting happened - the operation failed, or we saw package changes.

### DIFF
--- a/dnf5-plugins/automatic_plugin/automatic.cpp
+++ b/dnf5-plugins/automatic_plugin/automatic.cpp
@@ -425,25 +425,27 @@ void AutomaticCommand::run() {
         }
     }
 
-    for (const auto & emitter_name : config_automatic.config_emitters.emit_via.get_value()) {
-        std::unique_ptr<Emitter> emitter;
-        if (emitter_name == "stdio") {
-            emitter = std::make_unique<EmitterStdIO>(config_automatic, transaction, output_stream, success);
-        } else if (emitter_name == "motd") {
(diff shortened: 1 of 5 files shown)


AdamWill:
There don't appear to be any tests for this code, so, uh...how much do you trust my C++?

hint: the answer should be "not very far at all". Please review carefully. :D

I can imagine a more sophisticated implementation of this where you could configure it *per emitter* - or have a `changed_emit_via` list of emitters to do this for, or whatever. But my C++ is not good enough to *write* either of those in a not-dumb way (I don't know how to properly factor out the loop over the emitters into a shared function, which would be needed to do either of those without stupidly copy/pasting the loop). So, this!

jan-kolarik:
Hi @AdamWill, sorry for the delayed response. If I'm looking at this correctly, this behavior differs from dnf4, where no message was shown for an empty transaction. I believe we should make this the default behavior and introduce a config option for users who prefer explicit messages indicating that "everything is fine". Will check with the team and get back to you next week.

AdamWill:
sorry, didn't get back to it yet, will try and update this later today.

AdamWill:
OK, rebased and updated with the logic inverted but the default kept as `false` (so now we do not emit boring messages by default).

PR #1886, opened 2024-11-21
spec: toggle dnf5_obsoletes_dnf for RHEL 11

yselkowitz:
Fedora 41 has completed the migration to DNF5, so ELN (the future RHEL 11) is now ready to follow suit.

See also: https://github.com/rpm-software-management/dnf/pull/2161


evan-goode (approved):
Thanks!

PR #1888, opened 2024-11-22
Make dnf5 compatible with sdbus-cpp version 2

m-blaha:
The new and incompatible version of sdbus-cpp library has been released upstream.
I'd like to upgrade also version in Fedora rawhide, but to not break dnf5
building we fist need to adjust it.
This patch makes dnf5 buildable with both sdbus-cpp versions.

### DIFF
--- a/cmake/sdbus_cpp.cmake
+++ b/cmake/sdbus_cpp.cmake
@@ -1,0 +1,10 @@
+# Attempt to find sdbus-c++ version 2 (preferred)
+pkg_check_modules(SDBUS_CPP REQUIRED sdbus-c++>=1)
+
+if(SDBUS_CPP_VERSION LESS 2)
+    message(STATUS "Using sdbus-c++ version 1")
+else()
+    # Define macro for version 2 if found
+    add_definitions(-DSDBUS_CPP_VERSION_2)
+    message(STATUS "Using sdbus-c++ version 2")
+endif()
--- a/dnf5-plugins/needs_restarting_plugin/CMakeLists.txt
+++ b/dnf5-plugins/needs_restarting_plugin/CMakeLists.txt
@@ -7,7 +7,7 @@ add_library(needs_restarting_cmd_plugin MODULE needs_restarting.cpp needs_restar
 # disable the 'lib' prefix in order to create needs_restarting_cmd_plugin.so
 set_target_properties(needs_restarting_cmd_plugin PROPERTIES PREFIX "")
 
-pkg_check_modules(SDBUS_CPP REQUIRED sdbus-c++)
+include(sdbus_cpp)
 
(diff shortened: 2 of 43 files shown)


Conan-Kudo (approved):
This fixed compiling against sdbus-cpp v2 in openSUSE Tumblweed.

kontura (commented):
This looks reasonable to me but the dnf5daemon CI tests fail, I have verified that also locally.
I didn't investigate it further but the error is: `[System.Error.ENXIO] Failed to enter a variant (No such device or address)` when `sdbus-cpp` < 2.

With `sdbus-cpp` >= 2 (I used your build from https://copr.fedorainfracloud.org/coprs/mblaha/dnf5-sdbus-cpp-2/) I am getting some `[org.rpm.dnf.v0.rpm.Rpm.TransactionError] rpm transaction failed with code 6`.

jonathanspw:
Is this PR ready or should it remain a draft?

m-blaha:
Oh, sorry, I forget to switch it back.

kontura:
I am still getting some crashes with `sdbus-cpp > 2`.
Specifically it crashes and the CI hangs for me in:
```gherkin
distro-sync.feature: Scenario: updating a signed pkg
gpg-noeol.feature: Scenario: Install signed package from repository
gpg.feature: Scenario Outline: Install <offline> masterkey signed package and check GPG key was imported
gpg.feature: Scenario Outline: Install <offline> subkey signed package with masterkey signed dependency
```
It seems like some problem with signing.

Plus I am getting some crashes that result in failing tests in:
```
dnf/clean-cachefiles.feature
dnf/gpg.feature
dnf/install-path.feature
dnf/install-provides.feature
dnf/upgrade-dependent.feature
```

However since @m-blaha is the maintainer of `sdbus-cpp` in fedora I am thinking about merging this anyway. Perhaps fixing the crashes will be easier when we don't have to deal with so many patches?

m-blaha:
Hmm, maybe. The crashes (at least those I'm aware of) were all related to the client side of tha dnf5daemon. The server seems to work correctly.

kontura:
> The crashes (at least those I'm aware of) were all related to the client side of tha dnf5daemon.

I see, so is the plan to fix those in subsequent PR?

> The server seems to work correctly.

My bad, I didn't check why it hangs properly, the server indeed doesn't crash.
For me it gets stuck here: https://github.com/rpm-software-management/dnf5/blob/main/dnf5daemon-server/session.cpp#L193 (presumably it would timed out after 5 min).
As far as I can tell the `key_import` callback is never called in the client.

m-blaha:
Created a specialized tracking issue for fixing dnf5daemon-client with sdbus-cpp 2: https://github.com/rpm-software-management/dnf5/issues/2040

Issue #931, opened 2023-10-03
DNF reposync Plugin


j-mracek:

kontura:
- The `--source` option should be changed to `--srpm`. (We are trying to unify the options in dnf5 and we have decided to use `--srpm` everywhere. Old dnf sometimes used `--source` and sometimes `--srpm` which was confusing.)
- Related issue: https://github.com/rpm-software-management/dnf5/issues/380

Conan-Kudo:
Any progress on this? The `reposync` command is the only way to mirror RHEL content, for example.

Here's an active use of this: https://github.com/datto/rhel-reposync-playbook/blob/main/roles/install_reposync/files/rhel-reposync.sh

kontura:
We should also make sure to cover this issue: https://github.com/rpm-software-management/dnf-plugins-core/pull/566#pullrequestreview-2420626395

PR #1903, opened 2024-11-26
Implement reposync plugin

m-blaha:
Implements command `dnf5 reposync` for creating local copies of remote
repositories.
The syntax is mostly compatible with the dnf4 version of the command.

Resolves: https://github.com/rpm-software-management/dnf5/issues/931


dnf5-plugins/reposync_plugin/reposync.cpp
```diff
@@ -0,0 +1,420 @@
...
+                for (const auto & [md_type, md_location] : repo->get_metadata_locations()) {
+                    std::cout << join_url(repo_location, md_location) << std::endl;
+                }
+            }
+            for (const auto & [pth, pkg] : pkg_list) {
+                auto urls = pkg.get_remote_locations();
+                if (urls.empty()) {
+                    std::cerr << libdnf5::utils::sformat(_("Failed to get mirror for package: \"{}\""), pkg.get_name())
+                              << std::endl;
+                    continue;
```
kontura, line 398: The `continue` doesn't seem to be needed?

dnf5-plugins/reposync_plugin/reposync.cpp
```diff
@@ -0,0 +1,420 @@
...
+        // resolve '.', '..', and existing symlinks in the repo_path
+        repo_path = std::filesystem::weakly_canonical(repo_path);
+    } else {
+        repo_path = repo_download_path(repo);
+    }
+    auto & optional_metadata_option = get_context().get_base().get_config().get_optional_metadata_types_option();
+    if (!optional_metadata_option.get_value().contains(libdnf5::METADATA_TYPE_ALL)) {
+        optional_metadata_option.set(libdnf5::METADATA_TYPE_ALL);
+    }
+    repo.download_metadata(repo_path);
```
kontura, line 352: If I understand correctly we download the same metadata for the second time here. All enabled repositories were `expired` in the `configure` step so we could just copy those and download only the missing parts.

I think this would be a nice optimization but I wouldn't block this PR on that.

dnf5-plugins/reposync_plugin/reposync.hpp
```diff
@@ -0,0 +1,73 @@
...
+class ReposyncCommand : public Command {
+public:
+    explicit ReposyncCommand(Context & context) : Command(context, "reposync") {}
+    void set_parent_command() override;
+    void set_argument_parser() override;
+    void configure() override;
+    void run() override;
+
+private:
+    using download_list_type = std::vector<std::pair<std::filesystem::path, libdnf5::rpm::Package>>;
```
kontura, line 47: I may be missing something but would `std::map<std::filesystem::path, libdnf5::rpm::Package>` be a better fit here?

The paths from the vector are converted to an `unordered_set` twice in the code, using the map we could avoid that.

Though the order would be different.

kontura:
I have also noticed in the tests that there is an exit code change: https://github.com/rpm-software-management/ci-dnf-stack/pull/1599/files#diff-ea74dda041309ef66140b8a3856ddc3f3ace60e6e93733efcdc762410324fc39R390

Is this on purpose?
To be honest I think I would prefer the dnf4 behavior. The way I see it if any package from the repository failed to download the `reposync` command did fail.

dnf5-plugins/reposync_plugin/reposync.hpp
m-blaha: Thank you! The use of vector here is most likely the result of gradual development. Let me check whether the `std::map` would be a better fit.

doc/dnf5.8.rst
```diff
@@ -167,6 +167,9 @@ These are available after installing the ``dnf5-plugins`` package.
 :ref:`repoclosure <repoclosure_plugin_ref-label>`
     | Display a list of unresolved dependencies for repositories.
 
+:ref:`reposync <reposync_plugin_ref-label>`
+    | Synchronize packages and metadata of a remote DNF repository to a local directory.
+
```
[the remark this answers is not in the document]
m-blaha: Thanks.

m-blaha:
> I have also noticed in the tests that there is an exit code change: https://github.com/rpm-software-management/ci-dnf-stack/pull/1599/files#diff-ea74dda041309ef66140b8a3856ddc3f3ace60e6e93733efcdc762410324fc39R390
> 
> Is this on purpose? To be honest I think I would prefer the dnf4 behavior. The way I see it if any package from the repository failed to download the `reposync` command did fail.

Thanks for noticing! I did not take a note and forgot about it. The thing is, that currently PackageDownloader::download() does not indicate recoverable errors related to individual targets. It also looks like download() method does not honor `retries` config option. I'm thinking about creating a separate issue for this.

dnf5-plugins/reposync_plugin/reposync.cpp
```diff
@@ -0,0 +1,420 @@
...
+    if (arch_option.contains("src")) {
+        base.get_repo_sack()->enable_source_repos();
+    }
+
+    libdnf5::repo::RepoQuery repos_query(base);
+    repos_query.filter_enabled(true);
+
+    if (norepopath_option->get_value() && repos_query.size() > 1) {
+        throw libdnf5::cli::ArgumentParserConflictingArgumentsError(
+            M_("Can't use --norepopath with multiple repositories enabled"));
```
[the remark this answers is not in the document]
m-blaha: Thanks, fixed.

dnf5-plugins/reposync_plugin/reposync.cpp
m-blaha: Removed.

PR #1917, opened 2024-12-02
doc: Use PGP instead of GPG

ppisar:
The public-key technology is called Pretty Good Privacy. GPG is a synonymous for GnuPG, a piece of software which DNF5 does not use.

### DIFF
--- a/dnf5/commands/offline/offline.cpp
+++ b/dnf5/commands/offline/offline.cpp
@@ -384,7 +384,7 @@ void OfflineExecuteCommand::pre_configure() {
     // Additional removal could trigger unwanted changes in transaction.
     ctx.get_base().get_config().get_clean_requirements_on_remove_option().set(false);
     ctx.get_base().get_config().get_install_weak_deps_option().set(false);
-    // Disable gpgcheck entirely, since GPG integrity will have already been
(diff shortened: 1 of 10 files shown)


pmatilai:
Erm. PGP is the name of a proprietary implementation, just like GnuPG is the name of one open-source implementation.
OpenPGP is the name *technology*.

ppisar:
OpenPGP is a name invented for the purpose of IETF standardization coming years after creation of PGP. See <https://www.ietf.org/rfc/rfc1991.html>. It's like insisting on calling ODF "Open Document Format for Office Applications".

pmatilai:
Sure the OpenPGP name was invented later. 
It doesn't change the fact that PGP is the name of the  *proprietary program*, still sold even. I wouldn't want us to be associated with THAT.

pmatilai:
Besides, PGP is a trademard: https://trademarks.justia.com/744/67/pgp-74467812.html

That's one of the reasons people changed their PGP references to GPG back in the day. It wasn't much better, but at least GnuPG folks wouldn't come sue you.

You really don't want to go to using PGP, that's nothing but asking for trouble. There's a proper name for the standard, we should use it.

PR #1921, opened 2024-12-04
Fix libdnf5 actions plugin sign conversion compilation err

jrohel:
Fix compilation error:
error: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]


ppisar (approved):
Good.

PR #1929, opened 2024-12-05
Python API: Method `DownloadCallbacks.add_new_download` can return `None`

jrohel:
It's a backwards compatibility API.

Originally, return value support was not implemented in SWIG bindings for the `add_new_download` method and the method only accepted `None` as a return value.

Later, support was implemented for returning an integer, which is then passed in subsequent callbacks (`progress`, `end`, ...) as the `user_cb_data` argument. By expecting an integer instead of `None`, existing programs returning None are broken.

This PR adds the ability for Python programs to return None (it has the same meaning as returning the integer 0 in this case).

### DIFF
--- a/bindings/libdnf5/repo.i
+++ b/bindings/libdnf5/repo.i
@@ -61,10 +61,15 @@
 %typemap(directorout, noblock=1) void * {
     int swig_val;
     int swig_res = SWIG_AsVal_int($1, &swig_val);
-    if (!SWIG_IsOK(swig_res)) {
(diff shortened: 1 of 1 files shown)


ppisar:
That could work.

evan-goode (approved):
I confirm this fixes the Lorax bug: https://bodhi.fedoraproject.org/updates/FEDORA-2024-de59b25ef6.

evan-goode:
This fixes a backwards-incompatibility introduced in https://github.com/rpm-software-management/dnf5/pull/1849.

PR #1930, opened 2024-12-05
changes_from_dnf4: fix formatting of indented `list` points

kontura:
Currently the points are all rendered on a single line: https://dnf5.readthedocs.io/en/latest/changes_from_dnf4.7.html


ppisar (approved):
Good. Verified locally that it fixes the formatting.

PR #1931, opened 2024-12-05
doc: Use OpenPGP instead of PGP

ppisar:
"PGP" is a trademark and a name of the original implementation. Contemporary standard (RFC 4880) calls the technology "OpenPGP" instead.

This patch changes a documentation, the comments, and the messages. It keeps function and variable names untouched. Mainly because it would mean breaking API and ABI.

Related: https://github.com/rpm-software-management/dnf5/pull/1917


evan-goode (approved):
Thanks, I agree. OpenPGP is the correct name for the standardized interface, although it may be less familiar to users than "PGP" or "GPG".

evan-goode:
I fixed one more occurrence.

PR #1932, opened 2024-12-05
Update pre-commit hooks to latest versions in F41

evan-goode:
Update clang-format and autopep8 to the latest versions in Fedora 41. Update pre-commit-hooks to the latest upstream version.

autopep8 update was needed to fix https://github.com/hhatto/autopep8/issues/720.


kontura (approved):
Thanks!

PR #1933, opened 2024-12-05
Python API: `add_new_download` return `None` regression test

evan-goode:
`DownloadCallbacks.add_new_download` can now return either `None` or an `int`. This adds a regression test to check that `add_new_download` can still return `None`.


jrohel (changes requested):
Thank you for the unit test.

Please remove the unnecessary `test_instance` class variable.

evan-goode:
Done and rebased.

jrohel (approved):

jrohel:
Thanks

PR #1925, opened 2024-12-04
`MultiProgressBar` fixes and tests

kontura:
Fix couple of issues for `MultiProgressBar` printing.

Add testing for `MultiProgressBar` in interactive mode.

Issue #1865, opened 2024-11-15
Invalid UTF-8 sequence in trimmed scriptlet output


dansimko:
We have started to observe [Testing Farm jobs failing](https://artifacts.dev.testing-farm.io/ea6cae4d-85a8-493e-9824-363cf483b7e5/) with Fedora 41 and Rawhide failing on trimmed DNF output containing a `0xe2` byte, which cannot be interpreted as a valid UTF-8 sequence.

Relevant log excerpt:
```
[148/149] Installing python3-libdnf5-0: 100% | 129.3 MiB/s |   8.0 MiB |  00m00s
[149/149] Installing butane-0:0.22.0-1. 100% |   1.2 MiB/s |   7.9 MiB |  00m06s
>>> Running post-transaction scriptlet: libvirt-daemon-log-0:10.6.0-5.fc41.x86_6
>>> Finished post-transaction scriptlet: libvirt-daemon-log-0:10.6.0-5.fc41.x86_
>>> Scriptlet output:
>>> Created symlink '/etc/systemd/system/sockets.target.wants/virtlogd.socket' \xe2
>>> Created symlink '/etc/systemd/system/sockets.target.wants/virtlogd-admin.soc
>>> 
>>> Running post-transaction scriptlet: libvirt-daemon-lock-0:10.6.0-5.fc41.x86_
>>> Finished post-transaction scriptlet: libvirt-daemon-lock-0:10.6.0-5.fc41.x86
```

The issue appears to be related to changes made in dnf5 5.2.7.0's to scriptlet output printing.

jan-kolarik:
This might already be fixed by https://github.com/rpm-software-management/dnf5/pull/1936, but since it's currently workeded in the clients, it's hard to verify. A reproducer would be needed.

PR #1936, opened 2024-12-06
Add padding to ProgressBar messages to avoid overlapping

kontura:
Add padding to fully fill the terminal_width, this is because `MultiProgressBar` overrides its own messages, it doesn't clear the lines. If the message is short some leftover characters could be still present after it.

The combination of adding/removing scriptlet messages and just overwriting `MultiProgressBar` lines is making this quite complicated.

Another approach to fix the issues could be reverting https://github.com/rpm-software-management/dnf5/commit/8fbbeb20f517f6bb160c0bc6592e3725f041d2c7 it would make our code simpler but I think this should still be more efficient though I have not tested it.

Follow up for: https://github.com/rpm-software-management/dnf5/pull/1925

### DIFF
--- a/libdnf5-cli/progressbar/download_progress_bar.cpp
+++ b/libdnf5-cli/progressbar/download_progress_bar.cpp
@@ -20,6 +20,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5-cli/progressbar/download_progress_bar.hpp"
 
+#include "../utils/utf8.hpp"
+
 #include "libdnf5-cli/tty.hpp"
 
 #include <algorithm>
@@ -181,8 +183,10 @@ void DownloadProgressBar::to_stream(std::ostream & stream) {
         auto message_type = msg.first;
         auto message = msg.second;
 
+        const auto & prefix = ">>> ";
+
         stream << std::endl;
-        stream << ">>> ";
+        stream << prefix;
 
         color_used = false;
         if (tty::is_interactive()) {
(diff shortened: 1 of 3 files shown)


libdnf5-cli/progressbar/download_progress_bar.cpp
```diff
@@ -205,6 +205,12 @@ void DownloadProgressBar::to_stream(std::ostream & stream) {
             }
         }
 
+        // Add padding to fully fill the terminal_width, this is because MultiProgressBar
+        // overrides its own messages, it doesn't clear the lines.
+        // If the message is short some leftover characters could be still present after it.
+        if (message.length() < terminal_width - 4) {
```
evan-goode, line 211: `message.length()` does not always equal the displayed width of the message, e.g. message=`もで 諤奯ゞ`. I tried with 92ab80c2520e96a9e1dbcc10db536d74aa0a6bdf to add support for wide characters using our existing `libdnf5::cli::utils::utf8::width` helper but it's not quite correct:

1. I introduced a `double free or corruption` fault in the libdnf_cli tests, which seems to be triggered by calling `setlocale(LC_ALL, "C.UTF-8")` in `ProgressbarInteractiveTest::setUp`. Adding that call on the current main branch (without this PR's changes) also triggers the double free. I'm not sure what the issue is.
2. See https://github.com/rpm-software-management/dnf5/commit/92ab80c2520e96a9e1dbcc10db536d74aa0a6bdf#r150174705.

Thought I would share my progress and see if you could make anything of this.

m-blaha: Also loosely connected to https://github.com/rpm-software-management/dnf5/issues/1865 - trimming the message to the terminal width in the middle of a UTF multibyte character.

kontura: That is a good point, I like your improvements.

Interestingly I cannot reproduce the `double free or corruption` it is working for me fine.
I tried running the `libdnf5-cli` tests through `valgrind` and apart from an unrelated `versionlock_config` issue I didn't see any problems.

evan-goode: Interesting, want to cherry-pick https://github.com/rpm-software-management/dnf5/commit/92ab80c2520e96a9e1dbcc10db536d74aa0a6bdf here then? Then we can see whether the Packit CI tests will reproduce the double-free.

kontura: It worked here even with you commit.

evan-goode: Great, I think this is good to merge then.

evan-goode (approved):

Issue #1938, opened 2024-12-06
dnf remove doesn't show total space freed


Dapid:
I find useful to know how much space will be cleared when removing a package. `dnf` obviously knows it, since it is displaying it per package, but it would be nice to have a tally at the end, like in version 4:

```
# dnf remove python3.12
Package                                                         Arch           Version                                                          Repository                               Size
Removing:
 python3.12                                                     x86_64         3.12.7-1.fc41                                                    fedora                               31.5 KiB
Removing unused dependencies:
 python3.12-libs                                                x86_64         3.12.7-1.fc41                                                    fedora                               41.3 MiB
 python3.12-tkinter                                             x86_64         3.12.7-1.fc41                                                    fedora                                2.2 MiB

Transaction Summary:
 Removing:           3 packages
<---- dnf4 would tell me here how much I am freeing
Is this ok [y/N]:
```

ppisar:
I confirm this issue. The bug is somewhere in print_transaction_size_stats() in dnf5/main.cpp.

PR #1940, opened 2024-12-06
Fix reporting disk space to be freed on a pure package removal

ppisar:
There was a bug in testing whether a difference of an installation size and a removal size fits into int64_t. E.g. on a pure package removal, the difference was negative, overflowing resulting unsigned type, leading to a very large positive value, larger than could fit into int64_t. As a result an existing range check failed and no report was printed.

This patch fixes it by preventing from a subtraction with a negative rusult.

(A fix which would compute the difference before the branching would either loose precision, or decide the branching twice.)

Resolves: #1938

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -814,16 +814,15 @@ static void print_transaction_size_stats(Context & context) {
             dwnl_pkgs_size_unit));
     }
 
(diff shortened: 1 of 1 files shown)


ppisar:
@kontura, after a rebase, all tests pass.

kontura (approved):

PR #1948, opened 2024-12-11
Fix plural typo

buttonfly1000:
`Replacing:       {:4} package` -> `Replacing:       {:4} packages`


ppisar:
Thanks for the patch. The fix in the code is good. However, we do not edit translation catalogs. POT files are generated automatically and PO files exclusively by translators. Could you please remove from this pull request changes in the catalogs?

buttonfly1000:
> Thanks for the patch. The fix in the code is good. However, we do not edit translation catalogs. POT files are generated automatically and PO files exclusively by translators. Could you please remove from this pull request changes in the catalogs?

Thanks for the guidance. I removed the changes in the catalogs.

PR #1962, opened 2024-12-17
Incorrect library name in libdnf5-cli.pc

m-blaha:
SSIA


ppisar (approved):
Good. I will merge it after finishing a CI job.

Issue #1951, opened 2024-12-13
dnf5-automatic command_format does not pass `{body}` anymore


Szpadel:
With dnf5-automatic `{body}` inside `command_format` is not replaced with update summary.

`/etc/dnf/automatic.conf`:
```
[command]
command_format = "curl https://example.com -F title=dnf-automatic -F message={body}"
```

Results in `{body}` message being submitted instead of update content.

This looks like regression introduced with dnf5, as it worked correctly with dnf4

evan-goode:
Thank you for filing an issue! https://github.com/rpm-software-management/dnf5/pull/1972 should fix this.

PR #1972, opened 2024-12-20
automatic: Substitute variables in command_format

evan-goode:
`{body}` was not substituted correctly in `command_format` for the command emitter.

Resolves https://github.com/rpm-software-management/dnf5/issues/1951


ppisar (approved):
Good.

Issue #667, opened 2023-06-27
"dnf5 history" cannot search by a package name


j-mracek:
it seems like as of right now this is partially implemented, but not fully. notably:

    [adamw@xps13a blivet (prefer-gpt-x86_64)]$ dnf history list ostree
    Invalid transaction ID range "ostree", "ID" or "ID..ID" expected, where ID is "NUMBER", "last" or "last-NUMBER".
    [adamw@xps13a blivet (prefer-gpt-x86_64)]$ 

i.e. you can't do `history list` (or, in fact, `history info`) by package name. It seems to only expect history transaction IDs. querying by package name is very useful and missing it is somewhat significant.

_Originally posted by @AdamWill in https://github.com/rpm-software-management/dnf5/issues/140#issuecomment-1608319391_

j-mracek:
I am suggesting to modify the syntax with DNF5 to not guess what is the argument (name or transaction id).

In dnf4
`dnf history list 1` means to list transaction with id 1 (default) or list all transactions with package named 1. `1` is a valid name for a package.

ppisar:
Downstream bug report <https://bugzilla.redhat.com/show_bug.cgi?id=2083479>.

belegdol:
Without the ability to show a history for a specific package, the functionality is very limited. One cannot be reasonably expected to manually list the details for dozens of transactions, just to see if the package happened to be among the ones updated.

moschlegbz:
I totally second this, this feature was in dnf4 and was removed in dnf5. 
I wrote a script just to get around this, but it's very inefficient because I have to run "dnf history info <num>" on each transaction id in the history, check if "package name" is one of the packages and if so display the information.  Obviously a "dnf history list <package>" and "dnf history info <package>" would be much faster and better.

PR #1980, opened 2025-01-03
Add --contains-pkgs=.. option to history list and info

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/667

### DIFF
--- a/dnf5/commands/history/arguments.hpp
+++ b/dnf5/commands/history/arguments.hpp
@@ -70,6 +70,16 @@ public:
 
 std::function<std::vector<std::string>(const char * arg)> create_history_id_autocomplete(Context & ctx);
 
+class HistoryContainsPkgsOption : public libdnf5::cli::session::AppendStringListOption {
+public:
+    explicit HistoryContainsPkgsOption(libdnf5::cli::session::Command & command)
+        : AppendStringListOption(
+              command,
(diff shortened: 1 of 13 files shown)


kontura:
I updated the PR to keep the current `history info` behavior (showing only the last transaction) but I have added a specific hint when filtered out.

It still bothers me a little that all the other dnf5 `info` subcommands default to printing all elements so I am open to discussing this further.

m-blaha:
> I updated the PR to keep the current `history info` behavior (showing only the last transaction) but I have added a specific hint when filtered out.
> 
> It still bothers me a little that all the other dnf5 `info` subcommands default to printing all elements so I am open to discussing this further.

Thank you! I agree that the behavior of `dnf5 history info` is a bit confusing. It’s not consistent with other `* info` commands, nor is it compatible with what `dnf4` used to do.

It might be better to sacrifice compatibility with `dnf4 info` and have `dnf5 history info` default to showing information about all transactions. I do understand that the `dnf4 info` behavior, which prints only the last transaction, is quite handy. However, users would just need to adapt to using `dnf5 history info last` instead, which still sounds sane enough. This way we will at least keep consistency across all `* info` commands.

m-blaha (approved):
Thank you!

Issue #1956, opened 2024-12-15
dnf5 bash completion does not include nevra information like dnf4


grumpey:
With dnf4: 
```
sudo dnf4 install kernel-devel-
kernel-devel-6.11.10-300.fc41.x86_64          kernel-devel-matched-6.11.10-300.fc41.x86_64  
kernel-devel-6.11.4-301.fc41.x86_64           kernel-devel-matched-6.11.4-301.fc41.x86_64   
```
```
sudo dnf4 install hexchat-
hexchat-2.16.2-5.fc41.i686           hexchat-autoaway-2.0-15.fc41.x86_64  hexchat-devel-2.16.2-5.fc41.x86_64   
hexchat-2.16.2-5.fc41.x86_64         hexchat-devel-2.16.2-5.fc41.i686
```
```
 sudo dnf4 install kernel-core-6.11.
kernel-core-6.11.10-300.fc41.x86_64  kernel-core-6.11.4-301.fc41.x86_64
```

With dnf5: 
`sudo dnf install kernel-devel-matched`
`sudo dnf install kernel-core`

```
sudo dnf install hexchat-
hexchat-autoaway  hexchat-devel

```

With dnf5 if you add the version information it is removed when you try to tab to complete.

jrohel:
Current status of bash completion in dnf5

If we are working with installed packages - we want to remove (remove), reinstall (reinstall), swap (swap) there are two cases:
1. usually only **one package of a given name is installed** and bash completion **completes only the name**
2. **for multiple installed packages of the same name** (installonly packages - usually kernel) the **nevras are displayed** so that the user can choose which one he wants to work with.

Examples (some packages are installed in multiple versions):
```
# dnf5 remove kernel-
kernel-core-0:6.11.10-300.fc41.x86_64           kernel-modules-core-0:6.11.11-300.fc41.x86_64
kernel-core-0:6.11.11-300.fc41.x86_64           kernel-modules-core-0:6.12.6-200.fc41.x86_64
kernel-core-0:6.12.6-200.fc41.x86_64            kernel-modules-extra-0:6.11.10-300.fc41.x86_64
kernel-headers                                  kernel-modules-extra-0:6.11.11-300.fc41.x86_64
kernel-modules-0:6.11.10-300.fc41.x86_64        kernel-modules-extra-0:6.12.6-200.fc41.x86_64
kernel-modules-0:6.11.11-300.fc41.x86_64        kernel-srpm-macros
[...]

PR #1983, opened 2025-01-03
Bash completion: Always offer NEVRAs and NAMEs for packages

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1956

Originally, for packages, the NAME was offered in bash completion. In the case of multiple packages with the same NAME, there was an option to offer NEVRA instead of a NAME for those packages.

Now, bash completion always offers NEVRA for packages. There is an option to offer package NAME in addition to NEVRAs. This is useful, for example, when the user wants to install a package. Usually the user only wants to enter the package NAME and not the entire NEVRA. If multiple packages with the same NAME are available, the specific
package is then selected by the solver.

The DNF5 aplication now always offers both NEVRAs and NAMEs.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -1073,7 +1073,7 @@ std::vector<std::string> match_specs(
     bool installed,
     bool available,
     bool paths,
-    bool nevra_for_same_name,
+    bool only_nevras,
     const char * file_name_regex) {
     auto & base = ctx.get_base();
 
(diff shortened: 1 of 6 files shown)


evan-goode:
Nice, this works. 

> In the case of multiple packages with different nevras but the same name, there is an option to offer a name in addition to nevras for those packages.

I think I'd prefer also completing the name even if there is only one NEVRA available, i.e.

```
$ dnf5 --complete=2 dnf5 install hell
hello
hello-0:2.12.1-5.fc41.x86_64
```

mostly for the sake of consistency. It keeps the shell history a bit cleaner if the command is completed just as `dnf5 install hello`, and the user might intuitively expect a completion of just the package name and not the full NEVRA.

jrohel:
> I think I'd prefer also completing the name even if there is only one NEVRA available

OK. Sounds like a good idea. I've simplified the logic. Now NAMEs are always offered in addition to NEVRAs.

dnf5/context.cpp
```diff
@@ -1128,19 +1128,15 @@ std::vector<std::string> match_specs(
...
-                    matched_pkgs_query.remove(pkg);
+            if (!only_nevras) {
+                if (auto name = package.get_name(); name.length() >= pattern.length()) {
+                    // The output must not include a package name shorter than the length of the input patter
+                    // (we don't want to shorten the user-specified input).
+                    result_set.insert(std::move(name));
                 }
             }
+
+            result_set.insert(package.get_full_nevra());
```
m-blaha, line 1139: Does it make sense to skip the source rpm packages NEVRAS?

jrohel: > Does it make sense to skip the source rpm packages NEVRAS?

Now, all matching packages **from all allowed repositories** are offered. Including source packages.
Fedora source packages are in their own repository, which is disabled by default. If the source package repository is enabled, source packages are offered.

Examples:
```
# dnf5 download iftop<TAB><TAB>
iftop                              iftop-0:1.0-0.34.pre4.fc41.x86_64  
```
```
# dnf5 --enable-repo=fedora-source download iftop<TAB><TAB>
iftop                              iftop-0:1.0-0.34.pre4.fc41.src     iftop-0:1.0-0.34.pre4.fc41.x86_64  
```

What is meant by the question? Do you want the source packages to not be offered? It is true that NEVRAs for source packages only make sense for certain commands - for example, `download`.

m-blaha: I discovered it i copr repository, which contains both binary and source rpm. I was just thinking whether *.src packages are usable for the user. But it's really niche issue, as you said - most of the repositories do not contain source packages.

evan-goode: I think it's fine that .src.rpms are suggested too, the user probably knows what they're doing if they select a specific NEVRA to install/remove/download.

Issue #1982, opened 2025-01-03
Bash completion behaves incorrectly if there is a colon ':' in the word


jrohel:
Example:
`# dnf5 remove kernel-core-0:6.<TAB>`
result is:
`# dnf5 remove kernel-core-0:kernel-core-0:6.1`
instead of:
`# dnf5 remove kernel-core-0:6.1`

In other words, the prefix with the colon is repeated.

PR #1984, opened 2025-01-03
Fix colon in bash completion

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1982

Word completion may behave improperly if there is a colon in the word to be completed. The colon can be special to readline's word completion code: it can be one of the characters that breaks words for the completer.
    
The current set of completion word break characters is available in bash as the value of the `COMP_WORDBREAKS` variable.  Removing `:` from that value is enough to make the colon not special to completion: `COMP_WORDBREAKS=${COMP_WORDBREAKS//:}`

This fix is workaround for cases where `:` is a break character (it is contained in `COMP_WORDBREAKS`). A side effect in this case is the removal of the prefix containing the colon in the list of bash completion suggestions.

### DIFF
--- a/dnf5/bash-completion/dnf5
+++ b/dnf5/bash-completion/dnf5
@@ -14,6 +14,19 @@ _do_dnf5_completion()
     fi
 
     mapfile -t COMPREPLY <<<$("${1}" "--complete=${cword}" "${words[@]}")
+
+    # In Bash, with a colon in COMP_WORDBREAKS, words containing colons are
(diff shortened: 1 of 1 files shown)


evan-goode (approved):
Awesome, I ran into this while testing https://github.com/rpm-software-management/dnf5/pull/1983 but assumed it was a bug in my Bash configuration. I think this is a reasonable workaround. LGTM.

PR #1986, opened 2025-01-05
automatic: fix documentation on automatic.conf location in sample config

### LINKED ISSUE #1770: Please document order of precedence for automatic.conf
I had some unexpected behavior (updates were being downloaded and not applied) on a system and it looks like it came down to having both `/etc/dnf/automatic.conf` and `/etc/dnf/dnf5-plugins/automatic.conf`

`/usr/share/dnf5/dnf5-plugins/automatic.conf` only talks about `/etc/dnf/dnf5-plugins/automatic.conf`.
```
# This configuration file is managed by the dnf5-plugin-automatic package.
# Please do not edit it. To make changes in dnf5 automatic configuration
# edit /etc/dnf/dnf5-plugins/automatic.conf and make your adjustments there.
```
https://github.com/rpm-software-management/dnf5/blob/main/dnf5-plugins/automatic_plugin/config/usr/share/dnf5/dnf5-plugins/automatic.conf#L3

Meanwhile the man page only talks about `/etc/dnf/automatic.conf`
```
The operation of the tool is controlled by configuration files. Default values are set from ``/usr/share/dnf5/dnf5-plugins/automatic.conf`` config file. Host-specific overrides from ``/etc/dnf/automatic.conf`` are then applied.
```
https://github.com/rpm-software-management/dnf5/blob/main/doc/dnf5_plugins/automatic.8.rst?plain=1#L36

On some freshly installed systems I copied `/usr/share/dnf5/dnf5-plugins/automatic.conf` to `/etc/dnf/dnf5-plugins/automatic.conf` modified it and everything worked fine.

On the problematic one I did an upgrade and am not sure if automatic.conf was left over from dnf 4 automatic or I got my wires crossed reading the docs (I suspect left over since the modify date is April), but either way it led to confusion, as it seems like `/etc/dnf/automatic.conf` takes precedence over `/etc/dnf/dnf5-plugins/automatic.conf`.

exploide:
In 73ed42f the preferred location of the `automatic.conf` file changed from `/etc/dnf/dnf5-plugins/automatic.conf` to `/etc/dnf/automatic.conf`. However, the sample config file still had a comment that changes should go to the config file in the old location. This PR fixes the comment for consistency.

fixes #1770

### DIFF
--- a/dnf5-plugins/automatic_plugin/config/usr/share/dnf5/dnf5-plugins/automatic.conf
+++ b/dnf5-plugins/automatic_plugin/config/usr/share/dnf5/dnf5-plugins/automatic.conf
@@ -1,6 +1,6 @@
 # This configuration file is managed by the dnf5-plugin-automatic package.
 # Please do not edit it. To make changes in dnf5 automatic configuration
-# edit /etc/dnf/dnf5-plugins/automatic.conf and make your adjustments there.
+# edit /etc/dnf/automatic.conf and make your adjustments there.
 
 [commands]
 # Whether updates should be applied when they are available, by


kontura:
Nice, thank you.

Issue #1893, opened 2024-11-24
libdnf5::rpm::VersionlockCondition::VALID_COMPARATORS overlinkinkg in libdnf5-cli test violates ODR


Conan-Kudo:
When packaging dnf 5.2.7.0, I observed that the tests crash:

```
[  474s] + /usr/bin/ctest --output-on-failure --force-new-ctest-process -j4
[  474s] Test project /home/abuild/rpmbuild/BUILD/dnf5-5.2.7.0/build
[  474s]       Start  1: test_libdnf
[  476s]  1/17 Test  #1: test_libdnf .................................................   Passed    2.45 sec
[  476s]       Start  2: test_libdnf_cli
[  476s]       Start  8: test_python3_libdnf_cli
[  476s]  2/17 Test  #8: test_python3_libdnf_cli .....................................   Passed    0.06 sec
[  476s]  3/17 Test  #2: test_libdnf_cli .............................................***Exception: SegFault  0.06 sec
[  476s] UTF8Test::test_length_en (duration: 0ms) : OK
[  476s] UTF8Test::test_length_cs (duration: 0ms) : OK
[  476s] UTF8Test::test_length_cn (duration: 0ms) : OK
[  476s] UTF8Test::test_length_ja (duration: 0ms) : OK
[  476s] UTF8Test::test_width_en (duration: 0ms) : OK
[  476s] UTF8Test::test_width_cs (duration: 0ms) : OK
[  476s] UTF8Test::test_width_cn (duration: 0ms) : OK
[  476s] UTF8Test::test_width_ja (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_length_en (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_length_cs (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_length_cn (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_length_ja (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_width_en (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_width_cs (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_width_cn (duration: 0ms) : OK
[  476s] UTF8Test::test_substr_width_ja (duration: 0ms) : OK
[  476s] ProgressbarTest::test_download_progress_bar (duration: 0ms) : OK
[  476s] ProgressbarTest::test_multi_progress_bar (duration: 0ms) : OK
[...]

PR #1988, opened 2025-01-06
Copr plugin: Fix resource leak in load_all_configuration

jrohel:
I discovered a resource leak while working on https://github.com/rpm-software-management/dnf5/issues/1893.
There was a missing `globfree` function call in the `CoprConfig::load_all_configuration()` method.

One can say that adding the line `globfree(&glob_result);` after the `for` loop is enough. But it's not true. The `load_copr_config_file` method called in the loop can generate an exception. So I rewrote the code to use `std::filesystem::directory_iterator`.

### DIFF
--- a/dnf5-plugins/copr_plugin/copr_config.cpp
+++ b/dnf5-plugins/copr_plugin/copr_config.cpp
@@ -24,7 +24,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "libdnf5/utils/os_release.hpp"
 
+#include <cstdlib>
 #include <filesystem>
+#include <set>
 
 
 namespace dnf5 {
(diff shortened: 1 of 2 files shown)


dnf5-plugins/copr_plugin/copr_config.cpp
```diff
@@ -49,12 +51,23 @@ void CoprConfig::load_all_configuration() {
...
+
+    // Load configuration files from drop-in directory
+    {
+        // Create a set of configuration files sorted by name from the drop-in directory
+        const auto drop_in_dir = etc_dir / "dnf/plugins/copr.d/";
+        std::set<std::filesystem::path> drop_in_files;
+        std::error_code ec;
+        for (const auto & dentry : std::filesystem::directory_iterator(drop_in_dir, ec)) {
+            const auto & path = dentry.path();
+            if (dentry.is_regular_file() && path.extension() == ".conf") {
```
ppisar, line 63: I worry that is_regular_file() ignores symlinks. That would be a regression comparing to glob().

kontura: I believe symlinks are not ignored.

The [doc](https://en.cppreference.com/w/cpp/filesystem/directory_entry/is_regular_file) says that its effectively `std::filesystem::is_regular_file(status())` and [status](https://en.cppreference.com/w/cpp/filesystem/status) says that symlinks are followed.

Issue #1919, opened 2024-12-03
If baseurl contains a curly bracket, HTTP server returns 403 and skip_if_unavailable=true, "dnf makecache" crashes


ppisar:
With this repository definition:

~~~~
[foo]
baseurl = http://localhost/{}
enabled = 1
gpgcheck = 0
name = foo
skip_if_unavailable = 1
~~~~

DNF5 crashes:

~~~~
# ~test/dnf5/redhat-linux-build/dnf5/dnf5 --repo foo makecache
Updating and loading repositories:
 foo                                                           ???% |   0.0   B/s |   0.0   B |  00m00s
>>> Curl error (7): Could not connect to server for http://localhost/{}/repodata/repomd.xml [Failed to
>>> Curl error (7): Could not connect to server for http://localhost/{}/repodata/repomd.xml [Failed to
>>> Curl error (7): Could not connect to server for http://localhost/{}/repodata/repomd.xml [Failed to
>>> Curl error (7): Could not connect to server for http://localhost/{}/repodata/repomd.xml [Failed to
>>> Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tr
terminate called without an active exception
Aborted (core dumped)
~~~~

GDB shows two theads. A thread terminating with std::thread destructor, just when libdnf5::repo::RepoSack::Impl::update_and_load_repos() reaches an end of its body:

~~~~
(gdb) bt
#0  0x00007ffff7080b44 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff70272de in raise () at /lib64/libc.so.6
#2  0x00007ffff700f3d4 in abort () at /lib64/libc.so.6
#3  0x00007ffff7209069 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] ()
    at /lib64/libstdc++.so.6
#4  0x00007ffff721b36c in __cxxabiv1::__terminate(void (*)()) () at /lib64/libstdc++.so.6
#5  0x00007ffff7208c11 in std::terminate() () at /lib64/libstdc++.so.6
#6  0x00007ffff76adb7d in std::__terminate ()
    at /usr/include/c++/14/x86_64-redhat-linux/bits/c++config.h:2168
#7  std::thread::~thread (this=0x7fffffffcb20, __in_chrg=<optimized out>)
    at /usr/include/c++/14/bits/std_thread.h:182
[...]

PR #1993, opened 2025-01-08
repo: Fix logging metadata download errors handling

m-blaha:
Previously, if a download error message line contained `{}`,
logger.warning would interpret it as a format string without any
arguments passed, causing the libdnf5::utils::sformat function to crash.
This commit addresses the issue by ensuring error messages are handled
safely.

Fixes: https://github.com/rpm-software-management/dnf5/issues/1919


ppisar (approved):
Good.

PR #1994, opened 2025-01-08
Docs: Update template code, documentation page for the DNF5 Command Template

ferdnyc:
Recently on Fedora Discussions, [it was brought up / discovered](https://discussion.fedoraproject.org/t/dnf5-command-template-seems-to-have-bugs/141526) that the template source code in the DNF5 documentation had fallen a bit out of date, to the point where it wasn't even compilable because the internal APIs have changed so much in two years.

Since it actually says, right at the top of that page:

> #### Note
> This code is thought to be self explanatory. You should be able to copy the snippets, following the directory structure and naming shown above each one.

...It seemed clear that it was no longer remotely achieving that goal.

So, I went through and updated all of the included source code based on the current `dnf5/commands/` source and `dnf5/main.cpp`, built and tested the template source integrated into DNF5 (as a new `dnf5/commands/template/` directory, just like the note says), and also updated the sample output at the end to reflect reality.

I hope to do the same for the other two template documents, as well, but we'll see what happens. I wanted to at least submit this one for now, and I'll take the rest one step at a time.

### DIFF
--- a/doc/templates/command/template.cpp
+++ b/doc/templates/command/template.cpp
@@ -6,6 +6,13 @@ namespace dnf5 {
 
 using namespace libdnf5::cli;
 
+void TemplateCommand::set_parent_command() {
+    auto * arg_parser_parent_cmd = get_session().get_argument_parser().get_root_command();
+    auto * arg_parser_this_cmd = get_argument_parser_command();
+    arg_parser_parent_cmd->register_command(arg_parser_this_cmd);
+    arg_parser_parent_cmd->get_group("software_management_commands").register_argument(arg_parser_this_cmd);
+}
+
 void TemplateCommand::set_argument_parser() {
     // Context is the main object in dnf5.
     // It contains useful functions and pieces of information necessary to run
@@ -22,7 +29,7 @@ void TemplateCommand::set_argument_parser() {
     //
     // Set a description that would be displayed in the second column of the
     // help message would also add the command to the parser
-    cmd.set_short_description("A command that prints its name and arguments' name");
+    cmd.set_description("A command that prints its name and arguments' name");
 
     // Add foo and bar options
     //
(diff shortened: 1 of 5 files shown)


kontura (commented):
I have noticed only one minor detail.

This is really great, thank you!

kontura:
Plus it seems there is a couple formatting issues, see the `Pre Commit` action.

doc/templates/template-command.rst
```diff
@@ -75,7 +75,7 @@ Following this example you should have an output like this.
 
     $ dnf5 template --help
     Usage:
-      dnf5 template [GLOBAL OPTIONS] template [ARGUMENTS]
+      dnf5 [GLOBAL OPTIONS] template [ARGUMENTS]
```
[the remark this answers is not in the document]
ferdnyc: @kontura Yup, you called it, it should be `[OPTIONS]`.

doc/templates/command/template.cpp
```diff
@@ -22,31 +29,33 @@ void TemplateCommand::set_argument_parser() {
...
     // Add foo and bar options
     //
     // Option 1: as said in the header file here you should handle the pointer
     // by giving up the ownership to dnf5's parser.
     // Set the default value here.
     foo_option = dynamic_cast<libdnf5::OptionBool *>(
-        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+        parser.add_init_value(
+            std::unique_ptr<libdnf5::OptionBool>(
+                new libdnf5::OptionBool(false))));
```
ferdnyc, line 42: @kontura My only problem with this (which the pre-commit formatter wants to change back to the un-wrapped format:)

```c++
    foo_option = dynamic_cast<libdnf5::OptionBool *>(
￼       parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
```

...is that it won't fit when it's embedded on the documentation page, without forcing the reader to scroll horizontally.

Is there an override comment that can stop the formatter from re-wrapping it? I really think it reads better, for the documentation, if the line length is kept down to fit the page format.

doc/templates/command/template.cpp
```diff
@@ -6,6 +6,13 @@ namespace dnf5 {
 
 using namespace libdnf5::cli;
 
+void TemplateCommand::set_parent_command() {
+    auto * arg_parser_parent_cmd = get_session().get_argument_parser().get_root_command();
+    auto * arg_parser_this_cmd = get_argument_parser_command();
+    arg_parser_parent_cmd->register_command(arg_parser_this_cmd);
+    arg_parser_parent_cmd->get_group("software_management_commands").register_argument(arg_parser_this_cmd);
```
ferdnyc, line 13: I'd also prefer to be able to wrap this line, come to think of it.

ferdnyc: Although I suppose I could change it to this:

```suggestion
    auto group = arg_parser_parent_cmd->get_group("software_management_commands");
    group.register_argument(arg_parser_this_cmd);
```

ferdnyc: I did end up making that change, and I also shortened the rest of the variable names to keep the width down. My thinking being, these are local variables in a 5-line function, keeping track of exactly what they represent isn't really an issue. But, let me know if you feel it's a problem.

ferdnyc:
I have a feeling the formatter is still going to complain about that one line.

doc/templates/command/template.cpp
ferdnyc: @kontura Done, thanks! I think that looks much better. I also made some other changes, I'll add some review comments about those.

kontura (approved):
Thank you for working on this, it is much appreciated.

I have created an issue that could help with this in the future: https://github.com/rpm-software-management/dnf5/issues/2016.

PR #1995, opened 2025-01-08
libdnf5-cli unit tests: Use only public API methods

jrohel:
The libdnf5-cli progressbar unit tests unnecessarily used the protected method `DownloadProgressBar::to_stream`. The libdnf5-cli library publicly provides the `<<` operator (API) for `DownloadProgressBar`. This operator does the same thing and internally uses the `to_stream` method.


kontura:
Oh, I see, thank you.

PR #1998, opened 2025-01-09
Document `list` command changed handling of installed packages repos

kontura:
The change happened in `from_repo` attribute of `Package`.

In dnf4 there was: https://github.com/rpm-software-management/dnf/blob/master/dnf/package.py#L82-L94


jan-kolarik (approved):
Thanks, that should be useful for users!

PR #1999, opened 2025-01-09
I18N: Annotate indentation of the transaction summary

ppisar:
Translators are recommend to align the messages to prevent from a broken table:

    Podsumowanie transakcji:
     Instalowanie:         6 pakietów
     Aktualizowanie:         33 pakietów
     Zastępowanie:         48 pakietów
     Usuwanie:         199 pakietów

Relates: https://bugzilla.redhat.com/show_bug.cgi?id=2336561

### DIFF
--- a/libdnf5-cli/output/transaction_table.cpp
+++ b/libdnf5-cli/output/transaction_table.cpp
@@ -128,48 +128,61 @@ public:
         if (installs != 0) {
             std::fputs(
                 libdnf5::utils::sformat(
-                    P_(" Installing:      {:4} package\n", " Installing:      {:4} packages\n", installs), installs)
+                    // Keep the replaceble number aligned across all messages.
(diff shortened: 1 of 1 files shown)


kontura:
I saw there are macros to provide context for translatable messages: https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/utils/bgettext/README.md?plain=1#L56

It is used only on couple places, like: https://github.com/rpm-software-management/dnf5/blob/3fa8d63db9c060e1744cfd268cbba918941d2fd0/libdnf5/base/transaction.cpp#L1373

However I don't know how it will show up to the translators or even if it is better then than normal comments.

ppisar:
Gettext context is used to differentiate two messages which are worded the identically in English, but needs different translation in other languages. Using them for belonging-to-summary-transaction-table would be an abuse and more importantly would not deliver the annotation to the translators.

kontura:
I see, thanks.

Issue #753, opened 2023-07-20
Suppress header in `dnf list --upgrades` in `--quiet` mode


abitrolly:
The line "Available upgrades" is not suppressed when running `dnf list --upgrades --quiet`.

https://github.com/rpm-software-management/dnf5/blob/0019b127039c3c93ae89710a6793fd938005fc2a/dnf5/commands/list/list.cpp#L216

Is there any way to turn it off?

pkratoch:
Sorry for no comment for so long. This issue is valid. The `--quiet` option doesn't work yet, see https://github.com/rpm-software-management/dnf5/issues/570 .

abitrolly:
That's was an unfortunate choice to use CPP instead of more safe and modern Rust or more maintainable Go in 2021. I guess the choice was made to improve the speed, but we all know that it is algorithmic strength and parallelization that gives the best results.

Maybe the CPP choice was to get the minimal footprint, because `dnf5` is obligatory for minimal installs.

As a co maintainer of https://cli.urfave.org/ I could give you a patch in Go, but for CPP I have no idea how to add global flag and check for it here. Lemme ask LLM.

abitrolly:
Asked LLM, but it was unable to discover where `quiet` is defined, so here is is.

https://github.com/rpm-software-management/dnf5/blob/2d956584b88110046a80c85407e90ecd97a6ac6f/dnf5/main.cpp#L150-L163

And here is the commit that added it, with the `print_info` function that takes care of the check.

https://github.com/rpm-software-management/dnf5/commit/0caf87c0c0687784f55fdd20135a649de8c44a66

Need to check if it is still callable.

abitrolly:
`print_info` is callable, but the messed up thing is that the output is made by library! `libdnf5::cli::output::PackageListSections`

That's a poor design to me. How to handle user input and output is the responsibility of user facing client code. The library should just handle the data - JSON or whatever in and out.

abitrolly:
https://github.com/rpm-software-management/dnf5/blob/2d956584b88110046a80c85407e90ecd97a6ac6f/libdnf5-cli/output/package_list_sections.cpp#L41

kontura:
We have decided to reject this functionality.
See https://github.com/rpm-software-management/dnf5/pull/2001 for more details.

abitrolly:
First @pkratoch says that the issue is valid, I spend time, now it is not valid. :(

pkratoch:
[...]

Issue #570, opened 2023-05-28
Option -q (--quiet) does not work


xsuchy:
When I run `dnf5 -q -y`, then the output is still written on stdout. It seems like the -q is completely ignored.

This is a regression from the old DNF.

j-mracek:
Just a question - what `-q` should hide?

DNF4 provides strange outputs with -q
```
sudo dnf install acpi -q
Is this ok [y/N]:
```

xsuchy:
It should hide:

<details>
<summary>output of dnf upgrade</summary>
<pre>
 Copr repo for josm-latest owned by jje 100% |   6.8 KiB/s |   3.6 KiB |  00m01s Fedora 38 - x86_64                     100% |  20.8 KiB/s |  17.9 KiB |  00m01s Copr repo for droidcam owned by meeuw  100% |   3.3 KiB/s |   1.8 KiB |  00m01s Copr repo for beakerlib owned by prais ???% |   0.0   B/s |   0.0   B |  00m00s
>>> Curl error (6): Couldn't resolve host name for http://coprbe.devel.redhat.co
>>> Curl error (6): Couldn't resolve host name for http://coprbe.devel.redhat.co
>>> Curl error (6): Couldn't resolve host name for http://coprbe.devel.redhat.co
>>> Curl error (6): Couldn't resolve host name for http://coprbe.devel.redhat.co
>>> Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.x Copr repo for edgetx-companion-2.5 own 100% |  17.6 KiB/s |   3.3 KiB |  00m00s Copr repo for go-ipfs owned by elxreno 100% |   5.7 KiB/s |   3.3 KiB |  00m01s Copr repo for devel owned by @abrt     100% |  11.9 KiB/s |   3.6 KiB |  00m00s Copr repo for dnf5-unstable owned by r 100% |   1.5 KiB/s |   1.8 KiB |  00m01s Blue Jeans Network, Inc. - x86_64 soft 100% |   3.3 KiB/s |   2.9 KiB |  00m01s Copr repo for swid owned by adelton    100% |   4.8 KiB/s |   3.3 KiB |  00m01s Copr repo for nanoblogger owned by msu 100% |   5.4 KiB/s |   3.3 KiB |  00m01s Copr repo for ddashboard owned by mhla ???% |   0.0   B/s |   0.0   B |  00m00s
[...]

PR #2001, opened 2025-01-12
Make output less vebose in `--quiet` mode

abitrolly:
Makes `--quiet` option description more quiet.

Suppress header in `dnf list --upgrades` and similar one table command.

Closes https://github.com/rpm-software-management/dnf5/issues/753.

### DIFF
--- a/dnf5/commands/list/list.cpp
+++ b/dnf5/commands/list/list.cpp
@@ -178,6 +178,8 @@ void ListCommand::run() {
         config.get_color_list_available_reinstall_option().get_value(),
         config.get_color_list_available_upgrade_option().get_value());
 
+    auto heading = "";
+    
     switch (pkg_narrow) {
         case PkgNarrow::ALL: {
             libdnf5::rpm::PackageQuery available(base_query);
@@ -197,8 +199,9 @@ void ListCommand::run() {
         }
         case PkgNarrow::INSTALLED: {
             if (!ctx.get_quiet()) {
-                package_matched |= sections->add_section(_("Installed packages"), installed);
+                heading = _("Installed packages");
             }
+            package_matched |= sections->add_section(heading, installed);
             break;
         }
         case PkgNarrow::AVAILABLE: {
@@ -208,8 +211,9 @@ void ListCommand::run() {
                 base_query.filter_latest_evr();
             }
             if (!ctx.get_quiet()) {
-                package_matched |= sections->add_section(_("Available packages"), base_query);
+                heading = _("Available packages");
             }
+            package_matched |= sections->add_section(heading, base_query);
             break;
         }
         case PkgNarrow::UPGRADES:
@@ -218,8 +222,9 @@ void ListCommand::run() {
             base_query.filter_arch(std::vector<std::string>{"src", "nosrc"}, libdnf5::sack::QueryCmp::NEQ);
             base_query.filter_latest_evr();
             if (!ctx.get_quiet()) {
-                package_matched |= sections->add_section(_("Available upgrades"), base_query);
+                heading = _("Available upgrades");
             }
+            package_matched |= sections->add_section(heading, base_query);
             break;
         case PkgNarrow::OBSOLETES: {
             base_query.filter_priority();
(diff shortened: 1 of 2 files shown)


kontura:
This is really not the right approach, the sections also print the packages. With your patch it doesn't print anything at all when used with `-q`.

This also didn't work in old dnf.

Can I ask why do you want to hide the headers?

abitrolly:
> Can I ask why do you want to hide the headers?

@kontura I needed to get a list of updates and their versions and these headers were in the way https://gitlab.com/abitrolly/dnf-go-gui/-/blob/master/docs/DNF-API.md

Proper machine readable export, like CSV or JSON will remote the need.

> This is really not the right approach, the sections also print the packages. With your patch it doesn't print anything at all when used with `-q`.

That's unfortunate. I also don't see that patching core lib to handle different verbosity levels is the right thing.

kontura:
> That's unfortunate. I also don't see that patching core lib to handle different verbosity levels is the right thing.

Right, actually we have two libraries: `libdnf5` which is for the core functionality and then `libdnf5-cli` which we use to share command line functionality, formatting and such between tools.

> Proper machine readable export, like CSV or JSON will remote the need.

The list command is not intended to be parsed.
Try repoquery:
```
dnf5 repoquery --upgrades --queryformat="%{name}, %{evr}, %{arch}, %{repoid}\n"
```
Note that old dnf4 used to add the new line at the end automatically so you might have to check that if you want to use it also with dnf4.

I should also mention that in some situations the output of `repoquery` or `list` with `--upgrades` might differ from the actual transaction of the `upgrade` command. This is because they don't take obsoletes into account they just check which packages have upgrades available.
If you want the packages that will be upgraded by `upgrade` command you should use `check-upgrade` command. Unfortunately it doesn't have a proper `--json` flag for machine readable output yet, but I think it should have it in the future.

abitrolly:
@kontura I think I fixed the quiet flag for now.

If `check-upgrade` could construct plain 2D table with all fields in text form to be ready to be exported to CSV, maybe I could patch it..

kontura:
I am sorry I should have been more clear yesterday but I don't like these changes.

While it is true that dnf5 lacks the machine readable output for most commands in this situation you should use the `repoquery` command mentioned. It provides the same information and can be formatted exactly as you need.

We don't want to encourage users to parse the `list` commands.

Perhaps I can keep this open for a while to see if other developers have a different opinion but personally I would close it.

abitrolly:
@kontura that's just an example how to make quiet work again https://github.com/rpm-software-management/dnf5/issues/570

> I am sorry I should have been more clear yesterday but I don't like these changes.

You don't like how people use the output, or there are technical issues with making --quiet work here?

ppisar:
I agree with kontura: Text user interface is not a machine interface. Imposing that onus on the human output would make difficult to evolve it. Also human interface lacks many guarantees API has: Delimiting records, escaping, data types, data structures, versioning, locale independence etc. In short, it's unreasonable to handle a user interface as a machine interface. Both from DNF users and from DNF developers.

Ideally this should be handled by D-Bus interface.

abitrolly:
Well, maybe the feature is excessive, but does it work at least? Can other people use the PR as an example how to control output verbosity?

Realistically `dnf list` maybe useful for simple `cut` scripts. For Go I'd rather access DB directly. For programmatic API for `dnf5` I would take a look at https://www.nushell.sh/ handles data.

kontura:
> Well, maybe the feature is excessive, but does it work at least? Can other people use the PR as an example how to control output verbosity?

Yes, it works. I don't think there are any technical issue with your approach.

In my opinion for your use case it would be best to use the API directly via GO bindings.
Unfortunately they are not working now and we don't have the resources to fix them up. It would require some proficiency with GO and SWIG.

xyhhx:
> Can I ask why do you want to hide the headers?

@kontura there are plenty of cases where this would be desired. all package managers should have a way to totally suppress output when installing packages

my use case is thus: i'm on qubes os, and i want to build binaries from packages available on cargo in a networked vm, then transfer the results to an "air-gapped" vm. i _want_ to do something like so[^1]:

```sh
qvm-run-vm @dispvm:dvm-dev -- 'run0 dnf in -qy libseccomp-devel \
  && run0 cargo install --locked syd --features oci --root /opt/syd \
  && tar -C /opt/syd -czf - bin' \
> syd.tar.gz
```

but this will never work because dnf can't totally suppress its output when installing `libseccomp-devel` (or any packages)

[^1]: please don't comment on the quality of the script i'm running in the other vm, this is just a proof of concept

Issue #1895, opened 2024-11-24
Bugs in dnf versionlock makes it unusable


scaronni:
The plugin `versionlock` does not work properly. Taking into account the differences compared to the DNF4 plugin, it still does not behave correctly.

At Nvidia we're trying to get away from packages with the branch name in the actual package name and would like to switch to standard tools for doing branch locking. It works fine in DNF4, it does not in DNF5.

First issue: automatic filtering. Test is done on the same repository.

DNF4:

```
# dnf versionlock add \*nvidia\*565\*
Last metadata expiration check: 2:08:07 ago on Sun 24 Nov 2024 09:19:13 AM CET.
Adding versionlock on: nvidia-persistenced-3:565.57.01-1.fc40.*
Adding versionlock on: nvidia-driver-3:565.57.01-4.fc40.*
Adding versionlock on: nvidia-settings-3:565.57.01-1.fc40.*
Adding versionlock on: compat-nvidia-repo-3:565.57.01-2.fc40.*
Adding versionlock on: akmod-nvidia-3:565.57.01-1.fc40.*
Adding versionlock on: nvidia-driver-cuda-3:565.57.01-4.fc40.*
Adding versionlock on: nvidia-xconfig-3:565.57.01-1.fc40.*
Adding versionlock on: nvidia-kmod-common-3:565.57.01-2.fc40.*
Adding versionlock on: nvidia-libXNVCtrl-devel-3:565.57.01-1.fc40.*
Adding versionlock on: kmod-nvidia-3:565.57.01-1.fc40.*
Adding versionlock on: xorg-x11-nvidia-3:565.57.01-4.fc40.*
Adding versionlock on: libnvidia-fbc-3:565.57.01-4.fc40.*
Adding versionlock on: nvidia-libXNVCtrl-3:565.57.01-1.fc40.*
Adding versionlock on: libnvidia-cfg-3:565.57.01-4.fc40.*
Adding versionlock on: dkms-nvidia-3:565.57.01-1.fc40.*
Adding versionlock on: nvidia-driver-libs-3:565.57.01-4.fc40.*
Adding versionlock on: libnvidia-ml-3:565.57.01-4.fc40.*
Adding versionlock on: nvidia-driver-cuda-libs-3:565.57.01-4.fc40.*
Adding versionlock on: nvidia-modprobe-3:565.57.01-1.fc40.*
```

[...]

PR #2009, opened 2025-01-14
versionlock: Fix wildcards handling in `add` command

m-blaha:
There’s a bug in how the `dnf5 versionlock add <spec>` command handles
package specifications when the `spec` contains a wildcard and matches
multiple package names with the same version (EVR). In such cases, only
the first package is considered, while the rest are quietly skipped.
For example there are three packages matching 'aws-checksums*' pattern,
but only one is actually locked:

❯ dnf repoquery 'aws-checksums*'
Repositories loaded.
aws-checksums-0:0.2.2-1.fc40.i686
aws-checksums-0:0.2.2-1.fc40.x86_64
aws-checksums-devel-0:0.2.2-1.fc40.i686
aws-checksums-devel-0:0.2.2-1.fc40.x86_64
aws-checksums-libs-0:0.2.2-1.fc40.i686
aws-checksums-libs-0:0.2.2-1.fc40.x86_64

❯ dnf versionlock add 'aws-checksums*'
Repositories loaded.
Adding versionlock on "aws-checksums = 0.2.2-1.fc40".

With this patch, each group of packages with the same name is handled
separately, ensuring that all the packages are correctly locked:

❯ dnf5 versionlock add 'aws-checksums*'
Repositories loaded.
Adding versionlock on "aws-checksums = 0.2.2-1.fc40".
Adding versionlock on "aws-checksums-devel = 0.2.2-1.fc40".
Adding versionlock on "aws-checksums-libs = 0.2.2-1.fc40".

For: https://github.com/rpm-software-management/dnf5/issues/1895

### DIFF
--- a/dnf5/commands/versionlock/versionlock_add.cpp
+++ b/dnf5/commands/versionlock/versionlock_add.cpp
@@ -106,25 +106,35 @@ void VersionlockAddCommand::run() {
             std::cerr << libdnf5::utils::sformat(_("No package found for \"{}\"."), spec) << std::endl;
             continue;
         }
-        libdnf5::rpm::PackageQuery installed_query(query);
-        installed_query.filter_installed();
-        if (!installed_query.empty()) {
-            // if spec is installed, add only installed version
-            query = installed_query;
-        }
-
-        std::unordered_set<std::string> versions{};
+        // The spec can resolve to multiple package names. Handle each name in
+        // the query separately.
+        std::set<std::string> package_names;
         for (const auto & pkg : query) {
-            auto evr = pkg.get_evr();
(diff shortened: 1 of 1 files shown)


kontura (approved):
This looks good, I have created a simple test here: https://github.com/rpm-software-management/ci-dnf-stack/pull/1620

I am curious, I see this doesn't really work with architectures and the docs do say at the beginning that it works with `set of names and versions for packages`, on the other hand they specify `package-spec` as an argument.
From the description of the `add` subcommand I would expect that 
```versionlock add name*.x86_64```
would work just for x86_64.

Is the support for architectures being considered?

PR #2010, opened 2025-01-14
automatic: Translate end-of-lines in email emitter by DNF

ppisar:
The fix in 9a5cba8fbb0c95f5da1410803f9ff5093895252a (automatic: Fix end-of-lines in messages sent by email emitter) utilized cURL library for translating LF ending to CR-LF in content of SMTP DATA command. It fixed the problem with curl-8.11.1, but broke end-of-lines in e-mail headers with old curl-8.9.1.

I was unable to find what has changed in cURL, but the cause was that automatic plugin already separated headers by CR-LF, therefore cURL probably double-encoded them and that was again rejected by sendmail 8.18.1.

This patch reverts 9a5cba8fbb0c95f5da1410803f9ff5093895252a and instead performs the end-of-line normalization fully in dnf5::EmailMessage::str() method. Now the output of the method is completely valid e-mail message.

Implementation detail: I changed dnf5::Emitter::output_stream variable from const to non-cost to be able to call getline() on it. If is a problem, I will need to come with a more complicated solution because I can only obtain a new-line delimiter used in the std::stringstream object by getline().

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2335508

### DIFF
--- a/dnf5-plugins/automatic_plugin/email_message.cpp
+++ b/dnf5-plugins/automatic_plugin/email_message.cpp
@@ -23,18 +23,25 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/utils/format.hpp>
 
 #include <chrono>
+#include <string>
 
 namespace dnf5 {
 
 // TODO(mblaha): use some library to create an email instead of this template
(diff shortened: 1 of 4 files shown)


m-blaha (approved):
Thank you!

Issue #1779, opened 2024-10-17
Ruby bindings RFE: libdnf5 modules and classes should be namespaced.


jackorp:
While not necessarily a problem, the current lack of namespacing in Ruby does not seem too great to me.

For example `Common` module is a global constant upon requiring `libdnf5/common`. I think having a prefix so that it becomes for example `LibDNF5::Common` (or `Libdnf5::Common`, or other variations of capitalization) would be better.

I'd apply this to all modules present, but names such as `Logger`, `Common`, or `Base` are IMO more probably to have a clash in other projects than for example Rpm.

There could be a small problem with compatibility, but if `LibDNF5` is a module only used for namespacing (with no additional/new methods on it), then it can be included in the global namespace guilt-free:
```ruby
require 'libdnf5/base'
include LibDNF5

base = Base::Base.new()
```

jackorp:
Hmm already ran into a problem with the default Ruby Logger class.

I cannot work with rack and Sinatra gems. since there is a conflict between different Logger constant definitions at the top level if I also `require "libdnf5/base"`.

Simple example:
```console
$ sudo dnf install ruby-libdnf5 rubygem-sinatra
```

And a file `example.rb`
```ruby
#!/usr/bin/ruby
require 'sinatra'
require 'libdnf5/base'

get '/' do
  "Hello"
end
```
When I run `ruby example.rb` to start the server
I get an exception:
```
$ ruby example.rb
/usr/share/ruby/logger/version.rb:3:in `<top (required)>': Logger is not a class (TypeError)
/usr/lib64/ruby/vendor_ruby/libdnf5/logger.so:0: previous definition of Logger was here
	from /usr/share/ruby/logger.rb:17:in `require_relative'
	from /usr/share/ruby/logger.rb:17:in `<top (required)>'
	from <internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from <internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from /home/jprokop/.local/share/gem/ruby/gems/rack-2.2.10/lib/rack/logger.rb:3:in `<top (required)>'
	from <internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from <internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:136:in `require'
[...]

PR #2021, opened 2025-01-24
ruby bindings: Fix namespacing for libdnf5 classes/modules

jackorp:
See https://github.com/rpm-software-management/dnf5/issues/1779

The ruby/CMakeLists.txt also defined `-module` twice. Once "by hand" appending the raw `-module ${MODULENAME}` and then it is also added via the `set_property` for `SWIG_MODULE_NAME`. Once is enough, so I just deleted appending it to CLI and just adjusted the `set_property` call. This results in all constants in Ruby now prepended with `Libdnf5`, therefore preventing top-level constant collision (and allowing me to get on with my project).

By the way since setting SWIG_MODULE_NAME appends `-module` to swig CLI, it is overriding what is present in the files in the `%module` directive. Both Perl and Ruby are setting `-module` which makes specifying the module path via `%module` relatively useless even if there seems to be intent, judging by the fact that they are guarded in module files via `#if defined`.

JFTR not sure if there is or isn't possibility or some sane way of flattening the namespaces, Logger::Logger::$LOG_LEVEL is a bit unwieldy.

### DIFF
--- a/bindings/libdnf5/advisory.i
+++ b/bindings/libdnf5/advisory.i
@@ -3,7 +3,7 @@
 #elif defined(SWIGPERL)
 %module "libdnf5::advisory"
 #elif defined(SWIGRUBY)
-%module "libdnf5/advisory"
+%module "libdnf5::advisory"
 #endif
 
 #if defined(SWIGRUBY)
--- a/bindings/libdnf5/base.i
+++ b/bindings/libdnf5/base.i
@@ -3,7 +3,7 @@
 #elif defined(SWIGPERL)
 %module "libdnf5::base"
 #elif defined(SWIGRUBY)
-%module "libdnf5/base"
+%module "libdnf5::base"
 #endif
 
 %include <exception.i>
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -3,7 +3,7 @@
 #elif defined(SWIGPERL)
 %module "libdnf5::common"
 #elif defined(SWIGRUBY)
-%module "libdnf5/common"
+%module "libdnf5::common"
 #endif
 
 %include <cstring.i>
--- a/bindings/libdnf5/comps.i
+++ b/bindings/libdnf5/comps.i
@@ -3,7 +3,7 @@
 #elif defined(SWIGPERL)
 %module "libdnf5::comps"
 #elif defined(SWIGRUBY)
-%module "libdnf5/comps"
+%module "libdnf5::comps"
 #endif
 
 %include <exception.i>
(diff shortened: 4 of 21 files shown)


jrohel:
> As a result, all constants in Ruby are now appended with Libdnf5, which prevents top-level constants from colliding (and I can continue my project).

This seems like the right modification. In other languages, we use the `libdnf5` namespace. I'm wondering why we don't have that from the start for ruby as well. 

I'm also considering how many ruby API users we have. This is an API breaking change. It's easily fixable on the user side. But it is a change.

jackorp:
> This seems like the right modification. In other languages, we use the libdnf5 namespace. I'm wondering why we don't have that from the start for ruby as well.

Seems since "Initial project template" 5bb316d367d7e4f878db34458084cd1b00801ae7 `ruby/CMakeLists.txt` was wrong or MODULE_NAME had different value than intended judging by the following being present:
```
+    set_property(SOURCE ../../${LIBRARY_NAME}/${MODULE_NAME}.i PROPERTY SWIG_MODULE_NAME ${MODULE_NAME})
+    set(CMAKE_SWIG_FLAGS -module ${MODULE_NAME} -ruby)
+    swig_add_library(${TARGET_NAME} LANGUAGE ruby SOURCES ../../${LIBRARY_NAME}/${MODULE_NAME}.i)
```

For comparison, relevant section of perl5 cmake in that same commit:
```
+    set_property(SOURCE ../../${LIBRARY_NAME}/${MODULE_NAME}.i PROPERTY SWIG_MODULE_NAME ${LIBRARY_NAME}::${MODULE_NAME})
+    swig_add_library(${TARGET_NAME} LANGUAGE perl SOURCES ../../${LIBRARY_NAME}/${MODULE_NAME}.i)
```
perl does not have the `set()` directive and `set_property()` contains `${LIBRARY_NAME}::${MODULE_NAME}` instead of just `${MODULE_NAME}`

Not sure if there is more history of this somewhere but 2020 is the most I see for these files in this repository.

> I'm also considering how many ruby API users we have. This is an API breaking change. It's easily fixable on the user side. But it is a change.

That I do not know, but I can say confidently if they are using libdnf5, they aren't using the default/bundled `logger` gem that comes with Ruby (unless they are jumping through some serious hoops to make it work).

jrohel:
We will have to move libdnf5 classes/modules out of the global namespace. This will cause the API change. That's not good. But doing it later will be even more of a pain for all.
Ruby support is in a "best effort" state. From that perspective, it's best to make the change as soon as possible.
I will merge this PR.

PR #2029, opened 2025-02-02
ImplPtr: Support for empty source in copy assignment operator, unit tests

jrohel:
ImplPtr supports empty source (a source that owns nothing) in the copy and move constructor and in the move assignment operator. However, support for empty source in the copy assignment operator was missed.

Now there is support for empty source in the copy assignment operator as well. This makes the code consistent. In addition, the copy assigment operator can now be used for classes having an empty pointer to the implementation. For example, it is in the class as a placeholder for the future.

Unit tests included.


jan-kolarik (approved):
Thank you!

PR #2032, opened 2025-02-04
Classes common::Message and common::EmptyMessage

jrohel:
`common::Message`: A base abstract class for passing a message whose formatting, including localization (translation, argument format) is done at the destination.

`common::EmptyMessage`: class for passing an empty message

Closes: https://github.com/rpm-software-management/dnf5/issues/2034
Requires: https://github.com/rpm-software-management/dnf5/pull/2029
Requires: https://github.com/rpm-software-management/dnf5/pull/2031

### DIFF
--- a/bindings/libdnf5/common.i
+++ b/bindings/libdnf5/common.i
@@ -28,8 +28,11 @@
 }
 
 %{
+    #include "libdnf5/common/message.hpp"
     #include "libdnf5/common/weak_ptr.hpp"
 %}
+%include "libdnf5/common/message.hpp"
+
 %include "libdnf5/common/weak_ptr.hpp"
 #if defined(SWIGPYTHON)
 %extend libdnf5::WeakPtr {
--- a/include/libdnf5/common/message.hpp
+++ b/include/libdnf5/common/message.hpp
@@ -54,6 +54,26 @@ private:
     ImplPtr<Impl> p_impl;
 };
 
+
+/// Class for passing an empty message.
+class LIBDNF_API EmptyMessage final : public Message {
+public:
+    EmptyMessage();
+    EmptyMessage(const EmptyMessage & src);
+    EmptyMessage(EmptyMessage && src) noexcept;
+    ~EmptyMessage() override;
+
+    EmptyMessage & operator=(const EmptyMessage & src);
+    EmptyMessage & operator=(EmptyMessage && src) noexcept;
+
+    /// Returns empty string
+    ///
+    /// @param translate  ignored
+    /// @param locale     ignored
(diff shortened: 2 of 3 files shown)


jan-kolarik:
Hmm, sorry, but I am bit confused about the usage of the `Message` object. Could you provide some simple example of how would one use it e.g. for implementing some new callback method with passing this object? I am specifically interested how dnf5 app would prints the message.

jrohel:
In plugin:
```
class InfoMessage : public libdnf5::Message {
public:
    InfoMessage(arg_type arg) : arg(arg) {}

    std::string format(bool translate, const libdnf5::utils::Locale * locale) const override {
        return libdnf5::utils::format(locale, translate, M_("Some text with args {} "), arg);
    }

private:
  arg_type arg;
};
```
And pass the object to callback.

In app:
```
bool repokey_remove(const libdnf5::rpm::KeyInfo & key_info,  const libdnf5::Message & info) override {
    std::string info_message = info.format(true, nullptr);
}
```

jan-kolarik:
> In plugin:
> 
> ```
> class InfoMessage : public libdnf5::Message {
> public:
>     InfoMessage(arg_type arg) : arg(arg) {}
> 
>     std::string format(bool translate, const libdnf5::utils::Locale * locale) const override {
>         return libdnf5::utils::format(locale, translate, M_("Some text with args {} "), arg);
>     }
> 
> private:
>   arg_type arg;
> };
> ```
> 
> And pass the object to callback.
> 
> In app:
> 
> ```
> bool repokey_remove(const libdnf5::rpm::KeyInfo & key_info,  const libdnf5::Message & info) override {
>     std::string info_message = info.format(true, nullptr);
> }
> ```

All right, thanks for the example. Why is the translate argument, which determines whether to translate, handled on the app side? It seems to require the app to understand the message content and decide whether to translate it. I understand that translation itself happens on the app side, not the client, but shouldn’t the decision to translate be made by the message author, who knows its semantics? For instance, error codes or machine-generated messages might not need translation.

To put it differently, if the message author already determines which parts are translatable, what scenario would require the app to override that decision and not translate the message?

jrohel:
If it makes sense to translate the passed text, the message author implements `Message::format` to return the translated text if `translate == true`.

The application decides whether the translation is used and in what language. Why? Well, because the application doesn't have to be just dnf5, but any application using libdnf5.This arbitrary application might want to output the received information translated in the user's preferred language to the terminal. But it may want to write it untranslated to the log. In general, we don't know where and in what language the unknown application will want to write the message.

jan-kolarik (approved):

Issue #2042, opened 2025-02-07
PluginLibrary::PluginLibrary: Unhandled null pointer on plugin instantiation


jrohel:
The `libdnf_plugin_new_instance` and `dnf5_plugin_new_instance` return a null pointer if the plugin instantiation fails. This case is not handled. The null pointer is dereferenced.

PR #2043, opened 2025-02-07
plugins: Check if the plugin instantiation was successful

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/2042

Fixes crash if plugin instantiation fails and returns a null pointer.


kontura (approved):
Thanks!

PR #2051, opened 2025-02-10
expired-pgp-keys: Install the plugin by default on Fedora 42+

jan-kolarik:
Also add a dependency on the gpg command provider.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -21,6 +21,9 @@ Requires:       dnf-data
 Recommends:     (dnf5-plugins if dnf-plugins-core)
 Recommends:     bash-completion
 Requires:       coreutils
+%if 0%{?fedora} > 41
+Recommends:     libdnf5-plugin-expired-pgp-keys
+%endif
 
 %if 0%{?fedora} || 0%{?rhel} > 10
 Provides:       microdnf = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


dnf5.spec
```diff
@@ -632,6 +635,7 @@ Libdnf5 plugin that installs repository's Appstream data, for repositories which
 Summary:        Libdnf5 plugin for detecting and removing expired PGP keys
 License:        LGPL-2.1-or-later
 Requires:       libdnf5%{?_isa} = %{version}-%{release}
+Recommends:     gnupg2
```
m-blaha, line 638: Can the plugin work without `gnupg2`? If not, I suggest changing this from  weak dependency to requirement.

dnf5.spec
```diff
@@ -21,6 +21,9 @@ Requires:       dnf-data
 Recommends:     (dnf5-plugins if dnf-plugins-core)
 Recommends:     bash-completion
 Requires:       coreutils
+%if 0%{?fedora} > 41
```
m-blaha, line 24: Is there a reason to not install the plugin on older Fedoras?

dnf5.spec
jan-kolarik: The plugin itself can be loaded without the package, but if the `gpg` command is not available on the system, it won't basically do anything. I recall some discussion around that, possibly some other provider of `gpg` command being possibly deployed, but right now I cannot find any of those. I can eventually change it to hard dep...

m-blaha: In that case we can use `Requires: /usr/bin/gpg` which would be satisfied by any provider of the binary. But since there is none other and we cannot be sure that alternative version would work for our use-case, I would prefer keeping `Requires: gnupg2`.

I do not see much sense in ability to load a plugin which will do nothing. I find it confusing for the users.

dnf5.spec
m-blaha: OK, thanks for clarification.

dnf5.spec
jan-kolarik: Thanks, I agree. Will change this to `Requires: gnupg2`.

m-blaha (approved):
Thank you!

Issue #2047, opened 2025-02-09
The plugins `libdnf_plugin_new_instance` and `dnf5_plugin_new_instance` API functions discard exceptions


jrohel:
The `libdnf_plugin_new_instance` and `dnf5_plugin_new_instance` functions return a null pointer if the instantiation of the plugin fails. The exception with failure information is discarded. Thus, the caller (libdnf library or dnf5 application) will not know the reason for the failure.

The above functions are linked as "C" functions. It is necessary to provide caller access to the exception.

ppisar:
I confirm the libdnf_plugin_new_instance functions as defined e.g. in libdnf5-plugins/actions/actions.cpp translates the exception to returning nullptr. If we want to return an error message, we need to change the prototype: either add a new argument or change the return type.

ppisar:
To prevent from the API/ABI break we could add a new function and keep to old one.

PR #2052, opened 2025-02-10
Plugin API: *_plugin_get_last_exception: Return pointer to last exception

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/2047

To prevent function names from being mangled in the plugin API, these functions are linked as C language functions (extern "C"). We do not throw exceptions from these functions. Specifically, the `libdnf_plugin_new_instance` and
`dnf5_plugin_new_instance` functions return a null pointer on failure and discard the exception. Thus, the caller (libdnf library or dnf5 application) does not know the reason for the failure.

Changes:
- This modification extends the libdnf and dnf5 plugin APIs with new `libdnf_plugin_get_last_exception` and `dnf5_plugin_get_last_exception` functions. The plugin can save the exception instead of discarding it. The new functions allow access to the saved exception.

- All plugins have been modified to implement the new API functionality.

- Documentation string were improved.

Note:
The original functions are unchanged. Plugins that do not implement the new functions can still be used. API and ABI remain compatible.

### DIFF
--- a/bindings/libdnf5/plugin.i
+++ b/bindings/libdnf5/plugin.i
@@ -36,6 +36,7 @@
 %ignore libdnf_plugin_get_version;
 %ignore libdnf_plugin_new_instance;
 %ignore libdnf_plugin_delete_instance;
+%ignore libdnf_plugin_get_last_exception;
 %feature("director") IPlugin;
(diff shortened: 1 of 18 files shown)


m-blaha (approved):
Thank you!

Issue #1605, opened 2024-08-02
Crash if repo_gpgcheck=1 and gpgkey is set to a nonexistent file


ppisar:
Having a repository configured like:

[test]
name=test
baseurl=file:///root/repos/test/
enabled=0
gpgcheck=0
repo_gpgcheck=1
metadata_expire=5m
gpgkey=file:///root/repos/test/key

but /root/repos/test/key file not exisiting, "dnf5 --enablerepo=test upgrade" crashes like this:

~~~~
Updating and loading repositories:
 test                                                                                                                                                                     100% |   0.0   B/s |   1.7 KiB |  00m00s
>>> Librepo error: repomd.xml GPG signature verification error: Signing key not foundterminate called without an active exception
Aborted (core dumped)
~~~~

a the backtrace is:

~~~~
Stack trace of thread 44109:
#0  0x00007f7fbe680724 __pthread_kill_implementation (libc.so.6 + 0x72724)
#1  0x00007f7fbe627d0e raise (libc.so.6 + 0x19d0e)
#2  0x00007f7fbe60f942 abort (libc.so.6 + 0x1942)
#3  0x00007f7fbe8a5da9 _ZN9__gnu_cxx27__verbose_terminate_handlerEv.cold (libstdc++.so.6 + 0xa5da9)
#4  0x00007f7fbe8b7c2c _ZN10__cxxabiv111__terminateEPFvvE (libstdc++.so.6 + 0xb7c2c)
#5  0x00007f7fbe8a5951 _ZSt9terminatev (libstdc++.so.6 + 0xa5951)
#6  0x00007f7fbec5f7a9 _ZN7libdnf54repo8RepoSack4Impl21update_and_load_reposERNS0_9RepoQueryEb.constprop.0.cold (libdnf5.so.2 + 0x5f7a9)
#7  0x00007f7fbed2d9f4 _ZN7libdnf54repo8RepoSack10load_reposEv (libdnf5.so.2 + 0x12d9f4)
#8  0x000055a414685152 _ZN4dnf57Context4Impl10load_reposEb (dnf5 + 0xcc152)
#9  0x000055a4145f31df main (dnf5 + 0x3a1df)
#10 0x00007f7fbe611248 __libc_start_call_main (libc.so.6 + 0x3248)
#11 0x00007f7fbe61130b __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x330b)
[...]

PR #2054, opened 2025-02-11
Fix: RepoSack::update_and_load_repos: properly terminate thread_sack_loader on exception

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1605

The `OnScopeExit` class is used to ensure proper termination of the `thread_sack_loader` in case of unexpected exit of the `update_and_load_repos` method due to an exception.

If the `thread_sack_loader` is not terminated (joined) before the method exits, the program will crash. Before this fix, the program crashed, for example, on an exception generated during pgp key import.


m-blaha:
Hm, I'm not able to reproduce the original issue #1605 any more.

m-blaha:
Sorry, my system was misconfigured after some previous experiments. Now I can reproduce it and confirm that the fix works.

m-blaha (approved):
Thank you!

PR #2056, opened 2025-02-12
CMake: Use list(APPEND FOO) over set(FOO ${FOO} ...)

ferdnyc:
```cmake
set(FOO ${FOO} BAR BAZ)
set(FOO "${FOO};BAR;BAZ")

# They're both just funny ways to spell...
list(APPEND FOO BAR BAZ)
```

This PR replaces all remaining cargo-culted occurrences of `set()` being used to extend a list variable in CMake code, for which `list(APPEND)` is the appropriate command.

Not touched are calls like:

```cmake
set(FOO_STRING "${FOO_STRING} extended-string-stuff")
```

...which are not `list(APPEND)` equivalent.


kontura:
Thanks

Issue #1336, opened 2024-03-21
libdnf5 plugins: vague error message while further dnf5 actions are blocked


glum23:
When an incorrect (non-existent) plugin name is used in an /etc/dnf/libdnf5-plugins/ config file, dnf5 stops working and provides quite vague error message:
```
# dnf5 repo list
Cannot load plugins: actions.conf
# dnf5 install zsh
Cannot load plugins: actions.conf
# cat /etc/dnf/libdnf5-plugins/actions.conf
[main]
name = unknown
enabled = 1
```
It would be nice to 
1. provide more useful info like "Cannot load plugin 'unknown' enabled from /etc/dnf/libdnf5-plugins/actions.conf"
2. ignore the plugin instead of completely blocking any further dnf5 actions

In addition to that, it'd be nice to provide more docs related to dnf5 / libdnf5 plugins, their configs etc.

version: dnf5-5.1.15-20240321004900.23.ge0ae9be5.fc39

ppisar:
I agree with the better error message, I don't agree with the ignorance. Some plugins can be critical. E.g. the not yet ported  DNF4 plugin which verifies PGP key in DNS.

j-mracek:
This error cannot detect not ported DNF4 plugin, because they have a configuration file in a different path. Interesting thing is:
Missing configuration file => plugin is not loaded
Present configuration file, but plugin is not preset -> fail

In DNF4 - plugin configuration file was optional
In DNF4 - broken plugin triggers a warning

- This behavior was introduce after an incident where a third party plugin had broken DNF.

ppisar:
I used the PGP plugin as an example of a plugin where a broken plugin should prevent DNF from continuing. I did not mean that DNF5 should attempt to load DNF4 plugins.

PR #2057, opened 2025-02-12
load_plugins: Preserve original exception with failure information

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/1336

The original implementation threw an exception with a list of plugins that failed to load. We usually got the name of one configuration file. Information about the reason for the failure was not included in the exception.

Now the code throws an exception for the first failed plugin. Exception contains the full path (not just the filename) and includes the original exception from the source of the error.


kontura:
I like it, thank you!

Issue #1851, opened 2024-11-12
Incorrect source repository (from_repo) after system upgrade


m-blaha:
After system upgrade with dnf5, the `from_repo` attribute for all installed packages contains `@stored_transaction` value. This is confusing to the users. The attribute should contain id of the repository the package really came from.

See `dnf repoquery --installed --queryformat="%{full_nevra} %{from_repo}\n"` on the upgraded system.

ppisar:
Fedora bug report <https://bugzilla.redhat.com/show_bug.cgi?id=2346690>.

PR #2061, opened 2025-02-13
Use actual repository ID in stored transactions

m-blaha:
Currently, when replaying a stored transaction, all RPM files are placed into a single `@stored_transaction` repository. This can be confusing, especially during a system upgrade, which also uses stored transactions.

With this patch, local RPM files from stored transactions are placed to the repository they originally came from.

Fixes: https://github.com/rpm-software-management/dnf5/issues/1851

### DIFF
--- a/libdnf5/rpm/package_sack_impl.hpp
+++ b/libdnf5/rpm/package_sack_impl.hpp
@@ -127,6 +127,9 @@ public:
     /// And sets `considered_uptodate` to` true`.
     void recompute_considered_in_pool();
 
+    void register_local_rpm_id(const Id id);
+    bool is_local_rpm_id(const Id id);
+
 private:
     bool provides_ready{false};
 
@@ -160,6 +163,9 @@ private:
     libdnf5::solv::SolvMap cached_solvables{0};
     int cached_solvables_size{0};
     PackageId running_kernel;
+
+    // ids of packages created from local rpm files
+    std::unordered_set<Id> local_rpm_ids;
 
     friend PackageSack;
     friend Package;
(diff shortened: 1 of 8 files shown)


libdnf5/rpm/package_sack_impl.hpp
```diff
@@ -161,6 +164,9 @@ class PackageSack::Impl {
     int cached_solvables_size{0};
     PackageId running_kernel;
 
+    // ids of packages created from local rpm files
+    std::unordered_set<Id> local_rpm_ids;
```
kontura, line 168: I am not really sure about this.
In my mind the `Repo::Type::AVAILABLE` repos are not meant for locally added rpms. I would be interested in an opinion from @jrohel.

On the other hand I understand that our options are limited here.

Given that we are interested only in the repo ids would it be possible to perhaps not load the repo definitions at all and create only `Repo::Type::COMMANDLINE` repos with the specified id when needed by the `add_stored_transaction_package(...)`?
This would have the advantage that package replays with repo ids that are no longer defined on the system would still have the original repoid instead of `@stored_transaction`.

m-blaha: Yes, this is also possibility. Let me experiment with this approach a bit...

m-blaha: I've created an alternative PR which does not create repositories from the system configuration and creates `COMMANDLINE` type repos on the fly as they are needed: https://github.com/rpm-software-management/dnf5/pull/2093

m-blaha:
Close in favor of https://github.com/rpm-software-management/dnf5/pull/2093

PR #2063, opened 2025-02-14
Create progress bar in script callback if one doesn't already exist

imciner2:
Normally, the script start callback is called after package verification has been done, so there is an active progress bar (created in the verification stage), however package verification can be disabled. If verification is disabled, then the script start callback is the first callback used (to run pre-trans scripts), and there will be no active progress bar, leading to a crash.

To handle the case of disabled verification, create a new progress bar in the script start callback for it to use in the future.

Fixes https://github.com/rpm-software-management/dnf5/issues/1949


jrohel:
@imciner2
Your PR fixed segfault.
However, disabling package verification causes another problem with progress bars. I created an issue for it a while ago https://github.com/rpm-software-management/dnf5/issues/2079. It's a minor issue, but since I had a solution in mind, I prepared a PR with a fix https://github.com/rpm-software-management/dnf5/pull/2080.

PR #2066, opened 2025-02-14
Rename dnf5-makecache timer to dnf-makecache.

gordonmessmer:
Some maintainers of Fedora presets would prefer that the dnf-makecache timer use the unversioned name so that the presets (and Ansible playbooks, etc) do not need to be updated to account for renaming it.  This change would create a file conflict between the "dnf" and "dnf5" packages in Fedora, but it's already impossible to install them simultaneously because dnf5 "obsoletes" the old dnf package.


gordonmessmer:
Wrinkle: dnf5 does not obsolete dnf on F40, so merging this would create a package conflict where there currently isn't one.

Conan-Kudo:
I don't think we should just do it this way. It should be build-time configurable because different distributions are on different paths for dnf4 -> dnf5.

ppisar:
Neal is right. DNF5 still can be built without obsoleting DNF4 and upstream needs to account with it.

keszybz:
Meh. Keeping separate naming would be useful if users were supposed to install both in parallel. But they are not going to. Separate dbs and caches means using both in parallel would be annoying and inefficient. Having the ability to install both in parallel was useful in the transitory period, but we're now exiting that period, and we want to have the solution that allows the smoothest transition. We want users to be able to switch to new versions of software without having to tweaks things and deal with spurious incompatibilities.

If we can't have this upstream, let's please do it downstream in Fedora. *We* are not allowing users to install dnf5 and dnf-5 in parallel, and we should aim for a smooth transition.

Conan-Kudo:
> If we can't have this upstream, let's please do it downstream in Fedora. We are not allowing users to install dnf5 and dnf-5 in parallel, and we should aim for a smooth transition.

What? Yes we are. You can install dnf4 and dnf5 on the same system just fine.

keszybz:
```console
$ rpm -q --obsoletes dnf5
dnf < 5
microdnf < 4
yum < 5
```

gordonmessmer:
I'm testing the result of a change that would make the unit name configurable, as Neal suggested.  I'll update the PR shortly.

Issue #2023, opened 2025-01-26
RFE: Optionally handle command return code in Actions plugin


marbu:
## Description

Actions plugin allows to execute external program for given libdnf5 hooks, but there seems to be no way to propagate an error back when the external program exits with non zero return code.

It would be interesting to have an option (in [actions file format](https://dnf5.readthedocs.io/en/latest/libdnf5_plugins/actions.8.html#actions-file-format)) to make dnf5 to raise an error when command returns non zero return code. Right now, the error is ignored.

## Reasoning

I was exploring the actions plugin to see if I could implement etckeeper dnf5 integration this way, and it turns out that this error handing functionality is the only missing piece.

See https://bugzilla.redhat.com/show_bug.cgi?id=2326283#c3

## Example

It could workd like this: by specifying `raise_error=1` dnf5 would stop the transaction.

```
# cat /etc/dnf/libdnf5-plugins/actions.d/etckeeper.actions
pre_transaction:::raise_error=1:etckeeper pre-install
post_transaction:::raise_error=1:etckeeper post-install
```

jrohel:
This is a good request.
Successful execution of some plugins (actions) is mandatory for further running of the program.  Therefore it is necessary that the action has the possibility to interrupt the program.

PR #2068, opened 2025-02-14
libdnf actions plugin: Introduce "mandatory" actions, improvements for multi-threaded/multi-process apps

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/2023

A new optional boolean option "mandatory" has been added. The allowed values are "0" and "1". For backwards compatibility, the default value is "0". If an action is mandatory ("mandatory" is set to "1"), an exception is generated if the action process failed to run or ended with a non-zero return code.


jan-kolarik (approved):
LGTM

PR #2076, opened 2025-02-19
Document the entire API by namespace/module

kontura:
I made my own build of the docs so the result can be previewed: https://amatej-dnf5.readthedocs.io/en/latest/api/index.html

It would be ideal if the entire documentation was generated automatically but then the cpp and Python docs would look very different. Doing it by namespace/module still gives us enough control while there is not that many of them.
This could be revisited later.


jan-kolarik (approved):
Looks good, thanks for filling the gaps there!

PR #2078, opened 2025-02-20
Actions plugin documentation: extension in version 1.4.0

jrohel:
Adds a description of the `raise_error` option and a description of the `stop`, `error` and `log` commands in the standard action output format.


jan-kolarik (approved):
Thanks!

Issue #2079, opened 2025-02-21
Wrong total number of transaction progressbars when package verification is disabled


jrohel:
Package verification is enabled, the total number of progress bars is correct:
```
Running transaction
[1/4] Verify package files                                     100% |   2.0 KiB/s |   2.0   B |  00m00s
[2/4] Prepare transaction                                      100% | 111.0   B/s |   2.0   B |  00m00s
[3/4] Installing libpcap-14:1.10.5-1.fc41.x86_64               100% | 135.9 MiB/s | 417.5 KiB |  00m00s
[4/4] Installing iftop-0:1.0-0.34.pre4.fc41.x86_64             100% |   6.0 MiB/s |  98.6 KiB |  00m00s
Complete!
```

Package validation is disabled, the total number of progress bars is wrong (`%_pkgverify_level none` inside "/etc/rpm/macros.verify"):
```
Running transaction
[1/4] Prepare transaction                                      100% |  27.0   B/s |   2.0   B |  00m00s
[2/4] Installing libpcap-14:1.10.5-1.fc41.x86_64               100% | 135.9 MiB/s | 417.5 KiB |  00m00s
[3/4] Installing iftop-0:1.0-0.34.pre4.fc41.x86_64             100% |   1.8 MiB/s |  98.6 KiB |  00m00s
Complete!
```

imciner2:
Thanks for catching that, I never actually noticed this was wrong.

PR #2080, opened 2025-02-21
Fix total number of transaction progress bars

jrohel:
Before the transaction, the total number of steps needed and thus the progress bars needed is calculated. The original calculation assumed that if new packages enter the system through transactions, the first step would be to verify them. However, package verification can be disabled.

This modification will add the verification step to the total number of steps only if verification is performed.

Closes: https://github.com/rpm-software-management/dnf5/issues/2079


ppisar (approved):
I confirm this fixed the problem and it seems not introducing new bugs.

Issue #1497, opened 2024-05-16
`copr enable` does not respect `--installroot`


teohhanhui:
It does with `dnf`...

jan-kolarik:
@praiskup Could you please check this one?

praiskup:
This has been added to the CPT's TODO list (3-months schedule)

praiskup:
> Unless we want to change how the copr enable command behaves I think we can close this.

I think this is exactly the thing we want to do - `dnf copr` should manipulate with repofiles _in chroot_, not _on host_.  That said, not just `enable` but `list`, `delete`, etc.  For DNF5 at least (I thought DNF4 is fine per original report, but haven't tried myself).

nikromen:
I am working on this one

PR #2081, opened 2025-02-22
Fix: let copr plugin to respect the installroot option

nikromen:
This creates repofiles in `installroot` to respect its location. However `dnf4 copr` behaves the same way as `dnf5 copr` with `--use-host-config`:

```
[root@7223880ab7fc tmp]# dnf4 copr enable @copr/copr-ping --installroot=/tmp/dnf4 -y
...
[root@7223880ab7fc tmp]# tree
.
└── dnf4
    ├── usr
    │   └── lib
    │       └── sysimage
    │           └── rpm
    │               ├── rpmdb.sqlite
    │               ├── rpmdb.sqlite-shm
    │               └── rpmdb.sqlite-wal
    └── var
        ├── cache
        │   └── dnf
        │       └── expired_repos.json
        └── log
            ├── dnf.librepo.log
            ├── dnf.log
            └── dnf.rpm.log

10 directories, 7 files
[root@7223880ab7fc tmp]#
```

the repofile is created inside `/etc/yum.repos.d`, the installroot is ignored.

```
[root@7223880ab7fc tmp]# ls /etc/yum.repos.d/
_copr:copr.fedorainfracloud.org:group_copr:copr-ping.repo                      fedora-cisco-openh264.repo   fedora-updates.repo
_copr:copr.fedorainfracloud.org:packit:rpm-software-management-dnf5-2081.repo  fedora-updates-testing.repo  fedora.repo
```

But `dnf4 install buildtag --installroot=/tmp/dnf4` somehow uses the repofile in `/etc/yum.repos.d`. The same steps with dnf5 will result in 

```
[root@7223880ab7fc tmp]# dnf5 install buildtag --installroot=/tmp/dnf5
Repositories loaded.
No match for argument: buildtag
No repositories were loaded from the installroot. To use the configuration and repositories of the host system, pass --use-host-config.
```

because dnf5 by default looks to the installroot. I don't know whether we want to change the default behaviour of `dnf copr` between versions 4 and 5, but I'd expect the `dnf5 copr` command to create repofile inside `installroot`, so this PR should address this issue.  

Fix https://github.com/rpm-software-management/dnf5/issues/1497

### DIFF
--- a/dnf5-plugins/copr_plugin/copr_repo.cpp
+++ b/dnf5-plugins/copr_plugin/copr_repo.cpp
@@ -25,6 +25,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include <fnmatch.h>
 #include <libdnf5-cli/utils/userconfirm.hpp>
+#include <libdnf5/base/base.hpp>
 #include <libdnf5/conf/const.hpp>
 
 #include <filesystem>
@@ -35,13 +36,15 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace dnf5 {
 
 
-std::filesystem::path copr_repo_directory() {
+std::filesystem::path copr_repo_directory(libdnf5::Base * base) {
     std::filesystem::path result;
+
+    std::filesystem::path installroot = base->get_config().get_installroot_option().get_value();
     if (char * dir = getenv("TEST_COPR_CONFIG_DIR")) {
-        result = dir;
+        result = installroot.empty() ? dir : installroot / dir;
         return result / "yum.repos.d";
     }
-    return COPR_REPO_DIRECTORY;
+    return installroot.empty() ? COPR_REPO_DIRECTORY : installroot / COPR_REPO_DIRECTORY;
 }
 
 
@@ -329,7 +332,7 @@ CoprRepo::CoprRepo(
 
 
 std::filesystem::path CoprRepo::file_path() const {
-    std::filesystem::path path = copr_repo_directory();
+    std::filesystem::path path = copr_repo_directory(base);
     path /= copr_id_to_copr_file(id);
     return path;
 }
@@ -439,7 +442,7 @@ void CoprRepo::save_interactive() {
 }
 
 void CoprRepo::remove_old_repo() {
-    std::filesystem::path path = copr_repo_directory();
+    std::filesystem::path path = copr_repo_directory(base);
     path /= "_copr_" + get_ownername() + "-" + get_projectname() + ".repo";
 
     if (std::filesystem::exists(path)) {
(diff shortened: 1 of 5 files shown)


dnf5-plugins/copr_plugin/copr_repo.cpp
```diff
@@ -37,11 +38,14 @@ namespace dnf5 {
 
 std::filesystem::path copr_repo_directory() {
     std::filesystem::path result;
+    libdnf5::Base base;
```
jrohel, line 41: This line creates a new instance of the `Base` class with the default settings.

We need to use the `base` object from the current dnf5 application context. It is obtained in multiple places using the line `auto & base = get_context().get_base();`. So we need to pass it here.

nikromen: ah yes I was testing it yesterday whether it works and the installroot was for some reason empty... that explains it, thanks! however when I try to compile this, I get this error:

```
/builddir/build/BUILD/dnf5-5.2.10.0-build/dnf5-5.2.10.0/dnf5-plugins/copr_plugin/copr_repo.cpp:42:19: error: ‘get_context’ was not declared in this scope; did you mean ‘gettext’?
   42 |     auto & base = get_context().get_base();
      |                   ^~~~~~~~~~~
      |                   gettext
[ 77%] Building CXX object dnf5/CMakeFiles/dnf5.dir/commands/autoremove/autoremove.cpp.o
```

I have no idea why - this is the first time I've got my hands dirty with cpp so I suppose it's something simple like missing some definition in .hpp file? Or do I need to pass it from somewhere as function argument?

jrohel: @nikromen 
`get_context` is a method of class `Command`. It can be called from instances of the `Command` class and derived classes. The `copr_repo_directory` is not a class method. Therefore, `get_context` must be used in the caller (class method) and then `base` is passed to `copr_repo_directory` as an argument.

Example solution:
1. Add the base argument: `std::filesystem::path copr_repo_directory(libdnf5::Base & base)`.
2. In the methods of the classes derived from `Command` we will call `copr_repo_directory(get_context().get_base())`. 
3. In the methods of the `CoprRepo` class, we will call `copr_repo_directory(*base);` - not derived from the `Command` class, but containing the `base` item.

jrohel:
@nikromen 
There was a bug in the copr plugin. Therefore, unit tests of copr_plugin with your PR failed.  I fixed the copr plugin (PR https://github.com/rpm-software-management/dnf5/pull/2131). Now the unit tests pass.

However, I recommend changing `std::filesystem::path copr_repo_directory(libdnf5::Base * base)` to `std::filesystem::path copr_repo_directory(libdnf5::Base & base)`
Why?  Because pointer can be nullptr. But if the argument is a reference, it is clear that the function expects a reference to an existing object (not nullptr).

jrohel:
@nikromen I will merge it even in this state. My recommendation of using a reference instead of a pointer can be applied in the future, perhaps in more places.

PR #2087, opened 2025-02-25
offline transactions: Keep user informed about scriptlet execution

m-blaha:
Scriptlet execution can take a significant amount of time. Inform the user when
it starts.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1667

### DIFF
--- a/dnf5/commands/offline/offline.cpp
+++ b/dnf5/commands/offline/offline.cpp
@@ -133,7 +133,10 @@ public:
     void elem_progress(const libdnf5::base::TransactionPackage & item, uint64_t amount, uint64_t total) override {
         RpmTransCB::elem_progress(item, amount, total);
 
-        plymouth.progress(static_cast<int>(100 * static_cast<double>(amount) / static_cast<double>(total)));
+        if (total != 0) {
+            // Reserve the last 5% of the progress bar for post-transaction callbacks.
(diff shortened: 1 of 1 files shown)


evan-goode:
I believe this will resolve https://bugzilla.redhat.com/show_bug.cgi?id=2345905.

evan-goode (commented):
LGTM. Though personally, I would rather have the progress bar just hang at 95% than try to guess progress using the hardcoded estimated scriptlet count. A typical system-upgrade might have 40 scriptlets but a smaller offline transaction might have only a couple. I guess a progress bar that is incrementing steadily and then jumps to 100% is better than a progress bar that hangs at 95%... but to me it still feels like a hack that's not really worth it. Users at this point are pretty used to progress bars hanging and jumping around.

I tested a F41->F42 upgrade and the post-transaction scriptlets took ~40 seconds out of the ~7min total.

m-blaha:
Sure, it was just an idea worth trying. You are right that this hack is unnecessary and unreliable. I've replaced it with reserving the last 5% of the progress bar for the scriptlets execution.

m-blaha:
RFE for the plymouth: https://bugzilla.redhat.com/show_bug.cgi?id=2350048

evan-goode (approved):
Works excellent, thanks!

PR #2092, opened 2025-02-27
Rename dnf5-makecache timer to dnf-makecache when dnf5_obsoletes_dnf

gordonmessmer:
Some maintainers of Fedora presets would prefer that the dnf-makecache timer use the unversioned name so that the presets (and Ansible playbooks, etc) do not need to be updated to account for renaming it. This change would create a file conflict between the "dnf" and "dnf5" packages in Fedora, but it's already impossible to install them simultaneously because dnf5 "obsoletes" the old dnf package.

This PR will use the unit name "dnf-makecache" when "dnf5_obsoletes_dnf", and "dnf5-makecache" when it does not obsolete dnf.

@keszybz @Conan-Kudo

### DIFF
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -237,7 +237,7 @@ Changes to individual commands
 
 ``makecache``
   * Metadata is now stored in different directories, see the ``cachedir`` configuration option :ref:`changes <cachedir_option_conf_changes_ref-label>` for more details.
-  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in ``dnf5-makecache.timer`` and the ``ConditionACPower`` setting in ``dnf5-makecache.service``.
+  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in ``dnf-makecache.timer`` and the ``ConditionACPower`` setting in ``dnf-makecache.service`` (these files are named ``dnf5-makecache.timer`` and ``dnf5-makecache.service`` on some systems).
 
 ``mark``
   * Renaming subcommands to be more intuitive: ``install`` -> ``user``, ``remove`` -> ``dependency``.
@@ -397,7 +397,7 @@ Additionally, corresponding command-line options ``--skip-broken`` and ``--skip-
 
 Deprecation of the ``metadata_timer_sync`` option
 -------------------------------------------------
-The ``metadata_timer_sync`` configuration option is now obsoleted by the ``dnf5-makecache.timer`` systemd timer settings.
+The ``metadata_timer_sync`` configuration option is now obsoleted by the ``dnf-makecache.timer`` systemd timer settings (named ``dnf5-makecache.timer`` on some systems).
 
 Changes to individual options
 -----------------------------
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -237,7 +237,8 @@ Changes to individual commands
 
 ``makecache``
   * Metadata is now stored in different directories, see the ``cachedir`` configuration option :ref:`changes <cachedir_option_conf_changes_ref-label>` for more details.
-  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in ``dnf-makecache.timer`` and the ``ConditionACPower`` setting in ``dnf-makecache.service`` (these files are named ``dnf5-makecache.timer`` and ``dnf5-makecache.service`` on some systems).
+
+  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in |DNF_MAKECACHE_TIMER_NAME_INLINE_LITERAL| and the ``ConditionACPower`` setting in |DNF_MAKECACHE_SERVICE_NAME_INLINE_LITERAL|.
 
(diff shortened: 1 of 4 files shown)


evan-goode:
I'm not sure it's a good idea to change this now. Won't this mean that users who have enabled `dnf5-makecache.timer` would now have a broken symlink in `/etc/systemd/system/timers.target.wants`? Could we perhaps instead provide a symlink from `dnf-makecache.timer` to `dnf5-makecache.timer`?

Conan-Kudo:
> a symlink named dnf-makecache -> dnf5-makecache would still require Fedora to modify presets. The symlink alone is insufficient to enable the timer by default

This is not hard for us to do. In fact, enabling the `dnf5-makecache` unit in presets is absolutely the easiest thing for us to do in Fedora. So this is not a valid objection.

gordonmessmer:
> This is not hard for us to do. In fact, enabling the dnf5-makecache unit in presets is absolutely the easiest thing for us to do in Fedora. So this is not a valid objection.

Every solution involves a number of compromises, and I am simply listing them.  I don't think any description is "invalid" unless it's incorrect.

Having said that, I can't tell from your feedback if you are in favor of this change, against this change, or what solution you would consider acceptable.

Conan-Kudo:
My core issue is that we can't really pretend dnf5 is like dnf4 or yum. There are tons of breaking changes and behavioral differences, even if we ignore the not-yet-reimplemented features. I get people want to do this, but we decided as a project to explicitly not do this because it creates mismatched expectations for people.

keszybz:
I think that's a misunderstanding of how the users view this. From a user's POV, yum/dnf/dnf5 are the thing that you call to install/update/remove/list/query packages. And those commands operate on rpms and repositories in the same way and actually even use the same repo config data. The internal architecture may be quite different, but as a user, I really really don't care about this. If it's faster and slightly more reliable, then I'm happy. If the option was renamed without providing backwards compat (why ever do that?!), then I'm a bit annoyed. If some feature is gone, I might delay upgrade… But it's still the same tool, and I won't install or use two in parallel, because it just doesn't make sense to have two separate views of the installed package set. I think that for some reason you're concentrating on *developers*, who would care about the APIs and architecture and all that, but the name of the main binary is for *users*. Our goal when creating the distro is to have a smooth experience where you don't need to figure out things and solve problems after every upgrade. Every time a program gets renamed, or options are changed incompatibly, or users need to reconfigure things after an update, we move further from this goal. Sometimes we don't have a choice, but let's agree that this is not a good thing.

Or in other words: yes, I understand that the libraries and code are quite different, and many things changed with dnf5, and some features changed incompatibly. Let's describe those in the docs and have users adjust, but let's not make any additional incompatible changes.

gordonmessmer:
I don't think there are meaningful changes where the "makecache" timer is concerned.  From the POV of the distribution or its users, enabling or disabling "dnf-makecache" should turn on or off dnf's background metadata refresh, whatever the version installed.

If this change isn't acceptable, can you opine about what change would be?

evan-goode:
> I don't think there are meaningful changes where the "makecache" timer is concerned. 

Agreed.

As I understand it our options are:

1. dnf5-makecache.timer (status quo)

    - Breaks (many) existing user enables/disables of dnf-makecache.timer
    - Breaks (many) scripts/playbooks enabling dnf-makecache.timer
    - We should update Fedora presets to enable dnf5-makecache.timer

2. dnf-makecache.timer

    - Breaks (few) existing user enables of dnf5-makecache.timer
    - Breaks (few) scripts/playbooks enabling dnf5-makecache.timer

3. dnf-makecache.timer AND symlink dnf5-makecache.timer -> dnf-makecache.timer (current state of PR)

    - Breaks (few) existing user enables of dnf5-makecache.timer

4. dnf5-makecache.timer AND symlink dnf-makecache.timer -> dnf5-makecache.timer

    - Breaks (many) existing user enables/disables of dnf-makecache.timer
    - We should update Fedora presets to enable dnf5-makecache.timer

@gordonmessmer I get your point that option 3 is the most pragmatic decision. I'll double-check with the DNF team and then merge this if there are no further objections.

evan-goode:
Rebased and updated a (newly added since the original PR) mention of `dnf5-makecache.{timer,service}` in `doc/changes_from_dnf4.7.rst`.

doc/changes_from_dnf4.7.rst
```diff
@@ -237,7 +237,7 @@ Changes to individual commands
 
 ``makecache``
   * Metadata is now stored in different directories, see the ``cachedir`` configuration option :ref:`changes <cachedir_option_conf_changes_ref-label>` for more details.
-  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in ``dnf5-makecache.timer`` and the ``ConditionACPower`` setting in ``dnf5-makecache.service``.
+  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` setting in ``dnf-makecache.timer`` and the ``ConditionACPower`` setting in ``dnf-makecache.service`` (these files are named ``dnf5-makecache.timer`` and ``dnf5-makecache.service`` on some systems).
```
ppisar, line 240: This rename in the documentation should also subject of the %{with dnf5_obsoletes_dnf} condition.

evan-goode: Done.

evan-goode (approved):

Issue #2065, opened 2025-02-14
Repo with missing key and `repo_gpgcheck` enabled causes fail even with `skit_if_unavailable`


kontura:
Given `missing.repo`:
```
[missing]
name=missing
enabled=1
repo_gpgcheck=1
gpgkey=file:///nonexistent
metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
```

```
$ dnf5 makecache --refresh  --setopt=skip_if_unavailable=true; echo $?
Updating and loading repositories:
 Fedora 42 - x86_64                                                                                                                                                       100% |  44.6 KiB/s |  21.0 KiB |  00m00s
 Fedora 42 - x86_64 - Updates                                                                                                                                             100% | 123.9 KiB/s |  23.9 KiB |  00m00s
 missing                                                                                                                                                                  100% |  47.2 KiB/s |  25.9 KiB |  00m01s
>>> Status code: 404 for https://mirror.karneval.cz/pub/linux/fedora/linux/development/42/Everything/x86_64/os/repodata/repomd.xml.asc (IP: 2a02:8301:0:2::150) - https://mirror.karneval.cz/pub/linux/fedora/linu
>>> Status code: 404 for https://mirror.karneval.cz/pub/linux/fedora/linux/development/42/Everything/x86_64/os/repodata/repomd.xml.asc (IP: 2a02:8301:0:2::150) - https://mirror.karneval.cz/pub/linux/fedora/linu
>>> Status code: 404 for https://mirror.karneval.cz/pub/linux/fedora/linux/development/42/Everything/x86_64/os/repodata/repomd.xml.asc (IP: 2a02:8301:0:2::150) - https://mirror.karneval.cz/pub/linux/fedora/linu
>>> Status code: 404 for https://mirror.karneval.cz/pub/linux/fedora/linux/development/42/Everything/x86_64/os/repodata/repomd.xml.asc (IP: 2a02:8301:0:2::150) - https://mirror.karneval.cz/pub/linux/fedora/linu
>>> Librepo error: GPG verification is enabled, but GPG signature is not available. This may be an error or the repository does not support GPG verification: Status code: 404 for https://mirror.karneval.cz/pub/
cannot open file: (2) - No such file or directory [/nonexistent]
1
```
vs dnf4:
```$ dnf4 makecache --refresh  --setopt=skip_if_unavailable=true; echo $?
[...]

Issue #2064, opened 2025-02-14
Repo without any source is silently ignored with `skip_if_unavailable=1`


kontura:
Given an `empty.repo` configured as:
```
[empty]
name=empty
enabled=1
gpgcheck=0
```
(there is no baseurl, mirrorlist or metalink)

Running:
```
$ dnf5 makecache --refresh --repo empty --setopt=skip_if_unavailable=1
Updating and loading repositories:
Repositories loaded.
Metadata cache created.
```

It would be nice to let the user know the enabled repo wasn't updated, similarly to dnf4:
```
$ dnf4 makecache --refresh --repo empty --setopt=skip_if_unavailable=1
Error: Cannot find a valid baseurl for repo: empty
Ignoring repositories: empty
Metadata cache created.
```

This is tested in https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/repo-sync.feature

PR #2094, opened 2025-02-27
repo: ignore key download errors if `skip_if_unavailable`

evan-goode:
In `repo::RepoSack::update_and_load_repos`, log and ignore errors related to reading or downloading GPG keys if we are using `skip_if_unavailable=false`

Resolves https://github.com/rpm-software-management/dnf5/issues/2065


evan-goode:
Related: https://github.com/rpm-software-management/dnf5/issues/2064. It would still be nice if a message were printed to stderr (not just the log) listing repos that were skipped.

kontura (approved):
Thanks.

PR #2095, opened 2025-02-28
FindRuby no longer provides upper-case RUBY_* variables

kontura:
Since cmake 4.0.0 version.

https://cmake.org/cmake/help/git-stage/policy/CMP0185.html
https://cmake.org/cmake/help/latest/module/FindRuby.html


jrohel:
LGTM

Issue #1965, opened 2024-12-18
Document behavior of error messages in dnf automatic


jan-kolarik:
In the dnf4 vs dnf5 documentation. In dnf4, it was `send_error_messages`, with a default value of `False`. Now, we send them by default.

PR #2096, opened 2025-02-28
doc: Removal of send_error_messages in automatic

m-blaha:
Document difference between dnf4 and dnf5 versions of the dnf automatic
behavior regarding to the `emitters.send_error_messages` configuration
option.

Fixes: https://github.com/rpm-software-management/dnf5/issues/1965


kontura:
Thanks, the failing tests are not related.

PR #2100, opened 2025-03-03
Replace in-tree crc32() with a call to zlib

dcantrell:
Replace the in-tree crc32() function with calls to zlib.  The only place I found the crc32 code was in was the config-manager plugin.

Fixes: SWM-1579

### DIFF
--- a/dnf5-plugins/config-manager_plugin/CMakeLists.txt
+++ b/dnf5-plugins/config-manager_plugin/CMakeLists.txt
@@ -14,6 +14,9 @@ target_link_libraries(libdnf5-cli PUBLIC ${LIBFMT_LIBRARIES})
 find_package(CURL 7.62.0 REQUIRED)
 include_directories(${CURL_INCLUDE_DIR})
 
+find_package(ZLIB REQUIRED)
+
+target_link_libraries(config-manager_cmd_plugin PRIVATE ZLIB::ZLIB)
 target_link_libraries(config-manager_cmd_plugin PRIVATE libdnf5 libdnf5-cli)
 target_link_libraries(config-manager_cmd_plugin PRIVATE dnf5)
 
--- a/dnf5-plugins/config-manager_plugin/addrepo.cpp
+++ b/dnf5-plugins/config-manager_plugin/addrepo.cpp
@@ -25,6 +25,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <fmt/format.h>
 #include <libdnf5/repo/file_downloader.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
+#include <zlib.h>
 
 #include <regex>
 #include <string_view>
@@ -52,27 +53,6 @@ std::string get_url_part(const std::string & url, CURLUPart what_part) {
     }
     curl_url_cleanup(c_url);
     return ret;
-}
-
-
-// Computes CRC32 checksum of input string.
-// Slow bitwise implementation. Used to calculate the hash of the omitted middle part of a long `repoid` if the `repoid`
-// is generated from a URL. Not used for cryptography.
-uint32_t crc32(std::string_view input) {
-    const uint32_t polynomial = 0x04C11DB7;
-    uint32_t crc = 0;
-
-    for (auto ch : input) {
-        crc ^= static_cast<uint32_t>(ch) << 24;
-        for (int i = 0; i < 8; ++i) {
-            if ((crc & 0x80000000) != 0) {
-                crc = (crc << 1) ^ polynomial;
-            } else {
-                crc <<= 1;
-            }
-        }
-    }
-    return crc;
 }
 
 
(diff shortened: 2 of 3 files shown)


dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -116,9 +96,11 @@ std::string generate_repoid_from_url(const std::string & url) {
     // Limits the length of the repository id.
     // Copies the first and last 100 characters. The substring in between is replaced by a crc32 hash.
     if (ret.size() > 250) {
-        std::string_view tmp{ret};
+        unsigned int sz = static_cast <unsigned int> (ret.size());
+        char middle[sz - 200]{};
+        ret.copy(middle, sz - 200, 100);
```
ppisar, line 101: Do we have to copy the substring instead of passing `ret.c_str() + 100` to crc32()?

m-blaha:
There was issue https://github.com/rpm-software-management/dnf5/issues/267 on failing builds due to under-linking. Should we add the zlib dependency also to the cmake files and link the config-manager plugin with zlib?

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -116,9 +96,11 @@ std::string generate_repoid_from_url(const std::string & url) {
...
         ret = fmt::format(
-            "{}-{:08X}-{}", tmp.substr(0, 100), crc32(tmp.substr(100, tmp.size() - 200)), tmp.substr(tmp.size() - 100));
+            "{}-{:08X}-{}", ret.substr(0, 100), crc32(0, (unsigned char *) middle, sz - 200), ret.substr(sz - 100));
```
ppisar, line 103: crc32_z() has size_t type of the last argument. If you used that, you wouldn't have to use the unsafe static_cast.

dnf5-plugins/config-manager_plugin/addrepo.cpp
dcantrell: No, we don't.  I'll fix that.

dnf5-plugins/config-manager_plugin/addrepo.cpp
dcantrell: Ah yes, I'll change that over to crc32_z().

dcantrell:
> There was issue #267 on failing builds due to under-linking. Should we add the zlib dependency also to the cmake files and link the config-manager plugin with zlib?

Yes.  I included the edit to the spec file, but forgot the cmake changes.  I'll get on that.

dnf5-plugins/config-manager_plugin/addrepo.cpp
ppisar: I think that now you can change sz variable type to size_t and remove the static_cast<>.

dnf5.spec
```diff
@@ -158,6 +158,7 @@ BuildRequires:  pkgconfig(libsolvext) >= %{libsolv_version}
 BuildRequires:  pkgconfig(rpm) >= 4.17.0
 BuildRequires:  pkgconfig(sqlite3) >= %{sqlite_version}
 BuildRequires:  toml11-static
+BuildRequires:  zlib-level
```
ppisar, line 161: s/zlib-level/zlib-devel/

dcantrell: My fingers are too fat for the keyboard.

dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -116,9 +96,9 @@ std::string generate_repoid_from_url(const std::string & url) {
     // Limits the length of the repository id.
     // Copies the first and last 100 characters. The substring in between is replaced by a crc32 hash.
     if (ret.size() > 250) {
-        std::string_view tmp{ret};
+        size_t sz = ret.size();
         ret = fmt::format(
-            "{}-{:08X}-{}", tmp.substr(0, 100), crc32(tmp.substr(100, tmp.size() - 200)), tmp.substr(tmp.size() - 100));
+            "{}-{:08X}-{}", ret.substr(0, 100), crc32_z(0, ret.c_str() + 100, sz - 200), ret.substr(sz - 100));
```
ppisar, line 101: I tried to build it (because CI cloud is broken today) and it fails with GCC 15 on Fedora 43 with:

~~~~
/home/test/dnf5/dnf5-plugins/config-manager_plugin/addrepo.cpp:101:72: error: invalid conversion from  const char*’ to ‘const unsigned char*’ [-fpermissive]
  101 |             "{}-{:08X}-{}", ret.substr(0, 100), crc32_z(0, ret.c_str() + 100, sz - 200), ret.substr(sz - 100));
      |                                                            ~~~~~~~~~~~~^~~~~
      |                                                                        |
      |                                                                        const char*
In file included from /home/test/dnf5/dnf5-plugins/config-manager_plugin/addrepo.cpp:28:
/usr/include/zlib.h:1723:81: note: initializing argument 2 of ‘long unsigned int crc32_z(long unsigned int, const unsigned char*, size_t)’
 1723 | Z_EXTERN unsigned long Z_EXPORT crc32_z(unsigned long crc, const unsigned char *buf, size_t len);
      |                                                            ~~~~~~~~~~~~~~~~~~~~~^~~
~~~~

I think we will need to explicitly cast the second argument to "const unsigned char *" as it what zlib.h expects.

dnf5-plugins/config-manager_plugin/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ target_link_libraries(libdnf5-cli PUBLIC ${LIBFMT_LIBRARIES})
 find_package(CURL 7.62.0 REQUIRED)
 include_directories(${CURL_INCLUDE_DIR})
 
+find_package(ZLIB REQUIRED)
+
+target_link_libraries(config-manager_cmd_plugin PRIVATE zlib)
```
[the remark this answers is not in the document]
dcantrell: Fixed.

ppisar (commented):
Please reindent the function call so that clang-format <https://github.com/rpm-software-management/dnf5/actions/runs/13682883474/job/38259321117?pr=2100> is happy.

dcantrell:
> Please reindent the function call so that clang-format https://github.com/rpm-software-management/dnf5/actions/runs/13682883474/job/38259321117?pr=2100 is happy.

Done.

ppisar (approved):
Now it's perfect. Thanks.

Issue #812, opened 2023-08-10
Implement or document missing option - dnf [options] makecache --timer


j-mracek:

ppisar:
Relatated question is dnf-makecache.timer systemd unit. Is dnf5-automatic.timer with download_updates=False good replacement?

PR #2101, opened 2025-03-04
doc: Document dropping of makecache --timer option

m-blaha:
The option (along with the metadata_timer_sync config option) was dropped in favor of systemd settings.

Fixes: https://github.com/rpm-software-management/dnf5/issues/812

### DIFF
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -394,6 +394,9 @@ To address this, the functionality has been split into two configuration options
 
 Additionally, corresponding command-line options ``--skip-broken`` and ``--skip-unavailable`` have been introduced for commands where applicable.
 
+Deprecation of the ``metadata_timer_sync`` option
+-------------------------------------------------
+The ``metadata_timer_sync`` configuration option is now obsoleted by the ``dnf5-makecache.timer`` systemd timer settings.
 
 Changes to individual options
 -----------------------------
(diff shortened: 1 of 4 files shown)


doc/changes_from_dnf4.7.rst
```diff
@@ -387,6 +387,9 @@ To address this, the functionality has been split into two configuration options
 
 Additionally, corresponding command-line options ``--skip-broken`` and ``--skip-unavailable`` have been introduced for commands where applicable.
 
+Deprecation of the ``metadata_timer_sync`` option
+-------------------------------------------------
+The ``metadata_timer_sync`` configuration option is now deprecated in favor of systemd timer settings.
```
ppisar, line 393: I would rather phrase is as "is now obsoleted by dnf5-makecache.timer systemd timer settings". The reason is that the option does nothing. "Deprecated" would mean that users still can use it and that it has an effect.

doc/changes_from_dnf4.7.rst
```diff
@@ -230,6 +230,7 @@ Changes to individual commands
 
 ``makecache``
   * Metadata is now stored in different directories, see the ``cachedir`` configuration option :ref:`changes <cachedir_option_conf_changes_ref-label>` for more details.
+  * The ``--timer`` option has been dropped in favor of the systemd ``OnUnitInactiveSec`` and ``ConditionACPower`` options.
```
ppisar, line 233: Maybe you could here name the systemd unit name. (I know we have users willing to change a name of the unit, but that's a different problem.)

doc/changes_from_dnf4.7.rst
m-blaha: Make sense, thanks!

doc/changes_from_dnf4.7.rst
m-blaha: Thank you, added.

ppisar (approved):
The patches are good. The CI failure is an unrelated infrastructure failure. I built the code and documentation locally.

PR #2105, opened 2025-03-04
system-upgrade: Add --allowerasing switch

m-blaha:
This patch adds the --allowerasing option to the `dnf system-upgrade download`
command. It helps users resolve dependency issues during the system upgrade.
The use of this option is even documented in the Fedora documentation:
https://docs.fedoraproject.org/en-US/quick-docs/upgrading-fedora-offline/

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2349754

### DIFF
--- a/dnf5/commands/system-upgrade/system-upgrade.cpp
+++ b/dnf5/commands/system-upgrade/system-upgrade.cpp
@@ -81,6 +81,8 @@ void SystemUpgradeDownloadCommand::set_argument_parser() {
     no_downgrade_arg->set_const_value("true");
     no_downgrade_arg->link_value(no_downgrade);
     cmd.register_named_arg(no_downgrade_arg);
+
+    allow_erasing = std::make_unique<AllowErasingOption>(*this);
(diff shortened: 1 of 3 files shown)


dcantrell (approved):

kontura:
Should the option also be configured for the daemon?

m-blaha:
> Should the option also be configured for the daemon?

Good point!
Actually, the option is already present in the dnf5daemon-client - as a global one. But the problem is that it's not handled in any way, neither on the client side, nor on the server side. As this would be a bigger change in the daemon, I would prefer to track it in a separate issue - https://github.com/rpm-software-management/dnf5/issues/2109 .

kparal:
> The use of this option is even documented in the Fedora documentation:

The documentation also mentions `--best` and `--setopt=keepcache=1`. Should those be enabled as well? Or perhaps even some additional ones?

m-blaha:
These are global options available for all commands.

PR #2106, opened 2025-03-04
Goal: Handle INSTALL and INSTALL_BY_COMPS group actions as INSTALL

pkratoch:


pkratoch:
Hm, the test "Upgrade environments with conflicting requests (groups are not removed)" fails. There is a situation when a group is removed from one environment and added to another and it results in group-install action, but should be group-upgrade. So, there is still a bug. I will look into why that is tomorrow.

kontura (approved):
Thank you for fixing this.
It looks good to me.

pkratoch:
> Hm, the test "Upgrade environments with conflicting requests (groups are not removed)" fails. There is a situation when a group is removed from one environment and added to another and it results in group-install action, but should be group-upgrade. So, there is still a bug. I will look into why that is tomorrow.

Fixed.

Issue #2085, opened 2025-02-25
Crash while rendering transaction progressbar if terminal has less than 4 columns: terminate called after throwing an instance of 'std::length_error': what():  basic_string::_M_replace_aux


hughsie:
Since updating to dnf5-5.2.10.0-2.fc41 all my Fedora github-action CI builds are failing with:

    terminate called after throwing an instance of 'std::length_error'
      what():  basic_string::_M_replace_aux
    (core dumped) dnf install -y dist/*.rpm

This morning I tried grabbing the as-built-in-github docker container from the container store, and ran it _successfully_ with podman **interactively** on Fedora 41. It's certainly something to do with running without a tty, or with a "small" tty. I couldn't reproduce the bug either using `ENV COLUMNS` or with `stty cols 10 rows 100`

Frustratingly, it's also impossible to "dnf downgrade dnf5 -y" in CI as it also crashes during downgrading itself. 

This seems to me that DNF is trying to create a huuuge string -- i.e. something like https://github.com/rpm-software-management/dnf5/commit/8974d33bd97396c148181474dd9efc734a05a072 looks really _dodgy_ with regards to integer underflow.

kontura:
It might be possible to workaround this by redirecting the output to a file or by setting environment variable `FORCE_COLUMNS` which dnf5 reads.

Perhaps something like: `FORCE_COLUMNS=100 dnf downgrade dnf5 -y`?

Can you link the failing build?

hughsie:
@kontura here's a good example: https://github.com/fwupd/fwupd/actions/runs/13522031670/job/37783483182?pr=8501 -- using `FORCE_COLUMNS=100` is a good workaround, thanks.

@Luap99 I've had no luck with removing the tty on GitHub actions; same problem.

suve:
On Fedora's devel mailing list, František Šumšal suggested that running `stty cols 1 && dnf reinstall -y bash` can reliably trigger this error. I can confirm that indeed it does.

@hughsie's suspicions seem to be correct; gdb points to the following code as the culprit:
```
        const auto prefix_width = libdnf5::cli::utils::utf8::width(prefix);
        const auto message_width = libdnf5::cli::utils::utf8::width(message);
        if (message_width < terminal_width - prefix_width) {
            message.append(terminal_width - message_width - prefix_width, ' ');
        }
```
With the "append" line being what triggers the write and then the assertion.

Looking at local variables shows:
```
[...]

Issue #1829, opened 2024-11-04
`DownloadProgressBar` truncates messages to terminal width


kontura:
Longer messages are cut short, this could be confusing for users.

For example, test scenario: `Scenario: present user understandable message when there is a mismatch between available repodata and packages` in [metadata.feature](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/metadata.feature) produces output like:
```
      Updating and loading repositories:
       simple-base test repository            100% | 464.8 KiB/s |   2.3 KiB |  00m00s
      Repositories loaded.
      Total size of inbound packages is 6 KiB. Need to download 6 KiB.
      After this operation, 0 B extra will be used (install 0 B, remove 0 B).
      [1/1] labirinto-0:1.0-1.fc29.x86_64       0% |   0.0   B/s |   0.0   B |  00m00s
      >>> Interrupted by header callback: Inconsistent server data, reported file Cont
      >>> Interrupted by header callback: Inconsistent server data, reported file Cont
      >>> Interrupted by header callback: Inconsistent server data, reported file Cont
      >>> Interrupted by header callback: Inconsistent server data, reported file Cont
      >>> No more mirrors to try - All mirrors were already tried without success
      --------------------------------------------------------------------------------
      [1/1] Total                             100% |   0.0   B/s |   0.0   B |  00m00s
      Failed to download packages
       Librepo error: Cannot download x86_64/labirinto-1.0-1.fc29.x86_64.rpm: All mirrors were tried
```

I don't think this is user understandable message.

kontura:
Fedora report: https://bugzilla.redhat.com/show_bug.cgi?id=2335693

PR #2110, opened 2025-03-06
progressbar: Messages printing on narrow terminal

m-blaha:
Handle situations, when the message (or even the prefix) is wider than 
available terminal size.

Fixes: https://github.com/rpm-software-management/dnf5/issues/2085

### DIFF
--- a/libdnf5-cli/progressbar/download_progress_bar.cpp
+++ b/libdnf5-cli/progressbar/download_progress_bar.cpp
@@ -71,6 +71,12 @@ std::ostream & operator<<(std::ostream & stream, DownloadProgressBar & bar) {
 void DownloadProgressBar::to_stream(std::ostream & stream) {
     update();
 
+    if (!is_finished() && !tty::is_interactive()) {
+        // don't spam non-interactive terminal with progressbar's progress
+        // only print the final form
+        return;
+    }
+
     // set the default delimiters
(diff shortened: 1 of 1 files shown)


m-blaha:
This PR just fixes the problem with integer underflow. There are still remaining issues with the messages in progressbars - e.g. https://github.com/rpm-software-management/dnf5/issues/1829

libdnf5-cli/progressbar/download_progress_bar.cpp
```diff
@@ -71,6 +71,12 @@ std::ostream & operator<<(std::ostream & stream, DownloadProgressBar & bar) {
 void DownloadProgressBar::to_stream(std::ostream & stream) {
     update();
 
+    if (get_state() == ProgressBarState::STARTED && !tty::is_interactive()) {
```
ppisar, line 74: What are about other states? I guess ERROR and SUCCESS are final states. WARNING is probably nowhere used. Shouldn't READY be handled here? Wouldn't be better to compare against is_finished()?

m-blaha: I've changed the condition to `!is_finished()` anyway. It definitely makes sense.

ppisar (approved):
Good.

KKoukiou:
@ppisar can you please backport this fix also in f42?

Issue #291, opened 2023-02-14
DNF5 Python API raises only RuntimeError


poncovka:
It looks like the only type of exception raised by the DNF5 Python API is `RuntimeError`.

```
base = libdnf5.base.Base()
query = libdnf5.repo.RepoQuery(base)
query.get()
```

The example above prints the traceback below:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.11/site-packages/libdnf5/repo.py", line 1779, in get
    return _repo.SackQueryRepoWeakPtr_get(self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Query must contain exactly one object.
```

Based on my experiments, this is the only exception type I get no matter what I call, so I guess there is an issue with Python bindings. Luckily, Anaconda installer catches DNF exceptions very early and replaces them with its own set of exceptions, so we shouldn't be affected by this issue too much. However, I am sure there will be projects that won't be happy about this.

jan-kolarik:
This is the relevant block of code in SWIG interface where the mapping is implemented: https://github.com/rpm-software-management/dnf5/blob/5.1.14/bindings/libdnf5/base.i#L25.

PR #2118, opened 2025-03-09
Swig exceptions handling

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/291
and can solve https://github.com/ansible/ansible/issues/84634

Description:

Exceptions were passed as `SWIG_RuntimeError`. The original exceptions type was dropped and the information contained too. Only the `what()` message was retained. Nested exceptions were completely ignored.

**Now the original exceptions, including nested exceptions, are passed to the target language.** The exceptions implement the `rethrow_if_nested` method, which can be used to raise a nested exception if it is present. And the `format` method, which formats the exception error message, including the error messages of nested exceptions.

Contains unit tests in Python, Ruby and Perl.


jrohel:
Rebased

bindings/libdnf5/exception.hpp
```diff
@@ -0,0 +1,137 @@
...
+        swig_if_exception(e, libdnf5::OptionBindsError);
+
+        swig_if_exception(e, libdnf5::ReadOnlyVariableError);
+
+        swig_if_exception(e, libdnf5::FileSystemError);
+        swig_if_exception(e, libdnf5::SystemError);
+        swig_if_exception(e, libdnf5::RuntimeError);
+        swig_if_exception(e, libdnf5::Error);
+        swig_if_exception(e, libdnf5::UserAssertionError);
+        swig_if_exception(e, libdnf5::AssertionError);
```
jan-kolarik, line 116: If I understand correctly, adding a new exception in C++ means we also need to update the SWIG code and add the mapping there. That's quite an overhead, but if there’s no way around it, we should at least document the process - maybe in https://dnf5.readthedocs.io/en/latest/best_practices/index.html.

I was thinking of a possibly simpler approach: registering each exception in a map on the C++ side, then iterating over it in SWIG during module initialization to dynamically create exception types. Let me know please if this sounds feasible or if it wouldn't work for our use case.

jan-kolarik (commented):
Overall, the implementation and tests look great. I just mentioned one potential idea to consider, which might simplify the overhead when adding new exceptions.

bindings/libdnf5/exception.hpp
jrohel: I made macros to make adding an exception to the swig easy.

But you're right. It is still work to add an exception:
Add an exception handler here. In addition, add a line defining the nested exception to "exception.i". And if the new exception will be in a new header file, it will need to be appended in both "exception.i" and "exception.hpp". And it would also be useful to set "%ignore" to the constructor in "exception.i" for the new exception. 

> I was thinking of a possibly simpler approach: registering each exception in a map on the C++ side,

This is in C++. You mean the code directly in the libdnf5 library?
I don't know how much easier/more complicated it will be to dirty the libdnf5 library with code to support swig. And even then, it won't guarantee that it will be work-free on the swig side. And I mainly wanted to avoid dirtying the libdnf5 library API with something that will only support swig. 

I may try even smarter macros on the swig side in the future. That's our internal thing that can be done later.

jan-kolarik: OK, sounds like the current approach is good to go. Please just follow-up with the PR with documentation.

jan-kolarik:
Please rebase again, the failing tests should disappear then.

jan-kolarik:
Thanks! There are some failures of dnf5daemon tests, but they are unrelated, seems like some issues/changes in `polkit`?

ppisar:
The newly exposed exceptions are not subtypes of RuntimeError. That breaks API as it requires applications to catch different types of exceptions. See <https://bugzilla.redhat.com/show_bug.cgi?id=2365689>.

PR #2124, opened 2025-03-11
libdnf5 and libdnf5-cli: Move exception declarations to own header files, unification

jrohel:
- Some of the exceptions have already been declared in their own header files. But not all of them.  This PR unifies the approach. Separating exception declarations from other declarations simplifies their subsequent use, such as wrapping in SWIG.

- Exceptions in libdnf5 are declared as `class`. Only two (`libdnf5::OptionBindsError` and `lindnf5::rpm::NevraIncorrectInputError`) were declared as `struct`. Unification, they are now also declared as `class`.


m-blaha:
Looks like you forgot to include `transaction_item_errors.hpp` file.

jrohel:
@m-blaha Yes. You're right. Thank you. I added the file.

m-blaha (approved):
Thank you!

PR #2125, opened 2025-03-11
offline: only define systemd constants when building with systemd 

katexochen:
Fixing the following error when building with WITH_SYSTEMD=OFF

> /build/source/dnf5/commands/offline/offline.cpp:54:7: error: 'sdbus' does not name a type
>    54 | const sdbus::ObjectPath SYSTEMD_OBJECT_PATH{"/org/freedesktop/systemd1"};

### DIFF
--- a/dnf5/commands/offline/offline.cpp
+++ b/dnf5/commands/offline/offline.cpp
@@ -50,11 +50,13 @@ using namespace libdnf5::cli;
 
 const std::string & ID_TO_IDENTIFY_BOOTS = libdnf5::offline::OFFLINE_STARTED_ID;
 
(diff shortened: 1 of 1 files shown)


m-blaha:
Given that the `SYSTEMD_OBJECT_PATH` constant is used only in parts of the code guarded by `#ifdef WITH_SYSTEMD`, would it make sense to also define the constant conditionally using the same #ifdef?

m-blaha:
Actually all the `SYSTEMD_*` constants can be defined conditionally.

m-blaha (approved):
Thank you!

PR #2130, opened 2025-03-12
Fix builddep command in CONTRIBUTING.md

samzeter:
the command `dnf builddep dnf5.spec` results in the following error:

Repositories loaded.
error: Unable to open dnf5.spec: No such file or directory Failed to parse spec file "dnf5.spec".
Failed to parse some inputs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ To setup a development environment, complete the following steps:
 1. Install build requirements:
 
    ```
-   dnf builddep dnf5.spec #[--define '_without_<option> 1 ...]
+   dnf builddep dnf5 #[--define '_without_<option> 1 ...]
    ```
 
 2. Build DNF5:


samzeter:
Tbh, I'm not even sure #[--define '_without_<option> 1 ...] is necessary, nor do I understand what it is about.

ppisar:
The original command is fine. It is intended to process a spec file from this git repository. With your change, it would process a source package from a your version of your distribution. Probably something different than is required for building source code from this repository.

Maybe better change would be to "./dnf5.spec" or to the very explicit "dnf builddep --spec dnf5.spec".

ppisar:
> Tbh, I'm not even sure #[--define '_without_ 1 ...] is necessary, nor do I understand what it is about.

That's an optional argument passing an RPM macro to a spec parser. E.g. <code>--define '_without_plugin_rhsm 1' </code>. Look up for %bcond_without in the spec file for various build-time options.

PR #2131, opened 2025-03-12
Fix dnf5 copr_plugin: always set `base` in `CoprRepo`

jrohel:
Constructing incomplete (with `base` set to `nullptr`) instances of `CoprRepo` is not a good idea. It is only a matter of time before someone uses this invalid `base`.

In fact, this is why PR https://github.com/rpm-software-management/dnf5/pull/2081 causes segfault in copr_plugin unit tests.


evan-goode (approved):
LGTM

PR #2132, opened 2025-03-12
Fix a memory leak when looking up for an OpenPGP key in RPM database

ppisar:
headerGetAsString() returns a pointer which the caller must free. Also if the call fails, it will return NULL and later std::string == operator will exhibit an undefined behviour (confirmed with an address sanitizer).

There were two places of the flawed code in
libdnf5::rpm::rpmdb_lookup() and in expired-pgp-keys plugin. This patch fixes both. Though the latter will be reimplemented later.


jrohel (approved):
LGTM. Thanks

Issue #1727, opened 2024-09-24
Give a more useful error message on image-mode systems


jlebon:
Some of the image-mode Fedora variants have started shipping dnf5 for layering purposes (see e.g. https://github.com/coreos/fedora-coreos-tracker/issues/1687). However, this is now confusing users trying to use it client-side: https://discussion.fedoraproject.org/t/f41-kinoite-native-container-dnf5-pgp-key-import-error/131946.

We need better error messages, similar to what was done in dnf4, in dnf5:
- https://github.com/rpm-software-management/dnf/pull/2053
- https://github.com/rpm-software-management/dnf/pull/2110
- https://github.com/rpm-software-management/dnf/pull/2112

jan-kolarik:
Hi, I see, thanks for the report. We will plan this to have it resolved soon in a same or similar manner as with the dnf4 reports.

Malix-Labs:
I ran into a problem with DNF5 on Fedora 41 (through on Bluefin) that would have been more easily debugged by this :) 

https://universal-blue.discourse.group/t/bluefin-stable-stream-is-now-based-on-fedora-41/5061/7?u=malix

travier:
It would be great if we could get this in dnf5 soon as users are confused about it on image mode / bootc systems.

PR #2133, opened 2025-03-14
Throw error message on read-only image-mode systems

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/1727.

There is a bit more boilerplate here than is immediately necessary, but we'll soon want to port over the `--transient` functionality for DNF 4, so this PR is keeping those changes in mind.


libdnf5/utils/bootc.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace libdnf5::utils::bootc {
+
+bool is_bootc_system() {
+    const std::filesystem::path ostree_booted{"/run/ostree-booted"};
+    return std::filesystem::is_regular_file(ostree_booted);
+}
+
+bool is_writable() {
+    return access("/usr", W_OK) == 0;
```
ppisar, line 34: This does not take into account --installroot as DNF4 does: If the --installroot differ from /, DNF4 assumes the destination is writable. A reason is that people use dnf on bootc system to install operating systems into a directory. E.g. mock tool does that.

dnf5/main.cpp
```diff
@@ -1493,6 +1494,13 @@ int main(int argc, char * argv[]) try {
                                 "checks will be performed."));
                         }
                     }
+                    if (libdnf5::utils::bootc::is_bootc_system()) {
```
ppisar, line 1497: This does not take into account --downloadonly option as DNF4 does. DNF4 users complained that they want to download packages (into standard /var/cache path) even on a locked bootc system. Probably they use it for mirroring updates.

include/libdnf5-cli/exception.hpp
```diff
@@ -54,6 +54,14 @@ class InsufficientPrivilegesError : public Error {
 };
 
 
+/// Exception is thrown when trying to write to a read-only system
+class ReadOnlySystemError : public Error {
```
ppisar, line 58: This enhances API and thus we should bump minor version of DNF5.

dnf5/main.cpp
evan-goode: I think it does respect `downloadonly` (and `--store` too) since this logic is in the `else` block here, right?

ppisar: You are right. I overlooked the nesting.

include/libdnf5-cli/exception.hpp
evan-goode: Bumped.

libdnf5/utils/bootc.cpp
evan-goode: Non-/ installroot works as expected.

include/libdnf5-cli/exception.hpp
ppisar: Could you move that bumping commit before a commit which adds this class? Or simply squash the two commits?

evan-goode: OK, I put the bumping commit first since the "Add bootc utility functions" commit also enhances libdnf5.

ppisar:
Now the code looks good. I was able to partially tests it (with remounting /usr read-only and creating the file in /run), (Partially means that my DNF5 build got mad while loading repository data and an address sanitized started to throw errors even without your patches. So highly unrelated to your patches.) After finishing CI I will approve this pull request.

Conan-Kudo:
Why isn't this a plugin? We don't want to have a repeat of the modularity stuff where we need to figure out how to pull it out later.

dustymabe:
Not sure if there is a way, but it would be nice if we could do the check before we download many MiB of repo metadata:

```
[core@cosa-devsh ~]$ sudo dnf install htop 
 Fedora 42 - x86_64 - Updates Archive                                                                                                                                                               100% |  12.2 MiB/s |  10.4 MiB |  00m01s
 Fedora 42 openh264 (From Cisco) - x86_64                                                                                                                                                           100% |   8.9 KiB/s |   6.0 KiB |  00m01s
 Fedora 42 - x86_64 - Updates                                                                                                                                                                       100% |   3.2 MiB/s |   5.7 MiB |  00m02s
 Fedora 42 - x86_64                                                                                                                                                                                 100% |  15.2 MiB/s |  35.4 MiB |  00m02s
Repositories loaded.
Package                                                                          Arch              Version                                                                          Repository                                          Size
Installing:
 htop                                                                            x86_64            3.4.1-1.fc42                                                                     updates                                        456.4 KiB
Installing dependencies:
 hwloc-libs                                                                      x86_64            2.12.0-1.fc42                                                                    fedora                                           2.9 MiB

Transaction Summary:
 Installing:         2 packages

Total size of inbound packages is 2 MiB. Need to download 2 MiB.
After this operation, 3 MiB extra will be used (install 3 MiB, remove 0 B).
Error: this bootc system is configured to be read-only. For more information, run `bootc --help`.
```

PR #2137, opened 2025-03-17
spec: Move /usr/lib/sysimage/libdnf5 from dnf5 to libdnf5

ppisar:
That directory is used by libdnf5. dnf5daemon-client and dnf5daemon-server is independent on dnf5.


m-blaha (approved):
Thank you!

Issue #2134, opened 2025-03-14
[REGRESSION] dnf5 does not pull signing keys


madonuko:
`dnf5` `5.2.11.0` has a regression where it refuses to pull missing signing keys from repos during syncing. This worked fine on `5.2.10.0`.

## To Reproduce

1. `podman pull ghcr.io/terrapkg/builder:frawhide`, this pulls the latest podman image from [our builders](https://github.com/terrapkg/builder/pkgs/container/builder)
As of today, the sha is a5f3fd92ce13
2. `podman run -it ghcr.io/terrapkg/builder:frawhide`
3. In the shell, run `dnf in -y 'pkgconfig(astal-io-0.1)'`

```
Updating and loading repositories:
 Fedora - Rawhide - Developmental packa 100% |  24.9 KiB/s |  10.0 KiB |  00m00s
 packages for the GitHub CLI            100% | 115.4 KiB/s |   3.0 KiB |  00m00s
 Fedora - Rawhide - Developmental packa 100% |   1.7 MiB/s |   4.0 MiB |  00m02s
 Terra rawhide (Extras)                 100% |   3.8 KiB/s |   6.6 KiB |  00m02s
>>> Librepo error: repomd.xml GPG signature verification error: Signing key not 
 Terra rawhide                          100% |   4.6 KiB/s |   7.8 KiB |  00m02s
>>> Librepo error: repomd.xml GPG signature verification error: Signing key not 
Repositories loaded.
Failed to resolve the transaction:
No match for argument: pkgconfig(astal-io-0.1)
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
```

## Old Behaviour

This is generated using `podman run -it 8b8895c43b10` (an older builder image with `dnf5` `5.2.10.0`)

```
Updating and loading repositories:
 packages for the GitHub CLI                                               100% |  14.6 KiB/s |   3.0 KiB |  00m00s
 Fedora rawhide openh264 (From Cisco) - x86_64                             100% |   3.2 KiB/s | 986.0   B |  00m00s
 Terra rawhide (Extras)                                                    100% |   4.3 KiB/s |   6.6 KiB |  00m02s
>>> Librepo error: repomd.xml GPG signature verification error: Signing key not found                              
 Fedora rawhide openh264 (From Cisco) - x86_64                             100% |   4.2 KiB/s |   5.8 KiB |  00m01s
[...]

PR #2138, opened 2025-03-17
repo: fix bad GPG error handling

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/2134, which is a regression caused by commit 0f1242b0.

`std::rethrow_if_nested` is useless if we first slice the `std::nested_exception` to a `std::exception`. We must call `std::rethrow_if_nested` on a instance of a type derived from (or convertible to?) `std::nested_exception` [1].

[1] https://en.cppreference.com/w/cpp/error/rethrow_if_nested

We are still missing tests for repo_gpgcheck in ci-dnf-stack, but we should fix https://github.com/rpm-software-management/dnf5/issues/2134 and cut a new release ASAP.

### DIFF
--- a/libdnf5/repo/repo_sack.cpp
+++ b/libdnf5/repo/repo_sack.cpp
@@ -400,28 +400,31 @@ void RepoSack::Impl::update_and_load_repos(libdnf5::repo::RepoQuery & repos, boo
 
     auto handle_repo_exception = [&](const Repo * repo, std::exception_ptr ep, bool report_key_err) {
         // Use an exception_ptr to preserve the original type of the exception, in case we re-throw it.
-        std::exception e;
+        std::exception exception;
         try {
             std::rethrow_exception(ep);
-        } catch (const std::exception & exception) {
-            e = exception;
-        }
-        if (report_key_err) {
-            try {
-                std::rethrow_if_nested(e);
-            } catch (const LibrepoError & lr_err) {
-                if (lr_err.get_code() == LRE_BADGPG) {
-                    return true;
+        } catch (const RepoDownloadError & rd_err) {
+            exception = rd_err;
+            if (report_key_err) {
+                try {
+                    std::rethrow_if_nested(rd_err);
+                } catch (const LibrepoError & lr_err) {
+                    if (lr_err.get_code() == LRE_BADGPG) {
+                        return true;
+                    }
+                } catch (...) {
                 }
-            } catch (...) {
             }
+        } catch (const std::exception & e) {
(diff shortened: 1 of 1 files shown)


libdnf5/repo/repo_sack.cpp
```diff
@@ -400,28 +400,31 @@ void RepoSack::Impl::update_and_load_repos(libdnf5::repo::RepoQuery & repos, boo
...
-            try {
-                std::rethrow_if_nested(e);
-            } catch (const LibrepoError & lr_err) {
-                if (lr_err.get_code() == LRE_BADGPG) {
-                    return true;
+        } catch (const RepoDownloadError & rd_err) {
+            exception = rd_err;
+            if (report_key_err) {
+                try {
+                    std::rethrow_if_nested(rd_err);
```
kontura, line 410: I think there is now a different bug when the `handle_repo_exception` gets non-nested `RepoDownloadError`.
For example from here: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/repo/repo_downloader.cpp#L474-L475 so basically https://github.com/rpm-software-management/dnf5/issues/2064.

It produces logs like:
```
2025-03-18T13:17:01+0000 [40529] WARNING Error loading repo "empty" (skipping due to "skip_if_unavailable=true"):
2025-03-18T13:17:01+0000 [40529] WARNING  std::exception
```
I think it can also happen from: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/repo/librepo.cpp#L117-L119

However these seem like far less serious problems so if you don't want to deal with that now I guess we can merge this and fix that later.

kontura (approved):
I have verified it fixes the regression.

libdnf5/repo/repo_sack.cpp
kontura: Actually the issue was already present before.
Lets fix it later.

kontura: I think its not caused by the non-nested `RepoDownloadError` but rather by `format(...)`ing the sliced `std::exception`.

evan-goode: Yes, was just fixing that, it can wait for another PR.

evan-goode:
Thanks!

jodobear:
I just installed Fedora Server 42, upgraded it and tried to install VS Codium and have the same issue:

```bash
epomd.xml GPG signature verification error: Signing key not found                                                                                                             
 VSCodium
```

PR #2139, opened 2025-03-17
CI: Support releasing from stable branches (release-5.2.x.x)

evan-goode:
These changes will help us modify our workflow to release to stable Fedora versions from stable versioned branches (e.g. `release-5.2.x.x`) instead of releasing always from `main`.

For the backport bot, we'll need to create a GitHub App to use in the workflow since the default GitHub Actions bot has some anti-infinite-loop protections built-in and GHA CI jobs will not run on GHA-created PRs ([allegedly](https://github.com/NixOS/nixpkgs/pull/360260)).

Following this PR, I'll set up Packit on the `release-5.2.x.x` branch to release to Fedora 41 and 42.
We'll also need to set up a `release-5.2.x.x` branch and backport workflow on ci-dnf-stack.


.github/workflows/prepare-release.yml
```diff
@@ -15,6 +18,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ref: ${{ inputs.branch }}
 
       - name: Update translations from weblate
         uses: ./.github/actions/weblate-pull-translations
```
ppisar, line 24: This will replace released translations with translations for code from main branch. I'm not sure it is a good idea. It will replace not only *.po files, but also *.pot files, meaning that messages removed from main branch would disappear their translations from stable branch. Unless we are going to maintain stable branches in dnf5-l10n repository too, I would disable this step for non-main branch.

ppisar:
> > I have no idea how the token works. E.g. what ${{vars.RSM_CI_APP_ID}} refers to.
> 
> I think I'll need to [create a GitHub App](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app#registering-a-github-app) under the rpm-software-management org and that will give us a client ID. Then we can add a private key to the app following https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps.
> 
> Then we need to add these as variables/secrets on the dnf5 GitHub repository.

I was wondering what vars.RSM_CI_APP_ID and secrets.RSM_CI_APP_PRIVATE_KEY are and those are variables which will be defined in the repository settings. They will contain the ID and key you will set up as described in the links you posted. Now I understand it.

ppisar (approved):

.github/workflows/backport.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions: {}
+
+jobs:
+  backport:
+    name: Backport pull request
+    if: github.repository_owner == 'rpm-software-management' && github.event.pull_request.merged == true && (github.event_name != 'labeled' || startsWith('backport', github.event.label.name))
```
ppisar, line 13: Doesn't <code>github.event_name != 'labeled' || startsWith('backport', github.event.label.name<code> disregard the label value when the pull request is closed? I.e. it executes korthout/backport-action@v3 for pull requests without a label.

PR #2145, opened 2025-03-20
expired-pgp-keys: Drop braced initialization to fix clang build

jan-kolarik:


m-blaha (approved):
Thank you!

ppisar:
Will a compiler correctly find operator+ from std::string class if the left operand is not std::string?

Issue #2142, opened 2025-03-18
expired-pgp-keys: Respect the installroot option


jan-kolarik:
The current implementation doesn't support the `--installroot` option.

From the description of the original PR:

> We should probably move some code to libdnf5/rpm/rpm_signature.cpp, or expose more code from there.

Follow-up to: https://github.com/rpm-software-management/dnf5/pull/2135

m-blaha:
PR with the fix: https://github.com/rpm-software-management/dnf5/pull/2147

ppisar:
The rescheduled test finished. It failed for a different reason. Now OpenPGP vs PGP string is fine. But expected UserID does not match. The the tests expects there a repository identifier (?), but a user identity from the key is printed.

It's hard to verify how the key was generated from the CI logs.

ppisar:
Actually dnf-behave-tests/fixtures/gpgkeys/sign.sh indeed places a keyspecs file name into to the key ID. The actually printed ID "rpmsoftwaremanagement_test-utils (None) <rpmsoftwaremanagement#test-utils@copr.fedorahosted.org>" indeed denotes a key from COPR build system. So you were right, your fix should fix this latest failure.

PR #2147, opened 2025-03-20
expired-pgp-keys: Respect install root

m-blaha:
Resolves: https://github.com/rpm-software-management/dnf5/issues/2142

### DIFF
--- a/libdnf5-plugins/expired-pgp-keys/expired-pgp-keys.cpp
+++ b/libdnf5-plugins/expired-pgp-keys/expired-pgp-keys.cpp
@@ -169,8 +169,13 @@ void ExpiredPgpKeys::process_expired_pgp_keys(const libdnf5::base::Transaction &
     }
 
     // Iterate over all installed OpenPGP keys.
-    // TODO: Respect install root.
     auto ts = rpmtsCreate();
+    auto root_dir = config.get_installroot_option().get_value();
+    if (rpmtsSetRootDir(ts, root_dir.c_str()) != 0) {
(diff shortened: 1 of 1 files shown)


libdnf5-plugins/expired-pgp-keys/expired-pgp-keys.cpp
```diff
@@ -169,8 +169,13 @@ void ExpiredPgpKeys::process_expired_pgp_keys(const libdnf5::base::Transaction &
     }
 
     // Iterate over all installed OpenPGP keys.
-    // TODO: Respect install root.
     auto ts = rpmtsCreate();
+    auto root_dir = config.get_installroot_option().get_value();
+    if (rpmtsSetRootDir(ts, root_dir.c_str()) != 0) {
+        logger.error("Expired PGP Keys Plugin: Failed to set rpm transaction root dir \"{}\".", root_dir);
+        rpmtsFree(ts);
+        return;
```
ppisar, line 177: Shouldn't a plugin raise an exception on an error?

m-blaha: I was thinking about it, but since the plugin is not mandatory and the DNF5 command in which it was executed can still succeed, I prefer to log the error rather than abort the entire process.

ppisar: Ok. It's a pitty user cannot see the failure, but that's what the current callbacks offer.

ppisar (approved):

PR #2148, opened 2025-03-20
doc: Environment variables for a terminal and temporary files

ppisar:
The list of variables for temporary files is copied from std::filesystem::temp_directory_path() as it the fuction which uses them.

I intentionally left out FORCE_CULUMNS. We should rename it to DNF5_FORCE_COLUMNS before documenting it and make a transaction table to respect it.

Resolve <https://bugzilla.redhat.com/show_bug.cgi?id=2353349>

### DIFF
--- a/doc/dnf5.8.rst
+++ b/doc/dnf5.8.rst
@@ -176,6 +176,8 @@ Options
 
 Following options are applicable in the general context for any ``dnf5`` command:
 
+.. _assumeno_option_ref-label:
+
 ``--assumeno``
     | Automatically answer no for all questions.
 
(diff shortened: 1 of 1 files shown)


evan-goode (approved):
Looks good. The only remaining undocumented environment variable I see is `DNF_SYSTEM_UPGRADE_NO_REBOOT`, which skips rebooting or powering off when running `dnf5 offline reboot` or `dnf5 system-upgrade reboot`, and can be useful for debugging those commands, but ordinary users should have no use for it.

The name is carried over from DNF 4, but maybe we should rename it to `DNF5_SYSTEM_UPGRADE_NO_REBOOT` to match the others before documenting it.

ppisar:
DNF_SYSTEM_UPGRADE_NO_REBOOT is documented in dnf-plugins-core's system-upgrade manual. I would keep that variable name in DNF5 for compatibility (it has the same function and users don't need to change their scripts). and documented it there. We can add DNF5_SYSTEM_UPGRADE_NO_REBOOT later if somebody objects that the name is inconsistent. Do you agree?

ppisar:
A pull request with the documentation for DNF_SYSTEM_UPGRADE_NO_REBOOT is in #2151.

Issue #2026, opened 2025-01-28
'history undo/redo/rollback' does not print superuser privileges error like 'install' does


oturpe:
In I try to run `dnf install` as normal user, I get:
```
$ dnf upgrade
The requested operation requires superuser privileges. Please log in as a user with elevated rights, or use the "--assumeno" or "--downloadonly" options to run the command without modifying the system state.
```

That is fine. The same error is also printed for `upgrade` and `remove` subcommands. However, sub-subcommands of `history` similarly require superuser privileges, but do not have the same error. For instance:
```
$ dnf history undo 45
Updating and loading repositories:
Repositories loaded.
Package
Removing:
 ...
Removing unused dependencies:
 ...

Transaction Summary:
 Removing:          17 packages

After this operation, 36 MiB will be freed (install 0 B, remove 36 MiB).
Is this ok [y/N]: y
Running transaction
Failed to open lock file "/run/dnf/rpmtransaction.lock": (13) - Permission denied
```

Please add the warning, so it is more clear what went wrong.

alimirjamali:
PR submitted

PR #2149, opened 2025-03-20
Ask for superuser privileges on `history <undo|redo|rollback>`

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/2026

### DIFF
--- a/dnf5/commands/history/history_redo.cpp
+++ b/dnf5/commands/history/history_redo.cpp
@@ -34,6 +34,7 @@ void HistoryRedoCommand::set_argument_parser() {
     auto & ctx = get_context();
     transaction_specs->get_arg()->set_complete_hook_func(create_history_id_autocomplete(ctx));
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
+    create_store_option(*this);
 }
 
 void HistoryRedoCommand::configure() {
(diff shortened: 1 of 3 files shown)


alimirjamali:
Review notes:

1. This PR contains an experimental fix for #2145. If it is successful,  I will remove it from this PR, so the PR to #2145 could be amended with this approach (which might be a little be better).
2. The original issue is marked as "good first issue". Technically I am a third time contributor. But I wanted to see all of the CI/CD tests to be green for main. If any of them are red, it would confuse "real" first time contributors. As it confused me a little bit.

alimirjamali:
> If you are still developing the fix, please mark this pull request as a draft so that people don't spent time time on reading ever changing code.

The last push was 7 hours ago and I am not sure why one of the CI/CD tasks has been running since that time?

> I doubt that adding a --store option is the correct way of enabling a privilege escalation. However, I don't know DNF5 code much. Also if you want to use that dirty trick with --store, you should explain it in the commit message.

Using the `--store` trick was suggested by one of the comments in the original issue ([here](https://github.com/rpm-software-management/dnf5/issues/2026#issuecomment-2647339979)). Otherwise the other option is to hard code it in `cmd_requires_privileges()` function. Both options are possible.

kontura:
Since we already use this approach for a bunch of other commands and the history commands should have the `--store` option anyway I am merging this. Thank you!

If we want to make the privilege requirement more explicit/cleaner in the future we can do that for all commands.

PR #2152, opened 2025-03-21
Various fixes for the dnf5.spec file

m-blaha:
- fix building without man pages
- fix building without dnf5 plugins
- fix building without any libdnf5 plugin

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -979,6 +979,7 @@ popd
 %endif
 
 %find_lang dnf5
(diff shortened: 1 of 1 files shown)


ppisar (approved):
Looks good. I tried building the package with defaults build conditions and that passed without any changes.

When building without libdnf_cli, linker failed to link dnf5 executable: "/usr/bin/ld: cannot find -llibdnf5-cli: No such file or directory". But that's not an issue in spec file.

Issue #1789, opened 2024-10-21
`installonly_limit=1` is allowed contrary to documentation


pkratoch:
The documentation of the `isntallonly_limit` option states that value "`1` is explicitly not allowed since it complicates kernel upgrades due to protection of the running kernel from removal", but when the option is set to `1`, dnf proceeds with the operation.

```
$ sudo dnf5 --setopt=installonly_limit=1 install nano
Updating and loading repositories:
Repositories loaded.
Package "nano-7.2-5.fc39.x86_64" is already installed.

Package                                       Arch        Version                                        Repository                    Size
Removing:
 kernel                                       x86_64      6.10.10-100.fc39                               <unknown>                  0.0   B
 kernel                                       x86_64      6.10.12-100.fc39                               <unknown>                  0.0   B
 kernel-core                                  x86_64      6.10.10-100.fc39                               <unknown>                 68.7 MiB
 kernel-core                                  x86_64      6.10.12-100.fc39                               <unknown>                 68.7 MiB
 kernel-modules                               x86_64      6.10.10-100.fc39                               <unknown>                 61.0 MiB
 kernel-modules                               x86_64      6.10.12-100.fc39                               <unknown>                 61.0 MiB
 kernel-modules-core                          x86_64      6.10.10-100.fc39                               <unknown>                 36.1 MiB
 kernel-modules-core                          x86_64      6.10.12-100.fc39                               <unknown>                 36.1 MiB
 kernel-modules-extra                         x86_64      6.10.10-100.fc39                               <unknown>                  2.6 MiB
 kernel-modules-extra                         x86_64      6.10.12-100.fc39                               <unknown>                  2.6 MiB

Transaction Summary:
 Removing:          10 packages
```

There are also (currently disabled) tests for it in https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/installonly.feature

alimirjamali:
PR submitted

- I used the exact error messages as dnf4
- It currently allows overriding the invalid value of `installonly_limit` within `/etc/dnf/dnf.conf` with a command line option. Should it? Or should it throw `InvalidConfigError` exception while parsing the config file?

PR #2153, opened 2025-03-21
Do not allow `installonly_limit=1`

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/1789

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -246,6 +246,12 @@ void RootCommand::set_argument_parser() {
                 throw libdnf5::cli::ArgumentParserError(
                     M_("{}: Badly formatted argument value \"{}\""), std::string{"setopt"}, std::string(value));
             }
+
+            if (std::string(value) == std::string("installonly_limit=1"))
+                throw libdnf5::cli::ArgumentParserInvalidValueError(
+                    M_("Config error: Error parsing --setopt with key 'installonly_limit', "
+                       "value '1': value 1 is not allowed"));
+
             auto key = std::string(value, val);
             auto dot_pos = key.rfind('.');
             if (dot_pos != std::string::npos) {
@@ -1403,6 +1409,10 @@ int main(int argc, char * argv[]) try {
             }
 
             base.setup();
+
+            if (base.get_config().get_installonly_limit_option().get_value() == 1)
+                throw libdnf5::InvalidConfigError(M_(
+                    "Invalid configuration value: installonly_limit=1 in /etc/dnf/dnf.conf; value 1 is not allowed"));
 
             auto & coloring = base.get_config().get_color_option().get_value();
             if (coloring == "always") {


dnf5/main.cpp
```diff
@@ -246,6 +246,12 @@ void RootCommand::set_argument_parser() {
                 throw libdnf5::cli::ArgumentParserError(
                     M_("{}: Badly formatted argument value \"{}\""), std::string{"setopt"}, std::string(value));
             }
+
+            if (std::string(value) == std::string("installonly_limit=1"))
```
dcantrell, line 250: Is this the best way to check for the installonly_limit setting in the config file?  Ideally we have the config file data in a better form so we're not having to compare against strings.  What if the string is "installonly_limit = 1"?

dnf5/main.cpp
```diff
@@ -246,6 +246,12 @@ void RootCommand::set_argument_parser() {
...
+                throw libdnf5::cli::ArgumentParserInvalidValueError(
+                    M_("Config error: Error parsing --setopt with key 'installonly_limit', "
```
dcantrell, line 252: I recommend saying "Configuration" rather than "Config.  And is this actually a parse error?  It parsed, it's just a prohibited value.

dnf5/main.cpp
```diff
@@ -1404,6 +1410,10 @@ int main(int argc, char * argv[]) try {
 
             base.setup();
 
+            if (base.get_config().get_installonly_limit_option().get_value() == 1)
+                throw libdnf5::InvalidConfigError(M_(
+                    "Invalid configuration value: installonly_limit=1 in /etc/dnf/dnf.conf; value 1 is not allowed"));
```
dcantrell, line 1415: Would it be helpful here to state what values are allowed?

dcantrell (changes requested):

dnf5/main.cpp
m-blaha: There is one more issue with this approach - the check for the forbidden value will be executed only for `dnf5` command line tool. But the value `1` will still remain allowed for API users.
I think more appropriate place to fix this issue would be in the `from_string_func` parameter of the `installonly_limit` config option constructor (We can throw `OptionInvalidValueError` with proper description here): https://github.com/rpm-software-management/dnf5/blob/f9179343796d754d99eb62a3d2ad39157026a5bb/libdnf5/conf/config_main.cpp#L144-L153
If fixed here, I believe both ways how to set the `installonly_limit` value will be covered - i.e. `/etc/dnf/dnf.conf` and `--setopt=installonly_limit=X`. It will also sanitize the value for API users.

alimirjamali: > If fixed here, I believe both ways how to set the `installonly_limit` value will be covered - i.e. `/etc/dnf/dnf.conf` and `--setopt=installonly_limit=X`. It will also sanitize the value for API users.

Thanks for the guidance. I will readjust the code, test it and amend the PR as suggested.

alimirjamali: > I think more appropriate place to fix this issue would be in the `from_string_func` parameter of the `installonly_limit` config option constructor (We can throw `OptionInvalidValueError` with proper description here):

There is a problem with this approach. It works well if the `installonly_limit=1` is provided via `--setopt` CLI option. But does not work if it is within `/etc/dnf/dnf.conf`. Throwing `OptionInvalidValueError` error is ignored and it defaults to 3. I am investigating why it behaves like this. Here is the diff:

<details>

<summary>libdnf5/conf/config_main.cpp</summary>

```diff
diff --git a/libdnf5/conf/config_main.cpp b/libdnf5/conf/config_main.cpp
index a8047911..84a05144 100644
--- a/libdnf5/conf/config_main.cpp 
+++ b/libdnf5/conf/config_main.cpp
@@ -141,16 +141,20 @@ class ConfigMain::Impl {
     OptionStringList group_package_types{GROUP_PACKAGE_TYPES};
     OptionStringAppendSet optional_metadata_types{
         OptionStringAppendSet::ValueType{libdnf5::METADATA_TYPE_COMPS, libdnf5::METADATA_TYPE_UPDATEINFO}};
-    OptionNumber<std::uint32_t> installonly_limit{3, 0, [](const std::string & value) -> std::uint32_t {
-                                                      if (value == "<off>") {
-                                                          return 0;
-                                                      }
-                                                      try {
-                                                          return static_cast<std::uint32_t>(std::stoul(value));
-                                                      } catch (...) {
-                                                          return 0;
-                                                      }
-                                                  }};
+    OptionNumber<std::uint32_t> installonly_limit{
+        3, 0, [](const std::string & value) -> std::uint32_t {
+            if (value == "<off>") {
+                return 0;
+            }
+            if (value == std::string("1")) {
+                throw OptionInvalidValueError(M_("Invalid configuration value; value 1 is not allowed"));
+            }
+            try {
+                return static_cast<std::uint32_t>(std::stoul(value));
+            } catch (...) {
+                return 0;
+            }
+        }};
 
     OptionStringAppendList tsflags{std::vector<std::string>{}};
     OptionBool assumeyes{false};
```

</details>

m-blaha: Hm, I see. In the `Config::load_from_parser()` function, the `OptionError` exceptions are logged to dnf5.log and then the invalid line is ignored and the default value used instead:

[dnf5/main.cpp#L1387](https://github.com/rpm-software-management/dnf5/blob/54fd7dff3aadfd3edcefafd1f33d889eab7a00ec/dnf5/main.cpp#L1387) -> [libdnf5/base/base.cpp#L121](https://github.com/rpm-software-management/dnf5/blob/54fd7dff3aadfd3edcefafd1f33d889eab7a00ec/libdnf5/base/base.cpp#L121) -> libdnf5/conf/config.cpp#L39-L58

https://github.com/rpm-software-management/dnf5/blob/54fd7dff3aadfd3edcefafd1f33d889eab7a00ec/libdnf5/conf/config.cpp#L39-L58

m-blaha: Since this affects all config options, we should maybe file a new issue, something like: "Errors while reading the dnf.conf file are not printed to the user and are only logged in dnf5.log". @dcantrell what do you think?

Issue #1003, opened 2023-11-08
Unfriendly message when only proxy_username is used


j-mracek:
I tried incorrect option combination in dnf.conf => `proxy_username` without proxy_password. It correctly fails but the report is not user friendly in comparison to DNF4. DNF4 also provides more problems not the only the first one.

```
$ sudo dnf install acpi --refresh
Error: repo 'copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly': 'proxy_username' is set but not 'proxy_password'
Error: repo 'copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable': 'proxy_username' is set but not 'proxy_password'
Error: repo 'beaker-client': 'proxy_username' is set but not 'proxy_password'
Error: repo 'fedora': 'proxy_username' is set but not 'proxy_password'

$ sudo dnf5 install acpi --refresh
Updating and loading repositories:
Error checking if metalink "https://mirrors.fedoraproject.org/metalink?repo=fedora-38&arch=x86_64" is in sync for repository "fedora"
 Option value is not set
```

_Originally posted by @j-mracek in https://github.com/rpm-software-management/dnf5/issues/465#issuecomment-1801239922_

m-blaha:
Still valid.

alimirjamali:
PR Submitted.

Review notes:
- DNF4 error message and behavior is mimicked here. Technically we can set `proxy_password` to `""` if the `proxy_password` config is missing.
- DNF5 currently recognizes `proxy_username` and `proxy_password` only as CLI argument (with `--setopt`) or via `/etc/dnf/dnf.conf` file. Individual proxy username and password settings per repo is broken. This is addressed in [another issue](https://github.com/rpm-software-management/dnf5/issues/1144#issuecomment-1880819502).
- The PR throws `libdnf5::MissingConfigError`. Technically it is possible to throw `ArgumentParserMissingDependentArgumentError` if the proxy is provided via `--setopt`. But maybe keep this simple.

Issue #1144, opened 2024-01-06
org.rpm.dnf.v0.rpm.Repo list crash


anaselli:
repo_attrs is not well documented, anyway I got from the code the following available attirbutes:

```
"id",
"name",
"type",
"enabled",
"priority",
"cost",
"baseurl",
"metalink",
"mirrorlist",
"metadata_expire",
"cache_updated",
"excludepkgs",
"includepkgs",
"skip_if_unavailable",
"gpgkey",
"gpgcheck",
"repo_gpgcheck",
"proxy",
"proxy_username",
"proxy_password",
"repofile",
"revision",
"content_tags",
"distro_tags",
"updated",
"size",
"pkgs",
"available_pkgs",
"mirrors",
```
i tried all of them and **proxy_password** makes the list method crashing with the following BT:

```
Traceback (most recent call last):
  File "/home/angelo/src/manatools/dnfdragora/../test-dnf5daemon.py", line 225, in <module>
    test()
  File "/home/angelo/src/manatools/dnfdragora/../test-dnf5daemon.py", line 104, in test
    repo_list = iface_repo.list({"repo_attrs": [
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib64/python3.10/site-packages/dbus/connection.py", line 634, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.rpm.dnf.v0.Error: Option value is not set
```

ppisar:
Repository proxy options are not yet fully implemented in D-Bus interface. See repo_to_map():

~~~~
            case RepoAttribute::proxy_username:
                //                dbus_repo.emplace(attr, libdnf_repo->get_config().get_proxy_username_option().get_value());
                dbus_repo.emplace(attr, "user foo");
                break;
            case RepoAttribute::proxy_password:
                dbus_repo.emplace(attr, libdnf_repo->get_config().get_proxy_password_option().get_value());
                break;
~~~~

proxy_username just sends a dummy value and proxy_password attempts to retrieve the value, but get_value() throws an exception if a priority of the option is undefined:

~~~~
const std::string & OptionString::get_value() const {
    if (get_priority() == Priority::EMPTY) {
[...]

PR #2154, opened 2025-03-21
Improve missing `proxy_password` message

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/1003

### DIFF
--- a/libdnf5/repo/librepo.cpp
+++ b/libdnf5/repo/librepo.cpp
@@ -175,9 +175,12 @@ static void init_remote(LibrepoHandle & handle, const C & config) {
     handle.set_opt(LRO_PROXYAUTHMETHODS, proxy_auth_methods);
 
     if (!config.get_proxy_username_option().empty()) {
-        auto userpwd = config.get_proxy_username_option().get_value();
-        if (!userpwd.empty()) {
-            userpwd = format_user_pass_string(userpwd, config.get_proxy_password_option().get_value());
+        auto proxy_username = config.get_proxy_username_option().get_value();
+        if (!proxy_username.empty()) {
+            if (config.get_proxy_password_option().get_priority() == libdnf5::Option::Priority::EMPTY)
+                throw libdnf5::MissingConfigError(M_("'proxy_username' is set but not 'proxy_password'"));
+            auto proxy_password = config.get_proxy_password_option().get_value();
+            auto userpwd = format_user_pass_string(proxy_username, proxy_password);
             handle.set_opt(LRO_PROXYUSERPWD, userpwd.c_str());
         }
     }


libdnf5/repo/librepo.cpp
```diff
@@ -175,9 +175,12 @@ static void init_remote(LibrepoHandle & handle, const C & config) {
     handle.set_opt(LRO_PROXYAUTHMETHODS, proxy_auth_methods);
 
     if (!config.get_proxy_username_option().empty()) {
-        auto userpwd = config.get_proxy_username_option().get_value();
-        if (!userpwd.empty()) {
-            userpwd = format_user_pass_string(userpwd, config.get_proxy_password_option().get_value());
+        auto proxy_username = config.get_proxy_username_option().get_value();
+        if (!proxy_username.empty()) {
+            if (config.get_proxy_password_option().get_priority() == libdnf5::Option::Priority::EMPTY)
+                throw libdnf5::MissingConfigError(M_("'proxy_username' is set but not 'proxy_password'"));
```
m-blaha, line 181: Including the repository ID in the message would be really helpful for users. Without it, locating the correct config file for adjustments could be quite challenging. What do you think?

I’m also not sure if this is the best place for these checks. I understand that this PR is simply enhancing an existing check, so we could track this in a new issue. There’s also a similar check for the `minrate` and `throttle` options that should likely be moved as well. But I'm not blocking this PR on it.

alimirjamali: > Including the repository ID in the message would be really helpful for users. Without it, locating the correct config file for adjustments could be quite challenging. What do you think?

Then 3 scenarios of missing `proxy_password` has to be covered:
1. Within individual repository config files
2. within the main `/etc/dnf/dnf.conf` config file
3. If `proxy_username` is provided via `--setopt=proxy_username=user` but the `proxy_password` was missing.

We have `LibrepoHandle::init_remote(const libdnf5::repo::ConfigRepo & config)` overloaded function in the file I modified which we can use for the 1s scenario. But I guess the two other scenarios should be covered in `dnf5/main.cpp` file?

m-blaha: I believe the `LibrepoHandle::init_remote()` handles all three scenarios. This is how repo config works (see repo/config_repo.cpp):

0. there is a hard coded default `nullptr` for proxy_username in ConfigMain
1. read `proxy_username` value from the main config
2. in the constructor of ConfigRepo set the default value for proxy_username from the main config value
3. read repository specific overrides from /etc/yum.repos.d/*.repo files
4. override the value (either main or repository specific) from --setopt

So any time you use in your code `repoconfig.get_proxy_username_option().get_value()`, you do not read a value stored in particular config file, but the value calculated using these steps.

So I think we can keep the check where it is. I'd like just to enhance the error message.

alimirjamali: > So I think we can keep the check where it is. I'd like just to enhance the error message.

There are few problems with this approach. Please let me explain. I have this so far:

<details>

<summary>Patch diff</summary>

```diff
diff --git a/libdnf5/repo/librepo.cpp b/libdnf5/repo/librepo.cpp
index 060e20fa..bafc16a6 100644
--- a/libdnf5/repo/librepo.cpp
+++ b/libdnf5/repo/librepo.cpp
@@ -175,9 +175,20 @@ static void init_remote(LibrepoHandle & handle, const C & config) {
     handle.set_opt(LRO_PROXYAUTHMETHODS, proxy_auth_methods);
 
     if (!config.get_proxy_username_option().empty()) {
-        auto userpwd = config.get_proxy_username_option().get_value();
-        if (!userpwd.empty()) {
-            userpwd = format_user_pass_string(userpwd, config.get_proxy_password_option().get_value());
+        auto proxy_username = config.get_proxy_username_option().get_value();
+        if (!proxy_username.empty()) {
+            if (config.get_proxy_password_option().get_priority() == libdnf5::Option::Priority::EMPTY) {
+                if (auto * config_repo = dynamic_cast<const libdnf5::repo::ConfigRepo *>(&config)) {
+                    throw libdnf5::MissingConfigError(
+                        M_("'proxy_username' is set but not 'proxy_password' for repo \"{}\""),
+                        config_repo->get_name_option().get_value());
+                } else {
+                    throw libdnf5::MissingConfigError(
+                        M_("'proxy_username' is set but not 'proxy_password' for main config"));
+                }
+            }
+            auto proxy_password = config.get_proxy_password_option().get_value();
+            auto userpwd = format_user_pass_string(proxy_username, proxy_password);
             handle.set_opt(LRO_PROXYUSERPWD, userpwd.c_str());
         }
     }
```
</details>

Problems:
1. For any repo, since the final calculated value of `proxy_password` is evaluated, while the repo name will be printed with the above patch, it is not possible to inform user where the actual error happened (CLI/dnf.conf/repo.conf).
2. The major issue is that currently dnf5 totally ignores proxy config set in individual repo files. As I mentioned in the original issue review notes: https://github.com/rpm-software-management/dnf5/issues/1144#issuecomment-1880819502

m-blaha (approved):

Issue #2007, opened 2025-01-14
Assertion error in pool.hpp


kilinitt:
Hello,

I get the following assertion failure when:

> [root@a4030d980752 /]# dnf install ./rsync-3.3.0-3.fc41.aarch64.rpm
> Updating and loading repositories:
> Repositories loaded.
> terminate called after throwing an instance of 'libdnf5::AssertionError'
>   what():  libdnf5/./solv/pool.hpp:246: void libdnf5::solv::Pool::swap_considered_map(libdnf5::solv::SolvMap&): Assertion 'other_considered_map.allocated_size() == 0 || other_considered_map.allocated_size() >= get_nsolvables()' failed: The considered map is smaller than the number of solvables in the pool
> Aborted (core dumped)

It happens after a specific and somewhat set of packages installed and uninstaleld as the image is built.

Once in this state, it happens with any .rpm file that i've tried installing via dnf. What information can I provide for you to be able to diagnose the problem?

libdnf5 version is: libdnf5-5.2.8.1-2.fc41

m-blaha:
Hi,
It seems you’re able to reproduce the issue. If possible, could you share step-by-step instructions on how to reach this unfortunate state? That would be really helpful. Also getting a backtrace from the core dump might shed some light on the problem.

kilinitt:
Hi @m-blaha, yes I have a reproducible setup, but it's actually quite difficult to detail the steps that get us to this point. I saw the problem on an intel machine as I was building an image a few days ago, and it went away as I worked on the image, so I brushed it off to docker build issues. Then it surfaced on a different image on my ARM machine, which is the reproducible case that I have. Happy to share that image with you if it can be done privately. Please let me know. I will also try to get a trace from the core dump for you later today.

kilinitt:
I can certainly do that. Apologies it's been a hectic week. Plan to get it to you by Monday @m-blaha. Thank you.

m-blaha:
We still don’t have a reproducer, so fixing the issue will be challenging.

Let me summarize what we have so far:

Here’s the traceback provided by the reporter:
```
#0  0x000071c235063114 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x000071c235009f9e in raise () from /lib64/libc.so.6
#2  0x000071c234ff1942 in abort () from /lib64/libc.so.6
[...]

PR #2159, opened 2025-03-25
rpm: Ensure make_provides_ready always uses an up-to-date considered map

m-blaha:
The `considered_map` may not be up to date when `make_provides_ready()` is called. This patch resolves the following scenario (see related issue

- If modules are enabled and available, one of the last steps in `load_repos()` is applying module_filtering, which also recomputes the considered map.

- During the goal resolution, command-line packages are added, increasing the number of solvables in the pool. This invalidates the `considered_map`, but it is not recomputed.

- After adding command-line packages, excludes are reloaded. As part of setting versionlock excludes, `make_provides_ready()` is called (via the `PackageQuery` constructor).

- At this point, the considered map is invalid and smaller than the current number of solvables, leading to an assertion failure in `swap_considered_map`.

Debugging this issue was tricky because `considered_map` is a bitmap, where each solvable takes only one bit. Since the allocated size is rounded up to bytes, there is usually some extra space for a few additional command-line packages without requiring a resize. However, if the number of solvables before adding a package was a multiple of eight, adding another package triggers the assertion failure:

```
terminate called after throwing an instance of 'libdnf5::AssertionError'
  what():  libdnf5/./solv/pool.hpp:246: void libdnf5::solv::Pool::swap_considered_map(libdnf5::solv::SolvMap&): Assertion 'other_considered_map.allocated_size() == 0 || other_considered_map.allocated_size() >= get_nsolvables()' failed: The considered map is smaller than the number of solvables in the pool
Aborted (core dumped)
```

With this patch, `make_provides_ready()` always works with an up-to-date considered_map, ensuring correctness.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2007

### DIFF
--- a/libdnf5/module/module_sack.cpp
+++ b/libdnf5/module/module_sack.cpp
@@ -350,6 +350,9 @@ void ModuleSack::Impl::make_provides_ready() {
         return;
     }
 
+    // before swapping the considered map make sure it's uptodate
+    recompute_considered_in_pool();
+
     // Temporarily replaces the considered map with an empty one. Ignores "excludes" during calculation provides.
     Map * considered = pool->considered;
     pool->considered = nullptr;
--- a/libdnf5/rpm/package_sack.cpp
+++ b/libdnf5/rpm/package_sack.cpp
@@ -54,6 +54,9 @@ void PackageSack::Impl::make_provides_ready() {
         return;
     }
 
+    // before swapping the considered map make sure it's uptodate
+    recompute_considered_in_pool();
+
     // Temporarily replaces the considered map with an empty one. Ignores "excludes" during calculation provides.
     libdnf5::solv::SolvMap original_considered_map(0);
     get_rpm_pool(base).swap_considered_map(original_considered_map);


evan-goode (commented):
>However, if the number of solvables before adding a package was a multiple of eight, adding another package triggers the assertion failure:

Wow, nasty bug. Thanks for the investigation and detailed analysis.

Should we remove this line then? https://github.com/rpm-software-management/dnf5/blob/f9179343796d754d99eb62a3d2ad39157026a5bb/libdnf5/base/goal.cpp#L3132

I wonder if this change should also be made in `ModuleSack::Impl::make_provides_ready`? That function looks quite similar.

If a similar bug comes up in the future, then I would suggest refactoring to replace reads from `{module_,}sack->p_impl->pool->considered` with a getter that calls `recompute_considered_in_pool`. `recompute_considered_in_pool` is already lazy thanks to `considered_uptodate`.

m-blaha:
I've added similar fix also for module_sack, just to stay on safe side and use correct considered. Module sack does not use the `swap_considered_map()` call, thus it would not trigger the assert. But it's probably safer to have it fixed in case we would refactor the module sack code in the future.

evan-goode:
> I'm not sure, TBH. In goal.cpp the `recompute_considered_in_pool()` is called always, unconditionally, while the `make_provides_ready()` returns fast in case provides re-calculation is not needed, without calling `recompute_considered_in_pool()`.

OK right, agreed. LGTM.

m-blaha:
Still thinking whether the assert is actually useful.
In this particular case what the code does is swap **already invalid** considered with nullptr and then swap the nullptr back with original, invalid, considered. There is no need to have valid considered map for make_provides_ready to work correctly.

@evan-goode  What do you think about dropping the assert completely instead of the unneeded considered recalculation?

evan-goode:
That would be OK with me, given that `swap_considered_map` is only used here and is not part of the public libdnf5 API. You're right, there is no guarantee that `pool->considered` is always valid so we don't need to enforce it here.

PR #2164, opened 2025-03-27
comps: Respect the display_order with listing groups and environments

pkratoch:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1655

### DIFF
--- a/libdnf5/comps/environment/environment.cpp
+++ b/libdnf5/comps/environment/environment.cpp
@@ -35,6 +35,7 @@ extern "C" {
 }
 
 #include <libxml/tree.h>
+#include <limits.h>
 
 #include <set>
 #include <string>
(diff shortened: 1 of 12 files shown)


libdnf5/comps/environment/environment.cpp
```diff
@@ -158,6 +159,17 @@ std::string Environment::get_order() const {
 }
 
 
+int Environment::get_order_int() const {
+    try {
+        return std::stoi(get_order());
+    } catch (const std::invalid_argument &) {
+        return INT_MAX;
+    } catch (const std::out_of_range &) {
+        return INT_MAX;
```
kontura, line 168: I am thinking if it would be worth to log these conversion exceptions?

kontura: Ok, that makes sense.
Thanks for looking into it.

Issue #307, opened 2023-02-20
Enable parallel downloaing of repositories


kontura:
Librepo has an API for parallel downloading of repositories, added here: https://github.com/rpm-software-management/librepo/pull/95.

I believe this API is untested but it could be used by libdnf to achieve parallel downloading of repositories.

kontura:
> Why hasn't this been worked on? I have too many repositories, so it takes like 1-2 mins updating everyone of them 😢

We have actually just started looking into it.

Sergio00166:
making the repo data download in parallel will make dnf way faster bc is the slowest part, the rest is quite fast.

RoyaltyInTraining:
This issue is without a doubt my worst source of frustration with DNF. Most repositories connect **painfully slow**, and this gets worse with every single repo that is added to the system. I can't confirm this, but I suspect this issue also makes graphical package managers like Gnome Software unnecessarily sluggish.

Sergio00166:
Comparing others like pacman or apt are way faster than dnf bc of the slow (and secuencial) repo data download. That made me to think to move from fedora to ubuntu just for how slow dnf is.

kontura:
I have created first drafts for this functionality.
It is available in this COPR repository: https://copr.fedorainfracloud.org/coprs/amatej/dnf5-parallel/
Both new librepo and libdnf5 are needed.

I would be grateful for any testing but please keep in mind it is currently experimental.

Also it still sequentially checks if the repos are up to date but once/if we start the repository metadata download it should be done in parallel.

kontura:
Additional dnf5 [PR](https://github.com/rpm-software-management/dnf5/pull/2269) and librepo [PR](https://github.com/rpm-software-management/librepo/pull/338) were merged.

The final download ordering is (when there are no locally cached metadata):
1. Download all configured metalinks and mirrorlists in parallel
2. Download all configured repomds in parallel
3. Download the rest of configured metadata in parallel (primary, comps...) 

This is not perfect, all downloads from one step have to finish before the next step can start however changing this would require extensive refactoring with I believe only a small speed up.

Therefore I consider this issue resolved.
(There is one more outstanding [issue](https://github.com/rpm-software-management/dnf5/issues/2247) regarding some hopefully rare error messages.)

PR #2167, opened 2025-03-28
Enable parallel downloading of main repository metadata 

kontura:
Requires: https://github.com/rpm-software-management/librepo/pull/336

For https://github.com/rpm-software-management/dnf5/issues/307

I have also created a testing COPR: https://copr.fedorainfracloud.org/coprs/amatej/dnf5-parallel/

I am afraid this will be hard to review because the commits are quite interconnected and they don't really work separately.
I was considering making them more standalone but then the PR would contain code that is added just to be removed.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -86,6 +86,12 @@ jobs:
     fmf_ref: &CI_DNF_STACK_REF main
     tmt_plan: "^/plans/integration/behave-dnf5$"
     packages: [dnf5]
+    tf_extra_params:
+      environments:
+        - artifacts:
+            - type: repository-file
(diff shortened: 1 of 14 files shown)


kontura:
Since some of the outputs for errors are different here is a CI update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1656.
It also adds couple new tests.

evan-goode (commented):
I did not look at the code or do extensive tests, but this is awesome. Loading repositories takes 1/2 to 2/3 the time on my setup. For some users with many repositories, the performance improvement might be even greater than the change to not download filelists. Thank you!

m-blaha (approved):
Thank you!

PR #2169, opened 2025-03-31
Add missing rpm scriptlet types: sysusers and pre/postuntrans

pmatilai:
These scriptlet types were added in rpm 4.19 and lack of them causes "Running unknown scriptlet" messages, which some find a bit unnerving. With this patch, the output becomes something like

   >>> Running sysusers scriptlet: dbus-common-1:1.14.10-4.fc41.noarch
   >>> Finished sysusers scriptlet: dbus-common-1:1.14.10-4.fc41.noarch


keszybz (approved):
LGTM.

Conan-Kudo (changes requested):
This is effectively an API break. Are we okay with this?

decathorpe:
> This is effectively an API break. Are we okay with this?

Would it still be an API break if the enum variants were added at the end?
(That is assuming C++ enum classes work the same way as C enums, and I know nothing about C++ :))

Conan-Kudo:
They do work the same way, but if we don't care that this is a breakage, we can land it.

Conan-Kudo:
(Notably, if this isn't actually part of the public library headers, then it's fine to land.)

keszybz:
Hmm, so those headers are distributes in `libdnf5-devel`. So it looks like the new entries may need to be moved to the end of the enum.

pmatilai:
Oh. I pondered the enum placement from a "logical" standpoint only, and negletted to consider the ABI implications (note ABI, not API). But indeed, these need to be added to the end. I'll push an update shortly.

pmatilai:
Uh, so pre-commit is wanting to add an extra space before the comment to *all* the items in the enum. Touching unrelated lines is a DONT in where I come from :smile: but I guess the expectation here is just do what clang-format says and don't argue?

Conan-Kudo:
> Uh, so pre-commit is wanting to add an extra space before the comment to _all_ the items in the enum. Touching unrelated lines is a DONT in where I come from 😄 but I guess the expectation here is just do what clang-format says and don't argue?

Do it in a separate commit and it's fine.

jan-kolarik:
> Uh, so pre-commit is wanting to add an extra space before the comment to _all_ the items in the enum. Touching unrelated lines is a DONT in where I come from 😄 but I guess the expectation here is just do what clang-format says and don't argue?

It's probably due to the additional comma at `TRIGGER_POST_UNINSTALL,`.

pmatilai:
> It's probably due to the additional comma at TRIGGER_POST_UNINSTALL,.

Yup - for reasons like this I prefer having the trailing comma always there so you don't need to touch unrelated lines when adding more items :sweat_smile: 

What should I do with this? Extra commit with just the whitespace changes?

jan-kolarik:
> > It's probably due to the additional comma at TRIGGER_POST_UNINSTALL,.
> 
> Yup - for reasons like this I prefer having the trailing comma always there so you don't need to touch unrelated lines when adding more items 😅
> 
> What should I do with this? Extra commit with just the whitespace changes?

I would edit the original commit and force push it here.

pmatilai:
> I would edit the original commit and force push it here.

There you go. All those unrelated lines changed makes me hurt a little but if that's the standard then so be it :smile:

jan-kolarik:
> (To be honest, I think we should allow trailing commas, because then it makes the diff less bad)

They're allowed, see https://github.com/rpm-software-management/dnf5/blob/5.2.12.0/include/libdnf5/base/goal_elements.hpp#L91.

jan-kolarik (approved):
Thank you Panu for handling this!

Issue #1881, opened 2024-11-20
Invalid hint for unknown short argument available in other commands


jrohel:
PR https://github.com/rpm-software-management/dnf5/pull/1858 introduced a hint for an unknown argument available in other commands.

However, there is a problem with a group of several short arguments. For example:
```
# dnf -yl upgrade
Unknown argument "-yl" for command "dnf5". Add "--help" for more information about the arguments. 
The argument is available for commands: . (It has to be placed after the command.)
```
Instead of being told that the second argument 'l' is unknown (or in this case, that it is available for repoquery), the user is told the information for the first argument 'y', which is available globally (an empty command is displayed).

alimirjamali:
PR Submitted. This minor issue is relatively easy and straightforward to solve.

PR #2171, opened 2025-04-01
Fix invalid hint for unknown short argument

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/1881

### DIFF
--- a/libdnf5-cli/argument_parser.cpp
+++ b/libdnf5-cli/argument_parser.cpp
@@ -1050,6 +1050,7 @@ void ArgumentParser::CommandOrdinary::parse(const char * option, int argc, const
     size_t used_positional_arguments = 0;
     int short_option_idx = 0;
     for (int i = 1; i < argc;) {
+        char unused_short_option = 0;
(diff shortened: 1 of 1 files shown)


libdnf5-cli/argument_parser.cpp
```diff
@@ -1048,6 +1048,7 @@ void ArgumentParser::CommandOrdinary::parse(const char * option, int argc, const
         }
     }
     size_t used_positional_arguments = 0;
+    char unused_short_option = 0;
```
m-blaha, line 1051: The `unused_short_option` variable is used only within the following `for` loop. I’d prefer to define it inside the loop so that its value isn’t carried across iterations and is clearly reset for each processed argument.

m-blaha (changes requested):
Thanks for working on it!

alimirjamali:
> Thanks for working on it!

Thanks for the review. I applied the requested changes.

m-blaha (approved):

Issue #1151, opened 2024-01-09
dnf5 download exits successfully with a bad argument


kontura:
```bash
$ dnf5 download non-sense
Updating and loading repositories:
Repositories loaded.
```
vs. old dnf4:
```bash
$ dnf download non-sense
Last metadata expiration check: 0:00:29 ago on Tue Jan  9 07:43:39 2024.
No package non-sense available.
Exiting due to strict setting.
Error: No package non-sense available.
```

I believe we agreed that dnf5 should also be strict by default. The current behavior should happen if `--skip-unavailable` is specified.

j-mracek:
I think that strict behavior is a correct default.

alimirjamali:
PR Submitted. The issue is relatively straightforward to solve.

Issue #817, opened 2023-08-14
`dnf5 clean` to show cleaned space size


abitrolly:
```
$ dnf clean all        
Removed 232 files, 10 directories. 0 errors occurred. 
```
It is more interesting to know how much free space had been recovered.

evan-goode:
Thanks for the suggestion! Yep, another nice feature to have, but low priority for now. APT and Zypper don't seem to do this, but `nix-collect-garbage -d` does, which I like.

alimirjamali:
PR Submitted. It will print the results similar to this:
```bash
$dnf clean all
Removed 26 files, 15 directories (total of 49 MiB). 0 errors occurred.
```

And for logging, it logs the raw values.

PR #2174, opened 2025-04-02
Disallow `dnf download non-sense` without `--skip-unavailable`

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/1151

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -158,6 +158,7 @@ void DownloadCommand::set_argument_parser() {
     cmd.register_named_arg(resolve);
     cmd.register_named_arg(alldeps);
     create_destdir_option(*this);
+    auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     cmd.register_named_arg(srpm);
     cmd.register_named_arg(url);
     cmd.register_named_arg(urlprotocol);
@@ -210,6 +211,14 @@ void DownloadCommand::run() {
         pkg_query.filter_latest_evr();
         if (!arch_option.empty()) {
             pkg_query.filter_arch(std::vector<std::string>(arch_option.begin(), arch_option.end()));
+        }
+
+        if (!pkg_query.size() && !ctx.get_base().get_config().get_skip_unavailable_option().get_value()) {
+            // User tried `dnf5 download non-sense` or `dnf download non-sense-wildcard*`
+            throw libdnf5::cli::CommandExitError(
+                1,
+                M_("No package \"{}\" available; You might want to use --skip-unavailable option."),
+                option->get_value());
         }
 
         for (const auto & pkg : pkg_query) {
--- a/doc/commands/download.8.rst
+++ b/doc/commands/download.8.rst
@@ -50,6 +50,9 @@ Options
 ``--destdir=<path>``
     | Set directory used for downloading packages to. Default location is to the current working directory.
 
+``--skip-unavailable``
+    | Allow skipping packages that are not available in repositories. All available packages will be downloaded.
+
 ``--srpm``
     | Download the source rpm. Enables source repositories of all enabled binary repositories.
 


dnf5/commands/download/download.cpp
```diff
@@ -212,6 +213,12 @@ void DownloadCommand::run() {
             pkg_query.filter_arch(std::vector<std::string>(arch_option.begin(), arch_option.end()));
         }
 
+        if (!pkg_query.size() && !ctx.get_base().get_config().get_skip_unavailable_option().get_value()) {
+            // User tried `dnf5 download non-sense` or `dnf download non-sense-wildcard*`
+            throw libdnf5::cli::ArgumentParserInvalidValueError(
+                M_("No package \"{}\" available."), option->get_value());
```
ppisar, line 219: It basically works, but I don't like the output:

~~~~
$ ASAN_OPTIONS='detect_leaks=0' ./redhat-linux-build/dnf5/dnf5 download does-not-exist dontpanic 
Repositories loaded.
No package "does-not-exist" available.. Add "--help" for more information about the arguments.
test@fedora-43:~/dnf5 $ echo $?
2
~~~~

First, remove the trailing dot from the exception string. It is appended by the the exception handler.

Second, the exception is controlled with --skip-unavailable option and that is not clear from the message. Ideally I'd like to see something like "No package {} available; You might want to use --skip-unavailable option".

Third, I don't like the exception type. It assumes there was a syntax error. But this is more like GoalResolveError or libdnf5::cli::CommandExitError type. See dnf5/main.cpp:1533 for various exceptions DNF5 program can handle. If you managed to fake a transaction with libdnf5::GoalProblem::NOT_FOUND problem in the context, print_resolve_hints() would handle the --skip-unavailable hint for you.

m-blaha: We usually use `libdnf5::cli::CommandExitError()` - e.g.: https://github.com/rpm-software-management/dnf5/blob/8763cf9bd332d314f78a0bc5f998204480dad4d6/dnf5/commands/history/history_rollback.cpp#L58

alimirjamali: I am applying and testing `CommandExitError` (with exit code 1). It is compiling at the moment. Amendment in few minutes.

ppisar: That's much better. But now the full stop character is missing. A handler for CommandExitError exception does not append it:

~~~~
$ ASAN_OPTIONS='detect_leaks=0' ./redhat-linux-build/dnf5/dnf5 --no-plugins --disablerepo=\* --enablerepo=dnf-nightly download does-not-exist donpanic
Repositories loaded.
No package "does-not-exist" available; You might want to use --skip-unavailable option
test@fedora-43:~/dnf5 $ echo $?
1
~~~~

alimirjamali: > That's much better. But now the full stop character is missing. A handler for CommandExitError exception does not append it:

OK. I will amend it shortly.

BTW, I do not know why some of the unittest failed with ` Unknown argument "install" for command "group"` message. Seams to be present in other PRs as well and should be unrelated to this PR.

ppisar: The output is now perfect. Thanks.

ppisar (approved):
The change is good.

I believe the CI failure (group install) is unrelated. Though I will try to locate the breaking commit before merging this change.

alimirjamali:
> I believe the CI failure (group install) is unrelated. Though I will try to locate the breaking commit before merging this change.

The 1st set of errors (`dnf clean` test failures) are introduced by this PR? (cc. @m-blaha):

https://github.com/rpm-software-management/ci-dnf-stack/pull/1660

m-blaha:
> > I believe the CI failure (group install) is unrelated. Though I will try to locate the breaking commit before merging this change.
> 
> The 1st set of errors (`dnf clean` test failures) are introduced by this PR? (cc. @m-blaha):
> 
> [rpm-software-management/ci-dnf-stack#1660](https://github.com/rpm-software-management/ci-dnf-stack/pull/1660)

Actually, it should be fixed by this PR.

alimirjamali:
> Actually, it should be fixed by this PR.

Maybe not. When I look at the test results, the output is from this (`dnf5daemon-client/commands/clean` rather than `dnf5/commands/clean`):
https://github.com/rpm-software-management/dnf5/blob/main/dnf5daemon-client/commands/clean/clean.cpp#L85-L89

So the `dnf5daemon-client/command/clean` output is unrelated to #817

ppisar:
The dnf/comps-group-list.feature tests fail for dnf5daemon-client because dnf5daemon-client does not implement them. The failure is unrelated to this pull request.

PR #2179, opened 2025-04-04
Add group ordering by orider_int to dnf5daemon

kontura:
This is a follow up to: https://github.com/rpm-software-management/dnf5/pull/2164

It is also needed for: https://github.com/rpm-software-management/ci-dnf-stack/pull/1661

It also unifies duplicated handling of group attributes in the daemon which was need for it to compile after adding the `order_int`.


jan-kolarik:
Sorry, I won't make this one :-/

Issue #295, opened 2023-02-15
Search command output to use libsmartcols


jan-kolarik:
In order to have consistent CLI output across our commands and the support for parsing the output easily, it is desired to have the output formatted with the `libsmartcols`.

This is considered as a follow-up of https://github.com/rpm-software-management/dnf5/pull/202.

alimirjamali:
PR Submitted.

<details>

<summary>The output of `dnf search libdnf5` before patch</summary>

```
[user@personal build]$ /usr/bin/dnf5 search libdnf5
Updating and loading repositories:
Repositories loaded.
Matched fields: name (exact)
 libdnf5.i686: Package management library
 libdnf5.x86_64: Package management library
Matched fields: name, summary
 libdnf5-cli-devel.i686: Development files for libdnf5-cli
 libdnf5-cli-devel.x86_64: Development files for libdnf5-cli
 libdnf5-plugin-actions.x86_64: Libdnf5 plugin that allows to run actions (external executables) on hooks
 libdnf5-plugin-appstream.x86_64: Libdnf5 plugin to install repo Appstream data
 libdnf5-plugin-expired-pgp-keys.x86_64: Libdnf5 plugin for detecting and removing expired PGP keys
 libdnf5-plugin-rhsm.x86_64: Libdnf5 rhsm (Red Hat Subscription Manager) plugin
 perl-libdnf5-cli.x86_64: Perl 5 bindings for the libdnf5-cli library
 python3-libdnf5.x86_64: Python 3 bindings for the libdnf5 library
 python3-libdnf5-cli.x86_64: Python 3 bindings for the libdnf5-cli library
 python3-libdnf5-python-plugins-loader.x86_64: Libdnf5 plugin that allows loading Python plugins
 ruby-libdnf5-cli.x86_64: Ruby bindings for the libdnf5-cli library
Matched fields: name
 libdnf5-cli.i686: Library for working with a terminal in a command-line package manager
 libdnf5-cli.x86_64: Library for working with a terminal in a command-line package manager
 libdnf5-devel.i686: Development files for libdnf
 libdnf5-devel.x86_64: Development files for libdnf
 perl-libdnf5.x86_64: Perl 5 bindings for the libdnf library
 ruby-libdnf5.x86_64: Ruby bindings for the libdnf library

```
</details>

<details>

<summary>The output of `dnf search libdnf5` after `libsmartcols` patch</summary>

```
[...]

PR #2180, opened 2025-04-04
Use `libsmartcols` for `dnf search` output

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/295


alimirjamali:
Review consideration and advises required:

1. At the beginning of `print_search_results` function, we have a block with `No matches found` message. Do I have the permission to throw an error message with exit code 1 there?
2. For `--showduplicates` option, we could have two different approaches. One would be having a separate `Arch` column. The other would be adding the `.arch` to package name like before. Please advise.

libdnf5-cli/output/search.cpp
```diff
@@ -97,19 +106,47 @@ void print_search_results(const SearchResults & results) {
...
+
+    if (!libdnf5::cli::tty::is_interactive()) {
+        // Do not hard-code 80 as non-interactive screen width. Let libdnf5::cli::tty to decide
+        auto screen_width = size_t(libdnf5::cli::tty::get_width());
+        scols_table_set_termwidth(table.get(), screen_width);
+        // The below is necessary to make the libsmartcols' truncation work with non-interactive terminal
+        scols_table_set_termforce(table.get(), SCOLS_TERMFORCE_ALWAYS);
+    }
 
+    const std::vector<std::tuple<const char *, double, int>> columns = {
```
lnicola, line 119: Nit: this can be an array.

libdnf5-cli/output/search.cpp
```diff
@@ -97,19 +98,64 @@ void print_search_results(const SearchResults & results) {
...
+
+    char * output_buffer;
+    size_t output_size;
+    FILE * output_stream;
+    output_stream = open_memstream(&output_buffer, &output_size);
+    scols_table_set_stream(table.get(), output_stream);
+    scols_print_table(table.get());
+    char line[4096];  // The usual maximum length of terminal line
+    // Printing table header in bold text
+    if (!fgets(line, sizeof(line), output_stream)) {
```
dcantrell: Static buffer alert.  In place of fgets(), I recommend getline().  getline() will read a line of input up to NUL and you can pass a NULL pointer as the buffer and getline() will allocate memory to store the line.  As the caller you do need to free() this buffer, but it's more reliable than using fgets() and a static buffer.

When I see comments like "usual maximum length of a terminal line" I know that someone in the future will find a way to make it larger.

libdnf5-cli/output/search.cpp
```diff
@@ -97,19 +98,64 @@ void print_search_results(const SearchResults & results) {
...
+        std::cout << _("Unable to retrieve search results.") << std::endl;
+        return;
+    }
+    std::cout << bold << std::string(line) << reset;
+    // Printing the rest of table with highlighting (only for first 2 columns)
+    auto pkg_n_desc_width = scols_column_get_width(scols_table_get_column(table.get(), 0));
+    pkg_n_desc_width += scols_column_get_width(scols_table_get_column(table.get(), 1));
+    pkg_n_desc_width += 2;  // extra space between columns
+    while (fgets(line, sizeof(line), output_stream)) {
```
dcantrell: Same as above, use getline() in your read loop rather than fgets().  output_stream needs to be initialized to NULL.

dcantrell (changes requested):
Overall I am not opposed to this PR.  I would like to see getline() used in place of fgets().

Lastly, I would prefer libsmartcols support to be a build time configuration option.  We should continue to be able to build dnf5 without needing libsmartcols.

libdnf5-cli/output/search.cpp
alimirjamali: This is done

libdnf5-cli/output/search.cpp
alimirjamali: This is also done. However, while the code compiles and runs fine locally (and passes all local tests), the CI/CD workflow is all red. Maybe I should not have rebased.

alimirjamali:
> Lastly, I would prefer libsmartcols support to be a build time configuration option. We should continue to be able to build dnf5 without needing libsmartcols.

Regarding this, I also do not like `libsmartcols` for non-interactive output (e.g. piped through less). Maybe in that case, it would be better to revert to previous code.

libdnf5-cli/output/search.cpp
```diff
@@ -97,19 +98,67 @@ void print_search_results(const SearchResults & results) {
...
+    auto pkg_n_desc_width = scols_column_get_width(scols_table_get_column(table.get(), 0));
+    pkg_n_desc_width += scols_column_get_width(scols_table_get_column(table.get(), 1));
+    pkg_n_desc_width += 2;  // extra space between columns
+    while (getline(&line, &size, output_stream) != -1) {
+        auto pkg_n_desc = std::string(line).substr(0, pkg_n_desc_width);
+        auto rest = std::string(line).substr(pkg_n_desc_width, std::string::npos);
+        // nitpick: The highlight does not work if the matching word is truncated
+        std::cout << highlight(pkg_n_desc) << rest;
+    }
+    free(line);
```
dcantrell, line 160: I think this free(line) needs to be inside the while loop and you also should add "line = NULL" for the next getline() call.  My understanding here is that since this is C API we can't rely on C++ cleaning up the memory management, but once you hand it to std::string on lines 155 and 156, we're good.  However, I could be wrong here and something magic happens with memory management for the C API calls, but I'm pretty sure anything that directly mallocs needs a corresponding free.  getline() will malloc so you need to free it once you're done.

alimirjamali: > I think this free(line) needs to be inside the while loop and you also should add "line = NULL" for the next getline() call.

It `realloc`s on next call. Source: https://man7.org/linux/man-pages/man3/getline.3.html

dcantrell: You're right, my bad.  Call it bad memories from working with a non-compliant getline() implementation in a C library that will remain unnamed.  :)

dcantrell:
> > ? It grows the runtime dependencies
> 
> Does it? With what? libsmartcols is already used _**heavily**_ in dnf5. It drives both the `dnf5 list` / `dnf5 history *` output, as well as the summary of changes for any transaction, and then also drives the status display during that transaction.

I misspoke.  Dependencies is the wrong word here, I meant complexity I guess, or additional complexity.  Both in the code but also in the UI.  As noted in other comments, it could be possible to introduce additional options to control the output.  Doing that well is hard so that's what I was trying to think about...what are the ultimate benefits here for most users.

> (Of those, I personally find the install/upgrade/etc. transaction progress display perfectly reasonable, and think it's a good use of smartcols. I'm not crazy about its use for the list-based functionality for the same reasons I don't love it for search. But at least the list views wrap overflowed lines on narrow displays (in ugly ways), instead of compressing one column into a pencil.)

I agree with you here.  Getting the output right for most users is difficult.  If this is a significant enough improvement, then I see it being fine.  If there's not much gain then maybe we don't really need it.

Issue #1785, opened 2024-10-19
repository column overly truncated in transaction summary


decathorpe:
From today's F41 update:

```
Package                                          Arch   Version                 Reposit      Size
Removing:
 kernel                                          x86_64 6.11.1-300.fc41         fedora    0.0   B
 kernel-core                                     x86_64 6.11.1-300.fc41         fedora   70.9 MiB
 kernel-modules                                  x86_64 6.11.1-300.fc41         fedora   62.7 MiB
 kernel-modules-core                             x86_64 6.11.1-300.fc41         fedora   36.8 MiB
Upgrading:
 adobe-mappings-cmap                             noarch 20231115-1.fc41         updates  15.2 MiB
   replacing adobe-mappings-cmap                 noarch 20230622-4.fc41         fedora   14.4 MiB
 adobe-mappings-cmap-deprecated                  noarch 20231115-1.fc41         updates 582.1 KiB
   replacing adobe-mappings-cmap-deprecated      noarch 20230622-4.fc41         fedora  582.1 KiB
 amd-gpu-firmware                                noarch 20241017-1.fc41         updates  97.6 MiB
   replacing amd-gpu-firmware                    noarch 20240909-1.fc41         fedora   21.1 MiB
...
```

The "Reposit" column is so truncated that both the column label ("Reposit" instead of "Repository") that it's impossible to distinguish between packages that come from the "updates" repo and packages that come from the "updates-testing" repo (in this case, all of the updates listed as coming from "updates" are indeed from "updates-testing", but there's no way to tell from the UI).

I don't even understand *why* this column is getting truncated to be so narrow - the table is *already* wider than the actual terminal width, so the window needs to be resized to make everything fit correctly. (This is also a regression from dnf 4 - it always reflowed output to fit into the current terminal width without needing to resize, or truncating columns).

ppisar:
What is your terminal size?

decathorpe:
Usually I just use the default size (80x24). dnf5 output never fits into that, it looks like it renders stuff for 83 columns wide at minimum.

Example from today: Columns are truncated, output is still 3 characters too wide for the current terminal width (see screenshots attached below).

```
$ sudo dnf distro-sync --refresh
Updating and loading repositories:
 Visual Studio Code                     100% |  89.7 KiB/s |  93.7 KiB |  00m01s
 Fedora 41 - x86_64 - Test Updates      100% |   3.1 MiB/s |   7.6 MiB |  00m02s
[...]

PR #2181, opened 2025-04-05
Improve upgrade summary on smaller terminals

alimirjamali:
resolves: https://github.com/rpm-software-management/dnf5/issues/785
resolves: https://github.com/rpm-software-management/dnf5/issues/1785


m-blaha (commented):
Would you mind extending the commit message a bit? It would be really helpful if you could briefly describe what the change does and how it achieves that. Including a couple of examples showing the behavior of DNF5 before and after the change would also make it easier to understand the impact. Thanks a lot!

alimirjamali:
> Would you mind extending the commit message a bit? It would be really helpful if you could briefly describe what the change does and how it achieves that.

I updated the PR commit comments and added some comments to the code.

> Including a couple of examples showing the behavior of DNF5 before and after the change would also make it easier to understand the impact. Thanks a lot!

Is it OK if I make a PR to `ci-dnf-stack` and cross-link it to this PR for the examples? Since the output is going to be changed the unittests require update as well.

Thank you very much for the review.

PR #2185, opened 2025-04-08
dnfdaemon: Configure separate cache directory

m-blaha:
Introduce a separate dnfdaemon configuration file for daemon specific config overrides - currently it overrides only the `cachedir` option.

### DIFF
--- a/libdnf5/repo/repo.cpp
+++ b/libdnf5/repo/repo.cpp
@@ -520,6 +520,14 @@ bool Repo::clone_root_metadata() {
 
     auto repo_cachedir = p_impl->config.get_cachedir();
     auto base_cachedir = p_impl->config.get_basecachedir_option().get_value();
+    auto system_cachedir = p_impl->config.get_main_config().get_system_cachedir_option().get_value();
+
+    // return fast if the system_cachedir does not exist or is equivelent to base_cachedir
+    std::error_code ec;
+    if (!std::filesystem::exists(system_cachedir, ec) ||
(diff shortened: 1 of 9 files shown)


libdnf5/repo/repo.cpp
```diff
@@ -520,6 +520,14 @@ bool Repo::clone_root_metadata() {
...
     auto repo_cachedir = p_impl->config.get_cachedir();
     auto base_cachedir = p_impl->config.get_basecachedir_option().get_value();
+    auto system_cachedir = p_impl->config.get_main_config().get_system_cachedir_option().get_value();
+
+    // return fast if the system_cachedir does not exist or is equivelent to base_cachedir
+    std::error_code ec;
+    if (!std::filesystem::exists(system_cachedir, ec) ||
+        std::filesystem::equivalent(base_cachedir, system_cachedir, ec)) {
+        return false;
+    }
```
pkratoch, line 530: Hm, the `std::filesystem::equivalent` returns `false` in case of an error, which is good, but I cannot find the same information for `std::filesystem::exists`. I think the only way how it could negatively influence this would be if the `system_cachedir` did exist, but the there was an error during the `std::filesystem::exists` and the default value in that case was `false`. But I don't in which circumstances the error can happen (according to the documentation, it would be `std::bad_alloc`) and if it would still make sense to continue with trying to clone the `system_cachedir` in that case.

jrohel (approved):
Thanks

Issue #2187, opened 2025-04-08
Wrong documenation for libdnf5::rpm::rpmvercmp() and libdnf5::rpm::evrcmp()


nikita-dubrovskii:
[This](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/rpm/nevra.hpp#L189) looks wrong:
```
/// Compare alpha and numeric segments of two versions.
/// @return 1 if `lhs` < `rhs`, -1 if `lhs` > `rhs`, 0 if they are equal
LIBDNF_API int rpmvercmp(const char * lhs, const char * rhs);
```

Here are logs from simple Python libdnf5 test: 
```
-- dracut-0:105-2.fc42.x86_64 vs dracut-105-3.fc42.x86_64
rpmvercmp(0, 0) = 0
rpmvercmp(105, 105) = 0
rpmvercmp(2.fc42, 3.fc42) = -1
rpmvercmp(3.fc42, 2.fc42) = 1
```

I guess docs should be like [here](https://github.com/rpm-software-management/rpm/blob/master/rpmio/rpmvercmp.cc):
```
/* return 1: a is newer than b */
/*        0: a and b are the same version */
/*       -1: b is newer than a */
```

ppisar:
Indeed. The function is just a wrapper for RPM's rpmvercmp(), so it behaves alike and the documentation should be alike. Thanks for the report.

There is a similar issue for: libdnf5::rpm::evrcmp().

ppisar:
Fixed with #2188.

Issue #2176, opened 2025-04-03
dnf/comps-group-list.feature CI tests fail for dnf5daemon: Unknown argument "install" for command "group"


ppisar:
testing-farm:fedora-rawhide-x86_64:dnf5daemon-tests CI test <https://artifacts.dev.testing-farm.io/64030dd9-5788-4dba-9a3d-4d38c53f8466/work-behave-dnf5daemonrz4pslfu/plans/integration/behave-dnf5daemon/execute/data/guest/default-0/script-00-1/output.txt> fails like this:

~~~~
@dnf5daemon
Feature: Tests for group list and info commands # dnf/comps-group-list.feature:2

  Background: Enable repo and mark a group as installed  # dnf/comps-group-list.feature:4

  Scenario: All user visible groups are listed by default (installed group is not duplicated)  # dnf/comps-group-list.feature:8
    Given I use repository "comps-group-list"                                                  # dnf/steps/repo.py:183
Successfully changed to user polkitd
12:38:21.068: Loading rules from directory /etc/polkit-1/rules.d
12:38:21.068: Loading rules from directory /run/polkit-1/rules.d
12:38:21.068: Error opening rules directory: Error opening directory ?/run/polkit-1/rules.d?: No such file or directory (g-file-error-quark, 4)
12:38:21.068: Loading rules from directory /usr/local/share/polkit-1/rules.d
12:38:21.069: Error opening rules directory: Error opening directory ?/usr/local/share/polkit-1/rules.d?: No such file or directory (g-file-error-quark, 4)
12:38:21.069: Loading rules from directory /usr/share/polkit-1/rules.d
12:38:21.069: Finished loading, compiling and executing 1 rules
Entering main event loop
Connected to the system bus
12:38:21.071: Acquired the name org.freedesktop.PolicyKit1 on the system bus
    And I successfully execute dnf with args "group install test-group"                        # dnf/steps/cmd.py:191
      Assertion Failed: Command has returned exit code 2: dnf5daemon-client -y --installroot=/tmp/dnf_ci_installroot_7mi9cjzt --releasever=29 --setopt=module_platform_id=platform:f29 --setopt=cachedir=/tmp/dnf_ci_installroot_7mi9cjzt/var/cache/dnf group install test-group
      Captured stdout:
      [31m
[...]

PR #2188, opened 2025-04-08
nevra: Fix description of evrcmp and rpmvercmp functions

nikita-dubrovskii:
This PR adjsuts description of both functions to actual [one](https://github.com/rpm-software-management/rpm/blob/master/rpmio/rpmvercmp.cc):
```
/* return 1: a is newer than b */
/*        0: a and b are the same version */
/*       -1: b is newer than a */
int rpmvercmp(const char * a, const char * b);
```

Issue: https://github.com/rpm-software-management/dnf5/issues/2187


ppisar (commented):
The change looks good. I will approve after finishing CI tasks.

ppisar (approved):
The CI failure is unrelated (#2176).

PR #2192, opened 2025-04-11
nevra: Add `bool Nevra::operator<(const Nevra &) const`

nikita-dubrovskii:
Although a `template<> cmp_nevra` function exists, it cannot be used
in Python bindings. This commit adds a member function overloads for
all comparions operators to enable comparison in bindings.

Due to new API this also bumps version to 5.2.13.0


nikita-dubrovskii:
@ppisar mind a review?

test/libdnf5/rpm/test_nevra.cpp
```diff
@@ -413,3 +413,18 @@ void NevraTest::test_cmp_naevr() {
...
+
+
+void NevraTest::test_nevra_operator_less() {
+    TestPackage foo_0_1_1_noarch("foo-1-1.noarch");
+    TestPackage foo_1_1_1_noarch("foo-1:1-1.noarch");
+    TestPackage foo_0_2_1_noarch("foo-2-1.noarch");
+    TestPackage foo_0_1_1_1_noarch("foo-1.1-1.noarch");
+    TestPackage foo_0_1_1_s390x("foo-1-1.s390x");
+
+    CPPUNIT_ASSERT(!(foo_0_1_1_noarch < foo_0_1_1_noarch));
```
ppisar, line 425: The tests should cover all operators you add. Please add tests for >, <=, and >=.
I think it's not needed to check various cases of being-less-than because the operator is implemented with test_cmp_nevra() which already covers them.

test/libdnf5/rpm/test_nevra.cpp
```diff
@@ -413,3 +413,20 @@ void NevraTest::test_cmp_naevr() {
...
+
+void NevraTest::test_nevra_comparison_operators() {
+    TestPackage foo_0_1_1_noarch("foo-1-1.noarch");
+    TestPackage foo_1_1_1_noarch("foo-1:1-1.noarch");
+    TestPackage foo_0_2_1_noarch("foo-2-1.noarch");
+    TestPackage foo_0_1_1_1_noarch("foo-1.1-1.noarch");
+    TestPackage foo_0_1_1_s390x("foo-1-1.s390x");
+
+    CPPUNIT_ASSERT(foo_0_1_1_noarch < foo_0_1_1_s390x);
+    CPPUNIT_ASSERT(foo_0_1_1_1_noarch < foo_0_2_1_noarch);
```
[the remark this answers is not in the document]
nikita-dubrovskii: Got you!

include/libdnf5/rpm/nevra.hpp
```diff
@@ -65,6 +65,15 @@ struct LIBDNF_API Nevra {
     /// @return `true` if all Nevra attributes (`name`, `epoch`, `version`, `release` and `arch`) match.
     bool operator==(const Nevra & other) const;
 
+    /// Compares Nevra attributes in order: `name`, `epoch`, `version`, `release`, `arch`, and checks
+    /// if any is less than the corresponding attribute.
+    /// @return `true` if any of the attributes is less than the corresponding one. Return `false` otherwise.
+    bool operator<(const Nevra & other) const;
```
ppisar, line 71: This patch adds new functions to the public API. That means this patch needs increasing DNF5 version from  5.2.12.0 to  5.2.13.0. See VERSION.cmake file.

ppisar: Could you please reverse the commit order, or squash them into a single commit? I try to avoid a situation where one gets a checkout with the new operators, but without the bumped version.

nikita-dubrovskii: Done. I've also updated the `spec` file, but skipped the changelog section. I guess there should be github-action bot or smth like that to do it properly, or ?

ppisar: Yeah, thanks for the spec file. But still could you swap the two commits?

nikita-dubrovskii: It's just 1 now

ppisar:
DNF team discussed this pull request and recommend to set up a <Nevra,Nevra> instance of cmp_nevra() in Swig template instead of adding new operators.

nikita-dubrovskii:
> DNF team discussed this pull request and recommend to set up a <Nevra,Nevra> instance of cmp_nevra() in Swig template instead of adding new operators.

https://github.com/rpm-software-management/dnf5/pull/2203

ppisar:
Now when we have #2203 merged, we don't need the new operators.

PR #2196, opened 2025-04-15
repo: Log correct path when loading system repo

m-blaha:
The debug message in `SolvRepo::load_system_repo()` did not respect the
value of the `installroot` option, always logging misleading "/" as the
root directory for loading the rpmdb.
This patch updates the message to correctly reflect the configured
installroot path.


ppisar (approved):
Good.

Issue #784, opened 2023-07-31
When specified advisory is not found dnf5 should report an error


kontura:
dnf4 output:
```
evan@fedora ~ $ sudo dnf install --advisory=foo hello
Last metadata expiration check: 0:00:52 ago on Thu 27 Jul 2023 01:16:07 PM UTC.
No security updates for argument "hello"
All matches were filtered out by exclude filtering for argument: hello
Error: Unable to find a match: hello
```
dnf5 output:
```
amatej@fedora ~ $ sudo dnf5 install --advisory=foo hello
Updating and loading repositories:
Repositories loaded.
Nothing to do.
```
We should also add some message like: `All matches were filtered out by exclude filtering for argument: hello`.

jan-kolarik:
The given advisories filtering is used with several commands (install, check-upgrade, repoquery, etc.). We could either report the message always when doing the filtering on a single place (probably [here](https://github.com/rpm-software-management/dnf5/blob/5.1.0/dnf5/commands/advisory_shared.hpp#L100) with checking if query is empty) or only for specific command when checking that the method returned a value, but it is an empty list (f.e. [here](https://github.com/rpm-software-management/dnf5/blob/5.1.0/dnf5/commands/install/install.cpp#L106) for install).

j-mracek:
I think that the first option is better (`advisory_query_from_cli_input`). The second option might hide a problem when multiple types are selected. We have to be very careful, because no results with --security is OK and doesn't require any warning. I somehow feel that we should handle `--advisory=foo` as an argument with install command => according --skip-unavailable.

Issue #1075, opened 2023-12-08
Unify errors and exit codes of query commands


pkratoch:
The behavior of query commands should be unified in these cases:

1. Output is empty because of e.g. no repositories are enabled.
2. Output is empty because the argument doesn't match anything.
3. Some arguments match, but some don't.

These commands currently in all the cases return exit code 0 and don't print any error message:

- advisory list
- advisory info
- environment list
- environment info
- group list
- group info
- repo list
- repo info
- repoquery
- copr list (considering only 1st case, since this command doesn't accept arguments)

These commands are different:

- module list and module info
    - exit code 0
    - in case there is an unmatched argument, it prints a message on stderr
- provides (considering only 2nd and 3rd case, since this command doesn't work without arguments)
    - exit code 1
    - in case there is an unmatched argument, it prints a message on stderr

Search is not included, as it's a little bit different. Similar is also repoquery. It might make sense for these to behave differently.

pkratoch:
I am not sure about exit codes, but I think a message on stderr about unmatched arguments would be good in all cases. Not sure about repoquery, though.

PR #2201, opened 2025-04-16
Report specs filtered out by advisories

m-blaha:
Resolves two use-cases:

1. The user is informed if there is no existing advisory in the `--advisories=...` list. Additionally, the transaction commands that support this option (install, upgrade) will fail with exit code 1.

```
❯ dnf5 install --advisories=foo hello
Repositories loaded.
No advisory found matching the requested name: "foo"
```

2. If the user tries to install or upgrade a package not present in the requested (and existing) advisories, the command prints a warning and fails with exit code 1 (unless the `--skip-unavailable` switch is present).

```
❯ dnf install --advisories=FEDORA-2025-77ace1a41b hello
Repositories loaded.
No match for argument 'hello' in advisories
```

Resolves: https://github.com/rpm-software-management/dnf5/issues/784

Tests adjustments: https://github.com/rpm-software-management/ci-dnf-stack/pull/1671

### DIFF
--- a/dnf5/commands/advisory_shared.hpp
+++ b/dnf5/commands/advisory_shared.hpp
@@ -28,7 +28,9 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/advisory/advisory_query.hpp>
 #include <libdnf5/base/base_weak.hpp>
 #include <libdnf5/utils/bgettext/bgettext-lib.h>
+#include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 
+#include <iostream>
 #include <optional>
 
 
@@ -43,7 +45,8 @@ inline std::optional<libdnf5::advisory::AdvisoryQuery> advisory_query_from_cli_i
     bool advisory_newpackage,
     const std::vector<std::string> & advisory_severities,
     const std::vector<std::string> & advisory_bzs,
-    const std::vector<std::string> & advisory_cves) {
+    const std::vector<std::string> & advisory_cves,
+    const bool strict_names) {
     std::vector<std::string> advisory_types;
     if (advisory_security) {
         advisory_types.emplace_back("security");
@@ -66,6 +69,16 @@ inline std::optional<libdnf5::advisory::AdvisoryQuery> advisory_query_from_cli_i
         if (!advisory_names.empty()) {
             auto advisories_names = libdnf5::advisory::AdvisoryQuery(base);
             advisories_names.filter_name(advisory_names);
+            if (advisories_names.empty()) {
+                const BgettextMessage msg = M_("No advisory found matching the requested name: \"{}\"");
+                if (strict_names) {
+                    throw libdnf5::cli::CommandExitError(1, msg, libdnf5::utils::string::join(advisory_names, ", "));
+                } else {
+                    std::cerr << libdnf5::utils::sformat(
+                                     TM_(msg, 1), libdnf5::utils::string::join(advisory_names, ", "))
+                              << std::endl;
+                }
+            }
             advisories |= advisories_names;
         }
 
--- a/dnf5/commands/advisory_shared.hpp
+++ b/dnf5/commands/advisory_shared.hpp
@@ -31,6 +31,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 
 #include <iostream>
+#include <memory>
 #include <optional>
 
 
@@ -46,7 +47,7 @@ inline std::optional<libdnf5::advisory::AdvisoryQuery> advisory_query_from_cli_i
     const std::vector<std::string> & advisory_severities,
     const std::vector<std::string> & advisory_bzs,
     const std::vector<std::string> & advisory_cves,
-    const bool strict_names) {
+    const bool strict) {
     std::vector<std::string> advisory_types;
     if (advisory_security) {
         advisory_types.emplace_back("security");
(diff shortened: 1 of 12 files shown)


m-blaha:
The PR requires also change in the ci-dnf-stack. I'll prepare a PR for it.

dnf5/commands/advisory_shared.hpp
```diff
@@ -66,6 +69,16 @@ inline std::optional<libdnf5::advisory::AdvisoryQuery> advisory_query_from_cli_i
         if (!advisory_names.empty()) {
             auto advisories_names = libdnf5::advisory::AdvisoryQuery(base);
             advisories_names.filter_name(advisory_names);
+            if (advisories_names.empty()) {
+                const BgettextMessage msg = M_("No advisory found matching the requested name: \"{}\"");
```
kontura, line 73: I like that this gives a concrete message.
I think it would be good to do something similar for `advisory_bzs` and `advisory_cves`.

In addition would it make sense to check if the returned advisory query is not empty (if there is one)?
Unless I am missing something if the query is empty no packages will get past that filter. I also think it can make the other message a little bit confusing: `No match for argument 'pkg' in advisories` but all advisories are filtered out.

libdnf5/base/log_event.cpp
```diff
@@ -158,6 +158,8 @@ std::string LogEvent::to_string(
                     utils::sformat(_("Cannot perform {0}, no match for: {1}."), goal_action_to_string(action), *spec));
             }
             return ret.append(utils::sformat(_("No match for argument: {}"), *spec));
+        case GoalProblem::NOT_FOUND_IN_ADVISORIES:
+            return ret.append(utils::sformat(_("No match for argument '{0}' in advisories"), *spec));
```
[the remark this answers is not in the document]
m-blaha: The message not mentioning the `argument` will be needed for the UPGRADE_ALL. Will try to come up with something.

dnf5/commands/advisory_shared.hpp
m-blaha: Yes, adding such a warning for both bzs and cves makes sense in the current state.

However, if we eventually decide to change the behavior to use intersection between different advisory filters instead of the current union (as discussed in https://github.com/rpm-software-management/dnf5/issues/1401), this could become a bit confusing. For example, if an advisory that fixes a given BZXY is filtered out due to a severity mismatch, the message "No advisory matching BZXY" might be misleading.

But I agree that providing some form of feedback when no advisory matches the given filters would be helpful to users.

kontura: > However, if we eventually decide to change the behavior to use intersection between different advisory filters instead of the current union (as discussed in https://github.com/rpm-software-management/dnf5/issues/1401), this could become a bit confusing.

Since I believe we are sticking with the current compatible version do you plan to add the mentioned warnings for bzs and cves or are we fine without them? I wouldn't consider it a blocker.

However I did notice the CI tests update now needs an update after the messages changed.

m-blaha: I will add the warnings for BZs and CVEs.
With regard to https://bugzilla.redhat.com/show_bug.cgi?id=2362585, does it make sense for `advisory list/info UNKNOWN-ADVISORY` commands to also return exit code 1?

kontura: > Or maybe more generally - return exit code 1 in case an advisory filter was specified, but the returning query is empty

That is a good question. Looking at https://github.com/rpm-software-management/dnf5/issues/1075 if all the query commands always return zero maybe we should keep it consistent?

m-blaha: Thanks for the pointer! Let me summarize what should happen when the filters result in an empty query across the different locations where advisory filters are used:

1. `advisory XXX` commands – return code 0
2. `check-upgrade` command – return code 0
3. `repoquery` command – return code 0
4. `install` command – similar to installing an unknown package – return code 1
5. `upgrade` command – similar to upgrading an unknown package – return code 1

In each case, a specific warning should be printed to stderr to inform the user about non-existent name/BZ/CVE.

kontura: That sound very good to me. :+1: 

(I am also wondering if `check-upgrade` shouldn't behave more like `upgrade` with regards to exit codes, if `check-upgrade unknown-package` should rather return code 1, but that is a separate thing.)

m-blaha: I'm still struggling with the `upgrade` use-case. There are situations (especially when no package spec is given), when having the empty advisories query is completely OK - e.g. `dnf upgrade --security`. So I guess the upgrade command with an empty advisories query is more like upgrade with no upgrades available in repos, then upgrade of an unknown package. Maybe this even apply to the `install` command from the original #784 issue.

Currently I think that we should not touch the exit codes and just print a warning/info message that the user entered non-existent advisory or not fixed BZ/CVE.

I should have think about it more before filing this PR...

kontura: I though the exit codes would change only when the advisory name/BZ/CVE doesn't match anything. It seemed to me that matches the unknown package case well.
So the advisory query can still be empty (which is valid as you said, but informing the user is nice) however when a specific advisory is requested and its not there it is an error.

Due to the queries being ORed it can also happen that the query is not empty but there was a wrong advisory name, like: `dnf upgrade --security --advisory=wrong-name`. To me this seems analogous to an upgrade of multiple packages where one has a typo: `dnf upgrade htop zsh unknown-package`. This is by default an error (controlled by `--skip-unavailable`).

Although maybe I am missing something its getting quite complex.

m-blaha: Maybe I'm just overthinking it. Added BZ/CVE warning message.

kontura (approved):
Thanks!
I will wait with the merge until https://github.com/rpm-software-management/ci-dnf-stack/pull/1671 is updated.

m-blaha:
@kontura Thank you! I've just updated the https://github.com/rpm-software-management/ci-dnf-stack/pull/1671 CI stack PR to match the current state of this PR.

PR #2203, opened 2025-04-17
bindings: Add Python cmp_nevra binding

nikita-dubrovskii:
This would simplify comparison of Nevra objects:
```python
import libdnf5

def parse_nevra(nevra_str):
    form_nevra = libdnf5.rpm.VectorNevraForm(1, libdnf5.rpm.Nevra.Form_NEVRA)
    nevra = libdnf5.rpm.Nevra.parse(nevra_str, form_nevra)[0]
    return nevra

# This is how we can compare today (similar to C++ code)
def less(lhs, rhs):
    for attr in ["get_epoch", "get_version", "get_release"]:
        r = libdnf5.rpm.rpmvercmp(getattr(lhs, attr)(), getattr(rhs, attr)())
        if r != 0:
            return r == -1
    return False

if __name__ == "__main__":
    old = parse_nevra("foo-1.2.2-1.fc41.x86_64")
    new = parse_nevra("foo-1.2.3-2.fc41.x86_64")
    
    assert(less(old, new) and libdnf5.rpm.cmp_nevra(old, new) == True)
    assert(less(old, old) or libdnf5.rpm.cmp_nevra(old, old) == False)
    assert(less(new, old) or libdnf5.rpm.cmp_nevra(new, old) == False)
```

### DIFF
--- a/bindings/libdnf5/rpm.i
+++ b/bindings/libdnf5/rpm.i
@@ -61,6 +61,7 @@
 %template(VectorNevra) std::vector<libdnf5::rpm::Nevra>;
 %template(VectorNevraForm) std::vector<libdnf5::rpm::Nevra::Form>;
 %template(PairBoolNevra) std::pair<bool, libdnf5::rpm::Nevra>;
+%template(cmp_nevra) libdnf5::rpm::cmp_nevra<libdnf5::rpm::Nevra>;
 
 %include "libdnf5/rpm/versionlock_config.hpp"
 


bindings/libdnf5/rpm.i
```diff
@@ -61,6 +61,7 @@
 %template(VectorNevra) std::vector<libdnf5::rpm::Nevra>;
 %template(VectorNevraForm) std::vector<libdnf5::rpm::Nevra::Form>;
 %template(PairBoolNevra) std::pair<bool, libdnf5::rpm::Nevra>;
+%template(cmp_nevra) libdnf5::rpm::cmp_nevra<libdnf5::rpm::Nevra>;
```
ppisar, line 64: I think the Swig template identifier should held the specialization type as it needs to be unique, e.g. <code>%template(CmpNevraNevra) libdnf5::rpm::cmp_nevra<libdnf5::rpm::Nevra, libdnf5::rpm::Nevra></code>. But please ask @jrohel for the review. I have no experience with Swig.

nikita-dubrovskii:
@jrohel mind a review?

bindings/libdnf5/rpm.i
jrohel: I'm thinking. `cmp_nevra` is a function template. SWIG supports function overloading. So it shouldn't be a problem in the future to define multiple functions with the same name (multiple specializations of a function template under the same name and with different types of arguments).
For example:
```
%template(cmp_nevra) libdnf5::rpm::cmp_nevra<libdnf5::rpm::Nevra, libdnf5::rpm::Nevra>
%template(cmp_nevra) libdnf5::rpm::cmp_nevra<libdnf5::rpm::Package, libdnf5::rpm::Package>
```
The problem might be the target language. SWIG supports multiple languages. Libdnf5 has bindings for Python, Ruby and Perl. So I looked at support for these 3 languages. It looks like overloaded functions are supported for all 3 languages. So we don't need to have a unique name in this case.

Issue #1928, opened 2024-12-05
API: PackageDownloader::download() method does not honor "retries" config option


m-blaha:
According to dnf5.conf(5) the `retries` option should limit the number of retries for downloading packages. But it's not honored in any way.

PR #2210, opened 2025-04-22
Deprecate deltarpm support in dnf5

m-blaha:
Following options are deprecated: `retries`, `deltarpm`, `deltarpm_percentage`.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1928

### DIFF
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -397,6 +397,11 @@ Deprecation of the ``metadata_timer_sync`` option
 -------------------------------------------------
 The ``metadata_timer_sync`` configuration option is now obsoleted by the |DNF_MAKECACHE_TIMER_NAME_INLINE_LITERAL| systemd timer settings.
 
+Deprecation of the ``retries`` option
+-------------------------------------
+The ``retries`` configuration option is now deprecated. In dnf4, despite being documented as the number of retries for downloading packages, it was only used to limit the number of package re-downloads due to deltarpm errors. Deltarpm is currently not implemented in DNF5.
+
+
 Changes to individual options
 -----------------------------
 ``best``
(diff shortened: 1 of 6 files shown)


kontura (commented):
For the getters I think we should also add a warning to stderr via
`LIBDNF5_DEPRECATED("The option does nothing");` macro in their definitions.

I have also noticed the builds are failing because we use the deprecated methods which now produces a warning and we treat all warnings as error.
I have run into this in the past and solved it by: https://github.com/rpm-software-management/dnf5/commit/9f0ad0ed10b079411f1915535dc66b407d91c1b5 but I don't know if that is the best approach.

m-blaha:
Thank you! I added the LIBDNF5_DEPRECATED macro and switched off the warnings so hopefully now the builds should pass.

kontura:
Oh, now the tests fail because the stderr is full of the new deprecation warnings.
Could we stop using the getters? It would resolve both the build and run time warnings.

m-blaha:
> Oh, now the tests fail because the stderr is full of the new deprecation warnings. Could we stop using the getters (since they do nothing)? It would resolve both the build and run time warnings.

:) I'll take a look.

m-blaha:
Since these options are both `ConfigMain` and `ConfigRepo` options, and the Repo one "inherits" from the Main one using `OptionChild` template (for which we need to use the deprecated ConfigMain getter), we have several options here:
1. change the Repo options from OptionChild to the underlying OptionNumber/OptionBool which is API/ABI breaking change
2. drop LIBDNF5_DEPRECATED macro from ConfigMain definitions
3. we can also declare the ConfigRepo a friend class of ConfigMain to allow access to the `main_config.p_impl` and use directly `main_config.p_impl->retries` instead of the getter (not sure, whether this affects ABI stability)

From these I like the second option most. Do you see any other option?

kontura:
> From these I like the second option most. Do you see any other option?

Not really

Only the `ConfigRepo::Impl` would need the friend but we would need declaration of `ConfiMain::Impl`.
Not sure if that is worth it.

I agree that dropping the `LIBDNF5_DEPRECATED` seems like the best option here.

kontura (approved):
Thanks!

PR #2213, opened 2025-04-22
Explicit 64->32 bit conversion in repo_cache.cpp

evan-goode:
On Fedora 41 i686, this conversion causes a warning which prevents compiling, so we need to make it explicit with static_cast. I'm not sure why the warning is thrown on F41 but not F42+.

Unfortuntaely, this warning represents a real bug: get_bytes_removed() can overflow on 32-bit systems if over 4GiB is removed. We should use a 64-bit integer for get_bytes_removed(), but changing this would break ABI compatibility since the just-released DNF 5.2.13.0.

I have included this patch downstream for DNF 5.2.13.0 on Fedora 41. I think we should merge it upstream as well, for now. Perhaps we should then fix the underlying issue by breaking ABI and making `get_bytes_removed` always return a u64. If we followed procedure, I believe that would mean bumping to DNF 5.3. But the 5.2.13.0 release has only been out for a couple days, and the ABI would only change on 32-bit platforms.

### DIFF
--- a/include/libdnf5/repo/repo_cache.hpp
+++ b/include/libdnf5/repo/repo_cache.hpp
@@ -45,8 +45,9 @@ struct LIBDNF_API RepoCacheRemoveStatistics {
 
     std::size_t get_files_removed();  // Number of removed files and links.
     std::size_t get_dirs_removed();   // Number of removed directorires.
-    std::size_t get_bytes_removed();  // Number of removed bytes.
-    std::size_t get_errors();         // Numbes of errors.
+    std::size_t
+    get_bytes_removed();  // Number of removed bytes. Note: will overflow on 32-bit systems if the total space removed exceeds 4GiB.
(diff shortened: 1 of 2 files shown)


m-blaha:
I would prefer releasing 5.2.13.1 (or similar) with the fixed API/ABI to always use 64-bit (std::unintmax_t ?) for `get_bytes_removed()`. Without this intermediate fix.
Can we somehow mark 5.2.13.0 release as "broken"?

ppisar:
I'm also for changing the return type and declaring it was a bug in the interface. size_t is for measuring memory objects. That's why stat(3) uses `off_t` type. For simplicity I would use uintmax_t.

evan-goode:
OK good, I agree. Let's do https://github.com/rpm-software-management/dnf5/pull/2215 instead and quickly release 5.2.13.1.

PR #2214, opened 2025-04-23
expired-pgp-keys: Use a temporary GnuPG home directory

### LINKED ISSUE #2212: `libdnf5-plugin-expired-pgp-keys` dnf plugin creates ~/.gnupg directory
https://bugzilla.redhat.com/show_bug.cgi?id=2361475

ppisar:
If dnf5 was executed as a root with HOME or GNUPGHOME variables set to other user's directory, gpg program executed by expired-pgp-keys would create that directory, but owned by root.

This patch fixes it by directing gpg' home directory a to a tempory directory. It will be typically /tmp/libdnf5.XXXXXX, in case you will need to register it into SELinux default file context database.

The temporary directory is resued by all gpg invocation to avoid reinitializing it again and again. This patch also adds --quiet option to gpg to supress an informational notice about creating a keybox there.

Resolves #2212
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2361475


evan-goode (approved):
LGTM, thanks!

PR #2215, opened 2025-04-23
RepoCacheRemoveStatistics::get_bytes_removed return uintmax_t

evan-goode:
On 32-bit systems, size_t is not large enough to represent the amount of freed space if it exceeds 4GiB. We should use a uintmax_t here instead.

Supersedes https://github.com/rpm-software-management/dnf5/pull/2213.


m-blaha (approved):
Thank you!

Issue #2170, opened 2025-03-31
RFE: change scriptlet "names" to use the spec designators


keszybz:
Loosely inspired by https://github.com/rpm-software-management/dnf5/pull/2169…

Dnf output about scriptlets uses "random" names like "postun-transaction". As a packager, I'm always confused by this output, because I know what a `%postuntrans` scriptlet is. This is what all the spec files use and what is described in Packaging Guidelines.

Please consider changing the output to just use those section names, something like
>>> Running %pre scriptlet: dbus-common-1:1.14.10-4.fc41.noarch
>>> Finished %post scriptlet: dbus-common-1:1.14.10-4.fc41.noarch

In general, the goal would be to make it easy to look at the output and then at the package .spec file and connect the one with the other in an obvious way. In particular, for scriplets which can appear multiple times, for example, `%trigger`, including the additional info like what the trigger is for in the output would be great. I have no idea if this is possible.

evan-goode:
I agree, the "display names" that DNF uses for these do not seem to be used elsewhere. We should just use the RPM macro names. I filed https://github.com/rpm-software-management/dnf5/pull/2216.

> In particular, for scriplets which can appear multiple times, for example, %trigger, including the additional info like what the trigger is for in the output would be great. I have no idea if this is possible.

Yeah, I'm not sure this is currently possible. @dmnks, is anything exposed in the RPM API about the reason a trigger was triggered?

PR #2216, opened 2025-04-23
Change scriptlet names to match spec designators

evan-goode:
For clarity, the names of the scriptlet types should match how they are defined in spec files.

For https://github.com/rpm-software-management/dnf5/issues/2170.


ppisar:
The patch is good. I also find the spec identifiers easier to understand. (Though I'm not sure nonpackagers will agree.)

Nonetheless, dnf/scriptlets.feature needs adjustments. Please do that.

ppisar (approved):
Good. I tested it locally with https://github.com/rpm-software-management/ci-dnf-stack/pull/1673.

PR #2221, opened 2025-04-24
download: Do not print "Downloading Packages:" header in quiet mode

ppisar:
The "Downloading Packages:" header was printed irresspective of --quiet option:

    # dnf -q download dontpanic
    Downloading Packages:
    #

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2362032


kontura (approved):
Thanks.

PR #2222, opened 2025-04-25
doc: Clarify advisories filters documentation

m-blaha:
This patch slightly adjusts the wording from "Consider only content..."
to "Include content..." to avoid implying that the options produce an
intersection of advisories when combined. The new phrasing aligns with
what was used in the DNF4 documentation.

Additionally, the descriptions of these options have been moved to the
shared library to improve maintainability and reduce duplication.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1401

### DIFF
--- a/doc/_shared/options/advisories.rst
+++ b/doc/_shared/options/advisories.rst
@@ -1,0 +1,4 @@
+``--advisories=ADVISORY_NAME,...``
+    | Include content contained in advisories with specified name.
+    | This is a list option.
+    | Expected values are advisory IDs, e.g. `FEDORA-2201-123`.
--- a/doc/_shared/options/advisory-severities.rst
+++ b/doc/_shared/options/advisory-severities.rst
@@ -1,0 +1,4 @@
+``--advisory-severities=ADVISORY_SEVERITY,...``
+    | Include content contained in advisories with specified severity.
+    | This is a list option.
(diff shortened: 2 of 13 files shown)


kontura:
This is great. Thank you.

For https://github.com/rpm-software-management/dnf5/issues/1401 I would also like to fix descriptions in https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/advisory_shared.hpp but it will be a bit tricky since the options are shared for both commands operating on advisories and packages.
Basically it should make sense for both `dnf5 advisory list --help` and `dnf5 repoquery --help`.

I would be totally fine doing that separately if you prefer.

m-blaha:
The options documentation is also used for both advisories / packages commands, so I would probably use the same descriptions also for these `--help` strings. Thanks for noticing, this is a great point!

evan-goode (changes requested):
Nice. To be complete, the dnf5daemon options (`dnf5daemon-client/commands/advisory/arguments.hpp`) still have the old "Limit to packages in * advisories" text, should we change them now also?

m-blaha:
> Nice. To be complete, the dnf5daemon options (`dnf5daemon-client/commands/advisory/arguments.hpp`) still have the old "Limit to packages in * advisories" text, should we change them now also?

You're totally right, thanks for spotting it - fixed!

kontura (approved):

PR #2224, opened 2025-04-25
doc: "dnf4 config-manager" dropped --dump and --dump-variables options

ppisar:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2270024

### DIFF
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -171,6 +171,12 @@ Changes to individual commands
       * Before: ``--save --setopt`` Now: ``setopt``
 
   * Existing repository files are not modified; drop-in override files are created instead.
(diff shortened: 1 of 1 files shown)


kontura:
Since the `--dump` option used to also dump repo config, like `dnf4 config-manager fedora --dump` maybe we should also mention `dnf5 --dump-repo-config=fedora`?

I have also noticed that `dnf4 config-manager --dump-variables` was replaced by `dnf5 --dump-variables` we could potentially include that as well.

ppisar:
I did not know about "dnf4 config-manger --dump REPOID". I will add document it. And the --dump-variables also. Thanks for the feedback.

ppisar:
I incorporated your suggestions to the commit.

kontura (approved):

PR #2233, opened 2025-05-01
Allow for "yuh" to confirm an action (Is this ok [y/N]: yuh)

notChocoMilk:
image


ppisar:
I'm against this patch. The prompt is going to expect locale specific answers. Not random words somebody finds funny.

notChocoMilk:
smh no joy or whimsy at all

PR #2240, opened 2025-05-06
dnfdaemon: Split-off polkit policy for trusted transactions

m-blaha:
The polkit policy for executing RPM transactions has been split into two distinct policies:

- `org.rpm.dnf.v0.rpm.execute_trusted_transaction`: This policy applies
  when all packages within the transaction originate from trusted repositories.

- `org.rpm.dnf.v0.rpm.execute_transaction`: This is the default policy.

A polkit rule allows any active local member of the wheel group to
execute trusted transactions.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -777,6 +777,21 @@ Package management service with a DBus interface.
 %{_mandir}/man8/dnf5daemon-server.8.*
 %{_mandir}/man8/dnf5daemon-dbus-api.8.*
 %endif
+
+
+# ========== dnf5daemon-server-polkit ==========
+
+%package -n dnf5daemon-server-polkit
+Summary:        Polkit rule to allow wheel group members install trusted packages
+License:        GPL-2.0-or-later
+Requires:       polkit
+
+%description -n dnf5daemon-server-polkit
+Polkit rule to allow active local wheel group members install packages from
+trusted repositories using dnf5daemon-server.
+
+%files -n dnf5daemon-server-polkit
+%{_datadir}/polkit-1/rules.d/org.rpm.dnf.v0.rules
 %endif
 
 
--- a/dnf5daemon-server/polkit/CMakeLists.txt
+++ b/dnf5daemon-server/polkit/CMakeLists.txt
@@ -1,6 +1,11 @@
-SET(POLKIT_DIR /usr/share/polkit-1/actions)
+SET(POLKIT_DIR /usr/share/polkit-1)
 
 install (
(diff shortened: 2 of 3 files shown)


dnf5daemon-server/services/goal/goal.cpp
```diff
@@ -342,10 +342,32 @@ sdbus::MethodReply Goal::get_transaction_problems(sdbus::MethodCall & call) {
...
 }
 
+/// Returns true in case all inbound packages in the transaction comes from
+/// repository with gpg checks enforced.
+bool is_transaction_trusted(libdnf5::base::Transaction * transaction) {
+    for (const auto & tspkg : transaction->get_transaction_packages()) {
+        if (transaction_item_action_is_inbound(tspkg.get_action())) {
+            auto repo = tspkg.get_package().get_repo();
+            if (repo->get_type() == libdnf5::repo::Repo::Type::COMMANDLINE) {
+                // local rpm file installation is untrusted
```
ppisar, line 352: I heard rumors that RPM version 6 is going to enforce signatures <https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/K4SO2G63Z7EGLGTXKFQX52SNZ45372V7/>. That's maybe why Panu sent a patch #2232 for DNF5.

I don't mind this conservative approach. (It's maybe good as an proactive measure against tricking a user into downloading a package with a web browser and then pointing the D-Bus client onto the local RPM file.) I only say that your pul request could be unnecessary strict.

m-blaha: Thanks for pointing that out. For the time being, I'd rather stay on the safe side, but I'll definitely be monitoring the development in RPM.

dnf5.spec
```diff
@@ -777,6 +777,21 @@ Package management service with a DBus interface.
...
 %{_mandir}/man8/dnf5daemon-dbus-api.8.*
 %endif
+
+
+# ========== dnf5daemon-server-polkit ==========
+
+%package -n dnf5daemon-server-polkit
+Summary:        Polkit rule to allow wheel group members install trusted packages
+License:        GPL-2.0-or-later
+Requires:       polkit
```
mcrha, line 787: You might also add `BuildArch: noarch` , due to the package containing only a text file, nothing built nor arch specific.

m-blaha: That's a good point. Addressed by PR https://github.com/rpm-software-management/dnf5/pull/2283

Issue #2249, opened 2025-05-16
`repoquery` | Printing issues


Ricky-Tigg:
DNF5 v. 5.2.13.1
```
_@_:~$ dnf5 -q rq --installed --qf %{name} dnf5*
dnf5dnf5-plugins_@_:~$ 
```
Hello. Two printing issues:
- before Shell prompt
- one-line

Output used to be printed correctly in the past, at least on DNF4,  as follows:
```
_@_:~$ dnf4 -q rq --installed --qf '%{name}' dnf5*
dnf5
dnf5-plugins
_@_:~$ 
``` 
And as always, the issue seen in so many tickets, non up-to-date DN5 `man`-pages – here, missing alias `rq` matching `repoquery` in the `man dnf5-repoquery` related pages.

ppisar:
> DNF5 v. 5.2.13.1
> 
> ```
> _@_:~$ dnf5 -q rq --installed --qf %{name} dnf5*
> dnf5dnf5-plugins_@_:~$ 
> ```
> 
> Hello. Two printing issues:
> 
>     * before Shell prompt
> 
>     * one-line
> 
> 
> Output used to be printed correctly in the past, at least on DNF4, as follows:
> 
> ```
> _@_:~$ dnf4 -q rq --installed --qf '%{name}' dnf5*
> dnf5
> dnf5-plugins
> _@_:~$ 
> ```
> 
This was an deliberate change to align with `rpm --qf` command. It is documented in dnf5-changes-from-dnf4(7) manual.

> missing alias `rq` matching `repoquery` in the `man dnf5-repoquery` related pages.

DNF5 documentation only promotes full command names (repoquery) and does not document aliases which only exist for backward compatibility. The intention standardized on the full command names and drop the redundant aliases.

You can get a list of the aliases with "dnf5 --help" in Compatibility Aliases section.

ppisar:
To get the new line, use `\n` sequence in the query formatting string: `dnf5 repoquery --qf '%{name}\n'`

m-blaha:
On a positive note, it also provides users with the ability to generate more complex, multi-lined outputs from the repoquery command.

PR #2251, opened 2025-05-19
doc: Document '\n' sequence in repoquery --queryformat string

ppisar:
"dnf5 repoquery --qf" does not append newlines. This change was documented, but it was not mentioned how to achieve the newlines.

Related: #2249


kontura (approved):
The failing tests are not related, thank you.

Issue #1496, opened 2024-05-16
Support `--releasever=/` to detect releasever from host system when using `--installroot`


teohhanhui:
> Note: You may also want to use the command-line option `--releasever=<release>` when creating the installroot, otherwise the *$releasever* value is taken from the rpmdb within the installroot (and thus it is empty at the time of creation and the transaction will fail). If `--releasever=/` is used, the releasever will be detected from the host (`/`) system. The new installroot path at the time of creation does not contain the *repository*, *releasever* and *dnf.conf* files.

https://dnf.readthedocs.io/en/latest/command_ref.html#installroot-label

jan-kolarik:
Hi, have you tried using the `--use-host-config` option? But anyway, I see the reference to `--releasever=/` is also in the dnf5 docs as it was retaken from dnf4, so it either has to be additionally implemented or fixed in the docs.

PR #2256, opened 2025-05-21
doc: `--use-host-config` instead of `--releasever=/`

evan-goode:
In DNF5, we don't support `--releasever=/`, but we do have the new `--use-host-config` option which sources configuration, reposdir, and vars from the host system. `--use-host-config` is typically what you want when bootstrapping a new installroot that doesn't have a releasever set, so the docs should recommend `--use-host-config` instead of `--releasever=/`.

Resolves https://github.com/rpm-software-management/dnf5/issues/1496.

### DIFF
--- a/doc/misc/installroot.7.rst
+++ b/doc/misc/installroot.7.rst
@@ -52,7 +52,7 @@ relative to the host.
 
 Note: You may also want to use the command-line option ``--releasever=RELEASEVER`` when creating
 the installroot, otherwise the $releasever value is taken from the rpmdb within the installroot
-(and thus it is empty at the time of creation and the transaction will fail). If ``--releasever=/``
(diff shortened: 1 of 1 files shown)


evan-goode:
My vote would be to not implement `--releasever=/`, to me it's a confusing way to denote "automatically detect the releasever from /", and most of the time you will want `--use-host-config` anyway to use config/repositories from /.

This is a departure from DNF 4, but we already changed the behavior here with https://github.com/rpm-software-management/dnf5/pull/271, by not loading config and repos from / when using an installroot without `--use-host-config`.

Conan-Kudo (approved):

ppisar:
We should also document this change in doc/changes_from_dnf4.7.rst.

PR #2257, opened 2025-05-22
changes_from_dnf4: makecache doesn't always refresh

evan-goode:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2324177.

See also https://github.com/rpm-software-management/dnf/issues/2242.


Conan-Kudo (approved):
This is a bit unintuitive, but OK.

Issue #773, opened 2023-07-25
dnf5daemon: Add function to get list of recently updated packages into a D-Bus interface


mcrha:
In order to be feature complete with PackageKit (on the gnome-software side), it would be good to add a function to get list of recently updated packages with an information about the update itself (like the update details, but not necessarily).

This is also related to the https://github.com/rpm-software-management/dnf5/issues/258

mcrha:
And related to the https://github.com/rpm-software-management/dnf5/issues/140 , if the history is provided also in the D-Bus API.

mcrha:
Also, you might provide a date/time stamp, for which the changes are returned. From the gnome-software point of view, you do not want to show "something new had been installed" with repeated .rpm files for the 10th time, which had been updated few months ago. Thus something like "get changes/updates since YYYYMMDDhhmmss".

Regarding update details, the type and severity (when the update happened within an advisory) would help. Or add a reference to such advisory, thus one would be able to ask for advisory details in a second call.

jan-kolarik:
~~Dummy comment to kick-off our sync.~~

Issue #1175, opened 2024-01-15
Unify epoch usage in DNF5


j-mracek:
Epoch is not handled consistently in DNF, DNF5, RPM

RPM does not show 0 epoch for `rpm -q B`

RPM shows 0 epoch  `rpm -q B --qf "%{name}-%{evr}\n" when `Epoch: 0` in in spec.

`rpmbuild` does not add epoch to rpm file name.

DNF/DNF5 and command repoquery always shows 0 epoch.

DNF/DNF5 transaction table does not show 0 epoch.

DNF transaction summary does not show 0 epoch.

DNF history info does not show 0 epoch for transaction packages.

DNF5 history info shows 0 epoch for transaction packages.

DNF5 and `Package.to_string()` returns optional epoch as well as Python binding `str(Package)`

DNF5 uses full NEVRA in scriptlets: Running trigger-install scriptlet: man-db-0:2.11.2-2.fc38.x86_64

Createrepo_c always adds 0 epoch to metadata

_Originally posted by @j-mracek in https://github.com/rpm-software-management/dnf5/pull/1092#discussion_r1447308700_

Conan-Kudo:
We should just always show `Epoch: 0` when `%EPOCH` is null or zero.

kontura:
I also vote for always showing the epoch.

Conan-Kudo:
Note that createrepo also coerces this to zero, so we would be consistent with that.

j-mracek:
> Note that createrepo also coerces this to zero, so we would be consistent with that.

I forget to mention that in the issue description. Thank you for mentioning it.

Conan-Kudo:
This really should be prioritized, it screws with API consumers too, see: https://github.com/rpm-software-management/dnf5/issues/1359 from @anaselli

PR #2259, opened 2025-05-26
dnfdaemon: Add new History interface

m-blaha:
Create a new D-Bus interface to retrieve information from the dnf5
transaction history. The new History interface currently supports only
one method: "recent_changes". This method retrieves information from the
history database about changes in the installed package set since the
specified point in time.

Resolves: https://github.com/rpm-software-management/dnf5/issues/773

### DIFF
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
@@ -1,0 +1,79 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+
+<!--
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+-->
+
+<node>
+<!-- org.rpm.dnf.v0.History:
+   @short_description: Interface to transactions history database
+-->
+<interface name="org.rpm.dnf.v0.History">
+    <!--
+        recent_changes:
+        @options: an array of key/value pairs
+        @changeset: a dictionary {"<type_of_change>":[<package>,<package>,...]}
+
+        Get a list of packages that have been recently changed. The
+        "type_of_change" key in the output dictionary can be one of the
+        following: "installed", "removed", "upgraded", or "downgraded". The
+        packages in the output are represented as dictionaries containing the
+        requested set of attributes for the changed packages.
+
+        For installed, upgraded, and downgraded packages, the currently
+        installed version is returned.
+
+        For upgraded and downgraded packages, the "original_evr" attribute
+        is also included, containing the package EVR before the change.
+
+        If requested, upgraded packages may also include an "advisory_name"
+        field, indicating the name of the update that contains the upgraded
+        (i.e., currently installed) version of the package.
+
+        For removed packages, only the name, arch, and evr of the removed
+        version are provided.
+
+        Following options and filters are supported:
+
+            - upgraded_packages: boolean, default "true"
+                Whether to include upgraded packages in the output.
+            - downgraded_packages: boolean, default "true"
+                Whether to include downgraded packages in the output.
+            - installed_packages: boolean, default "true"
+                Whether to include installed packages in the output.
+            - removed_packages: boolean, default "true"
+                Whether to include removed packages in the output.
+            - include_advisory: boolean, default "true"
+                Whether to include "advisory_name" to upgraded packages attributes.
+            - package_attrs: list of strings, default ["name", "summary", "evr", "arch"]
+                List of package attributes that are returned for each changed
+                package. Supported attributes are the same as in list() method of
+                the Rpm interface.
+            - since: int64, unix timestamp
+                Get changes since a given point in time.
+
+        Unknown options are ignored.
+    -->
+    <method name="recent_changes">
+        <arg name="options" type="a{sv}" direction="in" />
+        <arg name="changeset" type="a{saa{sv}}" direction="out" />
+    </method>
+</interface>
+
+</node>
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
@@ -42,9 +42,11 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         For upgraded and downgraded packages, the "original_evr" attribute
         is also included, containing the package EVR before the change.
 
-        If requested, upgraded packages may also include an "advisory_name"
-        field, indicating the name of the update that contains the upgraded
-        (i.e., currently installed) version of the package.
+        If requested, upgraded packages may also include an "advisories" list
+        field, indicating the name of the updates that contain the upgraded
+        (i.e., currently installed) version of the package. If the "all_advisories"
+        option is set, advisories for all versions between the original and currently
+        installed versions are also included.
 
         For removed packages, only the name, arch, and evr of the removed
         version are provided.
@@ -60,7 +62,10 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
             - removed_packages: boolean, default "true"
                 Whether to include removed packages in the output.
             - include_advisory: boolean, default "true"
-                Whether to include "advisory_name" to upgraded packages attributes.
+                Whether to include "advisories" to upgraded packages attributes.
+            - all_advisories: boolean, default "false"
+                Only make sense with include_advisory=true. Include also advisories
+                for all versions between the original and currently installed.
             - package_attrs: list of strings, default ["name", "summary", "evr", "arch"]
                 List of package attributes that are returned for each changed
                 package. Supported attributes are the same as in list() method of
--- a/dnf5daemon-server/services/history/history.cpp
+++ b/dnf5daemon-server/services/history/history.cpp
@@ -1,0 +1,229 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "history.hpp"
+
+#include "dbus.hpp"
+#include "package.hpp"
+
+#include <libdnf5/advisory/advisory_package.hpp>
+#include <libdnf5/advisory/advisory_query.hpp>
+#include <libdnf5/common/sack/query_cmp.hpp>
+#include <libdnf5/rpm/nevra.hpp>
+#include <libdnf5/rpm/package_query.hpp>
+#include <libdnf5/transaction/rpm_package.hpp>
+#include <libdnf5/transaction/transaction_history.hpp>
+#include <libdnf5/utils/format.hpp>
+#include <sdbus-c++/sdbus-c++.h>
+
+void History::dbus_register() {
+    auto dbus_object = session.get_dbus_object();
+#ifdef SDBUS_CPP_VERSION_2
+    dbus_object
+        ->addVTable(sdbus::MethodVTableItem{
+            sdbus::MethodName{"recent_changes"},
+            sdbus::Signature{"a{sv}"},
+            {"options"},
+            sdbus::Signature{"a{saa{sv}}"},
+            {"changeset"},
+            [this](sdbus::MethodCall call) -> void {
+                session.get_threads_manager().handle_method(
+                    *this, &History::recent_changes, call, session.session_locale);
+            },
+            {}})
+        .forInterface(dnfdaemon::INTERFACE_HISTORY);
+#else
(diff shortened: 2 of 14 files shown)


mcrha:
The `since` argument does not seem to work. When I omit it, I get the history for the last entry (which is not that bad), but when I pass there `1692092114`, which is `Tue Aug 15 09:35:14 AM UTC 2023`, I get no results. I did several transactions today, to test the various sections:
```
 ID Command line                       Date and time          Action(s) Altered
141 dnf downgrade vim-minimal          2025-05-27 12:57:48    8
140 dnf remove nvidia-gpu-firmware     2025-05-27 12:57:11    1
139 dnf update yelp                    2025-05-27 12:56:28    4
138 dnf install doxygen                2025-05-27 09:44:44    6
137 dnf install gnutls-debuginfo ....  2025-05-22 09:16:28    2
```

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+        output["upgraded"] = out_upgraded;
+    }
+    if (downgraded_requested) {
+        output["downgraded"] = out_downgraded;
+    }
+    if (installed_requested) {
+        output["installed"] = out_installed;
+    }
+    if (removed_requested) {
```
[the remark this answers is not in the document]
m-blaha: I believe returning even an empty array makes the client's life easier. They don't need to check for the presence of the key.

m-blaha:
> The `since` argument does not seem to work. When I omit it, I get the history for the last entry (which is not that bad), but when I pass there `1692092114`, which is `Tue Aug 15 09:35:14 AM UTC 2023`, I get no results. I did several transactions today, to test the various sections:

Hm, that's weird. It seems to be working for me. Do you mean you don't get any results at all, or that the arrays are returned, but empty? Also, dnfdaemon expects the `since` parameter to be int64 (type `x`); otherwise, the call should end with an error.

m-blaha:
>             - since: uint64, unix timestamp

Oops, my bad! Will fix it.

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+        {"changeset"},
+        [this](sdbus::MethodCall call) -> void {
+            session.get_threads_manager().handle_method(*this, &History::recent_changes, call, session.session_locale);
+        });
+#endif
+}
+
+std::string get_evr(const libdnf5::transaction::Package & trans_pkg) {
+    auto epoch = trans_pkg.get_epoch();
+    if (epoch == "0") {
```
kontura, line 69: Personally I am not a fan of this behavior especially for this interface. I believe always including the 0 is more client friendly.

There is also an issue: https://github.com/rpm-software-management/dnf5/issues/1175
(We could discuss it [and the issue] on a meeting if you want.)

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    auto & base = *session.get_base();
+    libdnf5::transaction::TransactionHistory history(base);
+    std::vector<libdnf5::transaction::Transaction> transactions;
+
+    bool timestamp_used = options.contains("since");
+    int64_t timestamp;
+    if (timestamp_used) {
+        timestamp = dnfdaemon::key_value_map_get<int64_t>(options, "since");
+        transactions = history.list_all_transactions();
```
kontura: As a potential optimization we could create a new DB query to fetch only the newer that `since` transactions.
However I wouldn't block this PR on that, it is only nice to have.

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+    std::unordered_set<std::string> seen_na{};
+
+    dnfdaemon::KeyValueMapList out_installed;
+    dnfdaemon::KeyValueMapList out_removed;
+    dnfdaemon::KeyValueMapList out_downgraded;
+    // pair of (old, new)
+    std::vector<std::pair<libdnf5::transaction::Package, libdnf5::rpm::Package>> upgrades;
+    using Action = libdnf5::transaction::TransactionItemAction;
+    libdnf5::rpm::PackageSet upgrades_set{base};
+    for (auto & transaction : transactions) {
```
kontura, line 151: This is kind of merging the transactions right? It is like the output from `dnf4 history info` with multiple transactions. 
I did some testing and the only thing I found is a problem with installonly packages.
I tried running with ancient timestamp to merge all my transactions and it reported only one kernel install but I have 3 installed.

I know that to reuse `libdnf5::transaction::transaction_merge.hpp` we would have to expose additional things on the libdnf5 API and maybe do some clean up but perhaps it would be better than having similar logic in two places. What do you think?

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            auto installed_pkg = installed_na.find(na);
+            if (action == Action::INSTALL) {
+                // check if the package is still installed
+                if (installed_requested && installed_pkg != installed_na.end()) {
+                    out_installed.push_back(package_to_map(installed_pkg->second, pkg_attrs));
+                }
+            } else {
+                // package was REMOVEd of REPLACED
```
kontura: Typo: `of`.

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+    auto reply = call.createReply();
+    std::map<std::string, dnfdaemon::KeyValueMapList> output;
+    if (upgraded_requested) {
+        // inject advisory to upgraded packages
+        std::unordered_map<std::string, libdnf5::advisory::AdvisoryPackage> advisories_by_nevra;
+        if (include_advisory) {
+            auto advisories = libdnf5::advisory::AdvisoryQuery(base);
+            for (const auto & adv_pkg :
+                 advisories.get_advisory_packages_sorted(upgrades_set, libdnf5::sack::QueryCmp::EQ)) {
+                advisories_by_nevra.emplace(adv_pkg.get_nevra(), adv_pkg);
```
kontura: I don't know if it happens but I think technically there can be multiple advisories for the same nevra.
This would only keep the first one right?

dnf5daemon-server/services/history/history.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    auto reply = call.createReply();
+    std::map<std::string, dnfdaemon::KeyValueMapList> output;
+    if (upgraded_requested) {
+        // inject advisory to upgraded packages
+        std::unordered_map<std::string, libdnf5::advisory::AdvisoryPackage> advisories_by_nevra;
+        if (include_advisory) {
+            auto advisories = libdnf5::advisory::AdvisoryQuery(base);
+            for (const auto & adv_pkg :
+                 advisories.get_advisory_packages_sorted(upgrades_set, libdnf5::sack::QueryCmp::EQ)) {
```
kontura: I am not 100% sure what is the intent but this will only keep advisory packages if the nevras exactly match.
For example if an upgrade from `a-1` to `a-5` happened and there is an advisory for `a-4` it won't be included even though it was fixed by the update. If I understand correctly it would only include advisory for `a-5`.

kontura (commented):
The logic here seems quite complex to me and it integrates several libdnf5 components, I would really like if we had tests for this.

dnf5daemon-server/services/history/history.cpp
m-blaha: That's a good point. I've done this for consistency reasons, to have package `evr` and `original_evr` in the same format. Currently `libdnf5::rpm::Package::get_evr()` (which eventually calls `id2str(id2solvable(id)->evr)` from libsolv) does not include `0` (at least not always) epoch to it's output.

dnf5daemon-server/services/history/history.cpp
m-blaha: Oh, there used to be a TODO for this, but it got lost during refactoring. I'll add it back.

dnf5daemon-server/services/history/history.cpp
m-blaha: Thanks for bringing this up. It's reasonable and I'll look into it.

dnf5daemon-server/services/history/history.cpp
m-blaha: Good point. Looks like one advisory name per package in the output is not enough. I'll change it to return a list of fixed advisories.

dnf5daemon-server/services/history/history.cpp
m-blaha: I tried to leverage the existing `libdnf5::transaction::merge_transactions()` code, but it is too focused on transaction replay. The biggest problem is that all upgrades generate a problem like: `Action 'Replaced' 'acpi-0:1.0.1-0.x86_64' cannot be merged because it is not present at that point (present versions are: acpi-1.7-24.fc41.x86_64) -> skipping it.` As a result, I'm unable to retrieve the originally installed version of the package before the merged transaction.

I'll look into the `merge_transactions()` code to see whether there's a possibility to adjust it to only merge transactions and not consider applicability for the replay.

dnf5daemon-server/services/history/history.cpp
m-blaha: The goal is for GNOME Software to be able to display something like this:

```
System Updates

NetworkManager-1.50.3-1.fc41
Update to 1.50.3 stable release:
 * Wait configuring MPTCP endpoints until DAD has completed.
 * Properly autoconnect OVS ports at boot.
 * Allow configuring "shared" IPv6 method in nmtui.
```

The NEVRA comes from the history database, and the text is from the security advisory (in this case, FEDORA-2025-d952d39072).

I'm not sure how useful it would be to print also advisories for all versions between the original one and the currently installed version. @mcrha what do you think?

Additionally, it would require adding an `AdvisoryQuery::get_advisory_packages_sorted` overload that works with a different structure than `PackageSet` (since, for example, in Fedora only the latest update is present in the repository, so I cannot get the package ID for an older update).

dnf5daemon-server/services/history/history.cpp
m-blaha: Eventually, I don't feel it's a good idea to make `merge_transactions()` a public API. It's not a straightforward transaction merge - it also takes the currently installed packages into account and, in fact, assumes the current package set as the starting point for the merged transaction. For my use case, I need something different: I want to treat the currently installed packages as the desired final state and only return what needs to be installed, upgraded, downgraded, or removed to reach that state.

dnf5daemon-server/services/history/history.cpp
mcrha: While having a full history of the changes between the newly installed version and the previously installed version would be nice, I do not think it's a show stopper. It also looks like an expensive operation, thus if you'll make it possible, it'll be good to have an option to enable it - maybe something like `with_details` or `with_detailed_changes` or any better name, possibly defaulting to `false`. It could be as an addition for the `include_advisory`, I guess.

It would work as a follow-up work for me too, no need to block this pull request on it.

include/libdnf5/advisory/advisory_set.hpp
```diff
@@ -132,6 +132,14 @@ class LIBDNF_API AdvisorySet {
...
     class LIBDNF_LOCAL Impl;
     std::unique_ptr<Impl> p_impl;
+
+    /// Gather AdvisoryPackages for each Advisory in the set.
+    /// The AdvisoryPackages are sorted by name, arch and evr. name and arch are
+    /// compared as strings, evr uses libdnf5::rpm::evrcmp() function.
+    ///
+    /// @param only_applicable Whether to return only AdvisoryPackages from applicable AdvisoryCollections.
+    LIBDNF_LOCAL std::vector<AdvisoryPackage> get_advisory_packages_sorted_by_name_arch_evr_string(
+        bool only_applicable = false) const;
```
kontura: Can I ask why is this not in the `Impl`?
Given that it is private I think it would be cleaner to have this completely hidden.

dnf5daemon-server/services/history/history.cpp
kontura: I see, that makes sense.
Thank you for the investigation!

kontura (approved):
This looks good, I have only one question.

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
```diff
@@ -0,0 +1,79 @@
...
+            - installed_packages: boolean, default "true"
+                Whether to include installed packages in the output.
+            - removed_packages: boolean, default "true"
+                Whether to include removed packages in the output.
+            - include_advisory: boolean, default "true"
+                Whether to include "advisory_name" to upgraded packages attributes.
+            - package_attrs: list of strings, default ["name", "summary", "evr", "arch"]
+                List of package attributes that are returned for each changed
+                package. Supported attributes are the same as in list() method of
+                the Rpm interface.
```
mcrha, line 72: Would it be possible to add also `installtime` as a package attribute, which would be the latest one from the considered installations, please? It's not the same as the `buildtime`, which is already available. It could be only here, I guess the package listing cannot get the install time easily.

I just realized that the gnome-software uses that install time in its internals, in the code responsible for notifying the user about installation changes, thus it could be quite handy to have. Supposing you can get to the installation time, which I guess you can, when there's the `since` option (aka the same time as the `since` consults is fine by me).

m-blaha: You mean to add "installtime" to the default value of `package_attrs`?

Actually `since` is not compared to the install time of a package, but to the rpm transaction execution time. This way we can also get the original version of an upgraded package, and also list of removed packages.

mcrha: Not to the default list necessarily, but to the list of the available package attributes (there's no such thing at the moment), while it can be available only for this history interface, not generally, especially if it's hard to get from the internal structures of the RPM/DNF. 

Hmm, is there a difference for the "added/installed/updated" packages? Not for me, but I may miss some point. For me, the transaction execution time matches install/add/update time. I do not mind package Z had been installed 2 minutes later than package A due to other 25 packages being installed before that package Z and after the transaction start. A second precision is not relevant here ;)

include/libdnf5/advisory/advisory_set.hpp
m-blaha: I was originally thinking about adding the method to the public interface and only later I moved to private part since it is only needed from AdvisoryQuery which is a subclass (and already is a friend) of AdvisorySet.
I also wanted to keep it as close to similar `get_advisory_packages_sorted_by_name_arch_evr()` as possible.
What do you think about making it public?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
m-blaha: > Not to the default list necessarily, but to the list of the available package attributes (there's no such thing at the moment), while it can be available only for this history interface, not generally, especially if it's hard to get from the internal structures of the RPM/DNF.

Oh I see. the `get_install_time()` method of `libdnf5::rpm::Package` class is not exposed via D-Bus. I will fix it in the follow-up PR.

include/libdnf5/advisory/advisory_set.hpp
kontura: I guess most users would prefer this version over the already public one because of the string sorting.
I would be fine with making it public.

m-blaha: Thanks, made it public.

Issue #2255, opened 2025-05-20
Unknown method reset or interface org.rpm.dnf.v0.Goal.


AngryPenguinPL:
After migrating to dnf5 I noticed a problem in an external application like yumex (dnf5) where an error occurs while performing package installation/removal operations:

`dbus.exceptions.DBusException: org.freedesktop.DBus.Error.UnknownMethod: Unknown method reset or interface org.rpm.dnf.v0.Goal.`

It turns out that the reset() method in the Goal interface is missing.

Here is full log from launching yumex: 
```
 File "/usr/lib/python3.11/site-packages/yumex/utils/__init__.py", line 91, in target
    result = self.task_func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/yumex/backend/presenter.py", line 99, in depsolve
    return self.package_backend.depsolve(pkgs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/yumex/backend/dnf5daemon/__init__.py", line 640, in depsolve
    res, rc = self._build_transations(pkgs, opts)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/yumex/backend/dnf5daemon/__init__.py", line 243, in _build_transations
    self.client.session_goal.reset()
  File "/usr/lib64/python3.11/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
14:59:01 CRITICAL: (yumex.main) -  Uncaught exception: org.freedesktop.DBus.Error.UnknownMethod: Unknown method reset or interface org.rpm.dnf.v0.Goal.
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/yumex/ui/window.py", line 449, in action_dispatch
    self.on_apply_actions_clicked()
[...]

PR #2260, opened 2025-05-26
add reset method, when using sdbus

timlau:
The reset method is not setup when sdbus is used.

This make appliactions that use this API break

https://github.com/rpm-software-management/dnf5/issues/2255

### DIFF
--- a/dnf5daemon-server/services/goal/goal.cpp
+++ b/dnf5daemon-server/services/goal/goal.cpp
@@ -115,6 +115,16 @@ void Goal::dbus_register() {
                 [this](sdbus::MethodCall call) -> void {
                     session.get_threads_manager().handle_method(*this, &Goal::cancel, call, session.session_locale);
                 },
+                {}},
+            sdbus::MethodVTableItem{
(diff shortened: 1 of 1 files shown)


ppisar:
Could you make the commit message more descriptive? What name of D-Bus method is added, what D-Bus interface is added to, and what the method is good for.

m-blaha:
@ppisar It's not a new method. The issue is that, due to compatibility with both sdbus-cpp v1 and v2, the D-Bus methods have to be set up twice - conditionally compiled for the currently used version of sdbus-cpp. The reset method was simply forgotten for sdbus-cpp v2.

ppisar:
I see. Then could you please rephrase the commit message so that it's obvious it's a bug fix? The messages are used for writing release notes. E.g. "dnf5daemon: Fix a missing reset method if built with sdbus-s++ version 2"

timlau:
I have changed the commit message to the one you proposed

m-blaha (approved):
Thanks for enhancing the commit message!

PR #2261, opened 2025-05-26
pre-commit: Exclude .tito/tito.props from check-xml

m-blaha:
This file is not intended to be an XML file, although it is identified
as one by identify-2.6.12.
This should fix the failing github pre-commit checks.


jrohel:
Thanks

PR #2266, opened 2025-05-28
cmdline_aliases: Support values consumed by command aliases

jrohel:
Added new key `number_of_attached_values` - Number of arguments consumed by the command alias. **These arguments can be used as values for attached named arguments.** Default is 0.
The `value` for `attached_name_arg` can contain `${<index>}` placeholders, which are replaced by values consumed by the command alias (as defined by `number_of_attached_values`).

The parser now supports configuration file versions `1.0` and `1.1`. Using the new `number_of_attached_values` key requires a configuration file with `version = 1.1`. Using it in a configuration file with `version = 1.0` will still work, but the parser will log a warning. Older versions of DNF that support only version `1.0` will accept only configuration files with version `1.0`, and will log `number_of_attached_values` as an unknown attribute.

Example:
- Alias `whatrequires` for `repoquery --installed --whatrequires=<requires_first_argument>`

<pre>
['whatrequires']
type = 'command'
attached_command = 'repoquery'
descr = 'Alias for "repoquery --installed --whatrequires=<argument>"' complete = true
number_of_attached_values = 1
attached_named_args= [
    { id_path = 'repoquery.installed' },
    { id_path = 'repoquery.whatrequires', <b>value='${1}'</b> }
]
</pre>

Command `dnf whatrequires nano,wine` means the same as `dnf repoquery --installed --whatrequires=nano,wine`


doc/misc/aliases.7.rst
```diff
@@ -79,9 +79,13 @@ Keys:
     - ``descr`` - Description that will be shown in help.
     - ``group_id`` - A group this alias is part of if any.
     - ``complete`` - Whether bash autocompletion should be used for this alias, default is false.
+    - ``number_of_attached_values`` - Number of arguments consumed by the command alias. These arguments
+      can be used as values for attached named arguments. Default is 0. Added in config file version 1.1.
+      Ignored by versions of DNF that support only config file version 1.0.
     - ``attached_named_args`` - Options that will be used with the command. The format is an array of inline
       tables, each of which must contain an ``id_path`` key to specify the path to an option, and may contain also
-      a ``value`` key to specify the value of the option.
+      a ``value`` key to specify the value of the option. The ``value`` can contain ``${index}`` placeholders,
```
ppisar, line 87: Please document that the index counts from 1.

ppisar (commented):
Otherwise, it works pretty well. Thanks.

dnf5/cmdline_aliases.cpp
```diff
@@ -124,14 +124,22 @@ void load_aliases_from_toml_file(Context & context, const fs::path & config_file
...
+                    break;
+                }
+            }
+            if (!supported) {
+                const auto msg = fmt::format(
                     "Unsupported version \"{}\" in file \"{}\", \"{}\" expected",
                     version,
                     config_file_path.native(),
-                    CONF_FILE_VERSION);
+                    libdnf5::utils::string::join(CONF_FILE_SUPPORTED_VERSIONS, ", "));
```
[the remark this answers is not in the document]
ppisar: The problem of warning a user is independent of this issue. I don't say you need to address it. I just thought that it was your omission.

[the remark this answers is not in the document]
jrohel: You're right that the message is logged, but it's also sent to standard error. The issue is that none of the messages in this file are currently localized.

Although it's beyond the original scope of this PR, I'll go ahead and localize all our error output messages in this file.

doc/misc/aliases.7.rst
jrohel: OK, I've added it to the documentation.

libdnf5-cli/argument_parser.cpp
```diff
@@ -47,6 +48,41 @@ std::size_t replace_all(std::string & inout, std::string_view what, std::string_
...
+            throw ArgumentParserInvalidValueError(
+                M_("Misconfigured alias \"{}\": cannot convert \"{}\" to a number"),
+                std::string(argv[0]),
+                match[0].str());
+        } catch (const std::out_of_range & ex) {
+            index = -1;
+        }
+        if (index < 0 || index >= argc) {
+            throw ArgumentParserInvalidValueError(
+                M_("Misconfigured alias \"{}\": index \"{}\" out of range"), std::string(argv[0]), match[0].str());
```
ppisar, line 76: The exception message appends `Add "--help" for more information about the arguments` suffix:

~~~~
dnf5 requires 'perl(lib)' --installed
Misconfigured alias "requires": index "${11111111111111111111111111111111111111}" out of range. Add "--help" for more information about the arguments.
~~~~

That suffix is misleading in this case. We should point to dnf5-aliases manual instead (or add no suffix). Should we use a different exception type?

dnf5/cmdline_aliases.cpp
```diff
@@ -681,51 +609,52 @@ void load_aliases_from_toml_file(Context & context, const fs::path & config_file
...
                                 }
                             } else if (key == "complete") {
                                 complete = value.as_boolean();
                             } else if (key == "attached_named_args") {
                                 attached_named_args = &value.as_array();
+                            } else if (key == "number_of_attached_values") {
+                                if (version == "1.0") {
+                                    const auto location = value.location();
+                                    logger->warning(
+                                        "Used config file version \"1.0\" for attribute \"{}\" of command \"{}\" "
```
ppisar, line 631: I worry people will use number_of_attached_values but forget to change version to 1.1 and they won't get any hint why it is ignored. Could the warning be printed to stderr too?

Also changing `version = '1.0'` example in the manual to 1.1 could be helpful. Otherwise, people who copy-paste examples will stick at 1.0.

ppisar (commented):
Otherwise, it's better now.

dnf5/cmdline_aliases.cpp
ferdnyc: Meson prints a warning whenever a `meson.build` feature is used that requires a Meson version greater than the one declared in the `project(meson_version: '>= x.y.z')` call, or whenever a feature is used that's deprecated in all matching versions. It's quite helpful.

The equivalent would be something like,

> 'alias_file.conf': WARNING: "number_of_args" requires version 1.1 or higher configuration, file declares version 1.0.

or

> 'alias_file.conf': WARNING: "number_of_args" requires version 1.1 or higher configuration, version 1.0 assumed because file has no 'version'.

...depending on whether the file contains any `version =` line at all.

@ppisar
> I worry people will use number_of_attached_values but forget to change version to 1.1 and they won't get any hint why it is ignored. 

Based on @jrohel's PR summary, it sounds like it _wouldn't_ be ignored.

> Using it in a configuration file with `version = 1.0` will still work, but the parser will log a warning.

...So, that sounds like basically the same approach as Meson.

ferdnyc: Actually, wait, maybe I'm having pronoun trouble. When you say, "they won't get any hint why **it** is ignored", by "it" do you mean the `number_of_attached_args`, or the `version = 1.0`?

ppisar: My problem is that logger->warning() does not print to stderr. It only writes into a log file full of debugging messages. I understand that DNF tries to overcome as many errors as it can, but not telling about them to the user is not good.

If I designed DNF logging mechanism, I would write all messages through logger object. The logger would write everything to the log file and messages of selected level (error, warning, info, debug) would also write to stderr. What level is selected would be influenced with -v (errors, warning, info) and -q (only errors) command-line options or a configuration option. Default would be errors and warnings.

But since DNF does not have this kind of messaging control, I simply proposed here copying this warning to stderr.

Problem with malfunctioning alias configuration is that a user gets executed a different action than he wrote on the command line. The tool silently changes behavior and that is the whole point why I started this thread. Maybe I have a problem that this should be an error and not a warning.

If @jrohel things that not reporting this mistake to stderr is intended behavior, or out of scope of this pull request, I will be fine with it and accept the pull request. But I need a response to move forward.

dnf5/cmdline_aliases.cpp
```diff
@@ -124,37 +173,37 @@ void load_aliases_from_toml_file(Context & context, const fs::path & config_file
...
+            bool supported{false};
+            for (auto * supported_version : CONF_FILE_SUPPORTED_VERSIONS) {
+                if (version == supported_version) {
+                    supported = true;
+                    break;
+                }
+            }
+            if (!supported) {
+                const auto msg = fmt::format(
                     "Unsupported version \"{}\" in file \"{}\", \"{}\" expected",
```
ppisar: This message goes to stderr, but it's not internationalized.

doc/misc/aliases.7.rst
```diff
@@ -76,49 +77,87 @@ The ``command`` alias defines an alias for a command.
...
+    - ``descr.<locale>`` – Description that will be shown in help. ``descr.C`` is used if a description
+      for the appropriate locale is not found. Added in config file version 1.1.
+    - ``descr`` – Description that will be shown in help. Deprecated in version 1.1, but remains backward
+      compatible. Does not support locales.
     - ``group_id`` - A group this alias is part of if any.
     - ``complete`` - Whether bash autocompletion should be used for this alias, default is false.
-    - ``attached_named_args`` - Options that will be used with the command. The format is an array of inline
-      tables, each of which must contain an ``id_path`` key to specify the path to an option, and may contain also
-      a ``value`` key to specify the value of the option.
+    - ``number_of_attached_values`` - Number of arguments consumed by the command alias. These arguments
```
ppisar: Current code does not support `number_of_attached_values`. Please remove it from the documentation.

doc/misc/aliases.7.rst
```diff
@@ -76,49 +77,87 @@ The ``command`` alias defines an alias for a command.
...
+      can be used as values for attached named arguments. Default is 0. Added in config file version 1.1.
+      Ignored by versions of DNF that support only config file version 1.0.
+    - ``required_values`` - Values (positional arguments) consumed by the command alias. The format is
+      an array of tables; each must include either `values_help.<locale>` for the active locale or
+      `values_help.C`, and may optionally include `descr.<locale>`. The default is an empty array.
+      Added in config file version 1.1.
+    - ``attached_named_args`` - Options that will be used with the command. The format is an array of tables,
+      each of which must contain an ``id_path`` key to specify the path to an option, and may contain also
+      a ``value`` key to specify the value of the option. The ``value`` can contain ``${index}`` placeholders,
+      which are replaced by values consumed by the command alias (as defined by ``number_of_attached_values``).
```
ppisar: Another occurrence of the removed keyword.

doc/misc/aliases.7.rst
```diff
@@ -76,49 +77,87 @@ The ``command`` alias defines an alias for a command.
...
     - ``group_id`` - A group this alias is part of if any.
     - ``complete`` - Whether bash autocompletion should be used for this alias, default is false.
-    - ``attached_named_args`` - Options that will be used with the command. The format is an array of inline
-      tables, each of which must contain an ``id_path`` key to specify the path to an option, and may contain also
-      a ``value`` key to specify the value of the option.
+    - ``number_of_attached_values`` - Number of arguments consumed by the command alias. These arguments
+      can be used as values for attached named arguments. Default is 0. Added in config file version 1.1.
+      Ignored by versions of DNF that support only config file version 1.0.
+    - ``required_values`` - Values (positional arguments) consumed by the command alias. The format is
+      an array of tables; each must include either `values_help.<locale>` for the active locale or
```
ppisar: The code uses `value_help`, but here the documentation reads `values_help`. Please correct the typo.

doc/misc/aliases.7.rst
```diff
@@ -76,49 +77,87 @@ The ``command`` alias defines an alias for a command.
...
+      `values_help.C`, and may optionally include `descr.<locale>`. The default is an empty array.
```
ppisar: Also `values_help` typo.

ppisar (commented):
I found few omissions in the documentation that do not reflect the changed code and one message without internationalization. I marked to locations in-line.

ppisar (approved):
Good.

PR #2268, opened 2025-05-28
dnf5: install/upgrade/reinstall/downgrade: Argument "--from-repo"

jrohel:
The `--from-repo` argument allows the user to run actions on packages in the specified repositories. However, any dependency resolution takes into account packages from all allowed repositories.

Multiple repository ids can be specified, separated by commas, and globs can be used.

This represents a simplified usage compared to the "do" command. Unlike the "do" command, the option is applied to all arguments and cannot be defined multiple times with different values.

Example:
dnf install --from-repo=fedora pkg1 pkg2

The `--from-repo` argument implements part of the functionality of the `repo-packages` command known from dnf4, but more generally. The dnf4 `repo-packages` command allows only one repoid to be specified.

### DIFF
--- a/dnf5/commands/from_repo.cpp
+++ b/dnf5/commands/from_repo.cpp
@@ -1,0 +1,54 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "from_repo.hpp"
+
+#include <libdnf5/conf/option_string_list.hpp>
+#include <libdnf5/utils/bgettext/bgettext-lib.h>
+#include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
+
+
+namespace dnf5 {
+
+void create_from_repo_option(Command & command, std::vector<std::string> & from_repos, bool detect_conflict) {
+    auto & parser = command.get_context().get_argument_parser();
+    auto from_repo_opt = parser.add_new_named_arg("from-repo");
+    from_repo_opt->set_long_name("from-repo");
+    from_repo_opt->set_description(
+        _("The following items can be selected only from the specified repositories. All enabled repositories will "
+          "still be used to satisfy dependencies."));
+    from_repo_opt->set_has_value(true);
+    from_repo_opt->set_arg_value_help("REPO_ID,...");
+    from_repo_opt->set_parse_hook_func(
+        [&from_repos, detect_conflict](
+            libdnf5::cli::ArgumentParser::NamedArg *, [[maybe_unused]] const char * option, const char * value) {
+            if (!detect_conflict || from_repos.empty()) {
+                from_repos = libdnf5::OptionStringList(value).get_value();
+            } else {
+                if (from_repos != libdnf5::OptionStringList(value).get_value()) {
+                    throw libdnf5::cli::ArgumentParserConflictingArgumentsError(
+                        M_("\"--from_repo\" already defined with diferent value"));
+                }
+            }
+            return true;
+        });
+    command.get_argument_parser_command()->register_named_arg(from_repo_opt);
+}
+
+}  // namespace dnf5
--- a/dnf5/commands/from_repo.hpp
+++ b/dnf5/commands/from_repo.hpp
@@ -1,0 +1,36 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
+
+Libdnf is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef DNF5_COMMANDS_FROM_REPO_HPP
+#define DNF5_COMMANDS_FROM_REPO_HPP
+
+#include <dnf5/context.hpp>
+
+#include <string>
+#include <vector>
+
+
+namespace dnf5 {
+
+void create_from_repo_option(Command & command, std::vector<std::string> & from_repos, bool detect_conflict);
+
+}
+
+#endif  // DNF5_COMMANDS_FROM_REPO_HPP
--- a/doc/commands/downgrade.8.rst
+++ b/doc/commands/downgrade.8.rst
@@ -54,6 +54,8 @@ Options
 ``--no-allow-downgrade``
     | Disable downgrade of dependencies when resolving the requested operation.
 
+.. include:: ../_shared/options/from-repo.rst
+
 ``--downloadonly``
     | Download the resolved package set without executing an RPM transaction.
 
--- a/dnf5/commands/do/do.cpp
+++ b/dnf5/commands/do/do.cpp
@@ -19,6 +19,8 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "do.hpp"
 
+#include "../from_repo.hpp"
+
 #include <dnf5/shared_options.hpp>
 #include <libdnf5/conf/const.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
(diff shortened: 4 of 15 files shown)


dnf5/commands/from_repo.cpp
```diff
@@ -0,0 +1,54 @@
...
+    from_repo_opt->set_has_value(true);
+    from_repo_opt->set_arg_value_help("REPO_ID,...");
+    from_repo_opt->set_parse_hook_func(
+        [&from_repos, detect_conflict](
+            libdnf5::cli::ArgumentParser::NamedArg *, [[maybe_unused]] const char * option, const char * value) {
+            if (!detect_conflict || from_repos.empty()) {
+                from_repos = libdnf5::OptionStringList(value).get_value();
+            } else {
+                if (from_repos != libdnf5::OptionStringList(value).get_value()) {
+                    throw libdnf5::cli::ArgumentParserConflictingArgumentsError(
```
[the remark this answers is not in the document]
jrohel: In this case, the header file is included through a chain of several other headers.
It will be cleaner to include it directly here. Thanks for the suggestion.

dnf5/commands/from_repo.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <dnf5/context.hpp>
+
+#include <string>
+#include <vector>
+
+
+namespace dnf5 {
+
+void create_from_repo_option(Command & command, std::vector<std::string> & from_repos, bool detect_conflict);
```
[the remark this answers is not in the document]
jrohel: I’ve added a comment with a description. (By the way, the function is not part of the API, but having a description is still a good idea.)

doc/commands/downgrade.8.rst
```diff
@@ -54,6 +54,8 @@ Options
 ``--no-allow-downgrade``
     | Disable downgrade of dependencies when resolving the requested operation.
 
+.. include:: ../_shared/options/from-repo.rst
```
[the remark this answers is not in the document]
jrohel: Yes, thanks

doc/_shared/options/from-repo.rst
```diff
@@ -0,0 +1,3 @@
+``--from-repo=REPO_ID,...``
+    | The following items can be selected only from the specified repositories. All enabled repositories will still be used to satisfy dependencies.
```
ppisar: This is the same  text as in the --help output. Wouldn't be better to use different, probably more detailed, explanation what the option does? It would be also good to explain what happens if --from-repo names a repository which is not enabled.

E.g. "Packages or their provides directly named on the command line to operate on will only be looked up from these specified repositories. Enabling and disabling repositories still applies. However, dependencies of those packages will be resolved from any enabled repository."

dnf5/commands/from_repo.cpp
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright Contributors to the dnf5 project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
```
ppisar: There are more places where we should write "dnf5" instead of "libdnf".

dnf5/commands/from_repo.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+Libdnf is free software: you can redistribute it and/or modify
```
ppisar: And here.

dnf5/commands/from_repo.cpp
```diff
@@ -0,0 +1,55 @@
...
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
```
ppisar: Here.

dnf5/commands/from_repo.cpp
```diff
@@ -0,0 +1,55 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
```
ppisar: And here.

dnf5/commands/from_repo.hpp
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright Contributors to the dnf5 project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
```
ppisar: s/libdnf/dnf5/

dnf5/commands/from_repo.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+Libdnf is free software: you can redistribute it and/or modify
```
ppisar: Here.

dnf5/commands/from_repo.hpp
```diff
@@ -0,0 +1,40 @@
...
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 2.1 of the License, or
+(at your option) any later version.
+
+Libdnf is distributed in the hope that it will be useful,
```
ppisar: Here.

dnf5/commands/from_repo.hpp
```diff
@@ -0,0 +1,40 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
```
ppisar: And here.

ppisar (commented):
Almost perfect. I found few more places where "dnf5" should be written. Also I have a proposal for a different text in the manual.

Conan-Kudo:
> @Conan-Kudo
> 
> > I don't see why we wouldn't want to allow people to select multiple repos to source from? The limitatio from dnf repository-packages isn't necessary for this model.
> 
> What do you mean? "--from-repo=REPO_ID, ..." is a list of repository IDs. So it's not like dnf "repository-packages" where only one repository can be defined.
> 

I'm saying that append `--from-repo` should work too for this.

> Or do you mean a more general approach, as implemented in the "do" command? -> Allow multiple uses of "--from-repo=REPO_ID, ..." to be able to install multiple packages with different "--from-repo" sources?

Yes, this too.

jrohel:
@Conan-Kudo
>>    Or do you mean a more general approach, as implemented in the "do" command? -> Allow multiple uses of "--from-repo=REPO_ID, ..." to be able to install multiple packages with different "--from-repo" sources?

>Yes, this too.

Named arguments (--something) now have a global meaning - their position in the command line no matters. For example: `dnf install --skip-unavailable pkg1 pkg2` and `dnf install pkg1 pkg2 --skip-unavailable` are equivalent.

However, consider this command: `dnf install pkg1 pkg2 --skip-unavailable --from-repo=repoA`
If we applied for the `--from-repo` the same argument-scoping logic as in the `do` command, then `--skip-unavailable` would apply to all listed packages (pkg1, pkg2), but `--from-repo` would apply only to the packages that follow it - in this case, none.

There was concern in the team that such behavior could confuse users: one named argument applying globally, while another applies only to the following items. To avoid this inconsistency, I’ve restricted argument scoping in standard commands - all named arguments now apply globally, regardless of position.

The more advanced, position-sensitive behavior is reserved for the `do` command, which is intended to act as a “super-command” for complex operations not possible with the basic commands.

Yes, I’m aware there’s a precedent - the dnf5 `builddep` command (provided by a plugin) uses the `--srpm` and `--spec` options, which can be repeated and apply only to the packages that follow them (similar to how `--from-repo` works in the do command). However, there are concerns about extending this behavior further. Making option scoping position-dependent in more commands could lead to confusion and inconsistent user expectations.

ppisar:
After today's DNF team meeting:

* If the commands without package arguments gains --from-repo option, it will be a separate pull request.
* The mixed-in syntax for --from-repo option will be retained for _do_ command.

jrohel:
> Nice. How about `distro-sync`, `builddep`, `check-upgrade`, and `repoclosure` too, in order of decreasing importance?

`distro-sync`:
I created PR https://github.com/rpm-software-management/dnf5/pull/2303 for `distro-sync` command. 

`repoclosure`:
The repoclosure command includes a `--check` option that behaves similarly to `--from-repo`. However, there are a couple of differences:
- The existing `--check` option doesn't support globs in its repository ID specification, though this could be easily fixed.
- Unlike `--from-repo`, `--check` works without explicitly listed packages on the command line, which effectively mirrors using `--from-repo with a '*'` package specification.

So, the core functionality is already there in `--check`; it just needs glob support, which is a straightforward fix. We can implement `--from-repo` for consistency across other commands.

> Also, upgrade, distro-sync, check-upgrade, and repoclosure all have a list of package specs as optional arguments, not required. It would be great if --from-repo still worked when no pkgspecs are provided. So dnf5 distro-sync --from-repo=<repoid> with no pkgspecs could be a replacement for dnf4 repository-packages <repoid> upgrade.

I've examined the `repository-packages` command implementation in DNF4. It automatically inserts the `'*'` package specification if the user doesn't provide a list of packages on the command line. `'*'` works with current implementation of `--from-repo` too.

As I mentioned in previous comment, using `'*'` also works with several other commands that require a package specification, such as `install`.

evan-goode:
> If a user wants to execute a command on all packages filtered by --from-repo, they can specify them using '*'. This approach already works and can be used even with commands that typically don't accept execution without package arguments. For example: dnf install --from-repo=my_repo '*' would install all packages from the "my_repo" repository, with dependencies potentially being installed from all allowed repositories.

That's good to know.

> I've examined the repository-packages command implementation in DNF4. It automatically inserts the '*' package specification if the user doesn't provide a list of packages on the command line. '*' works with current implementation of --from-repo too.

I think it would be appropriate to do this too in DNF5, if only to make the command a little shorter. But I am also fine with leaving it out. `dnf5 install --from-repo=repo` with no pkgspecs makes less sense than `dnf5 upgrade --from-repo=repo` with no pkgspecs. We can require the user to specify `'*'` in both cases as long as we document it well.

> check-upgrade: It's still unclear to me what exactly --from-repo will filter.

Yeah, I suppose this would be equivalent to `dnf5 check-upgrade --repo=repo`, so we can skip it.

> So, the functionality is already there in --check; it just needs glob support, for which I prepared fix.

Great.

Issue #2062, opened 2025-02-13
Option to set up mounts inside installroots


evan-goode:
Currently, RPM scriptlets will fail inside installroots if they depend on common files and devices that are not present under the installroot, including:

- "virtual" devices in /dev (e.g. /dev/null, /dev/zero, /dev/urandom)
- /proc
- /tmp and /run

DNF could provide an option to automatically create these common directories and mount certain host paths inside the installroot, like https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/build/dnf-installroot.

Inspired by discussion on https://matrix.to/#/#bootc:fedoraproject.org:

@cgwalters:

> dnf doesn't do any container features with --installroot and punts that to higher level tools which isn't wrong, but also forces everything to wrap it in different ways. I crafted https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/build/dnf-installroot which is a demonstration minimal version that works with as few as possible privileges inside a default podman/docker container

@supakeen:

> The above dance is pretty much what every build tool has to do having a single approach to that is probably nice; e.g. I'd prefer dnf to be able to do the above itself.
There can however be a bunch more to it to make-things-work if one also wants to run things inside that root.

PR #2270, opened 2025-05-28
WIP: Implement --with-mounts for "dnf5 --installroot=ABSOLUTE_PATH"

dcantrell:
I need more eyes on this.  I've been poking at it for long enough that I want to throw it out to the team.  This patchset is not quite ready for merging.  I've tried to make it clear what I'm doing and the functionality is implemented in base.cpp during setup.  Maybe that should move elsewhere?  I don't know how we want to structure that.

Also, the option name I picked is --with-mounts.  I don't care what it's called.  I also don't care if we want to do away with the option entirely and just make it default when you're working with --installroot.  Let's just agree as a team.

Test cases are missing right now and that's because I am getting wildly different behavior from mount(8) depending on how I try to test this.  In some instances when I am root and pass --with-mounts, mount(8) will still fail and tell me permission denied.  This is confusing to me.  At this point, more eyes on it is better.

Please review, comment, and so on.  I wanted to post this before I head to Flock, but I will be looking at it during that time.

### DIFF
--- a/libdnf5/utils/fs/utils.hpp
+++ b/libdnf5/utils/fs/utils.hpp
@@ -43,11 +43,6 @@ void move_recursive(const std::filesystem::path & src, const std::filesystem::pa
 [[nodiscard]] std::vector<std::filesystem::path> create_sorted_file_list(
     const std::vector<std::filesystem::path> & directories, std::string_view file_extension);
 
-// mkdirp(path, mode) - Implements 'mkdir -p' like you have at the
-//                      shell.  The mode given is applied at each
-//                      level.
-[[nodiscard]] int mkdirp(const std::string path, mode_t mode);
-
 }  // namespace libdnf5::utils::fs
 
 #endif  // LIBDNF5_UTILS_FS_UTILS_HPP
--- a/libdnf5/utils/fs/utils.cpp
+++ b/libdnf5/utils/fs/utils.cpp
@@ -117,44 +117,4 @@ void move_recursive(const std::filesystem::path & src, const std::filesystem::pa
     return paths;
 }
 
-//
-// mkdirp(path, mode) - Implements 'mkdir -p' like you have at the
-//                      shell.  The mode given is applied at each
-//                      level.
-//
-[[nodiscard]] int mkdirp(const std::string path, mode_t mode)
-{
-    bool first = true;
-    std::string subdir;
-    struct stat sb;
-
-    // handle all the parent directories
-    for (auto c : path) {
-        if (c == '/') {
-            // this just means we've seen the root directory
-            if (first) {
-                subdir += c;
-                first = false;
-                continue;
-            }
-
-            // we have a subdir, check to see if it exists
-            // if it does not exist, create it
-            if ((stat(subdir.c_str(), &sb) != 0) && (mkdir(subdir.c_str(), mode) == -1)) {
-                return -1;
-            }
-        }
-
-        subdir += c;
-    }
-
-    // final directory
-    if ((stat(path.c_str(), &sb) != 0) && (mkdir(path.c_str(), mode) == -1)) {
-        return -1;
-    }
-
-    return 0;
-}
-
-
 }  // namespace libdnf5::utils::fs


libdnf5/utils/fs/utils.hpp
```diff
@@ -43,6 +43,11 @@ void move_recursive(const std::filesystem::path & src, const std::filesystem::pa
 [[nodiscard]] std::vector<std::filesystem::path> create_sorted_file_list(
     const std::vector<std::filesystem::path> & directories, std::string_view file_extension);
 
+// mkdirp(path, mode) - Implements 'mkdir -p' like you have at the
+//                      shell.  The mode given is applied at each
+//                      level.
+[[nodiscard]] int mkdirp(const std::string path, mode_t mode);
```
kontura, line 49: I think this should not be necessary, in C++ there is https://en.cppreference.com/w/cpp/filesystem/create_directory.html which we already use in a bunch of places.

dnf5/main.cpp
```diff
@@ -503,6 +503,14 @@ void RootCommand::set_argument_parser() {
     installroot->link_value(&config.get_installroot_option());
     global_options_group->register_argument(installroot);
 
+    auto withmounts = parser.add_new_named_arg("with-mounts");
+    withmounts->set_long_name("with-mounts");
+    withmounts->set_description(
+        _("mount pseudo filesystems in ABSOLUTE_PATH when using installroot"));
```
kontura, line 509: There is no built in way to handle the relationship that `--with-mounts` requires `--installroot=...` but we do have an `libdnf5::cli::ArgumentParserMissingDependentArgumentError` which I think it would be good to throw when the argument is used without installroot.
We already have options that do something similar.

Conan-Kudo:
I would be nervous about making this default, since it could screw up various rootfs/container builds, but I agree this is a useful option to have. I will take a deeper look into this next week, and probably @dcantrell up to discuss it more.

Conan-Kudo:
FYI @schaefi @daandemeyer, you may want to review this.

evan-goode:
This is for https://github.com/rpm-software-management/dnf5/issues/2062.

evan-goode:
I pushed a commit to fix a few things, maybe this fixes some of the mysterious mount(8) behavior?

I agree with previous commenters, this should not be the default behavior. Maybe `--installroot-mounts` would be better to convey that it only applies to non-/ installroots?

The mounts should be cleaned up when DNF5 exits. IMO the best way is to create a private mount namespace with `unshare` like https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/build/dnf-installroot does. We do this in DNF4 already to privately mount the OSTree transient overlay. Generally speaking, this should be OK to do globally unless DNF5 ever wants to make persisting changes to the mount table (probably never?).

> the functionality is implemented in base.cpp during setup. Maybe that should move elsewhere? I don't know how we want to structure that.

I think it's appropriate to do this in `Base::setup`. We could encapsulate it in e.g. `Base::Impl::setup_installroot_mounts` and call that from `Base::setup`. The soonest we touch the installroot is currently right after `base.setup()` is called in `main`, in `libdnf5::create_rotating_file_logger`.

libdnf5/base/base.cpp
```diff
@@ -214,6 +228,105 @@ void Base::setup() {
...
+        }
+
+        if (close(fd) < 0) {
+            warn("close");
+        }
+
+        // bind mount /proc
+        tpath = installroot_path / "proc";
+
+        if (libdnf5::utils::proc::call(PATH_TO_MOUNT, {"--bind", "/proc", tpath.c_str()})) {
```
evan-goode, line 284: Each of the calls to `mount` will want `--options-source=disable` and I think `--no-mtab`.

ppisar:
This is a slippery slope: At the beginning there was procfs only. Then came sysfs, usbfs, then devpts. Then came devtmpfs and devpts left. Then appeared efivarfs and cgroupfs. I want to say the set of file systems to mount will change and it probably depends on a distribution. Also the file systems sometimes need nondefault options, especially options related to SELinux in containers. Else time bind-mount are more suitable.

I worry this feature will be never-ending supply of this-does-not-work-for-me bug reports.

Clearly these file systems are only needed for executing scriptlets which instrospect the running system. That itself is questionable because a decision when you want it and when you don't want it is in hands of the user. (In the past we used --installroot as the decision marker, but now when people execute DNF from a container to manage host with --installroot, we are doomed.) I would rather keep this functionality out of scope of DNF.

evan-goode:
If tools like mkosi and KIWI aren't going to use this (for understandable reasons), then I generally agree with you @ppisar.

> Clearly these file systems are only needed for executing scriptlets which instrospect the running system. That itself is questionable because a decision when you want it and when you don't want it is in hands of the user. (In the past we used --installroot as the decision marker, but now when people execute DNF from a container to manage host with --installroot, we are doomed.) I would rather keep this functionality out of scope of DNF.

Yes, allowing scriptlets to detect properties of the host system is not a valid use case of this feature since we can't assume those properties would also apply to the installroot system. But certain things like reading from `/dev/zero` are benign and we may want to support them. The question is how many scriptlets actually rely on /dev, /proc, and others. Does anyone (@cgwalters?) have an example of a Fedora package that fails to install/remove correctly in an installroot without these mounts set up? Maybe authselect?

Another problem is making this feature visible to the user. If `--with-mounts` is off by default, the best we could do is print a suggestion if a scriptlet fails in a non-/ installroot saying "try again with `--with-mounts`" which is not the greatest UX.

libdnf5/utils/fs/utils.hpp
dcantrell: Fixed, pushing patch momentarily.

dnf5/main.cpp
dcantrell: Fixed, pushing patch momentarily.

libdnf5/base/base.cpp
```diff
@@ -38,14 +38,27 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 #include "libdnf5/conf/config_parser.hpp"
 #include "libdnf5/conf/const.hpp"
 #include "libdnf5/utils/bgettext/bgettext-mark-domain.h"
+#include "libdnf5/utils/bootc.hpp"
```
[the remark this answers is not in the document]
dcantrell: Yes, fixed.  Pushing patch momentarily.

libdnf5/base/base.cpp
```diff
@@ -214,6 +228,105 @@ void Base::setup() {
...
+
+            if (dir == "tmp") {
+                if (chmod(tpath.c_str(), S_ISVTX|S_IRWXU|S_IRWXG|S_IRWXO)) {
+                    libdnf_throw_assertion("chmod(2) failure: {}", strerror(errno));
+                }
+            }
+        }
+
+        // some things (e.g., authselect) rely on this mount
+        tpath = installroot_path / "run/ostree-booted";
```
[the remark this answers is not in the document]
dcantrell: I'm not sure here.  Would it exist on the host so we could check that way?

libdnf5/base/base.cpp
dcantrell: Fixed, pushing patch momentarily.

cgwalters:
@dcantrell Thank you for working on this though.

The question becomes though where the "simple dnf --installroot with container" logic lives...

PR #2282, opened 2025-06-05
automatic: email emitter: Do not use fractional seconds in e-mail Date headers

ppisar:
"email" emitter produced Date: header with fractional seconds:

    Date: Sat, 31 May 2025 10:28:45.162419022 +0000

That violates RFC 5322 and some SMTP implementations reject messages with this mistake.

The cause was in fmt::format("{%S}") which formats non-integral seconds as a fractional number
<https://fmt.dev/11.1/syntax/#chrono-format-specifications>.

This patch fixes it by using std::chrono::time_point with second precision before passing it to fmt::format().

### DIFF
--- a/dnf5-plugins/automatic_plugin/email_message.cpp
+++ b/dnf5-plugins/automatic_plugin/email_message.cpp
@@ -44,7 +44,8 @@ void EmailMessage::set_body(std::stringstream & body) {
 }
 
 std::string EmailMessage::str() {
-    const auto now = std::chrono::system_clock::now();
+    // RFC 5322 requires dates with integral seconds.
+    const auto now = std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
     std::string date = fmt::format("{:%a, %d %b %Y %H:%M:%S %z}", now);
 
     std::string to_str;
--- a/test/dnf5-plugins/CMakeLists.txt
+++ b/test/dnf5-plugins/CMakeLists.txt
@@ -1,1 +1,2 @@
+add_subdirectory(automatic_plugin)
 add_subdirectory(copr_plugin)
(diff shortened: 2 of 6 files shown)


dnf5-plugins/automatic_plugin/email_message.cpp
```diff
@@ -44,7 +44,8 @@ void EmailMessage::set_body(std::stringstream & body) {
 }
 
 std::string EmailMessage::str() {
-    const auto now = std::chrono::system_clock::now();
+    // RFC 5322 requires dates with integral seconds.
+    const auto now = std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
```
mandree, line 48: Works for me (tested only in isolation, not in dnf5 context). Thank you.

Test base - old code with `#if 1`, change that to `#if 0` to get new code.

```c++
// g++ main.cc -std=c++20 -O -lfmt -o main-1
#include <fmt/format.h>
#include <fmt/chrono.h>
#include <chrono>
#include <iostream>

std::string date() {
#if 1
    const auto now = std::chrono::system_clock::now();
#else
    // RFC 5322 requires dates with integral seconds.
    const auto now = std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
#endif
    std::string date = fmt::format("{:%a, %d %b %Y %H:%M:%S %z}", now);
    return date;
}

int main()
{
        std::clog << date() << "\n";
        return 0;
}
```

outputs,
with `#if 1`: `Thu, 05 Jun 2025 19:56:32.704055672 +0000`
with `#if 0`: `Thu, 05 Jun 2025 19:56:34 +0000`

m-blaha (approved):
Thank you, especially for adding a test!

PR #2283, opened 2025-06-09
Mark dnf5daemon-server-polkit as noarch

m-blaha:
Since it only provides a plain text file, there's no need to distribute
an arch specific package.


kontura:
Thanks, I believe the failing rawhide daemon tests are not related.

dcantrell (commented):
For completeness and posterity, I am adding my review.  :)

Issue #2273, opened 2025-05-29
automatic: command_email emitter fails to send an e-mail for multiple recipients


ppisar:
Based on <https://github.com/rpm-software-management/dnf/issues/2241>:

If /etc/dnf/automatic.conf has:

~~~~
[emitters]
emit_via = command_email
[command_email]
email_to = root,test
~~~~

a command incompatible with s-nail's implementation of mail command is executed:

~~~~
execve("/bin/sh", ["sh", "-c", "--", "mail -Ssendwait -s \"[fedora-43] dnf5-automatic: 49 new upgrades have been downloaded.\" -r \"root\" \"root test\""], ...)
~~~~

s-nail expects each recipient as a separate argument. mailx also supports these separate arguments.

DNF team discussed this incompatibility and concluded that it's a bug in DNF5 and expansion of `email_to` value in `command_format` will be changed to produce separate arguments.

PR #2287, opened 2025-06-10
automatic: Expand emit_via in command_email emitter to individual arguments

ppisar:
If /etc/dnf/automatic.conf has:

    [emitters]
    emit_via = command_email
    [command_email]
    email_to = root,test

a command incompatible with s-nail-14.9.25, a "mail" command implementation, was executed:

    execve("/bin/sh", ["sh", "-c", "--", "mail -Ssendwait -s "[fedora-43] dnf5-automatic: 49 new upgrades have been downloaded." -r "root" "root test""], ...)

The cause was that s-nail does not support multiple recipients in a single argument.

This patch changes how "{email_to}" expands in command_format formatting string. Now it expands into multiple, space separated arguments. The new syntax is also accepted by mailx tool.

This patch also amends a documentation to explain how to configure multiple recipients.

A test will be added to ci-dnf-stack repository.

Resolve: #2273
Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1685

### DIFF
--- a/dnf5-plugins/automatic_plugin/emitters.cpp
+++ b/dnf5-plugins/automatic_plugin/emitters.cpp
@@ -136,7 +136,7 @@ void EmitterCommandEmail::notify() {
         if (!email_to.empty()) {
             email_to += " ";
         }
-        email_to += email;
+        email_to += quote(email);
     }
     std::string subject = libdnf5::utils::sformat(
         _("[{}] dnf5-automatic: {}"), config_automatic.config_emitters.system_name.get_value(), short_message());
@@ -146,7 +146,7 @@ void EmitterCommandEmail::notify() {
         fmt::arg("body", quote(output_stream.str())),
         fmt::arg("subject", quote(subject)),
         fmt::arg("email_from", quote(email_from)),
-        fmt::arg("email_to", quote(email_to)));
+        fmt::arg("email_to", email_to));
 
     FILE * command_pipe = popen(command_string.c_str(), "w");
     if (command_pipe) {


doc/dnf5_plugins/automatic.8.rst
```diff
@@ -182,7 +182,7 @@ The command email emitter configuration. Variables usable in format string argum
 ``email_to``
     list, default: ``root``
 
-    List of recipients of the message.
+    Comma-separated list of recipients of the message.
```
jrohel, line 185: My impression is that dnf5's INI-style configuration files accept the comma `,`, space ` `, and newline `\n` as delimiters for `OptionStringList`.

jrohel: Dnf5's INI-style configuration files accept the comma `,`, space ` `, and newline `\n` as delimiters for `OptionStringList`.

My concern is that someone might, based on a "Comma-separated list" expectation, write something like: `email_to = First Last <name@domain.xx>, recipient2`.

ppisar: I simply described examples I saw in dnf5.conf manual because the list syntax was not documented anywhere.

If the rules are more complicated, I will rather document the syntax in dnf5.conf manual and leave the automatic manual unchanged.

Do you have an idea how escaping or quoting works in the parser?

ppisar: I removed the documentation from this pull request. I list syntax will be documented in a separate pull request once escaping is clarified.

jrohel: > Do you have an idea how escaping or quoting works in the parser?

I'm afraid that `email_to` uses the default `from_string` parser (in "conf/option_string_list.cpp"), which doesn't support escaping. In other words, it's currently not possible to enter an email in the format "Real Name <name@domain.xx>". Only `name@domain.xx` is allowed, and fortunately, this format doesn't contain spaces.

jrohel:
Thanks for PR.  I will merge it.

Issue #1609, opened 2024-08-05
Make system state *.toml files writing more robust


m-blaha:
If dnf5 runs out of disk space during a transaction, it is currently possible that the data written to the `/usr/lib/sysimage/libdnf5/*.toml` files are incomplete. To prevent this, we should consider writing the files to a temporary location first, and then move the complete files to the final destination.

See https://github.com/rpm-software-management/dnf5/issues/1550 for more context.

ppisar:
See #1753 for an analysis.

ppisar:
Downstream bug report <https://bugzilla.redhat.com/show_bug.cgi?id=2330306>.

PR #2289, opened 2025-06-11
Make system state files writing more robust

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/1609

### DIFF
--- a/libdnf5/system/state.cpp
+++ b/libdnf5/system/state.cpp
@@ -19,8 +19,10 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "system/state.hpp"
 
+#include "utils/fs/utils.hpp"
 #include "utils/string.hpp"
 
+#include "libdnf5/base/base.hpp"
 #include "libdnf5/common/exception.hpp"
 #include "libdnf5/utils/bgettext/bgettext-mark-domain.h"
 #include "libdnf5/utils/fs/file.hpp"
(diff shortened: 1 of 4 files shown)


libdnf5/system/state.cpp
```diff
@@ -566,22 +568,58 @@ static std::string toml_format(const toml::ordered_value & value) {
 #endif  // #ifdef TOML11_COMPAT
 
 
+static std::filesystem::path suffix_new(std::filesystem::path && p) {
```
m-blaha: Disclaimer: This comment expresses only my personal preference. Since your code is correct and working, feel free to keep it as it is and ignore this comment.

I feel that using rvalue references (here and in the following function) introduces unnecessary complexity and makes the code harder to understand, use, and maintain. How about passing the `p` parameters simply by value and letting the compiler handle optimizations, such as copy elision and return value optimization?

libdnf5/system/state.cpp
```diff
@@ -607,9 +645,48 @@ static T load_toml_data(const std::string & path, const std::string & key) {
 }
 
 
+// In a given directory find all files endig with ".new" suffix and return them
+static std::vector<std::filesystem::path> gather_suffix_new_files(const std::filesystem::path & directory) {
```
m-blaha, line 649: One more thing – all `std::filesystem::exists()`, `std::filesystem::directory_iterator()`, `std::filesystem::is_directory()`, and especially `is_regular_file()` can throw an exception (e.g., if the directory does not exist or permission is denied), which will result in a `StateLoadError` being thrown. In most cases, this is fine and expected, but, for example, the presence of any inaccessible file (not even one of the `*.new` or `*.toml` files) in the system state directory would unnecessarily prevent the system state from loading.

libdnf5/system/state.cpp
kontura: Right, perhaps I got a bit carried a way.
I have no problem with simplifying it, I will update it to pass by value. :+1:

libdnf5/system/state.cpp
kontura: > the presence of any inaccessible file (not even one of the *.new or *.toml files)

I might not understand it enough but how could that happen?

If a user has the privileges to read the system state directory (which is required) then they see all the contents (basically have access to all the information in https://en.cppreference.com/w/cpp/filesystem/directory_entry.html) no?

Although one similar case is when we have some `*.new` files (the previous system state update failed) and an unprivileged user tries to read it. It would fail on the remove or rename, I have added a log for that.

m-blaha: Yeah, sorry. Disregard this comment. I was thinking that when the directory contains an inaccessible file, then `entry.is_regular_file()` would throw, but that's not what's happening (experimentally verified). Having access to the system state directory seems to be enough to check the type of the entry.

PR #2291, opened 2025-06-12
repoclosure: skip rich dependencies

AdamWill:
Similar to autodetect_unsatisfied_installed_weak_dependencies (in fact I stole the code from there). repoclosure can give bogus failures for dependencies like (foo if bar). If nothing provides foo in the repo(s) under test, but also bar is not present, that should not be considered a repoclosure failure, but repoclosure reports it as an "unresolved dep". See:

https://github.com/rpm-software-management/dnf-plugins-core/issues/549

Ideally we should handle rich dependencies correctly, but I cannot see how to do that, so I'm sending this as an alternative that's at least less-bad than the current state.


evan-goode:
Thanks for the patch, I'm not convinced this is better than the status quo, however. It seems this would reduce false failures (due to e.g. `foo if bar` if bar missing) but would increase false successes (e.g. `foo if bar` if only foo missing).

Looks like @kontura may have a proper fix in mind: https://github.com/openSUSE/libsolv/pull/592

Issue #1837, opened 2024-11-05
Provide a way to list packages from specific repositories


birdie-github:
Which makes searching/grepping this list nearly impossible.

Using `dnf5-5.2.6.2-1.fc41.x86_64`.

Now it lists:
```
tmpwatch.x86_64                                 2.11-27.fc41                    fedora
tmt.noarch                                      1.38.0-1.fc41                   updates
```

It used to be:
```
tmpwatch.x86_64                                 2.11-27.fc41                    @fedora
tmt.noarch                                      1.38.0-1.fc41                   @updates
```

And now that I know how dnf list installed works, it should be `dnf list --installed`, there's an inconsistency:

```
xfconf.x86_64                                  4.18.3-3.fc41                      fedora
xfd.x86_64                                     1.1.2-2.fc34                       @commandline
```

So actually the AT symbol is still there for some "repos".

pkratoch:
I think the main problem is that there is currently no good way to list packages from a specific repository. Grepping using the '@' symbol is just a workaround, but an option that would filter the list would be better.

Btw. the command repository-packages that could be used to achieve that as well is not yet implemented: https://github.com/rpm-software-management/dnf5/issues/915

pkratoch:
I renamed the issue to better describe the use case rather than a specific implementation. Previously, it was "dnf list now omits the @ symbol for repos' names"

birdie-github:
Packages names never include the at symbol, so you could just reinstate it, please.

pkratoch:
> Packages names never include the at symbol, so you could just reinstate it, please.

I agree that package names never include the '@' symbol. I am only saying it's good to differentiate between what the use case is (in this case to find packages from a specific repository) and how to achieve the use case. There are multiple options of how to implement that and I would argue that adding the '@' symbol is not the best one. In any case, naming the issue based on the use case instead of a specific implementation seems better.

birdie-github:
> Instead of grepping for @, you can grep for the third column.

This is how I've been doing that already, thanks but the command syntax is just a bit more complicated than using the AT symbol.

Issue #381, opened 2023-03-16
repoquery: add filtering based on `From repository`.


Ricky-Tigg:
**v. 4.14.0** | Hello. Values reported by
- `dnf --installed repository-packages updates-testing list` 
- `dnf rq --installed --repo=updates-testing` 

while sent through pipe `wc -l`, are respectively **non-null** and **null**, despite they have similar goals. Since the former command executed as intended, the value of the latter command is expected not to differ from the value resulting from the former command.

ppisar:
`--repo` is a global option and is documented in dnf5(8) page.

Issue #915, opened 2023-09-25
Document `repository-packages` alternatives in `man dnf5-changes-from-dnf4`


j-mracek:
The command should be not implemented. Some sub-commands can be replaced by another syntax, som may be dropped without a replacement

Sub-commands:

- [x] #951
  - alias dnf5 list --repo <repoid> - document it
- [x] #952
  - alias dnf5 info --repo <repoid>  - document it
- [x] #953
- [x] #954
- [x] #955
- [x] #956
- [x] #957
- [x] #958
- [x] #959
- [x] #960
- [x] #961

Conan-Kudo:
We definitely need a way to force packages to be requested from a specific repository though. That is critical for enabling sticky vendors by default.

j-mracek:
Yes I agree.

evan-goode:
Some of the functionality of repository-packages could be replaced by a `--from-repo <repository ID>` flag. `--from-repo` would select packages from the given repository but (unlike e.g. `--disablerepo`) would still include packages from other repositories during dependency resolution. 

If we did it this way, DNF 4 commands with their DNF 5 equivalents might look like:

```
dnf4 repository-packages myrepo remove [<package-spec>...]
dnf5 remove --from-repo myrepo [<package-spec>...]

dnf4 repository-packages move-to [<package-spec>...]
dnf5 reinstall --from-repo myrepo [<package-spec>...]

dnf4 repository-packages upgrade [<package-spec>...]
dnf5 upgrade --from-repo myrepo [<package-spec>...]

dnf4 repository-packages install [<package-spec>...]
dnf5 install --from-repo myrepo [<package-spec>...]
```

Another possibility suggested in our dev meetings was to extend the pkgspec format to allow specifying a repository; for example `dnf5 install librepo@copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly`, but we would still need something like a `--from-repo` flag for cases where no pkgspecs are specified, as in `dnf repository-packages upgrade`. And IMO, changes to the pkgspec format are not as discoverable as adding a `--from-repo` flag.

As for `repository-packages reinstall`, `remove-or-reinstall`, `remove-or-distrosync`, and `reinstall-old`, I don't know how widely these are used or whether they are worth implementing. @jrohel could the [do command](https://github.com/rpm-software-management/dnf5/pull/2089) replace any of these assuming that the `--installed-from-repo` flag is added?

kontura:
[...]

PR #2296, opened 2025-06-16
Improve "--from-repo="; Introduce "--installed-from-repo=" support for repoquery/list/info/remove/reinstall/upgrade/downgrade/distro-sync/swap

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/2315
Closes: https://github.com/rpm-software-management/dnf5/issues/381
Closes: https://github.com/rpm-software-management/dnf5/issues/1837
Resolves a significant portion of https://github.com/rpm-software-management/dnf5/issues/915.

CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1705

libdnf5:
-  Goal::Impl::add_specs_to_goal: Apply to_repo_ids settings in actions UPGRADE_ALL, UPGRADE_ALL_MINIMAL, and DISTRO_SYNC_ALL
- Goal::Impl::add_install_to_goal: Fix to_repo_ids handling 
- rpm::PackageQuery: New filter `filter_from_repo_id` 
- Goal: Honor set_from_repo_ids in `add_remove_to_goal`, `add_reinstall_to_goal`, `add_up_down_distrosync_to_goal` 

dnf5:
- dnf5: "--from-repo=" option enables and validates source repositories 
- Introduce "--installed-from-repo=" support for repoquery/list/info/remove/reinstall/upgrade/downgrade/distro-sync/swap

### DIFF
--- a/dnf5/commands/upgrade/upgrade.cpp
+++ b/dnf5/commands/upgrade/upgrade.cpp
@@ -69,6 +69,7 @@ void UpgradeCommand::set_argument_parser() {
     allow_erasing = std::make_unique<AllowErasingOption>(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     create_allow_downgrade_options(*this);
+    create_installed_from_repo_option(*this, installed_from_repos, true);
     create_from_repo_option(*this, from_repos, true);
     create_destdir_option(*this);
     create_downloadonly_option(*this);
@@ -126,6 +127,7 @@ void UpgradeCommand::run() {
         }
     }
 
+    settings.set_from_repo_ids(installed_from_repos);
 
     auto advisories = advisory_query_from_cli_input(
         ctx.get_base(),
--- a/dnf5/commands/upgrade/upgrade.cpp
+++ b/dnf5/commands/upgrade/upgrade.cpp
@@ -126,7 +126,6 @@ void UpgradeCommand::run() {
         }
     }
 
-    settings.set_to_repo_ids(from_repos);
 
     auto advisories = advisory_query_from_cli_input(
         ctx.get_base(),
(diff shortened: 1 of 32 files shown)


dnf5/commands/upgrade/upgrade.cpp
```diff
@@ -146,6 +145,7 @@ void UpgradeCommand::run() {
     if (pkg_specs.empty()) {
         goal->add_rpm_upgrade(settings, minimal->get_value());
     } else {
+        settings.set_to_repo_ids(from_repos);  // Apply from_repos only if packages are specified
```
kontura: This is a small detail but personally I don't find these `Apply from_repos only if packages are specified` comments very helpful.
It mentions what is happening but I would like to know why.

libdnf5/rpm/package_query.cpp
```diff
@@ -1177,6 +1177,57 @@ void PackageQuery::filter_repo_id(const std::vector<std::string> & patterns, lib
...
+                        };
+                        break;
+                }
+            }
+        }
+    }
+
+    // TODO(jrohel): The optimization replaces the original query_result buffer. Is it OK?
+    // Or we need to use a slower version "*p_impl &= filter_result;"
+    p_impl->swap(filter_result);
```
kontura, line 1228: The only problem I could come up with is that I think it invalidates all iterators from this query.

Maybe that could be unexpected for some users? Especially since this is the only filter that does it.
Perhaps it would be better to be consistent and use the slower approach?

jrohel: @kontura
> The only problem I could come up with is that I think it invalidates all iterators from this query.

Good point.
The question is, do we want to guarantee that iterators remain valid even after applying filters? After all, filters change the query's content. We wouldn't want to run into occasional strange behavior if users start relying on such a guarantee.

> Maybe that could be unexpected for some users? Especially since this is the only filter that does it.

`filter_installed` (often used filter) did it for years.

kontura: I see, ok, lets keep it for now. We can revisit it in the future.

PR #2300, opened 2025-06-19
Docs: Introduce new format denoting package spec formats

kontura:
For: https://github.com/rpm-software-management/dnf5/issues/1629

Around half of the commands use `package-spec-NPFB` we might consider documenting it as the default meaning of `package-spec` and using just `package-spec`. However I am not sure if it is worth it I kind of prefer to be explicit.

If we could come up with some more self explanatory or better notation I am happy to update it.
I was thinking about something like: `package-spec.nevra.provides.file-provides.binaries` but it is too long.

Another possibility could be dropping `package-spec` and using directly: `NEVRA/PROVIDES/FILE-PROVIDES/BINARIES` but that is still quite long.

### DIFF
--- a/doc/commands/install.8.rst
+++ b/doc/commands/install.8.rst
@@ -38,6 +38,16 @@ and their dependencies are installed on the system.
 If the specified packages are already installed, DNF5 does not check their dependencies again and
 simply verifies that the packages themselves are present.
 
+When ``package-spec-NPFB`` to specify the exact version of the package is given, DNF will install the
+desired version, no matter which version of the package is already installed. The former version of
+the package will be removed in the case of non-installonly package.
+
+On the other hand if ``package-spec-NPFB`` specifies only a name and obsoletes are enabled, DNF also
+takes into account packages obsoleting that name.
+Note that this can lead to seemingly unexpected results if the version of an installed package is obsoleted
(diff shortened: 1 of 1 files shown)


doc/commands/install.8.rst
```diff
@@ -25,19 +25,29 @@
...
 simply verifies that the packages themselves are present.
 
+When ``package-spec-NPFB`` to specify the exact version of the package is given, DNF will install the
+desired version, no matter which version of the package is already installed. The former version of
+the package will be removed in the case of non-installonly package.
+
+On the other hand if ``package-spec-NPFB`` specifies only a name and obsoletes are enabled, DNF also
+takes into account packages obsoleting that name.
+Note that this can lead to seemingly unexpected results if a package has multiple versions and some older
+version is being obsoleted. It creates a split in upgrade-path and both ways are considered correct,
```
ppisar: Does the described phenomenon happens only if multiple version of the package are already installed (i.e. it is an install-only package)? If so, please then phrase is as _if a package has multiple versions installed_. Otherwise, I have a poor imagination and phrase is as _if a package has multiple versions available_.

kontura: It was indeed meant when multiple versions are **available** but I see now that it was confusing.
I tried to enhance the text.

ppisar (approved):

Issue #2250, opened 2025-05-18
Failing tests when packaging for Arch Linux


jelly:
I'm trying to update our package to 5.2.13.1 from 5.2.12.0 and `ctest` now errors out with three test failures. They also occur on 5.2.13.0

```
FAIL: test_missing_setup_goal_resolve (base.test_goal.TestGoal.test_missing_setup_goal_resolve)
----------------------------------------------------------------------
libdnf5.exception.UserAssertionError: /usr/src/debug/dnf5/dnf5-5.2.13.1/libdnf5/base/goal.cpp:3112: libdnf5::base::Transaction libdnf5::Goal::resolve(): API Assertion 'p_impl->base->is_initial
ized()' failed: Base instance was not fully initialized by Base::setup()

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/build/dnf5/src/dnf5-5.2.13.1/test/python3/libdnf5/base/test_goal.py", line 34, in test_missing_setup_goal_resolve
    with self.assertRaisesRegex(libdnf5.exception.UserAssertionError,
         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                '^libdnf5/base/goal.cpp:[0-9]+:.*libdnf5::Goal::resolve\\(\\):'
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                ' API Assertion \'p_impl->base->is_initialized\\(\\)\' failed:'
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                ' Base instance was not fully initialized by Base::setup\\(\\)'):
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: "^libdnf5/base/goal.cpp:[0-9]+:.*libdnf5::Goal::resolve\(\): API Assertion 'p_impl->base->is_initialized\(\)' failed: Base instance was not fully initialized by Base::setup\(\)
[...]

PR #2307, opened 2025-06-21
test: ease path assertions in tests

jelly:
Building dnf5 with Arch Linux's build tooling fails as an absolute path is recorded instead of an absolute path:

AssertionError: "^libdnf5/logger/global_logger.cpp:[0-9]+: libdnf5::GlobalLogger::GlobalLogger\(\): API Assertion 'librepo_logger == nullptr' failed: Only one GlobalLogger can exist at a time" does not match
"/usr/src/debug/dnf5/dnf5-5.2.13.0/libdnf5/logger/global_logger.cpp:33: libdnf5::GlobalLogger::GlobalLogger(): API Assertion 'librepo_logger == nullptr' failed: Only one GlobalLogger can exist at a time"

Closes: #2250


kontura:
LGTM

PR #2309, opened 2025-06-23
dnf5: builddep: Argument "--from-repo"

jrohel:
The `--from-repo` argument allows the user to run builddep on packages in the specified repositories. However, any dependency resolution takes into account packages from all allowed repositories.

Multiple repository ids can be specified, separated by commas, and globs can be used. Usage is similar to the `install` command.

**Important Note:** The `builddep` command resolves packages in two steps:
  1. It first resolves the package specification with the `with_src` attribute set to `false`. From these resolved packages, it extracts the names of their corresponding source packages.
  2. Then, it retrieves source packages that match the names identified in step 1.

Corresponding source packages might reside in different repositories than their binary counterparts (e.g., a binary package in "fedora" and its source in "fedora-source"). Therefore, to ensure reliable operation, you should specify both repositories, like this:
`--from-repo=fedora,fedora-source`.


dcantrell (approved):
I like this option.

evan-goode:
It would be nice if we could move `create_from_repo_option` to `shared_options.hpp`. Then it wouldn't need to be duplicated in the builddep plugin. `builddep.cpp` already uses `create_store_option` from `shared_options.hpp`.

jrohel:
Moving `create_from_repo_option` (and `create_installed_from_repo_option`) to `shared_options.hpp` will make the function part of the DNF5 API, allowing it to be used in DNF5 plugins, including the builddep plugin.
Initially, I intentionally kept the function out of the public API until its usage was verified. So, I can prepare a commit for the move.

jrohel:
@evan-goode
I've rebased the code and moved `create_from_repo_option` (and `create_installed_from_repo_option`) to the public API. I did this in a separate commit because the change requires adjusting include paths across many commands.

Issue #2317, opened 2025-06-25
regression: 5.2.14.0: D-Bus interface errors out


mcrha:
Running:
```
gdbus call --system --dest org.rpm.dnf.v0 --object-path /org/rpm/dnf/v0 --method org.rpm.dnf.v0.SessionManager.open_session {}
```

with `dnf5daemon-server-5.2.14.0-1.fc43` results in:

```
Error: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Failed to register object vtable (Invalid argument)
(According to introspection data, you need to pass 'a{sv}')
```

while calling the same command with `dnf5daemon-server-0:5.2.13.1-3.fc43.x86_64` works properly and returns:

```
(objectpath '/org/rpm/dnf/v0/4e121ffac8e4bd3b73b969d748f4082d',)
```

It seems like the problem is when no arguments are passed to the call, aka an empty array, which was not the problem in the previous versions.

Might it be related to the sdbus-cpp version 2? The [Fedora rawhide build](https://koji.fedoraproject.org/koji/buildinfo?buildID=2737778) uses `sdbus-cpp-0:2.1.0-2.fc43`. Like if the sdbus-cpp version 2 is more strict about the types.

mcrha:
Hmm, I just noticed the [dnf5-5.2.13.1-3.fc43](https://koji.fedoraproject.org/koji/buildinfo?buildID=2726131) is built against `sdbus-cpp-0:2.1.0-2.fc43`, thus the same sdbus-cpp, thus it might be something else, something in the dnf5 itself.

kontura:
I see, I also [though](https://github.com/rpm-software-management/dnf5/issues/2040#issuecomment-2934144465) the problem was with sdbus-cpp-2 (our rawhide CI has been failing with this for a while) but I have just tracked it down to https://github.com/rpm-software-management/dnf5/commit/5c8fbc99de9840fc208b44220721e507f2d13900 which seems to be the real cause.

PR #2318, opened 2025-06-25
dnfdaemon: Removed incorrect output parameter names

m-blaha:
The method does not return any output parameters.

Resolves: https://github.com/rpm-software-management/dnf5/pull/2317


ppisar (approved):
This change is good. Verified manually with the reproduced from the issue and by comparing with an XML specification.

Issue #2311, opened 2025-06-23
docs: readthedocs builds are failing


kontura:
https://app.readthedocs.org/api/v2/build/28605249.txt

It seems they have been failing since April.

PR #2319, opened 2025-06-25
readthedocs: add swig include to fix doc builds

kontura:
Since commit https://github.com/rpm-software-management/dnf5/commit/99303105878d98093c2b8fe4922ed2ad22883c94 the binding files include
`#include "bindings/libdnf5/exception.hpp"` when running swig to
generate docs we need to include a path from where it can be found.

The PR also includes couple auxiliary commits.

Closes: https://github.com/rpm-software-management/dnf5/issues/2311

I created the docs on my own readthedocs account to verify it works with the changes: https://app.readthedocs.org/projects/amatej-dnf5/builds/28636746/


m-blaha (approved):
Thans!

PR #2329, opened 2025-07-01
dnfdaemon-client: Call server methods asynchronously

m-blaha:
We need to call server methods without blocking the current thread. Otherwise, the dnf5daemon-client cannot handle signals reporting progress of repository/package downloading and RPM transactions. Unfortunately, the repository key import confirmation workflow also depends on signal handling.

Curiously, blocking calls used to work with sdbus-cpp v1, but with v2 any signal handler was blocked until the call finished.

Hopefully, thish should resolve dnfdaemon CI tests failures.


ppisar (approved):
Good. I tested locally that the transaction progress is reported again. Also the ci-dnf-stack tests for key import pass again.

PR #2330, opened 2025-07-01
actions: Replace `L10N_SSH_KEY` secret with RSM CI GitHub token

evan-goode:
This will reduce the number of secrets we have to manage. We already use the `RSM_CI_APP_ID` and `RSM_CI_APP_PRIVATE_KEY` in `do-release.yaml`.

### DIFF
--- a/.github/workflows/weblate-sync-pot.yml
+++ b/.github/workflows/weblate-sync-pot.yml
@@ -16,6 +16,15 @@ jobs:
       options: --user root
     timeout-minutes: 10
     steps:
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.RSM_CI_APP_ID }}
+          private-key: ${{ secrets.RSM_CI_APP_PRIVATE_KEY }}
+          permissions:
+            contents: "write"
+          repositories: "rpm-software-management/dnf5-l10n"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/weblate-sync-pot.yml
```diff
@@ -16,6 +16,12 @@ jobs:
       options: --user root
     timeout-minutes: 10
     steps:
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.RSM_CI_APP_ID }}
+          private-key: ${{ secrets.RSM_CI_APP_PRIVATE_KEY }}
+
```
m-blaha, line 27: Since the token is used to create a commit in the weblate repository (dnf5-l10n), shouldn't we create token with
```
permissions: '{"contents":"write"}'
```

m-blaha: Also we might need to specify the repository to grant access to. According to documentation if both owner and repositories are empty, access will be scoped to only the current repository.
```
repositories: '["rpm-software-management/dnf5-l10n"]'
```

evan-goode: Sorry for the delay, you're right. Fixed.

m-blaha (approved):
Thank you!

PR #2333, opened 2025-07-02
dnfdaemon: Make it possible to set 'interactive' option for some operations

mcrha:
Some of the operations can require root privileges, for which it can ask the user
through polkit. It's okay when the operation is interactive, aka issued
by a user, but when the operation is ran by a background service, then
the password prompt coming out of blue is wrong. This change allows to 
tell the daemon server whether it can ask for the root password or not.


mcrha:
It seems to work fine here, thus I'm marking this as ready for review.

mcrha:
After a little chat with @mblaha we decided to not change the current API, thus I added `..._with_options()` variants of the functions where it had been needed and was not available yet. A similar pattern can be found in various libraries, one example for all: `g_ptr_array_new_with_free_func()`.

I've been thinking of "marking" the counter part functions without the `_with_options()` suffix as deprecated, but I doubt it has any impact on the D-Bus API, thus it would be just a comment in the interface definition files. I added there `This is equivalent to call "cancel_with_options()" with empty options.` kind of comments instead and I declared the new `_with_options()` before the old functions, thus they are spotted before the old functions when reading the files.

Another option can be to have a session-only property whether it's interactive or not, but the resources the session may need, not only for its live time, but also when (re-) reading the repositories, would be unnecessarily too much.

PR #2334, opened 2025-07-03
config: Add 'uninstall' alias

fmuellner:
The asymmetry between "install"/"remove" can be confusing even
to long-time users[0].

Address this by aliasing "remove" to "uninstall".

[0] https://samthursfield.wordpress.com/2025/06/29/dnf-uninstall/


pkratoch:
I agree with Petr here. Let's not have excessive amount of aliases to maintain.

Issue #2015, opened 2025-01-20
Crash under dnf5daemon::DownloadCB::create_signal_download()


mcrha:
This is with git main at commit 50e84a8bd3256954f28ae79dc6b32a6b107b9f09. I tried to install a `gnucash` through the dnf5daemon's session and it crashed with the following backtrace:
```gdb
Thread 51 "dnf5daemon-serv" received signal SIGSEGV, Segmentation fault.

#0  0x00007ffff413a209 in message_append_basic.lto_priv () at /lib64/libsystemd.so.0
#1  0x00007ffff7f7c153 in sdbus::Message::operator<<(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () at /lib64/libsdbus-c++.so.1
#2  0x000000000046098d in dnf5daemon::DownloadCB::create_signal_download (this=this@entry=0x7fffe43bdef0, signal_name="download_end", user_data=user_data@entry=0x7fffb8455060)
    at .../dnf5/dnf5daemon-server/callbacks.cpp:52
#3  0x0000000000461564 in dnf5daemon::DownloadCB::end (this=0x7fffe43bdef0, user_cb_data=0x7fffb8455060, status=libdnf5::repo::DownloadCallbacks::TransferStatus::ERROR, msg=0x7ffff6b6706f "Interrupted")
    at .../dnf5/dnf5daemon-server/callbacks.cpp:86
#4  0x00007ffff646f637 in libdnf5::repo::PackageTarget::~PackageTarget (this=0x7fffb8011fa0, __in_chrg=<optimized out>) at .../dnf5/libdnf5/repo/package_downloader.cpp:61
#5  std::destroy_at<libdnf5::repo::PackageTarget> (__location=0x7fffb8011fa0) at /usr/include/c++/14/bits/stl_construct.h:88
#6  std::_Destroy<libdnf5::repo::PackageTarget> (__pointer=0x7fffb8011fa0) at /usr/include/c++/14/bits/stl_construct.h:149
#7  std::_Destroy_aux<false>::__destroy<libdnf5::repo::PackageTarget*> (__first=0x7fffb8011fa0, __last=0x7fffb80128a0) at /usr/include/c++/14/bits/stl_construct.h:163
#8  std::_Destroy<libdnf5::repo::PackageTarget*> (__first=<optimized out>, __last=0x7fffb80128a0) at /usr/include/c++/14/bits/stl_construct.h:196
[...]

PR #2336, opened 2025-07-03
dnfdaemon: Crash under dnf5daemon::DownloadCB::create_signal_download()

mcrha:
The compiler decides when the variables are freed and the end of the function and when the `user_data` is freed before the `downloader`, then the `downloader` uses just freed `user_data` in the callbacks.

Fixes https://github.com/rpm-software-management/dnf5/issues/2015

### DIFF
--- a/dnf5daemon-server/session.cpp
+++ b/dnf5daemon-server/session.cpp
@@ -325,20 +325,24 @@ bool Session::check_authorization(
 }
 
 void Session::download_transaction_packages() {
-    libdnf5::repo::PackageDownloader downloader(base->get_weak_ptr());
-
-    // container is owner of package callbacks user_data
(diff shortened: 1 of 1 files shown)


mcrha:
Yes, it fixes the crash according to my tests. I did double-check, just in case I'd do anything wrong before.

I guess it's because the call ended with an exception - the variables might be freed in different order than stated in the document in that case.

ppisar:
> I guess it's because the call ended with an exception - the variables might be freed in different order than stated in the document in that case.

You patch changed the order of the the two variables. That's probably why it fixed it. Nevertheless the nesting block is more explicit.

mcrha:
Hmm, it can be it. Either way, making it explicit this way (though weird on the first look), might be better.

PR #2339, opened 2025-07-04
swig: Add %thread directive for repo_sack.hpp for Python

pkratoch:
The purpose is to enable multithread support in python. With the %thread directive, the Python GIL is released before entering the methods from repo_sack.hpp and acquired when exiting them. In case the methods are called in a thread, another thread is not blocked by it. This is especially important in case of loading repositories which can take a lot of time.

I am not very sure with this solution, since the documentation of this feature is very limited and I am mostly relying on testing. That said, enabling the threads for RepoSack fixes an issue in Anaconda where the UI freezes when repositories are being loaded. I ran locally Anaconda kickstart tests as well as ci-dnf-stack and they pass.

### DIFF
--- a/bindings/libdnf5/repo.i
+++ b/bindings/libdnf5/repo.i
@@ -1,5 +1,6 @@
 #if defined(SWIGPYTHON)
-%module(package="libdnf5", directors="1") repo
+%module("threads"=1, package="libdnf5", directors="1") repo
+%nothread;
 #elif defined(SWIGPERL)
 %module "libdnf5::repo"
 #elif defined(SWIGRUBY)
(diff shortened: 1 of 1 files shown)


ppisar:
Do you mean this <https://swig.org/Doc4.3/Python.html#Python_multithreaded> documentation? I also don't know anything about. According to Swig sources, %thread and %nothread are aliased for %feature("nothread","0") and %feature("nothread").

In accordance with <https://sourceforge.net/p/swig/mailman/swig-user/thread/00FEC548-F626-4790-9373-24BB0769E81D%40eyesopen.com/>, I would excpect reenabling the locking after %include "libdnf5/repo/repo_sack.hpp" block in case we will later include another header file. But if I do it, e.g. placing the %nothread at then end of the file, unit tests will fail. As you have it is passes.

@jrohel, can you see something wrong in this patch? I'm keen to approve it.

pkratoch:
Yes, I would expect the same, but I didn't figure out why it behaves the way it does. I also tried to use the `%thread` and `%nothread` with individual methods, but that doesn't seem to have any effect. I am not happy about this, but the alternative would be to study directly the swig sources and I don't think it's worth the effort.

M4rtinK:
I did some testing from the Anaconda PoV - with this change in place the delays are no longer happening & I have not seen any regressions - both in brief manual testing and in our full package related kickstart test batch.

ppisar (approved):
Jarek seems to be occupied. Let's merge it.

Issue #1530, opened 2024-06-05
Segfault when accessing a group or environment in python


pkratoch:
Hi, I need help with this issue with comps in python when they are created in another function. This is the reproducer:

```
>>> import libdnf5
>>> base = libdnf5.base.Base()
>>> base.setup()
>>> base.get_repo_sack().load_repos(True)
>>> def get_env(base):
...     environments = libdnf5.comps.EnvironmentQuery(base)
...     return next(iter(environments))
... 
>>> get_env(base).get_environmentid()
Segmentation fault (core dumped)
```

This works:

```
>>> import libdnf5
>>> base = libdnf5.base.Base()
>>> base.setup()
>>> base.get_repo_sack().load_repos(True)
>>> environments = libdnf5.comps.EnvironmentQuery(base)
>>> env = next(iter(environments))
>>> env.get_environmentid()
```

I would think there is something wrong with the copy constructor / assignment operator, but there is no problem e.g. for advisories that have basically the same copy constructor.
Is there something obvious I am missing?

jan-kolarik:
Hmm, my first though would be that the environments list is destroyed when going out of the scope from the `get_env()` function. But if the same code works for AdvisoryQuery, we would need to dig deeper I guess ...

kontura:
> Hmm, my first thought would be that the environments list is destroyed when going out of the scope from the `get_env()` function. But if the same code works for AdvisoryQuery, we would need to dig deeper I guess ...

I think that is it. `AdvisoryQuery` works quite a bit differently (more like `PackageQuery`), it is just a `SolvMap` so it doesn't store the advisory objects.

On the other hand `EnvironmentQuery`, `GroupQuery` and `RepoQuery` ultimately inherit from `libdnf5::set` template which stores data in a `std::set<T>`. When it goes out of scope the set destroys its content.

The reproducer doesn't crash with `RepoQuery` because [it doesn't store the `Repo` objects in the query but rather a `RepoWeakPtr`](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/repo/repo_query.hpp#L44).

So to fix this I think the comps queries would have to also use a weak pointer [here](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/comps/environment/query.hpp#L34) instead of the actual objects.

j-mracek:
May be this is missing what we have for `packageset`
```
%rename(next) libdnf5::rpm::PackageSetIterator::operator++();
[...]

PR #2345, opened 2025-07-08
5.3.0.0: group/environment query uses weak pointers to data in comps sack

kontura:
For: https://github.com/rpm-software-management/dnf5/issues/1530

On several places the API (specifically I think in `Goal` and some parts of `libdnf-cli`) still accept a Group or an Environment by value. This is visible in the code as de-referencing (`*`) given group or environment when passing it as argument. I am debating whether to break the API on more places by changing it to accepting weak ptrs. In any case I would do that in a separate PR.

Technically it also drops the functionality of `libdnf5::sack::ExcludeFlags::USE_DISABLED_REPOSITORIES` which was probably working before (created https://github.com/rpm-software-management/dnf5/issues/2346 to investigate the flag).


evan-goode (commented):
The changes look good, thank you.

There is one ci-dnf-stack failure: it looks like the output order of `dnf5 group list` is different than expected. I don't see why, it looks like `comps_display_order_cmp` should work correctly with `GroupWeakPtr`. Maybe the `Group::get_order` is already wrong before reaching the sorting function? I didn't investigate completely.

For the 5.3.0.0 version bump, we will also want to add a `changes_dnf5.3.rst` like we did for 5.2.0.0. That can come later.

kontura:
> There is one ci-dnf-stack failure: it looks like the output order of `dnf5 group list` is different than expected. I don't see why, it looks like `comps_display_order_cmp` should work correctly with `GroupWeakPtr`. Maybe the `Group::get_order` is already wrong before reaching the sorting function? I didn't investigate completely.

I believe that `Group::get_order` works only if there is an order defined in the metadata. If there is none the order seems kind of random. I made a PR that adds the ordering for these tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1699

pkratoch:
Thank you for working on this!

I've run into a segfault with this reproducer:

```
import libdnf5

base = libdnf5.base.Base()
sack = base.get_repo_sack()
sack.create_repos_from_system_configuration()
base.setup()
sack.load_repos()

for env in sorted(libdnf5.comps.EnvironmentQuery(base), key=lambda x: int(x.get_order_int())):
    print(env.get_environmentid())
```
Not sure why yet, I'll try to look into it.

pkratoch:
Hm, well, it actually obviously should segfault without the patch, that's expected...
Without the sorted function, it can also be reproduced more simply:
```
envs = [env for env in libdnf5.comps.EnvironmentQuery(base)]
for env in envs:
    print(env.get_environmentid())
```
It's strange that it does print couple of the ids...

kontura:
This has been properly fixed in https://github.com/rpm-software-management/dnf5/pull/2351
I am closing this, for more explanation see https://github.com/rpm-software-management/dnf5/pull/2351#issuecomment-3112243158.

If it turns out we do actually want this I am keeping the changes in my branch.

PR #2349, opened 2025-07-10
Do not trim progress bar messages

m-blaha:
To properly fix the issue we will need to break the ABI. Taking this opportunity to also add pImpl to *ProgressBar classes.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1829


libdnf5-cli/progressbar/multi_progress_bar.cpp
```diff
@@ -139,7 +139,6 @@ std::ostream & operator<<(std::ostream & stream, MultiProgressBar & mbar) {
     text_buffer.str("");
     text_buffer.clear();
 
-    std::size_t last_num_of_lines_to_clear = mbar.p_impl->num_of_lines_to_clear;
     std::size_t num_of_lines_permanent = 0;
```
kontura: I think the `num_of_lines_permanent` is not used anymore.

kontura (commented):
I do like this simplification of refreshing the progress bars and I didn't notice any problems while testing it.
However I would like to note that this kind of goes back to the original implementation before https://github.com/rpm-software-management/dnf5/pull/1825 in the sense that it clears before writing. It does only one clear and one write so that is good but I am still little bit worried that it will reintroduce the flickering on some terminals, based on https://github.com/rpm-software-management/dnf5/pull/1825#issuecomment-2480519074.

include/libdnf5-cli/progressbar/download_progress_bar.hpp
```diff
@@ -58,14 +65,8 @@ class LIBDNF_CLI_API DownloadProgressBar : public ProgressBar {
...
 
-    // widgets
-    libdnf5::cli::progressbar::NumberWidget number_widget;
-    libdnf5::cli::progressbar::DescriptionWidget description_widget;
-    libdnf5::cli::progressbar::PercentWidget percent_widget;
-    libdnf5::cli::progressbar::ProgressWidget progress_widget;
-    libdnf5::cli::progressbar::SpeedWidget speed_widget;
-    libdnf5::cli::progressbar::SizeWidget size_widget;
-    libdnf5::cli::progressbar::TimeWidget time_widget;
+    class Impl;
```
ppisar: Missing LIBDNF_LOCAL visibility qualifier.

include/libdnf5-cli/progressbar/multi_progress_bar.hpp
```diff
@@ -67,14 +76,8 @@ class LIBDNF_CLI_API MultiProgressBar {
...
 private:
-    std::size_t total_bar_visible_limit{0};
-    std::vector<std::unique_ptr<ProgressBar>> bars_all;
-    std::vector<ProgressBar *> bars_todo;
-    std::vector<ProgressBar *> bars_done;
-    DownloadProgressBar total;
-    // Whether the last line was printed without a new line ending (such as an in progress bar)
-    bool line_printed{false};
-    std::size_t num_of_lines_to_clear{0};
+    class Impl;
```
ppisar: Missing LIBDNF_LOCAL visibility qualifier.

include/libdnf5-cli/progressbar/progress_bar.hpp
```diff
@@ -54,101 +56,75 @@ class LIBDNF_CLI_API ProgressBar {
...
-    bool auto_finish = true;
-
-    std::chrono::time_point<std::chrono::system_clock> begin = std::chrono::system_clock::from_time_t(0);
-    std::chrono::time_point<std::chrono::system_clock> end = std::chrono::system_clock::from_time_t(0);
-
-    // current (average) speed over the last second
-    std::chrono::time_point<std::chrono::system_clock> current_speed_window_start = std::chrono::system_clock::now();
-    int64_t current_speed = 0;
-    int64_t current_speed_window_ticks = 0;
+    class Impl;
```
ppisar: Missing LIBDNF_LOCAL visibility qualifier.

include/libdnf5-cli/progressbar/progress_bar.hpp
```diff
@@ -54,101 +56,75 @@ class LIBDNF_CLI_API ProgressBar {
...
+    void set_description(const std::string & value);
 
     // messages
-    void add_message(MessageType type, const std::string & message) { messages.emplace_back(type, message); }
+    void add_message(MessageType type, const std::string & message);
     /// remove the last message
     void pop_message();
-    const std::vector<Message> & get_messages() const noexcept { return messages; }
+    const std::vector<Message> & get_messages() const noexcept;
+    const std::string & get_message_prefix() const noexcept;
```
ppisar, line 100: get_message_prefix() is a new public method. Please increase version minor number.

libdnf5-cli/progressbar/progress_bar.cpp
```diff
@@ -20,36 +20,240 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        if (message.front() == '\n') {
+            ++msg_lines;
+            current_column = 0;
+            message.remove_prefix(1);
+            continue;
+        }
+
+        // calculate the display width of the character
+        wchar_t wc;
+        auto bytes_consumed = std::mbtowc(&wc, message.data(), message.size());
```
ppisar: std::mbtowc() is not thread-safe. I hope that libdnf5-cli is only used from dnf5 so it shouldn't be a problem for DNF5 now. But I recommend to use std::mbrtowc().

ppisar (commented):
Overall good. The code is good except for the small nitpicks I commented in-line.

It indeed fixes trimming long messages. I tested it on a repository with invalid primary.xml checksum.

The flickering is subjectively not worse than the old implementation.  A drawback would be a terminal which does not supoport the "Esc [ 0 J" sequence. But we already could have the same issue with other sequences.

(In my opinion, ultimate solution would be replacing the multiprogressbar with a single progressbar aggregating all downloads. But that's a different story.)

m-blaha:
All comments should be now resolved.

dnf5.spec
```diff
@@ -434,7 +434,7 @@ Requires:       libdnf5%{?_isa} = %{version}-%{release}
 Library for working with a terminal in a command-line package manager.
 
 %files -n libdnf5-cli -f libdnf5-cli.lang
-%{_libdir}/libdnf5-cli.so.2*
+%{_libdir}/libdnf5-cli.so.3*
```
ppisar, line 438: I see. The soname change is because of removing private widgets members from DownloadProgressBar class (they were references by in-line methods). Shouldn't we then bump major version number?

m-blaha: Yeah. All three *ProgressBar classes have their private members moved to pImpl, and removed inline methods.

PR #2351, opened 2025-07-14
Copy return value from SetConstIterator::operator*

kontura:
`SetConstIterator::operator*` returns a reference, this is visible in
cpp and the user behaves accordingly.

In Python the user expects automatic memory management, the object
returned from query iterator should be valid as long as there is a
reference to it.
To ensure this add swig typemaps that copy the returned object.
It works because none of the queries (RepoQeury, GroupQuery,
EnvironmentQuery) store the actual data. They just store some kind of
proxy (RepoWeakPtr or wrapped libsolv ids) so if we copy the RepoWeakPtr
it still points to the same Repo object.

The PackageQuery and AdvisoryQuery return by value even in cpp so they
don't have this problem.

I think this preserves the Python API.

Another approach would be to change it in [cpp](https://github.com/rpm-software-management/dnf5/blob/main/include/libdnf5/common/set.hpp#L41) while it might be easier to maintain (because we wouldn't have to add the typemaps for each instantiated `libdnf5::set` template) it would break API.


pkratoch:
Awesome, thank you for working on this!

So, if I understand it correctly, in case of Repos, it copies the RepoWeakPtr, but in case of Groups and Environments, it now copies the whole objects, but that doesn't really matter for them, because of basically how they are used?

And while I suppose it sill would be preferable to move the ownership of Groups and Environments to the CompsSack, doesn't this already actually resolve https://github.com/rpm-software-management/dnf5/issues/1530 ?

kontura:
> Awesome, thank you for working on this!
> 
> So, if I understand it correctly, in case of Repos, it copies the RepoWeakPtr, but in case of Groups and Environments, it now copies the whole objects, but that doesn't really matter for them, because of basically how they are used?

Yes, I think so.
Both `Group` and `Environment` are fully defined by the set of ids they contain. These ids are added only when the objects are created and they cannot be added from outside. There is no other operation which can change the object (it has a package/group cache but they can always be recomputed and will be always the same).
The way I look at it its not that different from the `RepoWeakPtr`, it also doesn't contain the real data (those are in the pool) and it is only a "pointer" (ids) to them.


> And while I suppose it sill would be preferable to move the ownership of Groups and Environments to the CompsSack, doesn't this already actually resolve #1530 ?

Yes this fully resolves #1530.

I am actually not so sure we should move the ownership of Group and Environment objects to `CompsSack` (personally I woudn't do it and I would just close https://github.com/rpm-software-management/dnf5/pull/2345). My thinking is that `CompsSack` already owns the data (in the comps pool). Changing the queries to use GroupWeakPtr/EnvironmentWeakPtr is just an additional pointless indirection. Like I mentioned above the libsolv ids already are a kind of pointer to the real data.

PR #2353, opened 2025-07-18
dnfdaemon: Enable Download only transaction execution

m-blaha:
With the new "downloadonly" option passed to the "do_transaction"
method, packages are now only downloaded to the cache, but the RPM
transaction is not executed. No special permissions are required for
download-only transactions.

Might be useful for gnome software to only prepare an offline upgrade. The
second call to the "goal.do_transaction()" without the "downloadonly" option
set should actually check for permissions and schedule the offline upgrade for
next reboot.


mcrha:
I agree this will help gnome-software (and other similar apps). The polkit rules might not always be enough, because the background update is tried to be prepared for any user. The https://github.com/rpm-software-management/dnf5/pull/2333 will help to not show the password prompt when it's a background operation, but being able to download the messages any time (by any user) is a good thing. The actual installation can be done either by the root, or the user will enter the password. It may also work for different sessions, as long as the downloaded packages will not be deleted prematurely. It'll save the download time, due to cached downloaded packages, which is good.

dnf5daemon-server/session.cpp
```diff
@@ -368,18 +376,10 @@ void Session::store_transaction_offline() {
...
     constexpr const char * comps_in_trans_dir{"./comps"};
     const auto & comps_location = offline_datadir / comps_in_trans_dir;
 
-    const std::filesystem::path state_path{offline_datadir / libdnf5::offline::TRANSACTION_STATE_FILENAME};
-    libdnf5::offline::OfflineTransactionState state{state_path};
-
-    auto & state_data = state.get_data();
-
     state_data.set_status(libdnf5::offline::STATUS_DOWNLOAD_INCOMPLETE);
     state.write();
```
pkratoch: Why is the state written twice? Did I miss something being changed in it?

dnf5daemon-server/session.hpp
```diff
@@ -91,7 +91,7 @@ class Session {
     /// download packages for the current transaction
     void download_transaction_packages();
     /// prepare the current transaction to run during the next reboot
-    void store_transaction_offline();
+    void store_transaction_offline(bool downloadonly);
```
pkratoch, line 94: Is a version bump needed if the `Session::store_transaction_offline` changed what arguments it takes? Just asking, because, I don't know what makes up the API for dnf5daemon.

pkratoch (commented):
Thanks for the work.

See the two comments. Also, the commit message summary is hard to understand - can you please change "Download only" to "download-only"?

dnf5daemon-server/session.hpp
m-blaha: Strictly it's not - the dnf5daemon-server does not provide API other than that on D-Bus. And D-Bus interface remains the same - the `do_transaction` method accepts a dictionary `options` which current PR enhances to support another key. But it might be useful for documentation to reference since which version of dnf5daemon  the `downloadonly` option is supported.

dnf5daemon-server/session.cpp
m-blaha: Good catch! The intention was to move setting `STATUS_DOWNLOAD_INCOMPLETE` before the download is initiated. In reality I only copied the lines. Will fix it.

PR #2363, opened 2025-07-24
Disable fastest_mirror callback for package downloading

kontura:
See the comment for more info.
For: https://bugzilla.redhat.com/show_bug.cgi?id=2381859

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1709

### DIFF
--- a/libdnf5/repo/package_downloader.cpp
+++ b/libdnf5/repo/package_downloader.cpp
@@ -174,8 +174,20 @@ void PackageDownloader::download() try {
             continue;
         }
 
+        // Disable fastest_mirror callbacks for package downloading for now
+        // TODO(amatej): There is a problem in the API, libdnf5::repo::DownloadCallbacks::fastest_mirror interface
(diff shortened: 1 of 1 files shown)


ppisar (approved):
It indeed fixes the crash.
I also wondered whether the fastest_mirror optin has any effect in DNF5 and was unable to find any.

kontura:
> I also wondered whether the fastest_mirror option has any effect in DNF5 and was unable to find any.

I think it actually works, it is passed to librepo we just don't show any feedback to the user.
Though I believe it ends up in the log (librepo logs its).

ppisar:
You are right. There are librepo entries in dnf5.log. Though according to the log, it uses second fastest mirror for me.

kontura:
> You are right. There are librepo entries in dnf5.log. Though according to the log, it uses second fastest mirror for me.

I think that is actually due to a new feature: https://github.com/rpm-software-management/librepo/pull/324

Issue #1579, opened 2024-07-11
RFE: dump dnf configuration options and variables


pkratoch:
In dnf4, there was a method `dump` for configuration that returned a string representing the values of all the configuration options and it was, therefore, possible to log all the configuration options e.g. before Anaconda was run. Similarly for all the variables. Is there a way to do that in dnf5?

ppisar:
~~~~
# dnf5 --help | grep dump
  --dump-main-config                     Print main configuration values to stdout
  --dump-repo-config=REPO_ID,...         Print repository configuration values to stdout. List option. Supports globs
  --dump-variables                       Print variable values to stdout
~~~~

ppisar:
I see. You mean C++ API. Not a command-line interface. I'm not sure we have it in DNF5? Would it be helpful? Cannot an application iterate over all items?

jan-kolarik:
C++ API is iterable, so it seems like an error in Python bindings for `libdnf5::Config::opt_binds()`.

pkratoch:
Related issue (but not quite the same): https://github.com/rpm-software-management/dnf5/issues/963

PR #2369, opened 2025-07-30
swig: Add iterator for config options for python

pkratoch:
Resolves: https://github.com/rpm-software-management/dnf5/issues/1579

Alternatively, it might be done by defining a class similar to `Iterator` in `common.i`, only for a map iterator, and then adding this iterator for `libdnf5::OptionBinds`, but I didn't find out any way how to create bindings for the template `std::map<std::string, libdnf5::OptionBinds::Item>`. Any suggestions are appreciated.

### DIFF
--- a/bindings/libdnf5/conf.i
+++ b/bindings/libdnf5/conf.i
@@ -151,5 +151,29 @@ create_config_option_attributes(ConfigMain)
 %}
 #endif
 
+// The following adds a config options iterator in Python.
+//
+%define add_config_iterator(ClassName)
+#if defined(SWIGPYTHON)
+%pythoncode %{
+import re
+
+def ClassName##__iter__(self):
+    options = {}
+    for attr in dir(ClassName):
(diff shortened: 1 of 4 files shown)


kontura:
> Alternatively, it might be done by defining a class similar to `Iterator` in `common.i`, only for a map iterator, and then adding this iterator for `libdnf5::OptionBinds`, but I didn't find out any way how to create bindings for the template `std::map<std::string, libdnf5::OptionBinds::Item>`. Any suggestions are appreciated.

I have put together a commit that does this: https://github.com/kontura/dnf5/commit/1152d8364330a18843635e563fa18058138db9c4 but it needs to create and export new `libdnf5::OptionBinds::Item` constructors. I like that this approach makes the bindings API same as the C++ API but the new public constructors are not great. I am not sure what is better.

What do you think?
Should we discuss this on a meeting or perhaps ask @jrohel?

pkratoch:
Closing this in favor of https://github.com/rpm-software-management/dnf5/pull/2384 .

Issue #1739, opened 2024-09-30
Document the migration process from dnf4 to dnf5


jan-kolarik:
We should document how the migration from dnf4 actually occurs: what parts are done automatically, if there are any manual steps users need to take to preserve existing configurations or system state, what precautions to take, what is not supported, etc. I see this as important, since I’ve already noticed several concerns regarding this both here and on other forums.

PR #2371, opened 2025-07-30
doc: Add a chapter about migrating to dnf5

dcantrell:
Try to retell the story of moving Fedora to dnf5 by default and what the migration process was like.  Note this is different from the chapter explaining the differences between dnf4 and dnf5.

Resolves: #1739
Resolves: SWM-4363

### DIFF
--- a/doc/migrating_to_dnf5.7.rst
+++ b/doc/migrating_to_dnf5.7.rst
@@ -1,0 +1,196 @@
+.. _migrating_ref-label:
+
+############################
+ Migrating from DNF to DNF5
+############################
+
+This chapter describes the changes that occurred on the system during
+the migration from `DNF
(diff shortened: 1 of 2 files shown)


doc/migrating_to_dnf5.7.rst
```diff
@@ -0,0 +1,196 @@
...
+than steps you can take to migrate a dnf4 system to dnf5.
+
+.. _introducing_microdnf_ref-label:
+
+Introducing microdnf
+====================
+
+DNF was implemented in Python which means that a DNF-based system
+needs to have the Python stack installed and required modules.  In
+most cases this is fine, but there are instances where people what an
```
pkratoch: instances where people what an -> instances where people want an

dcantrell: Thanks, fixed.

Issue #1001, opened 2023-11-07
Crash because of empty /var/cache/libdnf5/temporary_files.toml file?


praiskup:
I just tried my routine `dnf5 update --refresh`:

```
...
After this operation 66 KiB will be used (install 781 MiB, remove 781 MiB).
Is this ok [y/N]: y
Failed to download packages
 An error occurred when parsing the temporary files memory file at '/var/cache/libdnf5/temporary_files.toml': [error] toml::value::as_array(): bad_cast to array
 --> unknown file
   |
 1 | unknown line
   |  the actual type is empty
```

It was reproducible, and work-around was to remove the `temporary_files.toml` file.  DNF tooling:

```
dnf5                 x86_64       5.1.6-20231106004845.10.g8598828e.fc40  copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable                 1.3 MiB
dnf5-plugins         x86_64       5.1.6-20231106004845.10.g8598828e.fc40  copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable               578.9 KiB
libdnf5              x86_64       5.1.6-20231106004845.10.g8598828e.fc40  copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable                 2.5 MiB
libdnf5-cli          x86_64       5.1.6-20231106004845.10.g8598828e.fc40  copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable               530.1 KiB
```

j-mracek:
Do you now how the empty file was created?

j-mracek:
The issue is reproducible by creating the empty file using touch.

jan-kolarik:
Briefly looking into the code, I couldn't find a reason for the file to be empty. Nevertheless, I believe we should change the behavior to avoid throwing an exception when the memory file has an invalid format. Instead, we should simply ignore it and delete the file in such cases.

praiskup:
> Do you now how the empty file was created?

No :-( I don't even know what is it's purpose.

j-mracek:
I am also curious how it happened. The question is whether we should recover from the situation.
If `toml::format(toml::value({{FILE_PATHS_TOML_KEY, files}}))` can  return an empty string then it would be possible to experience the situation. I was unable to figure from code and from documentation whether it is possible.

What about to modify following line?
```
[...]

PR #2372, opened 2025-07-31
repo: Do not fail on empty temporary files toml

m-blaha:
Use toml::find_or to safely read file paths from temp files toml,
returning an empty vector if the "files" key is missing.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1001
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2330306

### DIFF
--- a/libdnf5/repo/temp_files_memory.cpp
+++ b/libdnf5/repo/temp_files_memory.cpp
@@ -46,7 +46,7 @@ std::vector<std::string> TempFilesMemory::get_files() const {
 
     try {
         auto toml_data = toml::parse(full_memory_path);
-        return toml::get<std::vector<std::string>>(toml_data[FILE_PATHS_TOML_KEY]);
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ __pycache__
 *.log
 *.tar.gz
 libdnf5/conf/config.h
+.vscode
```
dcantrell: Put the .gitignore change in a separate commit.

m-blaha: Thinking about it, it's not a project specific change, I'll move this ignore to my global .gitignore and remove the commit.

dcantrell: Sounds good.  I personally do not mind this kind of change in the project, but I like to have changes to .gitignore separate from changes that may be cherry picked to other branches or pulled out as patches by downstream projects.

dcantrell (approved):

Issue #1808, opened 2024-10-30
Package `python3-libdnf5-python-plugins-loader` doesn't contain a config file


kontura:
Since the config file is missing the plugin is not enabled when the package is installed.

The other libdnf5 plugins have it, for example: [here](https://github.com/rpm-software-management/dnf5/blob/main/libdnf5-plugins/actions/actions.conf) and [here](https://github.com/rpm-software-management/dnf5/blob/main/libdnf5-plugins/rhsm/rhsm.conf).

It is also not clear if the plugin works: https://github.com/rpm-software-management/dnf5/issues/1775

Issue #1775, opened 2024-10-15
libdnf5 plugins written in Python are broken


m-blaha:
The usage of Python libdnf5 plugins isn't very clear and is poorly documented, so there's a chance I may have made a mistake. Here’s what I tried to write a plugin in Python:

1. created `/etc/dnf/libdnf5-plugins/python_plugins_loader.conf` config file with this content to enable python_plugins_loader:
```
[main]
name = python_plugins_loader
enabled = 1
```

2. copied `libdnf5-plugins/python_plugins_loader/plugin.py` to a temporary directory
3. set environment variable `LIBDNF_PYTHON_PLUGIN_DIR` to this directory and tried to run dnf5:

```
❯ LIBDNF_PYTHON_PLUGIN_DIR=/home/mblaha/temp/xxx dnf5 repoquery acpi
Cannot load plugin "/etc/dnf/libdnf5-plugins/python_plugins_loader.conf": Cannot load plugin "/home/mblaha/temp/xxx/plugin.py": PyObject_CallObject(plugin_class_constructor, nullptr): TypeError("IPlugin.__init__() missing 1 required positional argument: 'data'")
Cannot load plugins: python_plugins_loader.conf
```
(to get a better error message I also needed to hack `libdnf5::Plugins::load_plugins()` method to not only log the plugin loading failure, but also print it to console. The thing is that the error occurs during base.setup() call, before the log file is created).

I think that the problem is somehow connected with the change introduced by commit 60338b986.

PR #2378, opened 2025-08-06
Update `python_plugins_loader` to match new API and require configs for python plugins

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/1808
Closes: https://github.com/rpm-software-management/dnf5/issues/1775

I have configured the `python_plugins_loader` to read configs from `python_plugins_loader.d` directory which is located in the same directory with the rest of plugin configs. For example:
```
libdnf5-plugins/
├─ python_plugins_loader.conf
├─ actions.conf
├─ python_plugins_loader.d/
│  ├─ python_plugin.conf
```

It also adds needed `add_plugin(...)` API to `libdnf5::Base`.
Very similar API existed previously and its removal was one of the reasons why the plugin broke.


libdnf5-plugins/python_plugins_loader/python_plugins_loader.cpp
```diff
@@ -232,35 +243,59 @@ void PythonPluginLoader::load_plugin_file(const fs::path & file_path) {
...
+    if (config_dir_path.empty()) {
+        throw std::runtime_error("PythonPluginLoader::load_plugins_from_dir() config_dir_path cannot be empty");
+    }
 
-    std::vector<fs::path> lib_names;
-    std::error_code ec;
-    for (auto & p : std::filesystem::directory_iterator(dir_path, ec)) {
-        if ((p.is_regular_file() || p.is_symlink()) && p.path().extension() == ".py") {
-            lib_names.emplace_back(p.path());
+    std::vector<std::filesystem::path> config_paths;
```
pkratoch: Is there a reason to replace the `fs` with `std::filesystem` here when there is `namespace fs = std::filesystem;` and `fs` is used elsewhere in this file?

include/libdnf5/base/base.hpp
```diff
@@ -134,6 +135,9 @@ class LIBDNF_API Base {
 
     libdnf5::BaseWeakPtr get_weak_ptr();
 
+    /// Returns true if the plugin is successfully added, false otherwise.
+    bool add_plugin(const std::string & config_file_path, libdnf5::plugin::IPlugin & iplugin_instance);
```
pkratoch: I think this should have a docstring.

kontura: Would something like this work?
```
/// @brief Add LIBDNF5 plugin instance that introduces additional logic into the library using hooks.
/// Returns true if the plugin is successfully added, false otherwise.
```

pkratoch: Sure. And there is also the `@return` keyword.

libdnf5-plugins/python_plugins_loader/python_plugins_loader.cpp
kontura: No reason, just a force of habit.
I have fixed it in a couple more places as well. :+1:

include/libdnf5/base/base.hpp
kontura: Fixed.
I have also added the missing parameter descriptions.

libdnf5-plugins/python_plugins_loader/python_plugins_loader.cpp
```diff
@@ -207,7 +207,7 @@ void PythonPluginLoader::load_plugin_file(const fs::path & file_path) {
     if (!plugin_instance) {
         fetch_python_error_to_exception("PyObject_CallObject(plugin_class_constructor, nullptr): ");
     }
-    PyObject * libdnf = PyDict_GetItemString(plugin_module_dict, "libdnf");
+    PyObject * libdnf = PyDict_GetItemString(plugin_module_dict, "libdnf5");
     if (!libdnf) {
         fetch_python_error_to_exception("PyDict_GetItemString(plugin_module_dict, \"libdnf\"): ");
```
jrohel: Update text of error message "libdnf" -> "libdnf5"

kontura: Updated and I have updated couple more places to use libdnf5.

jrohel:
I'm going through the code again and I've found a logic problem (in the order of evaluation). With C++ plugins, we first check if a plugin is enabled, and if so, it's loaded. However, the `python_plugins_loader` now has the reverse logic. It loads the Python plugin (which loads the Python interpreter, initializes the Python code like imports, and calls the plugin constructor). The plugin instance is then passed to `base.add_plugin`, where the check for whether the plugin is enabled happens. This is too late. I still want to review the code one more time to see if I've missed anything, but it looks like a change is needed. Additionally, `base.add_plugin` is an API.

kontura:
> I'm going through the code again and I've found a logic problem (in the order of evaluation). With C++ plugins, we first check if a plugin is enabled, and if so, it's loaded. However, the `python_plugins_loader` now has the reverse logic. It loads the Python plugin (which loads the Python interpreter, initializes the Python code like imports, and calls the plugin constructor). The plugin instance is then passed to `base.add_plugin`, where the check for whether the plugin is enabled happens. This is too late. I still want to review the code one more time to see if I've missed anything, but it looks like a change is needed. Additionally, `base.add_plugin` is an API.

That is a very good point.
I tried to fix the issue but I needed to add additional new API.
In total this now adds two new functions to `Base`:
```cpp
    std::tuple<std::string, libdnf5::ConfigParser, bool> load_plugin_config(const std::string & config_file_path);

    void add_plugin(const std::string & plugin_name, libdnf5::ConfigParser & plugin_config, libdnf5::plugin::IPlugin & iplugin_instance);
```

It has the extra benefit that the plugin config is parsed only once but we do need to pass it around quite a bit, it goes from `libdnf5::Base` -> `python_plugins_loader` -> `libdnf5::Base` again.

include/libdnf5/base/base.hpp
```diff
@@ -134,6 +135,20 @@ class LIBDNF_API Base {
...
+    /// @brief Load libdnf5 plugin config, extract name of the plugin and check if it is enabled
+    /// @param config_file_path Path to a plugin config
+    /// @return a tuple with plugin name, parsed config and a bool whether the plugin is enabled
+    std::tuple<std::string, libdnf5::ConfigParser, bool> load_plugin_config(const std::string & config_file_path);
+
+    /// @brief Add libdnf5 plugin instance that introduces additional logic into the library using hooks.
+    /// @param plugin_name Name of the new plugin
+    /// @param plugin_config Parsed config of the new plugin
+    /// @param iplugin_instance New libdnf5 plugin instance
+    void add_plugin(
```
jrohel, line 147: I suggest changing the API to this: `void add_plugin(std::string plugin_name, libdnf5::ConfigParser && plugin_config, libdnf5::plugin::IPlugin & iplugin_instance);`

 Specifically, pass plugin_config using `&&`. Inside the method, a `move` is performed. The call in `python_plugins_loader` will also need to be updated to work with the revised API and to handle memory correctly.

libdnf5-plugins/python_plugins_loader/python_plugins_loader.cpp
```diff
@@ -232,37 +245,51 @@ void PythonPluginLoader::load_plugin_file(const fs::path & file_path) {
         fetch_python_error_to_exception("PyDict_CallMethod(base_class, \"get_locked_base\", NULL): ");
     }
     UniquePtrPyObject add_plugin_string(PyUnicode_FromString("add_plugin"));
-    PyObject_CallMethodObjArgs(locked_base.get(), add_plugin_string.get(), plugin_instance, NULL);
+    UniquePtrPyObject plugin_name_string(PyUnicode_FromString(plugin_name.c_str()));
+    type = SWIG_TypeQuery("libdnf5::ConfigParser *");
+    PyObject * py_config_parser = SWIG_NewPointerObj(plugin_config, type, 0);
```
jrohel: Update the code to release the `plugin_config` Python wrapper after use and to work with the revised API using `&&` (move semantics).
```
    UniquePtrPyObject config_parser(
        SWIG_NewPointerObj(new libdnf5::ConfigParser(std::move(*plugin_config)), type, SWIG_POINTER_OWN));
    if (!config_parser) {
        fetch_python_error_to_exception("SWIG_NewPointerObj(plugin_config, type, SWIG_POINTER_OWN)");
    }
    PyObject_CallMethodObjArgs(
        locked_base.get(),
        add_plugin_string.get(),
        plugin_name_string.get(),
        config_parser.get(),
        plugin_instance,
        NULL);
```

jrohel:
@kontura
Thanks for the updates. I've also left some recommendations for changes in the API and `python_plugins_loader`. Once those are applied, I'll merge this.

I did a deeper review and found a couple of other issues, mainly with memory leaks and incomplete Python plugin management. However, I'm working on a solution and will submit it as a new PR that builds on top of this one.

PR #2382, opened 2025-08-08
Encode package location before downloading

m-blaha:
librepo expects that the download locations are properly URL-encoded.
Unfortunately, in repository metadata createrepo_c leaves the URL in the location unencoded, so
we need to encode it before passing it to librepo.

This PR copies dnf4 behavior.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2325962

### DIFF
--- a/common/utils/url.cpp
+++ b/common/utils/url.cpp
@@ -39,7 +39,10 @@ bool is_url(std::string path) {
 }
 
 std::string url_encode(const std::string & src) {
-    auto no_encode = [](char ch) { return std::isalnum(ch) != 0 || ch == '-' || ch == '.' || ch == '_' || ch == '~'; };
+    auto no_encode = [](char ch) {
+        return (ch >= '0' && ch <= '9') || (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') || ch == '-' ||
+               ch == '.' || ch == '_' || ch == '~';
+    };
 
     // compute length of encoded string
     auto len = src.length();
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -177,7 +177,7 @@ std::pair<std::string, size_t> Vars::substitute_expression(std::string_view text
 
             // Find the end of the variable name
             auto it = std::find_if_not(res.begin() + static_cast<long>(pos_variable), res.end(), [](char c) {
-                return std::isalnum(c) != 0 || c == '_';
+                return (c >= '0' && c <= '9') || (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || c == '_';
             });
             auto pos_after_variable = static_cast<size_t>(std::distance(res.begin(), it));
 


common/utils/url.cpp
```diff
@@ -34,4 +38,75 @@ bool is_url(std::string path) {
            path.starts_with("https://");
 }
 
+std::string url_encode(const std::string & src) {
+    auto no_encode = [](char ch) { return std::isalnum(ch) != 0 || ch == '-' || ch == '.' || ch == '_' || ch == '~'; };
```
evan-goode: `std::isalnum` is locale-dependent, but I think you could do:

```
std::locale c_locale("C");
std::isalnum(ch, c_locale);
```

https://en.cppreference.com/w/cpp/locale/isalnum.html

evan-goode: Oh, just realized you were only moving this around :) Still, we might as well fix it.

libdnf5/repo/package_downloader.cpp
```diff
@@ -186,9 +187,18 @@ void PackageDownloader::download() try {
         cached_handle.set_opt(LRO_FASTESTMIRRORCB, NULL);
         cached_handle.set_opt(LRO_FASTESTMIRRORDATA, NULL);
 
+        std::string encoded_location = pkg_target.package.get_location();
+        if (encoded_location.find("://") == std::string::npos) {
+            // TODO(mblaha): Here we rely on the fact that createrepo_c does not
```
evan-goode, line 192: This does really seem like a `createrepo_c` flaw... but I don't see a way to change it without adding a new repo config option as you mentioned. So I'm fine with making this change so we are consistent with DNF4.

common/utils/url.cpp
m-blaha: That's a good catch! I'll add a commit fixing it.

evan-goode (changes requested):
Now we have

```
/builddir/build/BUILD/dnf5-5.2.16.0-build/dnf5-5.2.16.0/common/utils/url.cpp: In function ‘std::string libdnf5::utils::url::url_encode(const std::string&)’:
/builddir/build/BUILD/dnf5-5.2.16.0-build/dnf5-5.2.16.0/common/utils/url.cpp:50:24: error: conversion to ‘unsigned char’ from ‘char’ may change the sign of the result [-Werror=sign-conversion]
   50 |         if (!no_encode(ch)) {
      |                        ^~
/builddir/build/BUILD/dnf5-5.2.16.0-build/dnf5-5.2.16.0/common/utils/url.cpp:59:23: error: conversion to ‘unsigned char’ from ‘char’ may change the sign of the result [-Werror=sign-conversion]
   59 |         if (no_encode(ch)) {
      |           
```

m-blaha:
Sorry, I should have caught it before submitting.

test/libdnf5/utils/test_url.cpp
```diff
@@ -0,0 +1,198 @@
+/*
+Copyright Contributors to the libdnf project.
+
+This file is part of libdnf: https://github.com/rpm-software-management/libdnf/
```
ppisar: Is this file really copied from libdnf, or is it a new file? In the latter case please refer to DNF5.

m-blaha: It's a new file. I keep forgetting to change it.

m-blaha: Fixed.

evan-goode (approved):
Thanks!

PR #2383, opened 2025-08-08
Add --destdir option support to all commands with --downloadonly

m-blaha:
Extends `--destdir` option support from `upgrade` and `download` commands to all other commands that support `--downloadonly` option. This provides consistent behavior across all package management operations.

When `--destdir` is specified with transaction command, `--downloadonly` is automatically enabled to ensure packages are downloaded to the specified directory instead of being installed. This maintains consistency with the existing `upgrade` command behavior and also with dnf4 behavior.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/doc/commands/group.8.rst
+++ b/doc/commands/group.8.rst
@@ -133,6 +133,10 @@ Options for ``install``, ``remove`` and ``upgrade``
     | Disable downgrade of dependencies when resolving the requested operation.
     | Used with ``install`` and ``upgrade`` commands.
 
+``--destdir=<path>``
+    | Set directory used for downloading packages to. Default location is to the repository cache directory. Automatically sets the --downloadonly option.
+    | Used with ``install`` and ``upgrade`` commands.
+
 ``--downloadonly``
     | Download the resolved package set without executing an RPM transaction.
     | Used with ``install`` and ``upgrade`` commands.
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -191,7 +191,7 @@ repository configuration file should aside from repo ID consists of baseurl, met
 ``destdir``
     :ref:`string <string-label>`
 
-    Redirect downloaded packages to provided directory.
+    Redirect downloaded packages to provided directory. If set, packages downloaded to this directory are not automatically removed after the next successful transaction. To redirect the download location while allowing DNF5 to manage downloaded packages, use the :ref:`cachedir <cachedir_options-label>` option.
 
     Default: <package repository :ref:`cachedir <cachedir_options-label>`>/packages
 
--- a/dnf5/commands/distro-sync/distro-sync.cpp
+++ b/dnf5/commands/distro-sync/distro-sync.cpp
@@ -57,6 +57,7 @@ void DistroSyncCommand::set_argument_parser() {
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     create_installed_from_repo_option(*this, installed_from_repos, true);
     create_from_repo_option(*this, from_repos, true);
+    create_destdir_option(*this);
     create_downloadonly_option(*this);
     create_offline_option(*this);
     create_store_option(*this);
@@ -66,6 +67,10 @@ void DistroSyncCommand::configure() {
     auto & context = get_context();
     context.set_load_system_repo(true);
     context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
+
+    if (!context.get_base().get_config().get_destdir_option().empty()) {
+        context.get_base().get_config().get_downloadonly_option().set(true);
+    }
 }
 
 void DistroSyncCommand::run() {
--- a/dnf5/commands/do/do.cpp
+++ b/dnf5/commands/do/do.cpp
@@ -173,6 +173,7 @@ void DoCommand::set_argument_parser() {
     advisory_enhancement = std::make_unique<EnhancementOption>(*this);
     advisory_newpackage = std::make_unique<NewpackageOption>(*this);
 
+    create_destdir_option(*this);
     create_downloadonly_option(*this);
     create_offline_option(*this);
     create_store_option(*this);
@@ -207,6 +208,10 @@ void DoCommand::configure() {
 
     context.set_load_available_repos(
         in_pkgs_count > 0 ? Context::LoadAvailableRepos::ENABLED : Context::LoadAvailableRepos::NONE);
+
+    if (!context.get_base().get_config().get_destdir_option().empty()) {
+        context.get_base().get_config().get_downloadonly_option().set(true);
+    }
 }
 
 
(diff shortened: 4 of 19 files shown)


m-blaha:
Oh, thanks for pointing that out! I'll look into that.

m-blaha:
There is yet another use case: `--offline`. Although offline transactions may not be applicable if another transaction occurs between the download and reboot, the removal of downloaded packages is unexpected. As you mentioned, all these use cases signalize that not removing transaction packages when `--destir` is used might be solution.

kontura:
Regarding the implementation I don't like that all of the `TempFilesMemory` logic lives in `PackageDownloader`. In my opinion it should be in `libdnf5::base::Transaction`. Especially since the other users of `PackageDownloader` have to explicitly turn it off by `PackageDownloader::force_keep_packages()`. The method feels like a hack to me.

I suppose its too late to change that now and we already expose `PackageDownloader::force_keep_packages()` so please ignore my rant.

m-blaha:
I'll think about it, thanks for feedback.

m-blaha:
we can handle --destdir inside the Transaction::download() and call downloader.force_keep_packages(true) there. That way all PackageDownloader users will use it consistently.

m-blaha:
> I like it, I would only document the behavior for the `destdir` config option. Perhaps the `destdir` docs should also mention that if the user wants dnf to keep managing the packages they should rather change the `cachedir` option.
> 
> If I understand correctly we will loose the capability to redirect **only** the packages from cache and have them still managed by dnf but I am personally fine with that. Probably we should also document it in the `changes_from_dnf4` doc?

Added commit with docs enhancement.

I'm not sure the changes_from_dnf4 doc is needed - you are right that dnf4 removes packages downloaded using `dnf4 install --downloadonly --destdir` after the next successful transaction, but it seems more like a bug than intentional behavior to me (although dnf4 says "The downloaded packages were saved in cache until the next successful transaction." in such situation).

doc/commands/group.8.rst
```diff
@@ -133,6 +133,10 @@ Options for ``install``, ``remove`` and ``upgrade``
     | Disable downgrade of dependencies when resolving the requested operation.
     | Used with ``install`` and ``upgrade`` commands.
 
+``--destdir=<path>``
+    | Set directory used for downloading packages to. Default location is to the repository cache directory. Automatically sets the --downloadonly option.
+    | Used with ``install`` and ``upgrade`` commands.
```
kontura, line 138: Since this is not using the shared `destdir.rst` its missing the `Packages downloaded to the destdir are not automatically removed after the next successful transaction.` part.

doc/dnf5.conf.5.rst
```diff
@@ -191,7 +191,7 @@ repository configuration file should aside from repo ID consists of baseurl, met
 ``destdir``
     :ref:`string <string-label>`
 
-    Redirect downloaded packages to provided directory.
+    Redirect downloaded packages to provided directory. If set, packages downloaded to this directory are not automatically removed after the next successful transaction. To redirect the download location while allowing DNF5 to manage downloaded packages, use the :ref:`cachedir <cachedir_options-label>` option.
```
kontura, line 194: I just realized that setting this option now also implies `downloadonly`.
I am not sure the config option now makes sense but I believe the docs should mention it.

m-blaha: Oh, right. This is ugly. So setting this config option in dnf.conf will turn all transaction operations to downloadonly. It already does for `dnf5 upgrade`. Definitely not intended...

m-blaha: I guess I need to re-think this "simple" change.

kontura: Indeed, though like you mentioned it already is a problem. 

Thinking aloud here.
The way I understand the user facing behavior in dnf4:
- Config option `destdir` is used to redirect where packages are downloaded.
- The behavior is split by https://github.com/rpm-software-management/dnf/pull/944: depending on if there is a transaction or not, dnf either manages (cleans up) the packages or doesn't.
- The `--destdir` option sets the config option. 
- Important use case is `dnf4 system-upgrade download --destdir ...` it downloads a lot of packages and users expect dnf will clean them up. (We will need something like this also for dnf5: https://github.com/rpm-software-management/dnf5/issues/2398.)
- For dnf4 `download`/`reposync`/`modulesync` packages downloaded with `--destdir` are not cleaned up (no transaction).
- But basically for all other commands when paired with `--downloadonly` (those have a transaction) it is unexpected that dnf will clean the packages after the next transaction.

Unfortunately dnf5 kind of inherited this bad dnf4 behavior.

In dnf5 I think we can still split the `system-upgrade` use-case to a different option. 

Then `destdir` can always mean packages are no longer managed by dnf (although this could break some usecases when `destdir` is set in the config file, but `dnf5 upgrade` already did?).
To be honest I would image almost no one sets `destdir` in a config file, maybe some specialized systems?

In my opinion ideally the `destdir` config option is removed and `--destdir` both implies `--downloadonly` and packages are not managed by dnf.

ppisar: I don't think DNF5 must follow DNF4. Especially to follow bad design.

The #2398 issue does not imply that the package storage for off-line upgrades needs to be configurable and especially not with a command-line option. Remember that the other DNF process running after a reboot needs to locate the same storage at the same place where the first process saved the packages before the reboot. I also warn against storing pointers to the storage to various temporary locations as every indirection increases a security risk that some of them could be tempered. Users needs to be confident that they get installed the updates they requested.

I agree with `--destdir` simplification @kontura proposes.

Though we should consider what kinds of upgrade we want to support. We have now:

* `dnf upgrade`
* `dnf upgrade --offline`
* `dnf upgrade --downloadonly`
* `dnf upgrade --destdir`

E.g. do we want to support a use case where you perform `dnf upgrade --destdir PATH` and then, possibly on another machine, do `dnf --setopt destdir=PATH upgrade` to apply the update as if it were `dnf upgrade` with preserving installed-from-repository history records? (I know it does not work now because --setopt destdir is equivalent to --destdir) Some kind of postponed, or air-gap update? Something like off-line update does but portable in time and space and not necessitating update on reboot? Something like package manifest will do but including the package payload?

kontura: > Some kind of postponed, or air-gap update? Something like off-line update does but portable in time and space and not necessitating update on reboot? Something like package manifest will do but including the package payload?

I think we have that, any command with a transaction can do `--store /path/where/to/store/the/transaction` and then (possibly on a different system): `$ dnf replay /path/to/stored/transaction`.

This also stores the rpm packages.
The offline update is a wrapper around this store-replay functionality.

ppisar: > I think we have that, any command with a transaction can do `--store /path/where/to/store/the/transaction` and then (possibly on a different system): `$ dnf replay /path/to/stored/transaction`.
> 
You are right. I missed it in the documentation.

PR #2384, opened 2025-08-11
Bindings: add iterator to `libdnf5::OptionBinds`

kontura:
This requires to instantiate `std::map<std::string, libdnf5::OptionBinds::Item>` template because `libdnf5::OptionBinds` uses its iterator.

To instantiate the template we need to add constructors to `libdnf5::OptionBinds::Item`. I think this is needed beacuse because `Item` contains an Impl ptr and swig doesn't know how to create that (because it is hidden).

Finally to wrap the `libdnf5::OptionBinds::Item` we need to flatten the class hierarchy.

This is an alternative approach to https://github.com/rpm-software-management/dnf5/pull/2369.

### DIFF
--- a/libdnf5/conf/option_binds.cpp
+++ b/libdnf5/conf/option_binds.cpp
@@ -58,6 +58,10 @@ OptionBinds::Item::Item(Option & option) : p_impl(new Impl(option)) {}
 
 OptionBinds::Item::~Item() = default;
 
+OptionBinds::Item::Item() = default;
+OptionBinds::Item::Item(const Item & src) = default;
+OptionBinds::Item & OptionBinds::Item::operator=(const Item & src) = default;
+
 Option::Priority OptionBinds::Item::get_priority() const {
     return p_impl->option->get_priority();
 }
--- a/bindings/libdnf5/conf.i
+++ b/bindings/libdnf5/conf.i
@@ -14,6 +14,12 @@
 %import "exception.i"
 %import "logger.i"
 
+// This is required because `libdnf5::OptionBinds::Item` is nested
+// inside `libdnf5::OptionBinds` class.
+// While swig claims it shoudn't be needed for Python it seems it is:
+// https://www.swig.org/Doc4.3/SWIGDocumentation.html#SWIGPlus_nested_classes
+%feature("flatnested");
+
 %{
     #include "bindings/libdnf5/exception.hpp"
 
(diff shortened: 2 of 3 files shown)


libdnf5/conf/option_binds.cpp
```diff
@@ -58,6 +58,10 @@ OptionBinds::Item::Item(Option & option) : p_impl(new Impl(option)) {}
 
 OptionBinds::Item::~Item() = default;
 
+OptionBinds::Item::Item() = default;
```
jrohel, line 61: I don't like this public constructor. It creates an invalid object. The rest of the code assumes that the `p_impl`  is always valid. I'm wondering if there's a clever way to solve this.

kontura: It would be great if we could avoid it.
I did find this swig issue: https://github.com/swig/swig/issues/2625 but reading through the associated PR it seems the fix doesn't work in some languages (Python).

jrohel: I tried using `%feature("valuewrapper")`, but it didn't solve the issue. I suspect the problem lies deeper, within the SWIG's source code. I found that SWIG has two implementations for this, and I can switch to the more efficient one by defining `#define SWIG_STD_PAIR_ASVAL` (in our "conf.i"). However, this led to a different compilation error, and I'm not confident it's the right way to proceed

jrohel: I've been investigating the situation. The root problem lies with the implementation of the wrapper for `std::pair` in SWIG, which requires a default constructor for the contained elements. Without changing the C++ API, we can circumvent this requirement by adding auxiliary code to the .i file so that the wrapped `std::pair` contains a **pointer** to `OptionBinds::Item` instead of `OptionBinds::Item` itself. I've included an example solution below. The question is whether we should proceed with this approach, or if we should accept the compromise and implement the default constructor.

Example solution:
```
%inline %{
class OptionBindsIterator {
public:
    OptionBindsIterator(std::map<std::string, libdnf5::OptionBinds::Item>::iterator it) : it(it) {}
    void next() { ++it; }
    std::pair<const std::string, libdnf5::OptionBinds::Item *> value() {
        auto & val = *it;
        return {val.first, &val.second};
    }
    bool operator==(const OptionBindsIterator & other) { return it == other.it; }

private:
    std::map<std::string, libdnf5::OptionBinds::Item>::iterator it;
};
%}

%extend libdnf5::OptionBinds {
    OptionBindsIterator begin() { return {$self->begin()}; }
    OptionBindsIterator end() { return {$self->end()}; }
    OptionBindsIterator find(const std::string & id) { return {$self->find(id)}; }
};
%template(PairConstStringItemPtr) std::pair<const std::string, libdnf5::OptionBinds::Item *>;
%ignore libdnf5::OptionBinds::add;
%ignore libdnf5::OptionBinds::begin;
%ignore libdnf5::OptionBinds::cbegin;
%ignore libdnf5::OptionBinds::end;
%ignore libdnf5::OptionBinds::cend;
%ignore libdnf5::OptionBinds::find;
%include "libdnf5/conf/option_binds.hpp"

add_iterator(OptionBinds)
```

kontura: That is very interesting, thank you for further looking into this!

I have created another PR that uses your approach: https://github.com/rpm-software-management/dnf5/pull/2527.
If I understand correctly the only downside of the new approach is only the added complexity, it shouldn't be visible to the user in any way. Due to this I would probably prefer it.

PR #2385, opened 2025-08-12
doc: fix references to dnf5.conf(5)

mmarchini:
Some doc pages were referring to dnf5-conf(5), which doesn't exist. Fix those references to point to dnf5.conf(5) instead.


kontura:
Thanks!

Issue #2370, opened 2025-07-30
Manifest plugin


evan-goode:
The manifest plugin for DNF4 (https://github.com/rpm-software-management/dnf-plugins-core/pull/582) should be ported to DNF5. We'll also use https://github.com/rpm-software-management/libpkgmanifest.

The plugin should be able to:

- Generate manifest files (lockfiles) listing all packages and dependencies based on:
  - a list of pkgspecs (`dnf manifest new pkgA pkgB`)
  - manifest input files (`--input /path/to/packages.input.yaml`)
  - or system state (`--system`)
- Download packages from the manifest, including non-system base architectures (`dnf manifest download`)
- Install all non-source packages from a manifest (`dnf manifest install`)

See also https://github.com/rpm-software-management/dnf5/issues/833.

rriemann:
Hi all,

this is a timely issue. I compare the German government's approach of building distroless micro images with Nix Flakes against an alternative with podman multi-stage Containerfiles and `dnf`.

Currently, Nix Flakes builds are reproducible and lead possibly in this comparison. So I cannot wait until I can use dnf manifest to also have reproducible builds with podman and dnf.

Read more about this and find also the containerfile at: https://fedoramagazine.org/build-smaller-containers/

Can you tell me when this will be available in Fedora?

PR #2389, opened 2025-08-15
Enable copying config from one `libdnf5::Base` to another

evan-goode:
Adds `void Base::set_config(ConfigMain config)` and adds missing copy constructors and/or copy assignment operators (`operator=`) for:

- `libdnf5::Config`
- `libdnf5::ConfigMain`
- `libdnf5::Option`
- `libdnf5::OptionBinds`
- `libdnf5::OptionEnum`

Correct me if I'm wrong, but ideally these classes should also have move constructors and move assignment operators (rule of five).

`set_config` is wanted by the DNF5 manifest plugin (https://github.com/rpm-software-management/dnf5/issues/2370) so we can easily create new `libdnf5::Base` objects to resolve transactions for multiple architectures (requiring distinct sets of repositories).

I think these changes are ABI-compatible since these classes were already non-trivial.

### DIFF
--- a/include/libdnf5/base/base.hpp
+++ b/include/libdnf5/base/base.hpp
@@ -82,6 +82,7 @@ public:
 
     /// @return a reference to configuration
     ConfigMain & get_config();
+    void set_config(ConfigMain config);
     LogRouterWeakPtr get_logger();
     comps::CompsSackWeakPtr get_comps_sack();
     repo::RepoSackWeakPtr get_repo_sack();
--- a/include/libdnf5/conf/option.hpp
+++ b/include/libdnf5/conf/option.hpp
@@ -54,7 +54,11 @@ public:
 
     explicit Option(Priority priority = Priority::EMPTY);
     Option(const Option & src);
+    Option(Option && other) noexcept;
+
     virtual ~Option();
+
+    Option & operator=(const Option & other);
 
     /// Makes copy (clone) of this object.
     // @replaces libdnf:conf/Option.hpp:method:Option.clone()
--- a/libdnf5/base/base.cpp
+++ b/libdnf5/base/base.cpp
@@ -272,6 +272,9 @@ bool Base::are_repos_configured() const noexcept {
 ConfigMain & Base::get_config() {
     return p_impl->config;
 }
+void Base::set_config(ConfigMain config) {
+    this->p_impl->config = std::move(config);
+}
 LogRouterWeakPtr Base::get_logger() {
     return {&p_impl->log_router, &p_impl->log_router_guard};
 }
--- a/libdnf5/conf/config.cpp
+++ b/libdnf5/conf/config.cpp
@@ -23,6 +23,15 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 namespace libdnf5 {
 
 class Config::Impl {
+public:
+    Impl() = default;
+    Impl(const Impl & other) : binds(other.binds) {}
+
+    Impl & operator=(const Impl & other) {
+        binds = other.binds;
+        return *this;
+    }
+
 private:
     friend Config;
 
@@ -33,8 +42,18 @@ OptionBinds & Config::opt_binds() noexcept {
     return p_impl->binds;
 }
 
-Config::~Config() = default;
 Config::Config() : p_impl(new Impl()) {}
+
+Config::Config(const Config & other) : p_impl(other.p_impl) {}
+
+Config::~Config() = default;
+
+Config & Config::operator=(const Config & other) {
+    if (this != &other) {
+        p_impl = other.p_impl;
+    }
+    return *this;
+}
 
 void Config::load_from_parser(
     const ConfigParser & parser,
--- a/libdnf5/conf/config_main.cpp
+++ b/libdnf5/conf/config_main.cpp
@@ -110,6 +110,7 @@ class ConfigMain::Impl {
     friend class ConfigMain;
 
     explicit Impl(Config & owner);
+    explicit Impl(Config & owner, Config & other);
 
     Config & owner;
 
@@ -470,7 +471,20 @@ ConfigMain::Impl::Impl(Config & owner) : owner(owner) {
 ConfigMain::ConfigMain() {
     p_impl = std::unique_ptr<Impl>(new Impl(*this));
 }
+
+ConfigMain::ConfigMain(const ConfigMain & other) : Config(other) {
+    p_impl = std::make_unique<Impl>(*other.p_impl);
+}
+
 ConfigMain::~ConfigMain() = default;
+
+ConfigMain & ConfigMain::operator=(const ConfigMain & other) {
+    if (this != &other) {
+        Config::operator=(other);
+        p_impl = std::make_unique<Impl>(*other.p_impl);
+    }
+    return *this;
+}
 
 OptionNumber<std::int32_t> & ConfigMain::get_debuglevel_option() {
     return p_impl->debuglevel;
--- a/include/libdnf5/conf/config.hpp
+++ b/include/libdnf5/conf/config.hpp
@@ -37,7 +37,11 @@ public:
     OptionBinds & opt_binds() noexcept;
 
     Config();
+    Config(const Config & other);
+
     virtual ~Config();
+
+    Config & operator=(const Config & other);
 
     virtual void load_from_parser(
         const ConfigParser & parser,
--- a/include/libdnf5/conf/config_main.hpp
+++ b/include/libdnf5/conf/config_main.hpp
@@ -40,7 +40,11 @@ namespace libdnf5 {
 class LIBDNF_API ConfigMain : public Config {
 public:
     ConfigMain();
+    ConfigMain(const ConfigMain & other);
+
     ~ConfigMain();
+
+    ConfigMain & operator=(const ConfigMain & other);
 
     OptionNumber<std::int32_t> & get_debuglevel_option();
     const OptionNumber<std::int32_t> & get_debuglevel_option() const;
--- a/include/libdnf5/conf/option_binds.hpp
+++ b/include/libdnf5/conf/option_binds.hpp
@@ -67,6 +67,8 @@ public:
     OptionBinds(const OptionBinds & src);
     ~OptionBinds();
 
+    OptionBinds & operator=(const OptionBinds & other);
+
     Item & add(
         std::string id,
         Option & option,
--- a/include/libdnf5/conf/option_enum.hpp
+++ b/include/libdnf5/conf/option_enum.hpp
@@ -40,6 +40,8 @@ public:
     OptionEnum(std::string default_value, std::vector<std::string> enum_vals);
     OptionEnum(std::string default_value, std::vector<std::string> enum_vals, FromStringFunc && from_string_func);
 
+    OptionEnum(const OptionEnum & other);
+
     ~OptionEnum() override;
 
     /// Makes copy (clone) of this object.
--- a/libdnf5/conf/option.cpp
+++ b/libdnf5/conf/option.cpp
@@ -35,9 +35,11 @@ private:
 
 Option::Option(Priority priority) : p_impl(new Impl(priority)) {}
 
+Option::Option(const Option & src) = default;
+
 Option::~Option() = default;
 
-Option::Option(const Option & src) = default;
+Option & Option::operator=(const Option & other) = default;
 
 Option::Priority Option::get_priority() const {
     return p_impl->priority;
@@ -69,5 +71,6 @@ void Option::assert_not_locked() const {
 const std::string & Option::get_lock_comment() const noexcept {
     return p_impl->lock_comment;
 }
+
 
 }  // namespace libdnf5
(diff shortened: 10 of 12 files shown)


fhbash (approved):
LGTM

> Correct me if I'm wrong, but ideally these classes should also have move constructors and move assignment operators (rule of five).

For my understanding we are good on rule of three here, since we manually defined the constructor, destructor and copy assigned the compiler will not create move constructor/destructor automatically, right?

libdnf5/base/base.cpp
```diff
@@ -272,6 +272,9 @@ bool Base::are_repos_configured() const noexcept {
 ConfigMain & Base::get_config() {
     return p_impl->config;
 }
+void Base::set_config(ConfigMain config) {
+    this->p_impl->config = std::move(config);
```
jrohel: I'm not sure in which cases we want to use set_config. Do we want to allow the user to change the settings at any time? For example, the user is only allowed to call base.setup() once. There's a check for libdnf_user_assert(!pool, "Base was already initialized");. Shouldn't there be the same check here? I'm wondering how risky it is to allow configuration to be changed at any time later on.

libdnf5/conf/config.cpp
```diff
@@ -23,6 +23,15 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
 
 class Config::Impl {
+public:
+    Impl() = default;
+    Impl(const Impl & other) : binds(other.binds) {}
+
+    Impl & operator=(const Impl & other) {
+        binds = other.binds;
+        return *this;
+    }
```
jrohel: Do we really need custom constructors and an assignment operator here? What's the difference between them and the default ones the compiler gives us? Moreover, if we don't define anything, the compiler will generate the necessary versions, including the "move" versions if they are needed.

libdnf5/conf/config.cpp
```diff
@@ -33,9 +42,19 @@ OptionBinds & Config::opt_binds() noexcept {
     return p_impl->binds;
 }
 
-Config::~Config() = default;
 Config::Config() : p_impl(new Impl()) {}
 
+Config::Config(const Config & other) : p_impl(other.p_impl) {}
```
jrohel: Do we really want to write our own copy constructor? Isn't it enough to just tell the compiler to generate a default one?
`Config::Config(const Config & other) = default;`

If we want, we can also have the move constructor generated.
`Config::Config(Config && other) = default;`

libdnf5/conf/config.cpp
```diff
@@ -33,9 +42,19 @@ OptionBinds & Config::opt_binds() noexcept {
...
+
+Config::~Config() = default;
+
+Config & Config::operator=(const Config & other) {
```
jrohel: Let the compiler generate it for us.
`Config & Config::operator=(const Config & other) = default;`

libdnf5/conf/config_main.cpp
```diff
@@ -110,6 +110,7 @@ class ConfigMain::Impl {
     friend class ConfigMain;
 
     explicit Impl(Config & owner);
+    explicit Impl(Config & owner, Config & other);
```
jrohel: Where is it implemented and used?

libdnf5/base/base.cpp
evan-goode: The motivation is to have some way to construct a new `libdnf5::Base` that copies the config from another `libdnf5::Base`.

But you bring up a good point re: locking. Rethinking this a bit, do you think it would be better if instead of `Base::set_config`, we had a `Config::load_from_config(const Config & src)` that loaded all option binds from `src`? In that function, we could be sure to not set config options that are locked.

Or we could drop this PR and have API consumers do this manually. I think I will move this back to draft and see whether I can find another approach for my use case.

jrohel: > The motivation is to have some way to construct a new libdnf5::Base that copies the config from another libdnf5::Base.

Okay. But it's not that simple. For example, `OptionBinds::Item` contains a pointer to `Option` (in `OptionBinds::Item::Impl`). The code copies the contents of `OptionBinds::Item` (the pointer value). So the copy contains a pointer to the `Option` in the source configuration. This means when we use `OptionBinds::Item` in the copy, we will be reading/modifying the source configuration. And if the source configuration has already been deallocated, we will be working with memory that contains who knows what.

And the copy constructor and assignment operator for `ConfigMain` also look suspicious.

The resulting code can be compiled, but it doesn't perform a deep copy. The copy contains a lot of pointers to the original.

evan-goode:
> The ConfigMain::Impl copy constructor clones each option from the other.owner.opt_binds()

Sigh, this doesn't work either. I think we'd need a virtual `operator=`. I'm just going to hardcode each option.

include/libdnf5/base/base.hpp
```diff
@@ -58,7 +58,7 @@ class InternalBaseUser;
 class LIBDNF_API Base {
 public:
     /// Constructs a new Base instance and sets the destination loggers.
-    Base(std::vector<std::unique_ptr<Logger>> && loggers = {});
+    Base(std::vector<std::unique_ptr<Logger>> && loggers = {}, ConfigMain config = {});
```
jrohel: This change is okay from an API perspective, but it breaks the ABI. The original constructor with one argument has been replaced by a new one with two.

To maintain ABI compatibility, we should keep the original constructor and simply add the new one.
```
    Base(std::vector<std::unique_ptr<Logger>> && loggers = {});
    Base(std::vector<std::unique_ptr<Logger>> && loggers, ConfigMain config);
```
The constructors' implementation also needs to be adjusted.

libdnf5/base/base.cpp
```diff
@@ -53,14 +53,16 @@ namespace libdnf5 {
 static std::atomic<Base *> locked_base{nullptr};
 static std::mutex locked_base_mutex;
 
-Base::Base(std::vector<std::unique_ptr<Logger>> && loggers) : p_impl(new Impl(get_weak_ptr(), std::move(loggers))) {}
+Base::Base(std::vector<std::unique_ptr<Logger>> && loggers, ConfigMain config)
+    : p_impl(new Impl(get_weak_ptr(), std::move(loggers), std::move(config))) {}
```
jrohel, line 60: We also need to implement the original constructor with one parameter to avoid breaking the ABI. Please see the text in the constructor's declaration for details.

Add:
```
Base::Base(std::vector<std::unique_ptr<Logger>> && loggers)
    : p_impl(new Impl(get_weak_ptr(), std::move(loggers), {})) {}
```

jrohel:
@evan-goode 
> I removed Base::set_config and instead added a (optional) ConfigMain argument to the Base constructor. Do you think this is better?

I'm not sure if this will cover all the necessary use cases, but in general, creating a `Base` constructor that allows passing the configuration sounds like a reasonable idea.

**However, be careful:** if we want to avoid breaking the ABI, we must add a new constructor and leave the old one unchanged. Please see my comments in the code for details.

libdnf5/conf/config_main.cpp
```diff
@@ -110,6 +110,9 @@ class ConfigMain::Impl {
     friend class ConfigMain;
 
     explicit Impl(Config & owner);
+    explicit Impl(Config & owner, const ConfigMain::Impl & other);
+
+    Impl & operator=(const Impl & other);
```
jrohel, line 116: Writing a custom copy assignment operator that internally copies all members except for one reference (Config & owner) is a solution, but I recommend avoiding it here. Every time a new member is added, we'd have to remember to synchronize the copy operator as well.

What if we changed this problematic reference to a pointer instead? Or better yet, remove it entirely, since it's only used in the constructor and not afterwards. This would allow us to use the default copy assignment constructor.

Essentially, my recommendation is to remove your implementation and use the default one.
```
Impl & operator=(const Impl & other) = default;
```

And remove the line `Config & owner;` and the `owner` initialization.

evan-goode: Good catch! Done. Though if we implement `load_values` we will probably need to enumerate each field, or somehow iterate over them via `owner.opt_binds()`.

libdnf5/base/base.cpp
evan-goode: Done.

include/libdnf5/base/base.hpp
evan-goode: Done, I did

```
Base(ConfigMain config, std::vector<std::unique_ptr<Logger>> && loggers = {});
```

so the `loggers` can stay optional.

test/libdnf5/conf/test_conf.cpp
```diff
@@ -71,3 +71,57 @@ void ConfTest::test_config_pkg_gpgcheck() {
     CPPUNIT_ASSERT_EQUAL(&config_repo.get_pkg_gpgcheck_option(), &config_repo.get_gpgcheck_option());
 #pragma GCC diagnostic pop
 }
+
+libdnf5::ConfigMain copy_construct_config_main(const libdnf5::ConfigMain & config) {
```
jrohel, line 77: Move the functions `copy_construct_config_main`, `assign_config_main`, and `test_copy_config_main` to anonymous namespace:

```
namespace {
... functions ...
}
```
We do not want to pollute the global namespace. All tests files are linked together. I remember problems where a class with the same name was defined in multiple test files within the global namespace.

test/libdnf5/conf/test_conf.cpp
```diff
@@ -71,3 +71,57 @@ void ConfTest::test_config_pkg_gpgcheck() {
...
+    return libdnf5::ConfigMain{config};
```
jrohel, line 78: OK. RVO (Return Value Optimalization) is mandatory from C++17. The compiler constructs the object directly in the target memory location.

test/libdnf5/conf/test_conf.cpp
```diff
@@ -71,3 +71,57 @@ void ConfTest::test_config_pkg_gpgcheck() {
...
+}
+
+libdnf5::ConfigMain assign_config_main(const libdnf5::ConfigMain & config) {
+    libdnf5::ConfigMain config_copy;
+    config_copy = config;
+    return config_copy;
```
jrohel, line 84: Hmm. NRVO (Named Return Value Optimization) is an optional optimization, so the object might be copied and then moved. In the absence of a move constructor, two copies would indeed occur. However, modern compilers usually perform NRVO. Moreover, If the code works with a copy followed by a move (or two copies), it will certainly pass the test with a single copy. Therefore, the test's logic is sound.

include/libdnf5/conf/option.hpp
```diff
@@ -56,6 +56,8 @@ class LIBDNF_API Option {
     Option(const Option & src);
     virtual ~Option();
 
+    Option & operator=(const Option & other);
```
jrohel: As the `Option` class is the base for all other options, copying it by itself is not a meaningful operation. **I suggest making the operator protected** to prevent it from being called inadvertently.

Now that I think about it, the `Option(const Option & src)` copy constructor should have been protected as well.

libdnf5/base/base.cpp
```diff
@@ -53,14 +53,19 @@ namespace libdnf5 {
...
 
 Base::~Base() = default;
 
-Base::Impl::Impl(const libdnf5::BaseWeakPtr & base, std::vector<std::unique_ptr<Logger>> && loggers)
+Base::Impl::Impl(const libdnf5::BaseWeakPtr & base, std::vector<std::unique_ptr<Logger>> && loggers, ConfigMain config)
     : rpm_advisory_sack(base),
       plugins(*base),
       log_router(std::move(loggers)),
+      config(std::move(config)),
```
jrohel, line 68: Since `Config` has no move constructor or move assignment operator, a deep copy is performed. `std::move` will be ineffective here, but it won't cause any harm.

libdnf5/base/base.cpp
```diff
@@ -53,14 +53,19 @@ namespace libdnf5 {
 static std::atomic<Base *> locked_base{nullptr};
 static std::mutex locked_base_mutex;
 
-Base::Base(std::vector<std::unique_ptr<Logger>> && loggers) : p_impl(new Impl(get_weak_ptr(), std::move(loggers))) {}
+Base::Base(std::vector<std::unique_ptr<Logger>> && loggers)
+    : p_impl(new Impl(get_weak_ptr(), std::move(loggers), {})) {}
+
+Base::Base(ConfigMain config, std::vector<std::unique_ptr<Logger>> && loggers)
+    : p_impl(new Impl(get_weak_ptr(), std::move(loggers), std::move(config))) {}
```
jrohel, line 60: We have an unpleasant situation here. The `config` argument was passed by value (which results in either an expensive deep copy or copy elision, but only in specific cases). Following this, we have `std::move`, which will be ignored because `config` lacks a move constructor and a move assignment operator. This means the subsequent pass by value will be a deep copy. And then, as noted in my other comment, another deep copy occurs in the `Base::Impl::Impl` constructor.

We could pass `config` to the `Impl` constructor by reference to save one deep copy: `Base::Impl::Impl(const libdnf5::BaseWeakPtr & base, std::vector<std::unique_ptr<Logger>> && loggers, const ConfigMain & config)`. And remove `std::move` from there.

We could also pass by reference to Base::Base(ConfigMain config, std::vector<std::unique_ptr<Logger>> && loggers) to save yet another copy. However, the original version clearly states a copy is made. This change is on you.

test/libdnf5/conf/test_conf.cpp
evan-goode: Done.

include/libdnf5/conf/option.hpp
evan-goode: OK sure, done.

Issue #2355, opened 2025-07-21
dnf5daemon: no way to list ongoing transactions


jelly:
For Cockpit to use dnf5daemon for system upgrades we would like to know if there isn't already an active operation happening. For example a user can start a system update and then close the Cockpit browser tab, if they re-open it we would like to show that this is ongoing.

As alternative solution we could subscribe to all signals when logging in to Cockpit and use that to detect if a "transaction" is ongoing.

m-blaha:
Detecting an ongoing transaction should not be a problem.

I am considering how to gather details about the transaction, such as the progress bar and the package currently being processed, if necessary. Is there a dnfdaemon session open throughout the entire transaction on the Cockpit side? Or will reopening the browser tab initiate another dnfdaemon session? What information do you need about the transaction to display to the user?

jelly:
> I am considering how to gather details about the transaction, such as the progress bar and the package currently being processed, if necessary. Is there a dnfdaemon session open throughout the entire transaction on the Cockpit side? 

My current JavaScript abstraction of dn5daemon opens a new session for every action such as `install()`. So the sessions are opened in demand and closed after the relevant action completed. 

> Or will reopening the browser tab initiate another dnfdaemon session? What information do you need about the transaction to display to the user?

It won't open a session when you log in, but for example the Software updates page (based on PacageKit) calls [GetTransactionList](https://packagekit.freedesktop.org/gtk-doc/PackageKit.html#PackageKit.GetTransactionList) when the page loads to then show if a software update is already in progress.

Scenario's we should support is:

* A user doing a dnf transaction outside of Cockpit, if we try to resolve/update a system it is nice to show that something is currently blocking. (Although I found out that you can run two transactions in parallel but they seem to wait on the other to complete first)
* For the Software updates page a user might upgrade the system, disconnect and then log in again, then it should ideally continue showing the installation progress.

PR #2391, opened 2025-08-18
Use transaction lock file to provide details abour currently running transaction

m-blaha:
This PR:

- Enhances the existing Locker class to be able to read and write the contents of the lock file.
- Adds previously missing unit tests for the Locker class.
- Introduces a private TransactionLocker class, which is specifically designed for transaction locking and can record essential details about the current transaction, including the initiator, the process PID, and a description of the transaction.
- provide a more informative error message to the user. Rather than simply stating, "Failed to obtain rpm transaction lock. Another transaction is in progress," the message should include additional information, such as the PID of the other process, its current status, and the command line used to initiate the transaction.

For: https://github.com/rpm-software-management/dnf5/issues/2355

### DIFF
--- a/libdnf5/base/active_transaction_info.cpp
+++ b/libdnf5/base/active_transaction_info.cpp
@@ -3,6 +3,8 @@
 
 #include "libdnf5/base/active_transaction_info.hpp"
 
+#include "libdnf5/utils/bgettext/bgettext-mark-domain.h"
+
 #include <toml.hpp>
 
 #include <ctime>
@@ -87,10 +89,10 @@ ActiveTransactionInfo ActiveTransactionInfo::from_toml(const std::string & toml_
         info.set_comment(toml::find_or<std::string>(data, "comment", ""));
         info.set_pid(toml::find_or<pid_t>(data, "pid", -1));
         info.set_start_time(toml::find_or<time_t>(data, "start_time", 0));
-    } catch (const toml::syntax_error &) {
-        // Return default/empty ActiveTransactionInfo on TOML syntax errors
-    } catch (const toml::type_error &) {
-        // Return default/empty ActiveTransactionInfo on TOML type errors
+    } catch (const toml::syntax_error & ex) {
+        throw ActiveTransactionInfoParseError(M_("Error parsing transaction info: {}"), std::string(ex.what()));
+    } catch (const toml::type_error & ex) {
+        throw ActiveTransactionInfoParseError(M_("Error parsing transaction info: {}"), std::string(ex.what()));
     }
 
     return info;
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -972,10 +972,15 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
     lock_file_path /= std::filesystem::path(libdnf5::TRANSACTION_LOCK_FILEPATH).relative_path();
     std::filesystem::create_directories(lock_file_path.parent_path());
 
+    auto logger = base->get_logger().get();
     TransactionLocker transaction_locker(lock_file_path, info);
     if (!transaction_locker.write_lock()) {
         auto content = transaction_locker.read_content();
-        concurrent_transaction = std::make_unique<ActiveTransactionInfo>(ActiveTransactionInfo::from_toml(content));
+        try {
+            concurrent_transaction = std::make_unique<ActiveTransactionInfo>(ActiveTransactionInfo::from_toml(content));
+        } catch (const ActiveTransactionInfoParseError & ex) {
+            logger->warning(ex.what());
+        }
         return TransactionRunResult::ERROR_LOCK;
     }
     transaction_locker.write_metadata();
@@ -1109,9 +1114,6 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
     }
     db_transaction.set_dt_start(dt_start);
     db_transaction.start();
-
-
-    auto logger = base->get_logger().get();
 
     int pipe_out_from_scriptlets[2];
     if (pipe2(pipe_out_from_scriptlets, O_CLOEXEC) == -1) {
--- a/bindings/libdnf5/base.i
+++ b/bindings/libdnf5/base.i
@@ -31,7 +31,6 @@
     #include "libdnf5/base/base.hpp"
     #include "libdnf5/base/solver_problems.hpp"
     #include "libdnf5/base/log_event.hpp"
-    #include "libdnf5/base/active_transaction_info.hpp"
     #include "libdnf5/base/transaction.hpp"
     #include "libdnf5/base/transaction_package.hpp"
     #include "libdnf5/base/goal.hpp"
--- a/bindings/libdnf5/exception.hpp
+++ b/bindings/libdnf5/exception.hpp
@@ -1,6 +1,7 @@
 #ifndef _BINDINGS_LIBDNF5_EXCEPTINON_HPP_
 #define _BINDINGS_LIBDNF5_EXCEPTINON_HPP_
 
+#include "libdnf5/base/active_transaction_info_errors.hpp"
 #include "libdnf5/base/transaction_errors.hpp"
 #include "libdnf5/common/exception.hpp"
 #include "libdnf5/comps/group/package_errors.hpp"
@@ -63,6 +64,7 @@ public:
 
 static void create_swig_exception(const std::exception & e) {
     do {
+        swig_if_exception(e, libdnf5::base::ActiveTransactionInfoParseError);
         swig_if_exception(e, libdnf5::base::TransactionError);
 
         swig_if_exception(e, libdnf5::comps::InvalidPackageType);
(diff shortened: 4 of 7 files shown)


libdnf5/base/active_transaction_info.cpp
```diff
@@ -0,0 +1,99 @@
...
+    ActiveTransactionInfo info;
+
+    try {
+        auto data = toml::parse_str(toml_string);
+        info.set_description(toml::find_or<std::string>(data, "description", ""));
+        info.set_comment(toml::find_or<std::string>(data, "comment", ""));
+        info.set_pid(toml::find_or<pid_t>(data, "pid", -1));
+        info.set_start_time(toml::find_or<time_t>(data, "start_time", 0));
+    } catch (const toml::syntax_error &) {
+        // Return default/empty ActiveTransactionInfo on TOML syntax errors
```
dcantrell: What happens here?

libdnf5/base/active_transaction_info.cpp
```diff
@@ -0,0 +1,99 @@
...
+    } catch (const toml::type_error &) {
+        // Return default/empty ActiveTransactionInfo on TOML type errors
```
dcantrell: What happens here?

libdnf5/base/transaction.cpp
```diff
@@ -158,6 +162,28 @@ static std::vector<std::pair<ProblemRules, std::vector<std::string>>> get_remova
...
+
+TransactionLocker::TransactionLocker(const std::filesystem::path & path, const ActiveTransactionInfo & info)
+    : Locker(path.string()),
+      transaction_info(info) {}
+
+void TransactionLocker::write_transaction_metadata() {
+    try {
+        write_content(transaction_info.to_toml());
+    } catch (const SystemError & e) {
+        //TODO(mblaha): log the error (would require base object)
```
dcantrell: At least in the interim should this just go to stderr?

dcantrell (changes requested):

libdnf5/base/active_transaction_info.cpp
m-blaha: Both these exceptions mean that the lock file exists, but contains damaged/incorrect data not digestible by TOML parser. Since data from the lock file are not crucial for dnf5 behavior (only contains details about a concurrently running transaction), instead of throwing an error an empty ActiveTransactionInfo object is returned.

libdnf5/base/transaction.cpp
m-blaha: Hm, we try to keep libdnf5 quiet. It does not print to stderr at all.

libdnf5/base/active_transaction_info.cpp
dcantrell: Maybe I'm missing something here, but that feels like it is obscuring what is happening by returning an empty ActiveTransactionInfo object.  Damaged or incorrect data in the lock file feels like it warrants a new exception type so that it is more clear what is happening and where.  Or at least to help a developer debugging something going wrong.

libdnf5/base/transaction.cpp
dcantrell: If libdnf5 is logging, stderr is not all that different, right?  If we are to log here but are not [yet], then stderr feels appropriate.  Either that or raise an exception.

libdnf5/base/active_transaction_info.cpp
m-blaha: I guess you are right. I originally thought that the parsing errors are not worth handling properly. But yeah, it definitely makes debugging harder and it's better to do the error handling right away than adding it later.

m-blaha: @dcantrell I've added a commit with parsing errors handling.

libdnf5/base/transaction.cpp
dcantrell: If the catch block is going to do nothing here, I suggest just removing the try/catch block entirely and leaving the function as just a call to write_content().

libdnf5/utils/locker.cpp
```diff
@@ -64,6 +64,58 @@ bool Locker::lock(short int type) {
...
+        throw SystemError(errno, M_("Failed to seek lock file \"{}\""), path);
+    }
+
+    // Write the content
+    ssize_t written = ::write(lock_fd, content.c_str(), content.size());
+    if (written == -1) {
+        throw SystemError(errno, M_("Failed to write to lock file \"{}\""), path);
+    }
+
+    // Ensure data is written to disk
```
dcantrell, line 148: The single call to fsync() does not technically ensure data is written to the file.  Annoyingly you have to fsync() both the fd for the file and a separate fd for the directory containing the file.

This may be splitting hairs because we're also at the mercy of the kernel VFS layer and ultimately the filesystem driver and the storage device driver, but at least in userspace the control given to guarantee data is on the device we have to do two fsync() calls.

libdnf5/utils/locker.cpp
```diff
@@ -64,6 +64,58 @@ bool Locker::lock(short int type) {
...
+}
+
+std::string Locker::read_content() {
+    if (lock_fd == -1) {
+        throw RuntimeError(M_("Cannot read content: no lock held on file \"{}\""), path);
+    }
+
+    std::string content;
+    content.reserve(1024);
+    char buffer[1024];
```
dcantrell: Style preference, I like using BUFSIZ for arbitrary buffers rather than hardcoding a value.  In theory it keeps the code using platform-preferred buffer sizes as defined by the C library at build time.

libdnf5/base/transaction.cpp
```diff
@@ -970,15 +970,24 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
...
     lock_file_path /= std::filesystem::path(libdnf5::TRANSACTION_LOCK_FILEPATH).relative_path();
     std::filesystem::create_directories(lock_file_path.parent_path());
 
-    libdnf5::utils::Locker locker(lock_file_path);
-    if (!locker.write_lock()) {
+    TransactionLocker transaction_locker(lock_file_path, info);
+    if (!transaction_locker.write_lock()) {
         return TransactionRunResult::ERROR_LOCK;
     }
+    transaction_locker.write_transaction_metadata();
```
mfocko: NGL it feels quite verbose

At least for naming:
```suggestion
    TransactionLocker lock(lock_file_path, info);
    if (!lock.write_lock()) {
        return TransactionRunResult::ERROR_LOCK;
    }
    lock.write_transaction_metadata();
```

As for the interface itself, I would say that it can be easily deduced from the type name, so even
- `write_lock()` -> `write()`
- `write_transaction_metadata()` -> `write_metadata()`

wouldn't reduce the clarity IMO

---

Now that I'm looking at it, why don't we merge the `write_lock()` and `write_transaction_metadata()` into one method? In this case, I guess it would even make sense to fail with `::ERROR_LOCK` when the metadata cannot be stored properly?

include/libdnf5/base/transaction.hpp
```diff
@@ -197,6 +198,9 @@ class LIBDNF_API Transaction {
     /// Retrieve captured RPM log messages
     std::vector<std::string> get_rpm_messages();
 
+    /// Retrieve metadat of concurrently running transaction
```
mfocko: just a typo

```suggestion
    /// Retrieve metadata of concurrently running transaction
```

libdnf5/base/transaction.cpp
```diff
@@ -985,6 +988,8 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
 
     TransactionLocker transaction_locker(lock_file_path, info);
     if (!transaction_locker.write_lock()) {
+        auto content = transaction_locker.read_content();
+        concurrent_transaction = std::make_unique<ActiveTransactionInfo>(ActiveTransactionInfo::from_toml(content));
```
mfocko: I guess we can guarantee only one concurrent transaction, because of the shared path?

dnf5/context.cpp
```diff
@@ -379,6 +421,18 @@ void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
     if (result != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
         print_error(libdnf5::utils::sformat(
             _("Transaction failed: {}"), libdnf5::base::Transaction::transaction_result_to_string(result)));
+
+        // For transaction lock errors, provide detailed information about the running transaction
+        if (result == libdnf5::base::Transaction::TransactionRunResult::ERROR_LOCK) {
+            auto * active_info = transaction.get_concurrent_transaction();
+            if (active_info) {
```
mfocko:
```suggestion
            if (auto * active_info = transaction.get_concurrent_transaction(); active_info) {
```

dnf5/context.cpp
```diff
@@ -532,6 +586,18 @@ void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
...
```
mfocko:
```suggestion
            if (auto * active_info = transaction.get_concurrent_transaction(); active_info) {
```

Issue #2184, opened 2025-04-08
dnf5 install --store option is not documented


ppisar:
According to --help output, `--store` option exists at `dnf5 install`, `dnf5 upgrade`, and probably more commands. However, that option is not described in their manual pages, i.e. dnf5-install(8). The manual only mentions `--offline`. The only place where `--store` is documented is dnf5-replay(8) manual.

Please document `--store` option in manual for all commands that support it. Also if `--store` implies `--offline`, it should be mentioned.

The --store option was added in 7bd215003c0837721669bee2f7063b56dc9bb2e6 commit ("Add `--store` option for storing arbitrary transaction"). @kontura, could you look at it?

kontura:
Sure.

PR #2393, opened 2025-08-19
Add missing shared transaction options and document them

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/2184

This is also fixes a `debuginfo-install` issue where it didn't check for privileges before running. Its fixed by adding the `--store` option to it.

### DIFF
--- a/doc/_shared/options/transaction.rst
+++ b/doc/_shared/options/transaction.rst
@@ -1,0 +1,6 @@
+``--offline``
+    | Store the transaction to be performed offline. See :ref:`offline command <offline_command_ref-label>`, :manpage:`dnf5-offline(8)`.
+
+``--store=PATH``
+    | Store the current transaction in a directory at the specified ``PATH`` instead of running it.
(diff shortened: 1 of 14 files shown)


doc/_shared/options/transaction.rst
```diff
@@ -0,0 +1,6 @@
+``--offline``
+    | Store the transaction to be performed offline. See :ref:`offline command <offline_command_ref-label>`, :manpage:`dnf5-offline(8)`.
+
+``--store=<path>``
```
ppisar: Please spell the replaceble `<path>` element as `PATH`. This the style we use everywhere else. Don't forget to update it in the following text.

ppisar (commented):
Otherwise, it's good.

doc/_shared/options/transaction.rst
kontura: Thanks, fixed.

PR #2399, opened 2025-08-20
Document API changes between DNF and DNF5

pkratoch:
It's not an exhaustive list, rather a comparison of the most common use cases.

### DIFF
--- a/doc/changes_from_dnf4.7.rst
+++ b/doc/changes_from_dnf4.7.rst
@@ -345,46 +345,28 @@ Changes to individual commands
 Changes on the API
 ==================
 
-PackageSet::operator[]
-----------------------
-It was removed due to insufficient O(n^2) performance.
-Use PackageSet iterator to access the data instead.
+``PackageSet::operator[]``
+  * It was removed due to insufficient O(n^2) performance. Use PackageSet iterator to access the data instead.
+
+``Package::get_epoch()``
+  * The return type was changed from ``unsigned long`` to ``std::string``.
 
 
-Package::get_epoch()
---------------------
-The return type was changed from ``unsigned long`` to ``std::string``.
+DNF: ``Package.size``, libdnf: ``dnf_package_get_size()``
+  * The return value was ambiguous, returning either package or install size. Use Package::get_download_size() and Package::get_install_size() instead.
+
+``dnf_sack_set_installonly``, ``dnf_sack_get_installonly``, ``dnf_sack_set_installonly_limit``, ``dnf_sack_get_installonly_limit``
+  * The functions were dropped as unneeded. The installonly packages are taken directly from main Conf in Base.
+
+``Query::filter()`` - ``HY_PKG_UPGRADES_BY_PRIORITY``, ``HY_PKG_OBSOLETES_BY_PRIORITY``, ``HY_PKG_LATEST_PER_ARCH_BY_PRIORITY``
+  * The priority filter was separated into a standalone method. Combine ``query.filter_priority()`` with ``query.filter_latest_evr()`` or another filter to achieve the original functionality.
+
+``Query::filter()`` - ``HY_PKG_LATEST``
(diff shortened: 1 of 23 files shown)


pkratoch:
I might extend this with comparison of setting up download and transaction callbacks, but I'd like to get review on this first to make sure it's going in the right direction, so, I would leave the callbacks for another PR.

doc/changes_from_dnf4.7.rst
```diff
@@ -338,46 +338,35 @@ Changes to individual commands
 Changes on the API
 ==================
 
-PackageSet::operator[]
-----------------------
-It was removed due to insufficient O(n^2) performance.
-Use PackageSet iterator to access the data instead.
+The differences between DNF4 and DNF5 API are not only in minor tweeks, but there are major architectural changes, therefore, it is not feasible to cover all the changes. The :ref:`tutorial_api_changes_ref-label` section describes the most common use cases and point out the main differences.
```
kontura: Should this be `... points out ...`?

doc/tutorial/api-changes/session.rst
```diff
@@ -0,0 +1,53 @@
...
+DNF5 C++:
+
+.. literalinclude:: ../tests/api_changes_from_dnf4/create_base.cpp
+    :language: c++
+    :linenos:
+    :lines: 2,4-7,14-
+
+Setting of configuration and variables should be done before calling the ``libdnf5::Base::setup`` in DNF5.
+
+In Python, there are shortcuts in Python for getting or setting the configuration options using the configuration class attributes (e.g. instead of ``get_skip_unavailable_option().get_value()`` and ``get_skip_unavailable_option().set()`` there is a property ``skip_unavailable``. To access all methods from the option, use standard getters as in C++ API.
```
kontura: The `in Python` is duplicated at the start.

doc/tutorial/api-changes/repos.rst
```diff
@@ -0,0 +1,39 @@
...
+In DNF5, the repositories must be loaded only after the ``libdnf5::base::Base::setup()`` call and they cannot be loaded repeatedly. If a new set of repositories is needed, a new Base object should be created.
+
+DNF4 Python:
+
+.. code-block:: python
+  :linenos:
+
+  import dnf
+
+  base = dnf.Base()
```
kontura: To match the other snippets I think the `base` creation shouldn't be included here.

test/python3/libdnf5/tutorial/api_changes_from_dnf4/load_repos.py
```diff
@@ -0,0 +1,12 @@
+repo_sack = base.get_repo_sack()
+
+# Optionally, read repositories from system configuration files.
+repo_sack.create_repos_from_system_configuration()
+
+# Optionally, create and configure a new repository.
+repo = repo_sack.create_repo("rpm-repeo1")
```
kontura: This is really a small detail but could we call the repo also `my_new_repo_id` here?

test/tutorial/test_api_changes_from_dnf4.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+
+void APIChangesTest::test_create_base() {
+#include "api_changes_from_dnf4/create_base.cpp"
+}
+
+
+void APIChangesTest::test_configure_base() {
+#include "api_changes_from_dnf4/create_base.cpp"
+// Comment necessary to have this order of includes.
```
kontura: Could we rather turn off the formatting for this section with `// clang-format off`? With an explanation that it otherwise wants to reorder the includes?

Or at least I would add the explanation, I was a bit confused by the comments.

pkratoch: I didn't realize there is this option. That's of course much better, thanks!

pkratoch:
Thanks for the review. Everything you mentioned should be fixed. I guess I reorganized the documentation so many times that I became blind to the small inconsistencies.

kontura (approved):

Issue #2403, opened 2025-08-23
Unsatisfied dependency with curly brackets malforms an error message: argument not found


suve:
Recently while doing some packaging work I made a silly mistake. Attached below is a minimal reproducer.

```spec
Name: repro
Summary: Error reproducer
License: WTFPL

# Whatever, just an empty file
Source0: empty.file

Version: 1.0
Release: 1%{?dist}

BuildArch: noarch

Requires: %{name}-%{data} = %{version}-%{release}

%description
Just a small error reproducer.

%files
```

Notice the **Requires** tag - instead of `%{name}-data`, the value is `%{name}-%{data}`; due to RPM leaving unknown macros as-is, when built, this creates a package that depends on `repro-%{data}`.

When trying to install this package, dnf gives a rather unhelpful and confusing error message:
```
$ dnf install /var/lib/mock/fedora-42-x86_64/result/repro-1.0-1.fc42.noarch.rpm
Updating and loading repositories:
Repositories loaded.
argument not found
```

For comparison, this is what happens when the **Requires** tag is fixed (i.e. changed to `%{name}-data`):
```
$ dnf install /var/lib/mock/fedora-42-x86_64/result/repro-1.0-1.fc42.noarch.rpm
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - nothing provides repro-data = 1.0-1.fc42 needed by repro-1.0-1.fc42.noarch from @commandline
You can try to add to command line:
  --skip-broken to skip uninstallable packages
```
As you can see, dnf prints a proper "could not satisfy dependency" error message in this case.

In case you're wondering, RPM handles this fine:
```
$ rpm -i /var/lib/mock/fedora-42-x86_64/result/repro-1.0-1.fc42.noarch.rpm
error: Failed dependencies:
	repro-%{data} = 1.0-1.fc42 is needed by repro-1.0-1.fc42.noarch
```

I'm on Fedora 42 with `dnf5-5.2.15.0-2.fc42`, `libdnf5-5.2.15.0-2.fc42` and `rpm-4.20.1-1.fc42`.

ppisar:
This does not happen with `a%`, it happens with `a%{}`. It also happens with `a{}`.

The error is from a formatting subroutine. It looks like a string injection error because `{}` is a formatting directive for fmt library DNF5 uses.

PR #2406, opened 2025-08-25
libdnf5: Don't double format a transaction error

ppisar:
If a package requires a dependency which cannot be satisfied, a transaction error is formatted and printed to an error output and a log file.

If the error message contained curly brackets, e.g. the dependency was "a{}", "argument not found" was printed instead of the transaction error:

    # dnf5 install ~test/rpmbuild/RPMS/noarch/test2-0-0.fc44.noarch.rpm
    Repositories loaded.
    argument not found

Expected output:

    Repositories loaded.
    Problem: conflicting requests
      - nothing provides a{} needed by test2-0-0.fc44.noarch from @commandline You can try to add to command line: --skip-broken to skip uninstallable packages

The cause was formatting the transaction error twice: An already formatted output from LogEvent.to_string() was passed again to logger.log() which expected a formatting string. The "argument not found" message was an exception from fmt library.

This patch fixes is by logging the already formatted transaction error as a literal string.

Resolve: #2403

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -356,9 +356,8 @@ void Transaction::Impl::add_resolve_log(
     const std::set<std::string> & additional_data,
     libdnf5::Logger::Level log_level) {
     resolve_logs.emplace_back(LogEvent(action, problem, additional_data, settings, spec_type, spec));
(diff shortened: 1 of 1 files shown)


jrohel:
LGTM. Thanks

PR #2407, opened 2025-08-26
dnf5daemon-server: Allow RPM key import without password prompt for wheel users

mcrha:
Similar to trusted RPM transaction, an RPM key import can be done silently, without asking root/admin password, when it's done by a user in the 'wheel' group.


fhbash (approved):
LGTM

AdamWill:
btw, it does occur to me to wonder if we should only allow import of "trusted" GPG keys this way, if there's a viable definition of "trusted" (key is on local filesystem as part of an installed package?)

PR #2409, opened 2025-08-26
Introduces `libdnf5` list plugins on `--version` parameter

fhbash:
Introduces `libdnf5` plugin listing on `--version`

Since `get_plugins_info()` from libdnf base requires `setup()` before, the `print_versions` calls need to postponed, which required an refactor of the `--version` parameter workflow.

Closes: #1335

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -234,6 +234,9 @@ public:
         return libdnf_plugins_enablement;
     }
 
+    void set_show_version(bool enable) { this->show_version = enable; }
+    bool get_show_version() const { return this->show_version; }
+
 private:
     Context & owner;
 
@@ -271,6 +274,7 @@ private:
 
     bool load_system_repo{false};
     LoadAvailableRepos load_available_repos{LoadAvailableRepos::NONE};
+    bool show_version{false};
 };
 
 // TODO(jrohel): Move logic into libdnf?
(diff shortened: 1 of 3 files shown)


dnf5/context.cpp
```diff
@@ -234,6 +234,9 @@ class Context::Impl {
         return libdnf_plugins_enablement;
     }
 
+    void set_show_version(bool enable) { this->show_version = enable; }
```
ppisar, line 237: Is there a reason why the setter refers to show_version with `this` pointer, but the getter does not?

dnf5/main.cpp
```diff
@@ -786,6 +794,22 @@ static void print_versions(Context & context) {
         const auto & api_version = iplugin->get_api_version();
         context.print_output(libdnf5::utils::sformat(_("  API version: {}.{}"), api_version.major, api_version.minor));
     }
+
+    first = true;
+    for (const auto & plugin_info : context.get_base().get_plugins_info()) {
+        if (first) {
+            first = false;
+            context.print_output(libdnf5::utils::sformat(_("\nLoaded libdnf5 plugins:")));
```
ppisar: Here the sformat() call is useless.

dnf5/context.cpp
fhbash: Ow.. I missed that.. fixing..

dnf5/main.cpp
fhbash: missed that too.. fixing..

ppisar (approved):
Good.

PR #2411, opened 2025-08-27
cli: Fix use-after-free issue in CommandAlias

m-blaha:
The raw pointer to the `value` variable, which is local to the if block,
was stored in the args array and later used in the
`target_arg.parse_long(option, args_count, args);` call. Unfortunately,
by this time, the `value` variable has already been destroyed.


ppisar (approved):
Good.

PR #2412, opened 2025-08-27
Add SPDX-License-Identifier, add "Copyright Contributors to the DNF5 project.", unify license header

jrohel:
- Copyright and Author lines were moved to the top of the file.

- The SPDX-License-Identifier was added after them, **followed by a blank line and the original license text**. In C and C++ files, the original license text was converted to single-line comments for a consistent style.

- Than "Copyright Contributors to the DNF5 project." line was added to files with an SPDX-License-Identifier,
but only if it's not already there.

- Added license header requirement to CONTRIBUTING.md 

**Based on the discussion, the original license texts were retained.**

Closes: https://github.com/rpm-software-management/dnf5/issues/410

<del>
Lines "This file is part of libdnf: https://github.com/rpm-software-management/libdnf/" and "This file is part of DNF5: https://github.com/rpm-software-management/dnf5/" were removed.

Especially the version with "libdnf" is massively used in the "DNF5" project. Which is not correct.
- Do we want to preserve the lines as is (with incorrect link to "libdnf" project)?
- Do we want to change link to "DNF5" project on all places?
- Do we want to stay with current PR approach - remove these lines?
</del>


CONTRIBUTING.md
```diff
@@ -116,6 +116,20 @@ To setup a development environment, complete the following steps:
      rpmbuild -ba --define "_sourcedir $(pwd)" dnf5.spec #[--with=<option>|--without=<option> ...]
      ```
 
+## License Header
+
+New C++ and C source files should start with the following header:
+```
+// SPDX-License-Identifier: THE-LICENSE-IDENTIFIER
```
evan-goode, line 124: I think we should add a note linking COPYING.md so contributors know which license identifier they should substitute.

Something like 

> Replace `THE-LICENSE-IDENTIFIER` with either `GPL-2.0-or-later` or `LGPL-2.1-or-later`, according to COPYING.md.

evan-goode: Should we also mention in COPYING.md that some files are licensed as LGPL-2.0-or-later?

evan-goode (changes requested):
Looks great overall, thank you!

CONTRIBUTING.md
jrohel: > Replace THE-LICENSE-IDENTIFIER with either GPL-2.0-or-later or LGPL-2.1-or-later, according to COPYING.md.

OK. Done.

But I think that the licensing situation is quite complex.

I assume that all code for the `libdnf5` and `libdnf5-cli` libraries must be under **LGPL**.

The DNF5 application is currently a mix of **GPL** and **LGPL**.
This could be a problem in the future for third-party DNF5 application plugins, as they would also have to be GPL. Is that what we want? Furthermore, it could be an issue for us if we want to move GPL licensed code from the DNF5 application into the `libdnf5` or `libdnf5-cli` libraries, where we need LGPL.

I'm not a lawyer, but I'm wondering if the DNF5 application should also be under LGPL for the reasons I've mentioned. We could start by stating that all new files for both the libraries and the DNF5 application must be LGPL. The remaining files could stay under LGPL or GPL. But I'm not sure about that.

evan-goode: Good points. The developer team agreed that all new files under DNF5 (libdnf5 or otherwise) should be licensed as LGPL-2.1-or-later so we can more easily move files as needed from the DNF5 section to the libdnf5 section, which is a common task for us. We'll update COPYING.md and CONTRIBUTING.md in a separate PR.

evan-goode (approved):

Issue #926, opened 2023-10-02
DNF local Plugin


j-mracek:
Plugin mostly important only for rawhide users (unstable branch without possibility to downgrade).

Do we want to implement it?

gjohnz:
I'd like to see python3-dnf-plugin-local included. It is a very useful plugin IMO. I've been using it since Fedora 29.

buckaroogeek:
I use the dnf local plugin extensively. My internet connection is wireless (rural area) and bandwidth is expensive. The local plugin allows me to easily re-use downloaded packages in multiple virtual and physical machines that run Fedora. Like @gjohnz, I have been using this since Fedora 33 (more or less).

10leej:
The lack of support for dnf5 is what blocks upgrading to fedora 41 for me. for many of the same reasons @buckaroogeek mentioned.

jsullivan3:
I put together [this package](https://github.com/jsullivan3/dnf-local-cache-actions/releases/tag/v0.1-1) based on `libdnf5-plugin-actions` functionality.  It seems to be solving the issue for me, and can read the configuration file from the `python3-dnf-plugin-local` package.  Hopefully you find it helpful too!

gjohnz:
I do get a permission error when using **dnf repolist** unless I use _sudo_
```

gjohnson@fedora=> dnf repolist
/usr/sbin/dnf-local-cache-actions: line 16: /var/log/dnf-local-cache-actions.log: Permission denied
/usr/sbin/dnf-local-cache-actions: line 21: /var/log/dnf-local-cache-actions.log: Permission denied
repo id                                                 repo name                                                            
fedora                                                  Fedora 41 - x86_64                                                   
fedora-cisco-openh264                  Fedora 41 openh264 (From Cisco) - x86_64                             
updates                                                 Fedora 41 - x86_64 - Updates
```

jsullivan3:
Oh cool!  Thank you very much @gjohnz!  I see what's going on there - the log file is owned by `root`, so a non-root user will be unable to write to it.  Probably tied to the logging issue I created earlier.  There seems to be another issue with the `/etc/dnf/plugins/local.conf` file being readable by only root, so even if the plugin can write to the log file, it won't be able to read the configuration.

The following workaround will make it usable for non-root users, though it unfortunately involves the rather yucky hack of making the log file world-writable:

```
sudo chmod 0644 /etc/dnf/plugins/local.conf
[...]

PR #2418, opened 2025-09-02
Add `local` plugin

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/926


doc/libdnf5_plugins/local.8.rst
```diff
@@ -0,0 +1,77 @@
...
+The repository is automatically added by the plugin with the following options::
+
+    [_dnf_local]
+    name = Local libdnf5 plugin repo
+    baseurl = <repodir>
+    skip_if_unavailable = true
+    cost = 500
+    metadata_expire = 0
+
+Note that the repository has ``pkg_gpgcheck`` verification enabled by default but doesn't specify any ``gpgkey``, it assumes all required keys were already imported.
```
evan-goode, line 25: Could this be a problem when using the same "local" repository across many hosts? Other "clients" may not have imported the required keys. Might be UX regression from the DNF4 plugin.

doc/libdnf5_plugins/local.8.rst
```diff
@@ -0,0 +1,77 @@
...
+
+The ``main`` section requires:
+
+``name``
+    :ref:`string <string-label>`
+
+    The plugin's name is ``local``
+
+``enabled``
+    Whether or where the plugin is enabled: ``true``, ``false``, ``host_only``, ``installroot_only``.
```
evan-goode: Should be `host-only` and `installroot-only`. Could we also briefly explain what `host-only` and `installroot-only` do?

doc/libdnf5_plugins/local.8.rst
```diff
@@ -0,0 +1,77 @@
...
+    :ref:`boolean <boolean-label>`
+
+    Whether running ``createrepo`` to generate repodata is enabled: ``true`` or ``false``.
+
+Additionally it can contain:
+
+``cachedir``
+    :ref:`string <string-label>`
+
+    If you want to speedup ``createrepo`` with the ``--cachedir`` option.
```
evan-goode: nit: `speed up`

libdnf5-plugins/local/local.cpp
```diff
@@ -0,0 +1,247 @@
+// Copyright Contributors to the DNF5 project
+// SPDX-License-Identifier: THE-LICENSE-IDENTIFIER
```
evan-goode: See pending https://github.com/rpm-software-management/dnf5/pull/2432. The header should say the name of the license, `LGPL-2.1-or-later`, not literally `THE-LICENSE-IDENTIFIER`. Glad we are fixing CONTRIBUTING.md, it was confusing :)

libdnf5-plugins/local/local.cpp
```diff
@@ -0,0 +1,247 @@
...
+            repodir = config.get_value("main", "repodir");
+        } else {
+            // default repo location
+            repodir = base.get_config().get_persistdir_option().get_value();
+            repodir = repodir / LOCATION_IN_PERSISTDIR;
+        }
+        repo->get_config().get_name_option().set("Local libdnf5 plugin repo");
+        repo->get_config().get_baseurl_option().set("files://" / repodir);
+        repo->get_config().get_skip_if_unavailable_option().set(true);
+        // Make local repo preffered over other repos (which by default have cost 1000)
```
evan-goode: nit: `preferred`

libdnf5-plugins/local/local.cpp
```diff
@@ -0,0 +1,247 @@
...
+        auto repo = repo_sack->create_repo(LOCAL_REPO_NAME);
+        if (config.has_option("main", "repodir")) {
+            repodir = config.get_value("main", "repodir");
+        } else {
+            // default repo location
+            repodir = base.get_config().get_persistdir_option().get_value();
+            repodir = repodir / LOCATION_IN_PERSISTDIR;
+        }
+        repo->get_config().get_name_option().set("Local libdnf5 plugin repo");
+        repo->get_config().get_baseurl_option().set("files://" / repodir);
```
evan-goode: The correct scheme is `file://`. This works since `a / b` where `b` is an absolute path will evaluate to just `b`. But I think the following would be more correct:

```
repo->get_config().get_baseurl_option().set("file://" + std::filesystem::absolute(repodir).string());
```

evan-goode (changes requested):

doc/libdnf5_plugins/local.8.rst
kontura: It is a problem, I wrote it because the dnf4 plugin had exactly such a bug reported: https://bugzilla.redhat.com/show_bug.cgi?id=1952302 (it should behave the same way as the dnf5 version).

Possibly we could turn off `pkg_gpgcheck` for the local plugin as suggested in the bz or make it an option?

Originally I was thinking we could store all the required keys but the local plugin doesn't have a repo file. Maybe we could create it somewhere but I would rather do that as an RFE.

doc/libdnf5_plugins/local.8.rst
kontura: Thanks that is a good catch.
Maybe we could have some central place where this is described.
For now I have added descriptions for the `host-only` and `installroot-only`.
I have also added one more commit that fixes this in a plugin tutorial.

libdnf5-plugins/local/local.cpp
kontura: Oh I see, thanks.

libdnf5-plugins/local/local.cpp
kontura: Fixed, thanks!

doc/libdnf5_plugins/local.8.rst
evan-goode: OK, if DNF4 has the same problem, then this doesn't need to block this PR.

>Possibly we could turn off pkg_gpgcheck for the local plugin as suggested in the bz or make it an option?

Maybe. I'd be hesitant to turn off `pkg_gpgcheck` for the local repository, that would probably introduce edge cases where sometimes packages never get gpgchecked at all.

> Originally I was thinking we could store all the required keys but the local plugin doesn't have a repo file. Maybe we could create it somewhere but I would rather do that as an RFE.

That sounds promising.

evan-goode (approved):
LGTM, thank you! Merge whenever you're ready.

PR #2422, opened 2025-09-04
feat: add field-specific search options (--name, --summary)

fhbash:
Add options to limit search scope to specific package metadata fields with proper validation to prevent conflicting option combinations.

Add new options on documentation

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2358652


fhbash:
> It looks good to me, but I feel the "-only" parts in the options are extraneous. I would prefer to see them shortened to --name and --sumary

Make sense.. I changed the options to `--name` and `--summary`

PR #2425, opened 2025-09-06
Manifest plugin for DNF5

evan-goode:
This PR ports the manifest plugin [from DNF4](https://github.com/rpm-software-management/dnf-plugins-core/pull/582) to DNF5. The arguments and behavior (should) be unchanged, except that the DNF5 plugin does not support modularity.

The plugin can generate [manifest files](https://github.com/rpm-software-management/libpkgmanifest/#manifest-files) listing all packages and dependencies based on package specs, [input files](https://github.com/rpm-software-management/libpkgmanifest/#input-files), or system state.

Supports downloading packages from the manifest, including non-system base architectures.

Enables installation of all non-source packages from the manifest.

Depends on https://github.com/rpm-software-management/dnf5/pull/2434 and `Requires: libpkgmanifest` which is currently only available in F43.

Since `download_callbacks.hpp` is not part of the public library, I copied the `DownloadCallbacks` implementation into the plugin. We could also discuss making it part of the public `dnf5` API, but it is a bit of a niche use case---this plugin must support loading a different sack or repositories for each architecture, which was possible in DNF4 but is not supported by `libdnf5::Base`, so we create a `Base` object for each architecture.

The `KeyImportRepoCB` implementation is also duplicated from `dnf5::Context`.

To do:

- [x] `--source argument`
- [x] Documentation
- [x] Add `--without` switch in `dnf5.spec` to omit the `manifest` plugin and not depend on libpkgmanifest


evan-goode:
Looking at https://github.com/rpm-software-management/dnf5/pull/2418, I wonder whether the manifest plugin should be its own subpackage instead of lumped in with `dnf5-plugins`.

ppisar:
> Looking at #2418, I wonder whether the manifest plugin should be its own subpackage instead of lumped in with `dnf5-plugins`.

A subpackage please. This plugin has a unique dependency on libpkgmanifest.

dnf5-plugins/manifest_plugin/manifest.cpp
```diff
@@ -0,0 +1,225 @@
...
+    base->setup();
+
+    return base;
+}
+
+void ManifestSubcommand::create_manifest_repos(
+    libdnf5::Base & base, libpkgmanifest::common::Repositories manifest_repos) const {
+    auto repo_sack = base.get_repo_sack();
+    for (const auto & manifest_repo : manifest_repos) {
+        auto repo = repo_sack->create_repo(manifest_repo.get_id());
```
kontura: I ran into a duplicate repoid error here when using this with the `local` plugin enabled. The `local` plugin always adds its repo so when it is also stored as `manifest_repo` it is duplicated.
I worry that any (even third party) plugin could create any repos.

Would it make sense to clear the `repo_sack` before we add the manifest repos?

dnf5-plugins/manifest_plugin/manifest.hpp
```diff
@@ -0,0 +1,115 @@
...
+class ManifestSubcommand : public Command {
+public:
+    explicit ManifestSubcommand(Context & context, const std::string & name) : Command(context, name) {}
+    void set_argument_parser() override;
+    void pre_configure() override;
+
+protected:
+    void create_manifest_repos(libdnf5::Base & base, libpkgmanifest::common::Repositories manifest_repos) const;
+    std::unique_ptr<libdnf5::Base> create_base_for_arch(const std::string & arch) const;
+    libpkgmanifest::manifest::Manifest parse_manifest(const std::string & arch) const;
```
kontura: I think this is not used.

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+    archs_arg->set_long_name("archs");
+    archs_arg->set_description("Explicitly specify basearchs to use");
+    archs_arg->set_has_value(true);
+    archs_arg->set_arg_value_help("<ARCH>,...");
+    archs_arg->link_value(archs_option);
+    cmd.register_named_arg(archs_arg);
+
+    source_option =
+        dynamic_cast<libdnf5::OptionBool *>(parser.add_init_value(std::make_unique<libdnf5::OptionBool>(false)));
+    auto * source_arg = parser.add_new_named_arg("source");
```
kontura: dnf4 was a little bit inconsistent with options dealing with source packages.
Some commands used `--source` and some `--srpm`.
I believe that we have decided to use `--srpm` in dnf5. Would it be ok to change this option to `--srpm` here as well to be consistent?
Note that some commands also have the `--source` alias due to compatibility.
Applies also for the `download` command.

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+        input = libpkgmanifest::input::Parser().parse_prototype(input_path_option->get_value());
+    } else if (pkg_specs.empty()) {
+        use_system_repository = true;
+        generate_system_snapshot = true;
+    }
+
+    if (input.has_value()) {
+        archs = input->get_archs();
+    } else if (archs_option->get_priority() > libdnf5::Option::Priority::DEFAULT) {
+        archs = std::vector<std::string>(archs_option->get_value().begin(), archs_option->get_value().end());
```
kontura: If we want to keep to commands structured like this would it make sense to make `--input=...` conflict with `--archs=...`? Since if `input` is set the `archs` won't be used? `--source` is possibly a similar case.

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+    per_arch_arg->set_long_name("per-arch");
+    per_arch_arg->set_description(_("Separate packages by basearch into individual manifest files"));
+    per_arch_arg->set_const_value("true");
+    per_arch_arg->link_value(per_arch_option);
+    cmd.register_named_arg(per_arch_arg);
+
+    archs_option = dynamic_cast<libdnf5::OptionStringList *>(
+        parser.add_init_value(std::make_unique<libdnf5::OptionStringList>(std::vector<std::string>())));
+    auto * archs_arg = parser.add_new_named_arg("archs");
+    archs_arg->set_long_name("archs");
```
kontura: Would it be possible to use just `--arch`? To make it consistent with the rest of dnf5 (reposync, download, repoquery)?
Applies also for the `download` command.

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+                repo_config.get_metalink_option().set(metalink);
+            } else if (!input_repo.get_mirrorlist().empty()) {
+                const auto & mirrorlist = vars->substitute(input_repo.get_mirrorlist());
+                repo_config.get_mirrorlist_option().set(mirrorlist);
+            } else if (!input_repo.get_baseurl().empty()) {
+                const auto & baseurl = vars->substitute(input_repo.get_baseurl());
+                repo_config.get_baseurl_option().set(baseurl);
+            } else {
+                throw libdnf5::RuntimeError(M_("Input repository \"{}\" has no baseurl"), input_repo.get_id());
+            }
```
kontura: I have noticed the logic here is quite similar to `ManifestSubcommand::create_manifest_repos(...)` I think we could re-use it.
The only differences seem to be:
- repo callbacks; but they are set anyway later
- and `pkg_gpgcheck` which will only take effect at installation time so it will not affect other commands. We could also move its setting specifically to the `install` subcommand.

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+        }
+    } else {
+        repo_sack->create_repos_from_system_configuration();
+
+        repo_sack->create_repos_from_paths(ctx.get_repos_from_path(), libdnf5::Option::Priority::COMMANDLINE);
```
kontura: Should this be called only once after the substitution?

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+        libdnf5::rpm::PackageQuery installed_query{base};
+        installed_query.filter_installed();
+        std::vector<libdnf5::rpm::Package> installed{installed_query.begin(), installed_query.end()};
+        return sort_pkgs(std::move(installed));
+    }
+
+    std::set<libdnf5::rpm::Package> resolved_pkgs;
+    libdnf5::Goal goal(base);
+    if (input.has_value()) {
+        goal.set_allow_erasing(input->get_options().get_allow_erasing());
```
kontura: Why isn't this set below with the rest of the input packages?
Do we want to set this when both `input` and `pkg_specs` are used?

dnf5-plugins/manifest_plugin/manifest_new.cpp
```diff
@@ -0,0 +1,400 @@
...
+        if (is_from_system_repo) {
+            rq.filter_id(dnf_pkg.get_from_repo_id(), libdnf5::sack::QueryCmp::EQ);
+        } else {
+            rq.filter_id(dnf_pkg.get_repo_id(), libdnf5::sack::QueryCmp::EQ);
+        }
+        if (rq.size() > 0) {
+            dnf_repo = *rq.begin();
+            manifest_pkg.set_repo_id((*dnf_repo)->get_id());
+        } else {
+            manifest_pkg.set_repo_id(BOOTSTRAP_REPO_ID);
```
kontura: Could we add some description to this `BOOTSTRAP_REPO_ID`? Or add some documentation for this?

dnf5-plugins/manifest_plugin/manifest.cpp
```diff
@@ -0,0 +1,225 @@
...
+    dnf_nevra.set_arch(manifest_nevra.get_arch());
+    return dnf_nevra;
+}
+
+std::pair<std::filesystem::path, std::string> splitext(const std::filesystem::path & path) {
+    const std::string extension = path.extension().string();
+    std::string root = path.string();
+    root.erase(root.length() - extension.length());
+    return {root, extension};
+}
```
kontura: Unless I missed something this is only ever used to get the `root` path. Could we simplify it and return just the `root`?

kontura (commented):
If I understood correctly I personally am not much of a fan of the way the commands are structured and in my opinion it also doesn't fit very well with the rest of dnf5.

Specifically I mean two things:
1. the `new` command kind of feels like a combination of two commands. Its positional arguments, the `--archs` and the `--source` options are incompatible with the `--input` option. I suppose they could conflict but an option conflicting with positional arguments seems odd.
2. More importantly in my opinion both `download` and `install` subcommands should have the `manifest` path as a positional argument. It is what they operate on, it doesn't feel like an option.

dnf5-plugins/manifest_plugin/manifest.cpp
evan-goode: Yeah, this plugin does not mix with the `local` plugin... or potentially many other plugins.

> Would it make sense to clear the repo_sack before we add the manifest repos?

Possibly, though I think we'd need new methods in `RepoSack` or `Base` and we'd want to disallow clearing the sack after `Base::repos_configured` is set to `true`.

With the local plugin specifically, the local repo should really not be written to the manifest file during `manifest new`. A hack would be to ignore the `_dnf_local` repository.

evan-goode:
> If I understood correctly I personally am not much of a fan of the way the commands are structured and in my opinion it also doesn't fit very well with the rest of dnf5.

Sure, the command structure here is copied from the DNF4 manifest plugin, but that plugin has few users. We might as well take this chance to revise it. If we do want to release the DNF4 manifest plugin in RHEL, before doing so we could change it to match the command structure we decide on here for DNF5.

> Specifically I mean two things:
> 
> 1. the `new` command kind of feels like a combination of two commands. Its positional arguments, the `--archs` and the `--source` options are incompatible with the `--input` option. I suppose they could conflict but an option conflicting with positional arguments seems odd.

I agree, I don't like the "hidden control flow" where the command does a completely different thing if the infile is present vs absent. We could create a `manifest resolve` command that operates on the input files, and keep `manifest new` to operate on pkgspecs and has the `--use-system` flag.

>  2. More importantly in my opinion both `download` and `install` subcommands should have the `manifest` path as a positional argument. It is what they operate on, it doesn't feel like an option.

This I think was a deliberate decision inspired by other package managers that support lockfiles. e.g. Cargo will default to `Cargo.toml`/`Cargo.lock`, npm to `package.json`/`package-lock.json`, Nix to `flake.nix`/`flake.lock`. There's some discussion at https://github.com/rpm-software-management/libpkgmanifest/issues/49 about the ergonomics of a single canonical name for RPM infiles/lockfiles. @dcantrell any opinion?

dnf5-plugins/manifest_plugin/manifest.hpp
evan-goode: Thanks!

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Renamed the `--source` option to `--srpm`.

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Done

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Fixed, thanks

dnf5-plugins/manifest_plugin/manifest.cpp
evan-goode: Sure, done and renamed the function to `strip_extension`.

evan-goode: Though now I guess we can just use `path.stem()`. Fixed again.

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Good catch, thanks.

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Done.

dnf5-plugins/manifest_plugin/manifest.cpp
kontura: Right, I think I see what you mean.
Since the manifests are meant to be used across machines there is no guarantee that such plugin repos will be available though I could image in some very controlled environments they might be.

I would be fine with leaving this up to the user, keeping the logic as is.

kontura: Personally I would rather avoid hard-coding ignoring the `_dnf_local`.

Mayybe a nice to have thing might be wrapping the exception with some explanation? It could say something like: `The stored manifest repository '_dnf_local' conflicts with already present repository of the same id (likely added by a libdnf5 plugin)`.

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: To be honest, I copied that logic from the DNF4 plugin without understanding it. After `BOOTSTRAP_REPO_ID`, is set, it's not updated, so `"bootstrap"` gets written to the manifest file and `manifest install`, etc. fails. Looking at the libpkgmanifest source, there are no references to this "bootstrap" repository so I'm not sure how it's supposed to work.

I have removed that logic. Now, if any package makes it into `add_pkgs_to_manifest` without its repository loaded, we print an error:

```
            throw libdnf5::RuntimeError(
                M_("Package \"{}\" wanted from repository \"{}\", but no such repository was found."),
                dnf_pkg.get_nevra(),
                repo_id);
```

which I think is appropriate behavior.

dnf5-plugins/manifest_plugin/manifest_new.cpp
evan-goode: Now `manifest resolve` doesn't have an `--archs` or `--arch` option. Both subcommands still have `--srpm` since `--srpm` does have an effect even when operating on an infile.

evan-goode:
> Splitting the commands sounds good to me. 👍

Done, now there is a `manifest resolve` and a `manifest new`. man page excerpt:

```
``new``
    Create a manifest file based on a list of packages or the installed
    packages on the system.

    If given a list of packages, all packages and their dependencies will be
    resolved and pinned in the manifest file.

    If a list of packages is not given, all installed packages on the system
    will be pinned in the manifest file.

``resolve``
    Create a manifest file from a provided input file.

    Resolves the list of packages specified in the input file and pins all
    packages and their dependencies in the manifest file.
```

> I see, we could still keep the default though, so doing `$ dnf5 manifest install` uses `packages.manifest.yaml` as the current version. However if a user wants to specify a different manifest yaml they would do `$ dnf5 manifest install path-to-my-manifest.yaml` instead of using `--manifest=path-to-my-manifest.yaml`.

I'd rather keep it as a non-positional option so that `dnf5 manifest install --manifest=mymanifest` can match `dnf5 manifest new --manifest=mymanifest pkgspec1 pkgspec2`. The positional arguments are used in `manifest new` so we have to use `--manifest` there, and it should be consistent across the subcommands.

dnf5.spec
```diff
@@ -907,7 +911,29 @@ automatically and regularly from systemd timers, cron jobs or similar.
 %exclude %{_bindir}/dnf-automatic
 %endif
 
+# ========== dnf5-manifest plugin ==========
+
+%if %{with plugin_manifest}
+%package -n plugin-manifest
```
kontura: One more thing, could we call the plugin package `dnf5-plugin-manifest`?
I think it would require dropping the `-n`.

To match the rest, like `dnf5-plugin-automatic`.

kontura (approved):
Otherwise I think we can proceed with a merge.
(The failing pre-commit hook is unrelated and should already be fixed in main.)

dnf5-plugins/manifest_plugin/manifest.cpp
evan-goode: Done.

dnf5.spec
evan-goode: Ah, fixed.

PR #2426, opened 2025-09-08
Document API changes in callbacks between DNF and DNF5

pkratoch:
This is a follow-up to https://github.com/rpm-software-management/dnf5/pull/2399

### DIFF
--- a/test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
+++ b/test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
@@ -1,0 +1,54 @@
+/* includes; won't compile in tests, in the docs we leave out the comment lines to show them
+# include <libdnf5/repo/download_callbacks.hpp>
+*/
+
+class PackageDownloadCallbacks : public libdnf5::repo::DownloadCallbacks {
+private:
+    using TransferStatus = libdnf5::repo::DownloadCallbacks::TransferStatus;
+
+    /// List of descriptions of downloads.
+    std::forward_list<std::string> download_descriptions;
+
+    void * add_new_download([[maybe_unused]] void * user_data, const char * description, double total_to_download) {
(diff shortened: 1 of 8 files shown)


test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+        return 0;
+    }
+};
+
+// Set the download callbacks.
+base.set_download_callbacks(std::make_unique<PackageDownloadCallbacks>());
+
+// Download the packages.
+transaction.download();
```
ppisar, line 54: Could declare a type of all external variables (`base`, `transaction`)? It will help readers to understand what the variables stands for.

test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
```diff
@@ -0,0 +1,50 @@
+class PackageDownloadCallbacks : public libdnf5::repo::DownloadCallbacks {
```
ppisar, line 5: Please explicitly include all header files needed for all used types.

test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
pkratoch: This is done the same way as the rest of the tutorials - the documentation is meant to be read in sequence. Otherwise, the code snippets would get longer and longer.

test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
pkratoch: Good point, I added them.

test/tutorial/api_changes_from_dnf4/download_callbacks.cpp
ppisar: Indeed.

PR #2430, opened 2025-09-10
builddep: Fix RPM error message capture by configuring log mask

fhbash:
Configure RPM log mask to `RPMLOG_UPTO(RPMLOG_PRI(RPMLOG_DEBUG))` to ensure detailed RPM error messages are captured and displayed when parsing spec files.


fhbash:
Closing since I can't reproduce it on the latest dnf5 version

PR #2432, opened 2025-09-11
Updating license files, licenses clarification

jrohel:


lgpl-2.0.txt
```diff
@@ -0,0 +1,480 @@
+                  GNU LIBRARY GENERAL PUBLIC LICENSE
```
ppisar, line 1: Could you also update dnf5.spec to package this file with %license macro in the subpackages which use files license like that? And also correct their License RPM tags?

Conan-Kudo:
What stuff do we have that's LGPL-2.0 that hasn't just been updated to LGPL-2.1?

Conan-Kudo:
Those all seem like mistakes. All this code is new, there's no reason to have LGPL-2.0-or-later over LGPL-2.1-or-later.

PR #2434, opened 2025-09-12
Add `ConfigMain::load_from_config`

evan-goode:
Supersedes https://github.com/rpm-software-management/dnf5/pull/2389.

Adds a new method, `ConfigMain::load_from_config` which enables copying config values from one `ConfigMain` to another.

```
/// For each option, set the option value to the value from `other`. An option
/// will only be set if its priority in `other` is greater or equal to the
/// current priority and the option in `other` is not empty.
void load_from_config(const libdnf5::ConfigMain & other);
```

An outstanding question: should we also load the `locked` status of each config option from the source config?

### DIFF
--- a/include/libdnf5/conf/config_main.hpp
+++ b/include/libdnf5/conf/config_main.hpp
@@ -321,6 +321,12 @@ public:
         libdnf5::Logger & logger,
         Option::Priority priority = Option::Priority::MAINCONFIG) override;
 
+
+    /// For each option, set the option value to the value from `other`. An option
+    /// will only be set if its priority in `other` is greater or equal to the
+    /// current priority and the option in `other` is not empty.
+    void load_from_config(const libdnf5::ConfigMain & other);
+
 private:
     class LIBDNF_LOCAL Impl;
     std::unique_ptr<Impl> p_impl;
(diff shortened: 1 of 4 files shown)


jrohel:
> This PR also adds a new constructor for `libdnf5::Base` that accepts a `ConfigMain`.

Why?

> Since we still need to be able to pass a `ConfigMain` to the `Base` constructor, I added move constructors/move `operator=` (and some copy constructors/copy operator=) to make `ConfigMain` moveable (but not copyable). 

Why?

You've already created `void load_from_config(const libdnf5::ConfigMain & other)` instead of a copy constructor and an assignment operator. So let's just use that. There's a `void Base::load_config()` method that loads the `ConfigMain` configuration from files. Why don't we just add `void load_config(const ConfigMain & source_config)`?
The implementation could look like this:
```
void Base::load_config(const ConfigMain & source_config) {
    p_impl->config.load_from_config(source_config);
}
```

And an example of use:
```
new_base.load_config(source_base.get_config());
```

A copy and a move aren't necessary. Is there another reason to complicate this at the moment? Also, in the method `Base(ConfigMain config, std::vector<std::unique_ptr<Logger>> && loggers = {}),` `config` is passed by value (a copy is made). Therefore, SWIG requires a copy when using this constructor and a move is insufficient. But as I noted, that constructor is likely not needed.

Issue #2420, opened 2025-09-03
Memory corruption in dnfdaemon


mcatanzaro:
Using dnf5daemon-server-5.2.16.0-3.fc43 on Fedora Workstation 43 with gnome-software-49~beta-6.fc43, I see three dnfdaemon crashes in my coredumpctl: one from yesterday, and two from Thursday last week. Here is yesterday's crash. I'm including all threads just in case, since the crash occurs on a secondary thread:

```
(gdb) t a a bt
Thread 6 (Thread 0x7fe3d0579800 (LWP 16937)):
#0  __syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56
#1  0x00007fe3d1e7ac3c in __internal_syscall_cancel (a1=a1@entry=140731916146112, a2=a2@entry=3, a3=<optimized out>, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=7) at cancellation.c:49
#2  0x00007fe3d1e7ac84 in __syscall_cancel (a1=a1@entry=140731916146112, a2=a2@entry=3, a3=<optimized out>, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=7) at cancellation.c:75
#3  0x00007fe3d1ef419e in __GI___poll (fds=fds@entry=0x7ffeb3ded5c0, nfds=nfds@entry=3, timeout=<optimized out>) at ../sysdeps/unix/sysv/linux/poll.c:29
#4  0x00007fe3d28d90df in poll (__fds=0x7ffeb3ded5c0, __nfds=3, __timeout=<optimized out>) at /usr/include/bits/poll2.h:44
#5  sdbus::internal::Connection::waitForNextEvent (this=this@entry=0x5648fc86d1d0) at /usr/src/debug/sdbus-cpp-2.1.0-3.fc43.x86_64/src/Connection.cpp:816
#6  0x00007fe3d28d9190 in sdbus::internal::Connection::enterEventLoop (this=0x5648fc86d1d0) at /usr/src/debug/sdbus-cpp-2.1.0-3.fc43.x86_64/src/Connection.cpp:151
#7  0x00005648c27b1661 in SessionManager::start_event_loop (this=0x7ffeb3ded970) at /usr/src/debug/dnf5-5.2.16.0-3.fc43.x86_64/dnf5daemon-server/session_manager.cpp:201
[...]

PR #2440, opened 2025-09-18
daemon: Protect libdnf5 access in D-Bus services with mutex

m-blaha:
This change fixes race conditions in various libdnf5/libsolv calls when multiple threads access them concurrently.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2420

### DIFF
--- a/dnf5daemon-server/services/advisory/advisory.cpp
+++ b/dnf5daemon-server/services/advisory/advisory.cpp
@@ -164,27 +164,28 @@ sdbus::MethodReply Advisory::list(sdbus::MethodCall & call) {
     KeyValueMap options;
     call >> options;
 
-    session.fill_sack();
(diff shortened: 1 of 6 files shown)


evan-goode:
So when is `LOCK_LIBDNF5()` needed and when is it not? I see `Goal::resolve` calls it while resolving the transaction but then `Goal::reset` does not use the lock. So can the goal be reset while is it's being resolved (even though it's unlikely a D-Bus caller would do that)?

I wonder whether it'd be better to do the locking on `get_session()` itself (and have callers always use `get_session()`) so that it's not the caller's responsibility to remember to lock. But I'm not very familiar with dnf5daemon so take that with a grain of salt. That may break opportunities for concurrency where we actually want them.

Issue #2398, opened 2025-08-19
dnf system-upgrade downloads several gigabytes to /usr/lib


jwakely:
Using the system state dir (`/usr/lib/sysimage/libdnf5/offline`) for packages downloaded during a system upgrade requires several gigabytes of free space on the root partition, which was not required by dnf4.

This causes problems for users without large amounts of free space, e.g. https://discussion.fedoraproject.org/t/dnf-system-upgrade-package-location/153958 (and I encountered the same problem myself).

Wouldn't something under /var be better for those packages? Alternatively, could something like `--downloaddir` be supported?

jwakely:
Looking at `man dnf.conf` again, it seems that `system_cachedir` would be a better default than `system_state_dir`. Short-lived storage for downloaded packages is not related to "the state of the system", which seems to be what `system_state_dir` is for, no? The downloaded packages are just needed for the duration of the offline transaction and then are removed - that's not really something "stateful".

PR #2442, opened 2025-09-18
offline: Store downloaded packages in /var/lib/dnf

pkratoch:
The persistdir is better than system_state_dir, which should be used only for storing system state, or than system_cachedir, which might not be guaranteed to survive reboot.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2398


dnf5daemon-server/session.cpp
m-blaha: Shouldn't here be also used offline_destdir? `dest_dir = offline_destdir / "packages";` ?

pkratoch: Thanks, fixed.

PR #2446, opened 2025-09-19
versionlock: Correct a help text for "dnf5 versionlock add --help"

ppisar:


ppisar:
I don't like describing versionlock with "versionlock" phrase. This definition by circle does not help explaining what it does. It becomes even more obvious when trying to translate the message. Don't you think we should use a more accessible language?

Describing package-spec-N as package-spec-N has the same problem, It only repeats what is on the left side:

~~~~
Arguments:                      
  package-spec-N                List of package-spec-N to add versionlock for
~~~~

What about this text:

~~~~
Arguments:                      
  package-spec-N                List of packages to lock their version to
~~~~

ppisar:
> List of packages to lock to their current versions

That's perfect. I used that and amended other occurrences you pointed and I found.

PR #2447, opened 2025-09-19
fix: unify epoch display across all DNF5 commands

### LINKED ISSUE #1175: Unify epoch usage in DNF5

fhbash:
Always show epoch (including 0) in package NEVRA strings for consistency. Updates Package::to_string(), transaction table, daemon API, and transaction replay.

Fixes inconsistent epoch handling where some commands showed epoch and others didn't.

Closes #1175

### DIFF
--- a/dnf5daemon-server/package.cpp
+++ b/dnf5daemon-server/package.cpp
@@ -329,10 +329,19 @@ std::string package_to_json(const libdnf5::rpm::Package & libdnf_package, const 
                 add_string_list(json_pkg, cattr, reldeplist_to_strings(libdnf_package.get_supplements()));
                 break;
             case PackageAttribute::evr:
-                json_object_object_add(json_pkg, cattr, json_object_new_string(libdnf_package.get_evr().c_str()));
+                // Always show epoch in EVR (epoch:version-release)
+                {
+                    std::string evr_with_epoch = libdnf_package.get_epoch();
+                    if (evr_with_epoch.empty()) {
+                        evr_with_epoch = "0";
+                    }
+                    evr_with_epoch += ":" + libdnf_package.get_version() + "-" + libdnf_package.get_release();
+                    json_object_object_add(json_pkg, cattr, json_object_new_string(evr_with_epoch.c_str()));
+                }
                 break;
             case PackageAttribute::nevra:
-                json_object_object_add(json_pkg, cattr, json_object_new_string(libdnf_package.get_nevra().c_str()));
+                json_object_object_add(
+                    json_pkg, cattr, json_object_new_string(libdnf_package.get_full_nevra().c_str()));
                 break;
             case PackageAttribute::full_nevra:
                 json_object_object_add(
--- a/include/libdnf5/rpm/package.hpp
+++ b/include/libdnf5/rpm/package.hpp
@@ -532,7 +532,7 @@ public:
     /// @since 5.0.5
     libdnf5::BaseWeakPtr get_base() const;
 
-    /// Return NEVRA -> 0 epoch is not shown in string
+    /// Return NEVRA -> always includes epoch (including 0 if epoch is empty)
     std::string to_string() const;
 
     /// Provide descriptive information about instance including NEVRA and ID
--- a/libdnf5-cli/output/transaction_table.cpp
+++ b/libdnf5-cli/output/transaction_table.cpp
@@ -336,7 +336,13 @@ TransactionTable::Impl::Impl(ITransaction & transaction) {
         // column. Thus adding a indentation manualy.
         scols_line_set_data(ln, COL_NAME, (" " + pkg->get_name()).c_str());
         scols_line_set_data(ln, COL_ARCH, pkg->get_arch().c_str());
-        scols_line_set_data(ln, COL_EVR, pkg->get_evr().c_str());
+        // Always show epoch in EVR (epoch:version-release)
+        std::string evr_with_epoch = pkg->get_epoch();
+        if (evr_with_epoch.empty()) {
+            evr_with_epoch = "0";
+        }
+        evr_with_epoch += ":" + pkg->get_version() + "-" + pkg->get_release();
+        scols_line_set_data(ln, COL_EVR, evr_with_epoch.c_str());
         if (tspkg->get_action() == libdnf5::transaction::TransactionItemAction::REMOVE) {
             scols_line_set_data(ln, COL_REPO, pkg->get_from_repo_id().c_str());
         } else {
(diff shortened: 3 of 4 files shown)


include/libdnf5/rpm/package.hpp
```diff
@@ -146,13 +146,15 @@ class LIBDNF_API Package {
     std::string get_arch() const;
 
     /// @return RPM package EVR (Epoch:Version-Release). If the Epoch is 0, it is omitted from the output.
+    /// @deprecated Use separate components or full NEVRA for consistent epoch display
```
kontura: Do we really want to deprecate these?
I would be fine with leaving them on the API.

kontura (commented):
I like this change but personally I would treat it as an API break.

include/libdnf5/rpm/package.hpp
fhbash: I don't have strong opinion on this, my thought to deprecated is warning API users is better to use full nevra, but I'm good to leave them like it was if it is the case

dcantrell (changes requested):
I also agree that this constitutes an API break.  Epoch is an odd one in the RPM space.  Many packagers and users do not understand it and that has led to some packages receiving Epoch updates when they should not have.  Most packages have '0' as the Epoch and then you have a number of them with crazy high epoch values.

I think if dnf5 were to start showing epoch all the time it would lead to further misuse of epoch among users and package maintainers.  I would like to see this functionality added to dnf5 as an option you can turn on via the config file, but leave the default as-is.  That brings in the new functionality but does not break the API as we have it now.  Minimal disruption.

include/libdnf5/rpm/package.hpp
fhbash: Reverted these changes as suggested.

dnf5daemon-server/package.cpp
```diff
@@ -329,10 +329,19 @@ std::string package_to_json(const libdnf5::rpm::Package & libdnf_package, const
...
                 break;
             case PackageAttribute::evr:
-                json_object_object_add(json_pkg, cattr, json_object_new_string(libdnf_package.get_evr().c_str()));
+                // Always show epoch in EVR (epoch:version-release)
+                {
+                    std::string evr_with_epoch = libdnf_package.get_epoch();
+                    if (evr_with_epoch.empty()) {
+                        evr_with_epoch = "0";
+                    }
+                    evr_with_epoch += ":" + libdnf_package.get_version() + "-" + libdnf_package.get_release();
```
dcantrell, line 338: This will lead to a NEVRA string starting with ":" if epoch is empty, or an exception if string concatenation can't turn an empty string in to "".

libdnf5-cli/output/transaction_table.cpp
```diff
@@ -336,7 +336,13 @@ TransactionTable::Impl::Impl(ITransaction & transaction) {
         // column. Thus adding a indentation manualy.
         scols_line_set_data(ln, COL_NAME, (" " + pkg->get_name()).c_str());
         scols_line_set_data(ln, COL_ARCH, pkg->get_arch().c_str());
-        scols_line_set_data(ln, COL_EVR, pkg->get_evr().c_str());
+        // Always show epoch in EVR (epoch:version-release)
+        std::string evr_with_epoch = pkg->get_epoch();
+        if (evr_with_epoch.empty()) {
+            evr_with_epoch = "0";
+        }
+        evr_with_epoch += ":" + pkg->get_version() + "-" + pkg->get_release();
```
dcantrell, line 344: See above about the leading ":".

dcantrell (approved):
Looks good to me now, thanks.

kontura:
Note also that the CI tests are failing (see the [checks](https://github.com/rpm-software-management/dnf5/pull/2447/checks)), looks like some of that are valid problems, in particular I think there is a problem with replaying transactions. However some other tests will simply need to be updated as they expect nevras without 0 epoch.

The CI is here: https://github.com/rpm-software-management/ci-dnf-stack/

Feel free to ping me if you would like some help with that.

dnf5daemon-server/package.cpp
fhbash: With this change, I believe we'll always have epoch = 0

libdnf5-cli/output/transaction_table.cpp
fhbash: See above comment

libdnf5/transaction/transaction_sr.cpp
```diff
@@ -346,8 +346,8 @@ TransactionReplay to_replay(libdnf5::transaction::Transaction & trans) {
     for (const auto & pkg : trans.get_packages()) {
         PackageReplay package_replay;
 
-        // Use to_nevra_string in order to have nevra wihtout epoch if it is 0
-        package_replay.nevra = rpm::to_nevra_string(pkg);
+        // Use to_full_nevra_string to always include epoch (including 0 if epoch is empty)
+        package_replay.nevra = rpm::to_full_nevra_string(pkg);
```
kontura: I think this is the only problematic change left.
The `PackageReplay` assumes that the nevra doesn't include epoch if its 0: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/transaction/transaction_sr.hpp#L64-L65
You either have to go through all the places where it is used and check if its ok with always present 0 (it is causing problems at least with `rpm_reason_overrides`) or it might be ok to drop this change. 
This will not affect any dnf CLI text output and we could file a separate issue to change this later on. I think it is easier to have more smaller changes than one big one.

fhbash: get it..I believe you're right, since removing this will not affect the cli output, I believe it's will be better to have this change in another Issue/PR.

I created this new issue https://github.com/rpm-software-management/dnf5/issues/2477  to work with it in the next near future, and I'll remove this change for this PR.

thanks!

kontura (commented):
Thanks.

I think that now https://github.com/rpm-software-management/ci-dnf-stack/pull/1765 needs an update. I believe that all the PR has to do is just a few smaller changes adding the 0 epoch in a couple places in these 4 files: install-package-ursine.feature, transaction-output.feature, replay-errors.feature and replay.feature.

kontura:
Thank you!

PR #2450, opened 2025-09-22
GHA: Fix weblate-sync-pot workflow

evan-goode:
This should fix the `weblate-sync-pot.yml` workflow which has been failing since... https://github.com/rpm-software-management/dnf5/pull/2330. Sorry for all the "PR run failed" emails that has caused.


m-blaha (approved):
Thank you!

PR #2453, opened 2025-09-24
Add environment install, remove and upgrade

pkratoch:
Resolves: https://github.com/rpm-software-management/dnf5/issues/724

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -413,6 +413,8 @@ Package management library.
 %{_libdir}/libdnf5.so.2*
 %dir %{_prefix}/lib/sysimage/libdnf5
 %attr(0755, root, root) %ghost %dir %{_prefix}/lib/sysimage/libdnf5/comps_groups
+%attr(0755, root, root) %ghost %dir %{_prefix}/lib/sysimage/libdnf5/comps_groups/environments
+%attr(0755, root, root) %ghost %dir %{_prefix}/lib/sysimage/libdnf5/comps_groups/groups
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/environments.toml
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/groups.toml
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/modules.toml
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -1170,10 +1170,13 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
 
         // Set correct system state for groups in the transaction
         auto comps_xml_dir = system_state.get_group_xml_dir();
-        std::filesystem::create_directories(comps_xml_dir);
+        auto comps_xml_dir_groups = comps_xml_dir / "groups";
+        auto comps_xml_dir_environments = comps_xml_dir / "environments";
+        std::filesystem::create_directories(comps_xml_dir_groups);
+        std::filesystem::create_directories(comps_xml_dir_environments);
         for (const auto & tsgroup : groups) {
             auto group = tsgroup.get_group();
-            auto group_xml_path = comps_xml_dir / (group.get_groupid() + ".xml");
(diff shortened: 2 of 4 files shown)


jrohel:
@pkratoch When I read Jan Kolarik's text in the issue: 'I remember we've already discussed in team that we should implement the env install and keep group install working with both groups and envs.' I thought that `environment install` should only accept environments. However, the implementation seems like it will behave the same as `group install` -meaning it accepts both. It's odd, but on the other hand, at least the environment and group operations will be consistent with each other. So, given the current situation, it's OK with me.

dnf5/commands/group/arguments.hpp
```diff
@@ -70,6 +70,16 @@ class GroupSpecArguments : public libdnf5::cli::session::StringArgumentList {
 };
 
 
+class CompsSpecArguments : public libdnf5::cli::session::StringArgumentList {
+public:
+    CompsSpecArguments(libdnf5::cli::session::Command & command, int nargs)
+        : StringArgumentList(
+              command, "group-spec|environment-spec", _("Pattern matching group or environment IDS."), nargs) {}
```
ppisar: Could you spell it as "IDs" or better as "identifiers"?

dnf5/commands/environment/environment_install.hpp
```diff
@@ -0,0 +1,30 @@
+// Copyright Contributors to the DNF5 project
+// SPDX-License-Identifier: GPL-2.0-or-later
```
ppisar: I think the team decided that new files in dnf5 will use LGPL-2.1-or-later license.
Or did you base this new file on group/group_install.hpp?

doc/commands/group.8.rst
```diff
@@ -40,7 +40,8 @@ related to them and it is also used for groups installation.
 To query environments use separate ``environment`` command.
 Note: ``dnf-4`` listed both environments and groups with the ``group`` command.
 
-Optional ``group-spec`` arguments could be passed to filter only groups with given names.
+Optional ``group-spec`` arguments can be passed to filter groups with given IDs or names. The install, remove and
+upgrade commands take both ``group-spec`` and ``environment-spec``.
```
ppisar: This needs to document that if there is environment and group of the same identifier, only the group will be subject of the command.

doc/commands/environment.8.rst
```diff
@@ -40,7 +40,7 @@ The ``environment`` command in ``DNF5`` offers several queries for getting infor
 about environments and groups related to them. You can install environments
 with the ``install`` command as ``install @environment-id``.
 
-Optional ``environment-spec`` arguments could be passed to filter only environments with given names.
+Optional ``environment-spec`` arguments can be passed to filter environments with given IDs or names.
```
ppisar, line 45: This dnf5-environment manual should document the install, remove, and upgrade command as dnf5-group manual does. It should also document that it accepts identifiers of both environments and grops and that environments takes precedence.

dnf5/commands/environment/environment_install.hpp
pkratoch: Alright. I don't remember exactly the outcome of the discussions, so I followed the instructions in https://github.com/rpm-software-management/dnf5/blob/main/CONTRIBUTING.md#license-header and https://github.com/rpm-software-management/dnf5/blob/main/COPYING.md which mention LGPLv2.1+ as an alternative in cases that didn't seem to apply. But I can change it.

doc/commands/environment.8.rst
pkratoch: Yeah, I forgot about this. I had to make quite big changes, both to the environment and to the group documentation, because the group command only ever mentioned groups even though environments could be installed as well.

dnf5/commands/environment/environment_upgrade.hpp
```diff
@@ -0,0 +1,30 @@
+// Copyright Contributors to the DNF5 project
+// SPDX-License-Identifier:LGPL-2.1-or-later
```
ppisar: Please add a space after the colon.

ppisar:
I believe the tests pass. Please correct the white space and I will approve this code PR.

dnf5/commands/environment/environment_upgrade.hpp
pkratoch: Fixed.

ppisar (approved):
Thanks for the tough work.

Issue #2183, opened 2025-04-07
`dnf replay` broken between 5.2.10.0 and 5.2.12.0


mtalexan:
The following is defined to work, and more complex versions of it seem to be a specifically supported workflow:
```console
dnf install --assumeyes --store=mytransaction cowsay

# disconnect from all networks

dnf replay --assumeyes mytransaction
```

Up to and including 5.2.10.0 of `dnf5` this worked as expected, using the RPMs downloaded into the `mytransaction/packages/` folder that are also referenced within the `mytransaction/transaction.json` file itself.

However in  5.2.12.0 it does not work.  Instead the `dnf replay` command ignores that the packages are already downloaded and available and attempts to download them again fresh.

mtalexan:
I can verify that if I run the `dnf install --assumeyes --store=mytransaction cowsay` command in two different (current/near current) Fedora system/containers that have the same repos enabled but have dnf5 `5.2.12.0` and `5.2.10.0` respectively, they produce the same `transaction.json` contents for both commands, and populate the `mytransaction/packages/` folder with the same RPMs. So this is purely a `dnf replay` bug.  

---

It can also be easier to test the `dnf replay` command by simply disabling all the repos rather than completely cutting network.  Or if you're doing it manually, just don't include the `--assumeyes` option and visually verify that it has 0 MB to download (or not) when it prompts you.

mtalexan:
Version `5.2.11.0` also has the same bug as `5.2.12.0`.

Tested on Fedora 41 using the available `dnf5` RPMs.  

<details>
<summary>`dnf --version` output for broken 5.2.12.0 (`dnf-5.2.12.0-1.fc41.x86_64`)</summary>
dnf5 version 5.2.12.0
dnf5 plugin API version 2.0
libdnf5 version 5.2.12.0
libdnf5 plugin API version 2.2

Loaded dnf5 plugins:
  name: builddep
  version: 1.0.0
  API version: 2.0

  name: changelog
  version: 1.0.0
  API version: 2.0

  name: config-manager
  version: 0.1.0
  API version: 2.0

  name: copr
  version: 0.1.0
  API version: 2.0

  name: needs_restarting
  version: 1.0.0
  API version: 2.0

  name: repoclosure
  version: 1.0.0
  API version: 2.0

  name: reposync
  version: 1.0.0
  API version: 2.0
</details>

<details>
[...]

PR #2456, opened 2025-09-25
Fix handling of repository IDs in stored transactions

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/2183

If there's an existing repository in the sack with a type different than `Repo::Type::COMMANDLINE`, it cannot be used to add an RPM file.

This happens for the `replay` command which can load normal available repositories.
In such cases, create and use new repos: `@stored_transaction (repo_id)`.

This has the advantage that both `offline` and `system upgrade` commands still use the original `repo_id` (because they don't load available repos). This is good because the user doesn't care that dnf is using stored transactions to implements these commands, it is an implementation detail.
On the hand when `replay` command is used the user sees both that the package was installed from a stored transaction and if present from which repo it originally came. All information is propagated to the user.

It means that the `from_repo` will contain `@stored_transaction (repo_id)` which is not a real repository, although it kind of is for the stored transactions. Still this could be confusing.

In addition this implementation "breaks" when the user disables all repos during `replay`. I am not sure if we can differentiate between `offline`/`system upgrade` and `replay` with disabled repos.

@m-blaha could you take a look? What do you think about this approach?


m-blaha:
> In addition this implementation "breaks" when the user removes repos used in the `replay`. I am not sure if we can differentiate between `offline`/`system upgrade` and `replay` with removed repos.

"breaks" here means, that the repository id is not mangled, right? So instead of `@stored_transaction (fedora)`, only `fedora` would be used (the same way as offline transaction would).

The patch is definitely step in the right direction, making the `replay` command usable again.

(I have to confess with something - I'm still thinking about duplicated repository ids mentioned in https://github.com/rpm-software-management/dnf5/issues/2183#issuecomment-2788500257 . I forget to investigate whether this duplicity can actually break something in our workflows. But it still can be done later.).

m-blaha (approved):

PR #2461, opened 2025-10-02
Prefix repository names for `--store`d transactions

kontura:
This fixes an issue of duplicate repository ids for stored transactions. 

Fixes: https://github.com/rpm-software-management/dnf5/issues/2183

This is yet another alternative to https://github.com/rpm-software-management/dnf5/pull/2456 and https://github.com/rpm-software-management/dnf5/pull/2189

In comparison to https://github.com/rpm-software-management/dnf5/pull/2456 this approach is always consistent, all stored and replayd transaction will have the repos prefixed, offline transactions are not affected.

This approach break ABI of `serialize` but the method is marked as experimental so in my opinion it is ok.

From my point of view this is the best solution so far surpassed only by somehow using only the original ids (like possibly duplicating them).

### DIFF
--- a/include/libdnf5/base/transaction.hpp
+++ b/include/libdnf5/base/transaction.hpp
@@ -173,8 +173,8 @@ public:
     /// @param comps_path If provided it is assumed all comps in this transaction are present there and
     ///                   the resulting serialized transaction will contain paths to those xml comps files
     ///                   (they can be stored using the `store_comps` method).
-    /// @param repo_prefix If provided each repository id used in this transaction will be wrapped in
-    ///                    brackets and prefix by repo_prefix.
+    /// @param repo_prefix If provided each repository id (except for @System repo) used in this transaction
+    ///                    will be wrapped in brackets and prefix by repo_prefix.
     std::string serialize(
         const std::filesystem::path & packages_path = "",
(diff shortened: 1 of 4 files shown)


ppisar:
I can see the prefix is presented to the users as well as stored to nevras.toml. I don't think this is a good idea. What about off-line updates? Will they also change the repository identifier? Do you remember how users were upset when loosing the repository origin after migrating from DNF4?

If we do this to work around assumptions in libsolv, we should remove the prefix before writing it to nevras.toml and possibly also when presenting it to the users when replying the transaction.

(Ideally replying a transaction should not load repositories into libsolv at all. Then we wouldn't have duplicate repository identifiers.)

kontura:
> > (Ideally replying a transaction should not load repositories into libsolv at all. Then we wouldn't have duplicate repository identifiers.)
> 
> This is also a valid solution. Btw, is it currently possible to store a transaction that contains both, downloaded and repository packages?

If I understand correctly that repository packages mean packages in the stored transaction but that are not downloaded then only by manually editing the stored transaction json.

In fact dnf cli cannot create a stored transaction without downloaded packages, it is possible only using the (experimental) API or by hand.

Indeed we could just drop that, don't load repositories during replay and be done. I just think it is a nice feature.

Issue #1730, opened 2024-09-25
`needs-restarting` should be able to list non-systemd processes that need restarting


evan-goode:
`dnf4 needs-restarting`, called without any args, will list processes on the system that have mmapped deleted files and need to be restarted.

Going forward, Cockpit will be using this feature of `dnf4 needs-restarting`, so DNF 5 should support this feature too.

I propose a new flag, `dnf5 needs-restarting --processes` that lists processes that need restarting. `dnf5 needs-restarting --processes` should support an additional `--exclude-services` option (as in https://github.com/rpm-software-management/dnf-plugins-core/pull/556) to filter out processes that are not managed by systemd (and would therefore already be listed by `dnf5 needs-restarting --services`).

`dnf5 needs-restarting --processes` should use the [new approach](https://github.com/rpm-software-management/dnf5/pull/887) used by `dnf5 needs-restarting --services` to determine whether a process is using outdated files, instead of looking at smaps like `dnf4 needs-restarting` does:

```
for each $process:
    find the $package that provides the binary for $process, if any
        recursively get all $dependencies of $package
        if $package or any $dependency has been updated since $process started
            recommend restart of $process
```

ppisar:
What about processes of executables that are not packaged. I would expect "dnf5 needs-restarting --processes" to catch them with inspecting memory mappings.

evan-goode:
> What about processes of executables that are not packaged? I would expect "dnf5 needs-restarting --processes" to catch them with inspecting memory mappings.

Well, yes, DNF 4 supported that. I am just not a fan of tracing file dependencies by the crude approach of inspecting mmaps, and I think we should not support it in DNF 5. There are just so many opportunities for false negative results, for example if a process reads the old version of a file into memory and then closes it, or if (early in its lifetime) a process checks what version of one of its dependencies is installed and then (later in its lifetime) changes how it calls the dependency accordingly. When the package metadata is available to determine more formally what files a given binary depends on, we are much more equipped to recommend restart/no-restart.

ppisar:
> I am just not a fan of tracing file dependencies by the crude approach of inspecting mmaps, and I think we should not support it in DNF 5.

Then it needs to be documented that DNF5 only handles processes of program owned by an installed RPM package.

[...]

Issue #2341, opened 2025-07-04
Please provide a machine readable API for `dnf needs-restarting`


jelly:
This request is based off a closed request for [dnf4](https://issues.redhat.com/browse/RHEL-56139).

The current output is very oriented to human text formatting, and is hard to parse mechanically, from e.g. Ansible or Cockpit. In Cockpit we used tracer's python API, but that's not available for `dnf needs-restarting`.

The hardest is kernel updates: After that, `dnf needs-restarting` is empty, but `--reboothint` has

```
Core libraries or services have been updated since boot-up:
  * dbus
  * kernel
  * kernel-core

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943
```

Parsing that mechanically is very error-prone. Can you please add a `--json` or similarly structured output? E.g.

```
{"type": "reboot", "packages": ["dbus", "kernel", "kernel-core"], "documentation": "https://access.redhat.com/solutions/27943"}
```

(or split it up by package to be more consistent with the other formats below -- that's not very  important)

This also applies to other updates. For example, the process list looks like this:

```
2489 : /usr/bin/podman --log-level=info system service
```

This is fairly easy to machine read still, but having this output more structured as JSON would be helpful:

```
[{"type": "process", "pid": 2489, "cmdline": ["/usr/bin/podman", "--log-level=info", "system", "service"]}, {...}, ...]
```

Likewise, {{--services}} could look like

```
[{"type": "unit", "unit": "podman.service"}, {...}, ...]
```

Thanks!

m-blaha:
Thanks for reaching out. We've already implemented JSON output for several commands, so adding this for `needs-restarting` is perfectly consistent with the rest of the dnf5.

PR #2463, opened 2025-10-06
Add --processes and --exclude-services options to the needs-restarting plugin

dcantrell:
Add the proposed --processes and --exclude-services options to the needs-restarting plugin.  Also update the man page for the plugin and write out a framework for tests for the plugin.

Please check the logic here.  The idea was similar to how the services check works, but it's checking running processes and those executables.  Running processes that no longer exist on the system shouldn't be reported, but maybe we do want to report those.

Fixes: #1730


dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -277,7 +299,18 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
         const auto start_timestamp_us =
             uint64_t{unit_proxy->getProperty("ActiveEnterTimestamp").onInterface(SYSTEMD_UNIT_INTERFACE)};
 
-        unit_file_to_service.insert(std::make_pair(fragment_path, Service{unit_name, start_timestamp_us}));
+        services.push_back(SystemdService{unit_name, start_timestamp_us, unit_object_path, fragment_path});
```
mfocko, line 331: ~~I think you should be able to use `emplace_back` here too, the type is clear:~~
```suggestion
        services.emplace_back(unit_name, start_timestamp_us, unit_object_path, fragment_path);
```

_edit_: never mind, there's no constructor for it…

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +353,223 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+                        std::string cgroup_procs_path = "/sys/fs/cgroup" + control_group + "/cgroup.procs";
+                        std::ifstream cgroup_file(cgroup_procs_path);
+                        if (cgroup_file.is_open()) {
+                            std::string pid;
+                            while (std::getline(cgroup_file, pid)) {
+                                if (!pid.empty()) {
+                                    service_pids.insert(pid);
+                                }
+                            }
+                        }
```
mfocko:
```suggestion
                        if (std::ifstream cgroup_file(cgroup_procs_path); cgroup_file.is_open()) {
                            std::string pid;
                            while (std::getline(cgroup_file, pid)) {
                                if (!pid.empty()) {
                                    service_pids.insert(pid);
                                }
                            }
                        }
```

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +353,223 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+    while ((entry = readdir(proc_dir)) != nullptr) {
+        // Check if the directory name is a number (PID)
+        if (entry->d_type != DT_DIR) {
+            continue;
+        }
+
+        const char * name = entry->d_name;
+        bool is_pid = true;
+        for (const char * c = name; *c != '\0'; ++c) {
+            if (*c < '0' || *c > '9') {
```
mfocko: Why not
```suggestion
            if (!std::isdigit(*c)) {
```

doc/dnf5_plugins/needs_restarting.8.rst
```diff
@@ -20,14 +20,14 @@
...
-#########################
+########################
+needs-restarting Command
+########################
 
 Synopsis
 ========
 
-``dnf5 needs-restarting [--services | -s]``
+``dnf5 needs-restarting [-p | --processes [-e | --exclude-services]] [-s | --services] [-r | --reboothint]``
```
mfocko, line 30: `-r` was not documented before? 👀

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: Thanks, modifying that and then running it through clang-format.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: No reason, other than I'm fighting 65 million years of C programming muscle memory.  :)

dcantrell: Making this change.

doc/dnf5_plugins/needs_restarting.8.rst
dcantrell: It was not as far as I could tell.  It was a no-op, so that may have been deliberate, but even in that case I think it should be documented both in the --help output and the man page since that is common for command line tools still carrying no-op options for backwards compatibility (e.g., see anything in GNU coreutils).

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +355,223 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+            if (install_time > process_info.start_time) {
+                processes_needing_restart.insert(exe_path);
+                // Track this package as updated so we can check reverse dependencies
+                updated_packages.add(dep);
+                break;
+            }
+        }
+    }
+
+    // Also check reverse dependencies: if any process depends on an updated package,
```
evan-goode: This section seems to do the same thing as the previous section.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +355,223 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+                        uint32_t{unit_proxy->getProperty("MainPID").onInterface(SYSTEMD_UNIT_INTERFACE)};
+                    if (main_pid > 0) {
+                        service_pids.insert(std::to_string(main_pid));
+                    }
+                } catch (const sdbus::Error & ex) {
+                    // Some services might not have a MainPID
+                    logger->debug("Failed to get MainPID for {}: {}", service.name, ex.what());
+                }
+
+                // Also get control group PIDs
```
evan-goode, line 422: The DNF4 implementation does this backwards, it iterates over processes and calls `GetUnitByPID` whereas this iterates over services and calls `MainPID`/`ControlGroup`. Without looking at the systemd source, I'm not 100% sure these are equivalent. Can a PID belong to a service according to `GetUnitByPID` without being a `MainPID` or belonging to the `ControlGroup`?

evan-goode: It may not matter for our purposes.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +355,223 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+                    for (int i = 0; i < 20; ++i) {
+                        iss >> field;
+                    }
+                    // Field 22 is starttime in clock ticks since boot
+                    unsigned long long starttime_ticks;
+                    if (iss >> starttime_ticks) {
+                        // Convert ticks to seconds and add to boot_time
+                        long ticks_per_second = sysconf(_SC_CLK_TCK);
+                        process_start_time =
+                            boot_time +
```
evan-goode, line 517: Ah, once again I am spending time thinking about the minutiae of needs-restarting boot time logic...

The process start time should always be measured from the "kernel boot time" (`btime` field of `/proc/stat`) rather than possibly systemd's `UnitsLoadStartTimestamp`. See https://github.com/rpm-software-management/dnf-plugins-core/blob/61bbaec7393f4910d7ac974b50198320e6ece353/plugins/needs_restarting.py#L301.

This may also be a good time to update this to match https://github.com/rpm-software-management/dnf-plugins-core/pull/560 regarding using btime instead of /proc/uptime.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -27,14 +28,17 @@
...
 
 #include <ctime>
 #include <fstream>
 #include <iostream>
+#include <sstream>
+#include <unordered_set>
 #include <vector>
 
-const SDBUS_SERVICE_NAME_TYPE SYSTEMD_DESTINATION_NAME{"org.freedesktop.systemd1"};
+const sdbus::ServiceName SYSTEMD_DESTINATION_NAME{"org.freedesktop.systemd1"};
```
evan-goode: Compiles on F42 if we use `std::string` here instead of `sdbus::ServiceName`.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: Yep, 100%.  Leftovers from various attempts and seeing what works and copying and pasting around.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: For me it would not build on F42 for me if this is changed to std::string.  But, I will commit it with std::string and assume the error is in my environment.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: I think technically yes, a PID can belong to a service without being a MainPID (e.g., httpd workers or postfix processes) as well as not belonging to the ControlGroup which may simply not exist for the service.  But this is just based on my understanding of how the system _probably_ works, not what systemd is doing.  However, I think you're right and that it doesn't matter in this case.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: @evan-goode I rewrote what's in this PR to match what dnf4 does.  But maybe I'm confused.  You're saying we should favor btime always, but in dnf4 it first asks systemd and failing that it falls through to st_time on /proc/1 and reading btime from /proc/stat and then taking the max (shouldn't it be min?) of those.  Do we want the systemd communication here or just strip that and leave it as reading /proc/1 and btime from /proc/stat and going with that?

dcantrell: Pushing what I have now.

evan-goode: Thanks for updating `get_boot_time` to get the `btime` field of /proc/stat.

>  You're saying we should favor btime always, but in dnf4 it first asks systemd...

What I meant was: specifically for the purpose of determining when a _process_ started, we need to measure relative to the kernel `btime` rather than any boot time we get from systemd or elsewhere. That's because the `starttime` field of /proc/pid/stat (`man 5 proc_pid_stat`) is measured by number of clock ticks after the btime. See https://github.com/rpm-software-management/dnf-plugins-core/pull/560.

IIRC the btime is updated when the system clock changes. The comment in `get_boot_time` says:

```
btime field of /proc/stat
   Reflects the time when the kernel started. Works for machines
   without RTC iff the current time is reasonably correct. Does not
   work on containers which share their kernel with the host---there,
   the host kernel uptime is returned.
```

so the btime is accurate, unless we're in a container, then the container's "boot" time will be after the kernel btime. But that doesn't matter for determining process age, since the /proc/pid/stat starttime will also be relative to the older kernel btime.

Isn't needs-restarting fun?

dcantrell: > so the btime is accurate, unless we're in a container, then the container's "boot" time will be after the kernel btime. But that doesn't matter for determining process age, since the /proc/pid/stat starttime will also be relative to the older kernel btime.

OK, gotcha now.

> Isn't needs-restarting fun?

It's something.  :smiley:

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -88,6 +114,48 @@ void NeedsRestartingCommand::configure() {
...
+            btime_stream >> tmp >> btime;
+            return btime;
+        }
+    }
+
+    if (btime == 0) {
+        logger->debug("Error reading /proc/stat, using {} as the system boot time.", btime);
+    }
+
+    logger->debug("Using {} as the system boot time.", btime);
```
evan-goode: Let's remove this line

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -320,11 +382,205 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+    for (const auto & [exe_path, process_info] : running_processes) {
+        // Recursively get all dependencies of the package that
+        // provides the executable (and include the package itself)
+        const auto & deps = recursive_dependencies(process_info.package, installed);
+        for (const auto & dep : deps) {
+            // If any dependency (or the package itself) has been
+            // updated since the process started, the process needs restart
+            const auto install_time = static_cast<time_t>(dep.get_install_time());
+            if (install_time > process_info.start_time) {
+                processes_needing_restart.insert(exe_path);
```
evan-goode: Maybe we could show the entire cmdline here? There may be multiple different processes with the same executable (e.g. electron, bash) and the user may expect to see each one.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: By this I assume you mean line 155, in which case done.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: Instead of the entire command line, would the PID make more sense?  I'm thinking about cases where needs-restarting is used in scripts and since this plugin outputs a list of processes to restart, the PID might be most helpful.  The entire command line is more distinguishing but that could also be duplicated and would require further iteration by the admin to gather a list of PIDs.  That is, if they are doing things that way (e.g., give me a list of PIDs to SIGHUP kind of thing).

evan-goode: Oh yeah, PID would be good too. I'd prefer `$pid\t$cmdline` but even PID alone is more useful than just executable name.

See also https://github.com/rpm-software-management/dnf5/issues/2341

dcantrell: I like that.  PID in column one then tab then the full command line.  That's still readable by a script but also moderately useful to a human at the terminal.  I'll work that up.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -456,7 +456,7 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
 
     // Map of executable paths to their process start times and packages
     struct ProcessInfo {
-        const char * pid;
+        std::string pid;
```
evan-goode, line 459: I made this an owned `std::string` since the `char * entry->d_name` doesn't live long enough.

evan-goode (approved):

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
dcantrell: Ah yes, thanks.  The C to C++ types still trip me up.  Thanks.

PR #2470, opened 2025-10-14
doc: document output of `--json` for `repo` command

### LINKED ISSUE #2428: Document JSON output of various commands
E.g. `dnf5 repo info --json` produces a JSON output. But meaning of the output fields is not documented.

It would be great to document JSON output for every command it can produce it. It should cover names of the fields, types the values, meaning of the values, optionality of the fields.

Based on issue #2423.

mfocko:
TODO:

- [ ] There's also
      https://github.com/rpm-software-management/dnf5/blob/2d2b92692411cbaedd589aedc37079f53198fbb4/dnf5daemon-server/package.cpp#L212
      I'm not sure how relevant is this for the documentation…
- [x] Additionally I've found `advisory` command that also supports `--json` switch, but it's not even mentioned in the docs in general that it's supported, so I'm not sure whether I should proceed with that too, or not.
- [ ] One more additional question, since there **is** an interface for printing:
     https://github.com/rpm-software-management/dnf5/blob/c87ee298457b1f19b156a776d4cb1872ae5ff7ba/include/libdnf5-cli/output/repo_info.hpp#L38
     I wonder why the JSON output is also not part of the interface

Fixes #2428

### DIFF
--- a/doc/commands/repo.8.rst
+++ b/doc/commands/repo.8.rst
@@ -73,3 +73,65 @@ Examples
 ``dnf5 config-manager setopt repo_id.enabled=0``
     | Persistently disable repository using the config-manager plugin command.
     | See :manpage:`dnf5-config-manager(8)` for more details.
+
+JSON Output
+===========
+
+* ``dnf5 repo list --json``
+
+  The command returns a JSON array of objects, each describing one repository.
+  Each object contains the following fields:
+
+  - ``id`` (string) - Repository ID.
+  - ``name`` (string) - Repository name.
+  - ``is_enabled`` (boolean) - Repository status, either ``true`` (enabled),
+    or ``false`` (disabled).
+
+* ``dnf5 repo info --json``
+
+  The command returns a JSON array of objects, each describing one repository.
+  Each object contains the following fields:
+
+  - ``id`` (string) - Repository ID.
+  - ``name`` (string) - Repository name.
+  - ``is_enabled`` (boolean) - Repository status, either ``true`` (enabled),
+    or ``false`` (disabled).
+  - ``priority`` (integer) - Repository priority.
+  - ``cost`` (integer) - Repository cost.
+  - ``type`` (string) - Repository type.
+  - ``exclude_pkgs`` (array of strings) - List of excluded packages.
+  - ``include_pkgs`` (array of strings) - List of included packages.
+  - ``timestamp`` (integer) - Timestamp of the last metadata update,
+    UNIX time.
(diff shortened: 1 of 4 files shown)


doc/commands/repo.8.rst
```diff
@@ -73,3 +73,57 @@ Examples
...
+    UNIX time.
+  - ``metadata_expire`` (integer) - Metadata expiration time. If not set, value
+    is taken from global config.
+  - ``skip_if_unavailable`` (boolean) - Whether to skip the repository
+    if it is unavailable.
+  - ``repo_file_path`` (string) - Path to the repository file.
+  - ``base_url`` (array of strings) - List of base URLs.
+  - ``metalink`` (string) - Metalink URL.
+  - ``mirrorlist`` (string) - Mirrorlist URL.
+  - ``gpg_key`` (array of strings) - List of GPG keys.
```
ppisar: Please write "OpenPGP" instead of "GPG" to be consistent with the rest of DNF5.

doc/commands/repo.8.rst
```diff
@@ -73,3 +73,57 @@ Examples
...
+  - ``repo_gpgcheck`` (boolean) - Whether to perform GPG check
+    of the repository metadata.
+  - ``pkg_gpgcheck`` (boolean) - Whether to perform GPG check
+    of the packages.
+  - ``available_pkgs`` (integer; 64-bit unsigned) - Number of available packages
```
ppisar: Is there a reason why to claim it is 64-bit? JSON integers are unlimited.

doc/commands/repo.8.rst
```diff
@@ -73,3 +73,57 @@ Examples
...
+  - ``exclude_pkgs`` (array of strings) - List of excluded packages.
+  - ``include_pkgs`` (array of strings) - List of included packages.
+  - ``timestamp`` (integer; 64-bit) - Timestamp of the last metadata update,
+    UNIX time.
+  - ``metadata_expire`` (integer) - Metadata expiration time. If not set, value
+    is taken from global config.
+  - ``skip_if_unavailable`` (boolean) - Whether to skip the repository
+    if it is unavailable.
+  - ``repo_file_path`` (string) - Path to the repository file.
+  - ``base_url`` (array of strings) - List of base URLs.
```
ppisar: Here could be noted that the URLs are "effective" or have "expanded variables". In contrast to what one writes into repository configuration.

ppisar (commented):
It looks good. But please squash the two commits. We do not want to have garbage in git history.

PR #2475, opened 2025-10-20
dnf5: Argument "--from-vendor=" for commands

jrohel:
The new `--from-vendor` argument allows users to run actions on available packages sourced only from the specified vendors.

Key behavior notes:
- Vendor Bypass: The vendor change check is bypassed for packages explicitly targeted by this option.
- Dependency Resolution: For items required only to satisfy dependencies, the vendor is ignored, or existing vendor change policies (controlled by 'allow_vendor_change=0') are used.
- Syntax: Multiple vendors can be specified, separated by commas, and globs can be used (e.g., 'vendorA, vendorB*, vendorC').

Applicable commands:
`do`, `install`, `upgrade`, `reinstall`, `downgrade`, `distro-sync`, `swap`, `buildep`, `download`

Usage is similar to `--from-repo`:
- `do` command: The argument applies only to the immediately following package arguments and may be defined multiple times.
- Other commands: The argument applies to all package arguments and cannot be defined multiple times with different values.

### DIFF
--- a/doc/_shared/options/from-vendor.rst
+++ b/doc/_shared/options/from-vendor.rst
@@ -1,0 +1,5 @@
+``--from-vendor=VENDOR,...``
+    | List of vendors separated by commas or spaces. Globs are supported (e.g., ``--from-vendor='Fedora\ project, vendorA, vendorB*, vendorC'``).
+    | Packages (or their provides) explicitly specified on the command line will only be looked up from the specified vendors.
+    | The vendor change check is **bypassed** for these packages.
+    | The vendor is ignored or vendor change policies (if ``allow_vendor_change=0``) will still be used for packages that satisfy dependencies.
--- a/doc/commands/distro-sync.8.rst
+++ b/doc/commands/distro-sync.8.rst
@@ -56,6 +56,8 @@ Options
 
 .. include:: ../_shared/options/from-repo.rst
 
+.. include:: ../_shared/options/from-vendor.rst
+
 ``--downloadonly``
     | Download the resolved package set without executing an RPM transaction.
 
--- a/doc/commands/do.8.rst
+++ b/doc/commands/do.8.rst
@@ -87,6 +87,8 @@ Options
 
 .. include:: ../_shared/options/from-repo.rst
 
+.. include:: ../_shared/options/from-vendor.rst
+
 ``--downloadonly``
     | Download the resolved package set without executing an RPM transaction.
 
--- a/doc/commands/downgrade.8.rst
+++ b/doc/commands/downgrade.8.rst
@@ -60,6 +60,8 @@ Options
 
 .. include:: ../_shared/options/from-repo.rst
 
+.. include:: ../_shared/options/from-vendor.rst
+
 ``--downloadonly``
     | Download the resolved package set without executing an RPM transaction.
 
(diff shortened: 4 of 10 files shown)


doc/commands/distro-sync.8.rst
```diff
@@ -56,6 +56,8 @@ Options
 
 .. include:: ../_shared/options/from-repo.rst
 
+.. include:: ../_shared/options/from-vendor.rst
```
evan-goode, line 59: Did you forget to commit `from-vendor.rst`?

libdnf5/base/goal.cpp
```diff
@@ -1451,6 +1477,34 @@ std::pair<GoalProblem, libdnf5::solv::IdQueue> Goal::Impl::add_install_to_goal(
         query |= installed;
     }
 
+    if (!to_vendors.empty()) {
+        query.filter_vendor(to_vendors, sack::QueryCmp::GLOB);
+        if (query.empty()) {
+            transaction.p_impl->add_resolve_log(
+                action,
+                GoalProblem::NOT_FOUND_IN_REPOSITORIES,  // TODO(jrohel): NOT_FOUND_FROM_VENDOR
```
evan-goode, line 1496: How much additional effort would it be to add `NOT_FOUND_FROM_VENDOR`? The error message

```
# dnf5 distro-sync --from-vendor=nonexistent mypackage
No match for argument 'mypackage' in repositories ''
```

is not ideal.

evan-goode (changes requested):
Works great! It's a bit counterintuitive that escaping the spaces is needed, since many vendors have spaces:

```
--from-vendor 'Fedora\ Project'
```

Let's make sure this is clear in the documentation.

doc/commands/distro-sync.8.rst
jrohel: Yes, I did. I have added the missing file now.

doc/_shared/options/from-vendor.rst
```diff
@@ -0,0 +1,5 @@
+``--from-vendor=VENDOR,...``
+    | List of vendors separated by commas or spaces. Globs are supported (e.g., ``--from-vendor='Fedora\ project, vendorA, vendorB*, vendorC'``).
+    | Packages (or their provides) explicitly specified on the command line will only be looked up from the specified vendors.
+    | The vendor change check is **bypassed** for these packages.
+    | The vendor is ignored or vendor change policies (if allow_vendor_change=0) will still be used for packages that satisfy dependencies.
```
evan-goode: nit: allow_vendor_change=0 should be double-backticked

jrohel: Thanks, I added backticks.

libdnf5/base/goal.cpp
jrohel: I know. I divided the work on sticky vendor support into several parts. I designed and implemented "Supports vendor change policies" (https://github.com/rpm-software-management/dnf5/pull/2469). This PR addresses the "--from-vendor" functionality. Additionally, I'm tackling other minor issues I encountered along the way (https://github.com/rpm-software-management/dnf5/pull/2485 and https://github.com/rpm-software-management/dnf5/pull/2502). And now I'm getting to the next part. And that is the reporting of vendor problems.

evan-goode: OK, we can add it later.

Issue #1581, opened 2024-07-15
`group upgrade` should upgrade group packages even if they are not installed as part of the group


kontura:
`$ dnf5 group upgrade test-group` should upgrade group packages that belong to `test-group` even if they are not installed as part of the group.

This would match dnf4 behavior.

PR #2482, opened 2025-10-23
When upgrading group, upgrade all packages that are part of it

pkratoch:
Instead of upgrading only packages that were installed with the group, upgrade all packages that are part of the group even if they were not originally installed with it.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1581

### DIFF
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -2684,9 +2684,6 @@ void Goal::Impl::add_group_upgrade_to_goal(
     comps::GroupQuery available_groups(base);
     available_groups.filter_installed(false);
 
-    rpm::PackageQuery query_installed(base);
-    query_installed.filter_installed();
-
     for (auto installed_group : group_query) {
         auto group_id = installed_group.get_groupid();
         // find available group of the same id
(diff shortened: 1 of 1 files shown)


libdnf5/base/goal.cpp
```diff
@@ -2725,9 +2722,9 @@ void Goal::Impl::add_group_upgrade_to_goal(
             old_set.emplace(pkg.get_name());
         }
         // set of package names that are part of the available version of the group
-        std::set<std::string> new_set{};
+        std::vector<std::string> new_set{};
```
mfocko:
```suggestion
        std::vector<std::string> new_set;
```

No need for the brackets, default constructor should do :)

libdnf5/base/goal.cpp
```diff
@@ -2725,9 +2722,9 @@ void Goal::Impl::add_group_upgrade_to_goal(
...
         for (const auto & pkg : available_group.get_packages()) {
```
mfocko: It would probably be better to provide size hint to avoid repeated reallocations, but not sure if it's worth it:
```suggestion
        const auto & pkgs_from_available_group = available_group.get_packages();

        std::vector<std::string> new_set;
        new_set.reserve(pkgs_from_available_group.size());
        for (const auto & pkg : pkgs_from_available_group) {
```

Checked locally, starts with no space pre-allocated after construction and then behaves as classical dynamic array. I'm for pre-allocating the memory.

pkratoch: Makes sense, I changed it according to your suggestion. Thanks!

Issue #1483, opened 2024-05-11
[Feature Request][dnf5][librepo] Allow filtering metalinks by location/domain


bc-lee:
As I understand, Fedora's dnf uses metalinks by default. It downloads the metalink file and then selects the fastest mirror from the list. This is great for most users, but it would be beneficial to have the option to filter mirrors by location/domain. For example, this would allow users to exclude mirrors known to be out of sync or slow.

From a configuration standpoint, adding options in the `main` section of `/etc/dnf/dnf.conf` would be ideal, as users want to control all repositories at once. An example configuration is shown below:

```
[main]
gpgcheck=True
...
metalink_exclude_domain=^mirror1\.example\.com$
# XX is an artificial country code for testing
metalink_exclude_location=^XX$
```

To achieve this functionality, `librepo` would need to be updated to allow filtering by location/domain as well.

ppisar:
As far as I know Fedora monitors mirrors and excludes bad ones from metalink document. However, I can see that this feature can be beneficial if there is a network problem between the host and the mirror. Or when the metalink generator does not monitor mirrors properly.

PR #2484, opened 2025-10-24
Add metalink exclude options to configuration

bc-lee:
This update introduces two new configuration options: `metalink_exclude_domain` and `metalink_exclude_location`. These options allow users to specify domains and locations to exclude when using metalink sources.

Requires: https://github.com/rpm-software-management/librepo/pull/357
Fixes: https://github.com/rpm-software-management/dnf5/issues/1483


ppisar:
CMakeList needs to check for availability of this feature (by probing the interface or checking a minimal version) and the spec file also needs specify the minimal librepo version for build- and run-time.

Issue #2483, opened 2025-10-24
Missing Escape Mechanism for Delimiters in Option Lists


jrohel:
For example, the command `dnf --setopt=reposdir='/home/ninja/my repos/' repo list` currently fails.

Instead of correctly interpreting the full path `"/home/ninja/my repos/"` as a single value for reposdir, the parser incorrectly splits it into two separate paths: `"/home/ninja/my\"` and `"repos/"`. This breakage occurs because the space is treated as a delimiter, even when enclosed in quotes.

Currently, both space (` `) and comma (`,`) act as delimiters within option lists. This prevents us from correctly including these characters within a option value.

PR #2485, opened 2025-10-24
Implement escape character support in OptionStringContainer::from_string

jrohel:
The parsing logic in `from_string` has been rewritten to handle the escape character (`\`).

Unit tests for escape character were added.

Closes: https://github.com/rpm-software-management/dnf5/issues/2483

### DIFF
--- a/test/libdnf5/conf/test_option.cpp
+++ b/test/libdnf5/conf/test_option.cpp
@@ -494,6 +494,41 @@ void OptionTest::test_options_string_list_custom_delimiters() {
     CPPUNIT_ASSERT_EQUAL((std::vector<std::string>{"Dfirstx", "DsecondX"}), option3.get_value());
 }
 
+void OptionTest::test_options_string_list_escape() {
+    OptionStringList option("Dfirstx\\,DsecondX");
+    CPPUNIT_ASSERT_EQUAL((std::vector<std::string>{"Dfirstx,DsecondX"}), option.get_value());
+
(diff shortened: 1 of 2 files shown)


jrohel:
I re-committed the tests. Only a change in a comment. These are unit tests, not CI tests.

libdnf5/conf/option_string_list.cpp
```diff
@@ -197,35 +197,70 @@ void OptionStringContainer<T, IsAppend>::test(const ValueType & value) const {
...
+        }
 
-    for (str = value.data();; str = nullptr) {
-        token = strtok_r(str, get_delimiters(), &saveptr);
-        if (token == nullptr) {
-            break;
+        if (*it == '\\') {
+            escape = true;
+            continue;
+        }
```
m-blaha, line 225: The doc string says that the backslash "can be used to escape delimiters and spaces to include them literally". But in fact it is used to escape **any** character (not only delimiters). And there is a bit unclear edge case with a trailing backslash, which is ignored - `from_string("fir\st,second\") => ["first", "second"]`.
I think the current implementation is valid, but the documentation should be clarified.

jrohel: Thanks. I changed the doc string to reflect that any character is escaped.
I also added a commit mentioning the escaping in man pages for the `list` option type.

m-blaha (approved):

Issue #2129, opened 2025-03-12
Downloaded packages unexpectedly removed by a parallel transaction


m-blaha:
If two transactions run in parallel - such as when DNF Automatic is configured - the RPM packages downloaded by one transaction may be removed by the other.

1. The first transaction is in the download phase:

```
❯ sudo dnf upgrade
Package                                            Arch   Version                   Reposit      Size
Upgrading:
 SDL2_image                                        x86_64 2.8.6-1.fc40              updates 217.9 KiB
   replacing SDL2_image                            x86_64 2.8.2-4.fc40              <unknow 213.3 KiB
 cmake                                             x86_64 3.30.8-1.fc40             updates  32.4 MiB
   replacing cmake                                 x86_64 3.30.7-1.fc40             updates  32.4 MiB
...
...
Transaction Summary:
 Upgrading:         54 packages
 Replacing:         54 packages

Total size of inbound packages is 186 MiB. Need to download 186 MiB.
After this operation, 283 KiB extra will be used (install 619 MiB, remove 619 MiB).
Is this ok [y/N]: y
[ 1/54] SDL2_image-0:2.8.6-1.fc40.x86_64         100% | 266.3 KiB/s | 109.4 KiB |  00m00s
[ 2/54] cmake-filesystem-0:3.30.8-1.fc40.x86_64  100% | 305.4 KiB/s |  16.8 KiB |  00m00s
[ 3/54] cmake-rpm-macros-0:3.30.8-1.fc40.noarch  100% | 159.4 KiB/s |  16.3 KiB |  00m00s
...
```

2. A second transaction is started in another terminal window and successfully completes before the first transaction finishes downloading.

  
```
❯ sudo dnf install acpi
Package                     Arch     Version                     Repository          Size
Installing:
 acpi                       x86_64   1.7-23.fc40                 fedora          47.5 KiB

Transaction Summary:
 Installing:         1 package

Total size of inbound packages is 27 KiB. Need to download 0 B.
After this operation, 47 KiB extra will be used (install 47 KiB, remove 0 B).
Is this ok [y/N]: y
[1/1] acpi-0:1.7-23.fc40.x86_64                  100% |   0.0   B/s |   0.0   B |  00m00s
-----------------------------------------------------------------------------------------
[1/1] Total                                      100% |   0.0   B/s |   0.0   B |  00m00s
Running transaction
[1/3] Verify package files                       100% | 500.0   B/s |   1.0   B |  00m00s
[...]

Issue #1640, opened 2024-08-20
dnf5 deletes cache of running dnf5


dschwoerer:
As a user with a slow download link, I can start 
`dnf5 install big-package(s)`
While the download is running, I realise I also need another package:
`dnf5 install small-package`
Then the second instance finishes the download first, and after a successful transaction deletes all the files, which in turn breaks the first instance.

dnf5 should either not start another instance, or it should not delete the cache files of a different, running instance.

This is especially annoying as downloading some big packages with many dependencies can take quite some time, and installing a small script requires a new attempt. If I remember correctly the download keeps running, but the install fails afterwards.

evan-goode:
I can reproduce this issue, e.g. with `dnf5 install -y blender` and `dnf5 install -y hello`. 

DNF 4 is not affected by this bug; the second DNF 4 process will wait until the first DNF 4 process has finished before starting to download RPMs for its transaction.

> dnf5 should either not start another instance, or it should not delete the cache files of a different, running instance.

I'd prefer if DNF 5 took the same approach that DNF 4 does. The implementation will likely be simpler and we will not stress the mirrors by having multiple DNF 5 processes try to download at once.

ppisar:
Locking itself won't fix it. Unless you mean a big lock which would prevent from concurrent "dnf install --downloadonly" commands.

The original report highlights DNF's (mis)feature of removing "unneeded" downloaded packages from a cache after finishing a transaction. "Unneeded" now means all packages. This needs to change to "installed in the successful transaction". That means that downloaded, but never used packages will remain. Garbage-collecting them should be a job of an explicit "dnf clean packages" command. Alternatively, DNF could automatically remove them after lapsing a configurable, long enough, e.g. 6 months, expiration period.

PR #2488, opened 2025-10-27
Improve sharing of temporary_files.toml among multiple processes

evan-goode:
Related: https://github.com/rpm-software-management/dnf5/issues/2129.

This PR implements Petr's suggestion here:

https://github.com/rpm-software-management/dnf5/issues/1640#issuecomment-2934132836

> The original report highlights DNF's (mis)feature of removing "unneeded" downloaded packages from a cache after finishing a transaction. "Unneeded" now means all packages. This needs to change to "installed in the successful transaction". That means that downloaded, but never used packages will remain. Garbage-collecting them should be a job of an explicit "dnf clean packages" command.

After deleting the inbound RPMs, we also remove old entries in `temporary_files.toml` corresponding to files that no longer exist. This prevents `temporary_files.toml` from growing too large if the RPMs are removed by some other means, e.g. `dnf5 clean all`.

I added some locking for `temporary_files.toml`, keeping in mind that multiple processes might try to read/write it simultaneously.

### DIFF
--- a/libdnf5/base/transaction.cpp
+++ b/libdnf5/base/transaction.cpp
@@ -902,15 +902,6 @@ void Transaction::Impl::process_scriptlets_output(int fd) {
     close(fd);
 }
 
-static bool contains_any_inbound_package(std::vector<TransactionPackage> & packages) {
-    for (const auto & package : packages) {
-        if (transaction_item_action_is_inbound(package.get_action())) {
-            return true;
-        }
-    }
-    return false;
-}
-
 Transaction::TransactionRunResult Transaction::Impl::test() {
     return this->_run(std::make_unique<libdnf5::rpm::TransactionCallbacks>(), "", std::nullopt, "", true);
 }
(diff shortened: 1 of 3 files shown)


libdnf5/base/transaction.cpp
```diff
@@ -1258,24 +1258,46 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
...
         auto keepcache = config.get_keepcache_option().get_value();
         auto any_inbound_action_present = contains_any_inbound_package(packages);
         if (!keepcache && any_inbound_action_present) {
+            // Find inbound package files
+            std::set<std::string> inbound_package_files;
+            for (const auto & tspkg : packages) {
+                if (transaction_item_action_is_inbound(tspkg.get_action())) {
+                    inbound_package_files.insert(tspkg.get_package().get_package_path());
+                }
+            }
```
pkratoch: We could check if `inbound_package_files` is empty instead of calling `contains_any_inbound_package` to not iterate over the packages twice.

libdnf5/base/transaction.cpp
```diff
@@ -1258,24 +1258,46 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
...
+            // Get all temp files
             libdnf5::repo::TempFilesMemory temp_files_memory(base, config.get_cachedir_option().get_value());
-            auto temp_files = temp_files_memory.get_files();
-            for (auto & file : temp_files) {
+            const auto & temp_files = temp_files_memory.get_files();
+            const std::set<std::string> temp_files_set(temp_files.begin(), temp_files.end());
+
+            // Remove all nonexistent files from the temp files listing
+            std::vector<std::string> removed;
+            for (const auto & file : temp_files_set) {
```
pkratoch: What is the `temp_files_set` for? Why not iterate directly the `temp_files`?

libdnf5/base/transaction.cpp
```diff
@@ -1258,24 +1258,46 @@ Transaction::TransactionRunResult Transaction::Impl::_run(
...
+            }
+
+            // Get all temp files
             libdnf5::repo::TempFilesMemory temp_files_memory(base, config.get_cachedir_option().get_value());
-            auto temp_files = temp_files_memory.get_files();
-            for (auto & file : temp_files) {
+            const auto & temp_files = temp_files_memory.get_files();
+            const std::set<std::string> temp_files_set(temp_files.begin(), temp_files.end());
+
+            // Remove all nonexistent files from the temp files listing
```
pkratoch: This comment is a bit confusing here, because first the existing files are removed and the removal of nonexistent files from the listing happens later. I suggest to add a comment that describes the removal of existing files for inbound packages and move this one to the `temp_files_memory.remove_files(removed);` line.

libdnf5/base/transaction.cpp
evan-goode: Thanks, that was left over from an previous commit.

libdnf5/base/transaction.cpp
evan-goode: Nice, done. I removed `contains_any_inbound_package`.

libdnf5/base/transaction.cpp
evan-goode: Done, I also modified it to always clean out entries of nonexistent files even when there are no inbound RPMs to remove or `keepcache` is true.

evan-goode:
I changed it to remove `temporary_files.toml` when there are no longer any entries. Now that test is passing.

And I removed the locking of `temporary_files.toml`, I realized that it will be better to lock the entire cachedir instead of that single file. We can do that (and address the problem of two processes using the same package file) in a follow-up PR.

I dropped the commit to add the `keep_file` option for `Locker` but kept the new functions for blocking locks, I believe we will want those shortly.

Issue #2487, opened 2025-10-27
Preserve ACL and mode while rotating logs


ppisar:
When /var/log/dnf5.log is rotated by DNF5, POSIX access control list is lost. I.e. The new, empty /var/log/dnf5.log file does not have the same ACL as the old, big one.

dnf5-5.3.0.0-0.20251026011528.5.2.17.0+85.gd7e1192b.fc44.x86_64 is affected.

Corresponding DNF4 bug report: https://github.com/rpm-software-management/dnf/issues/2279

Issue #1820, opened 2024-11-01
Log rotation doesn't keep file permissions


pkratoch:
In dnf4, when the log was rotated, it kept the file permissions of the log file and the new log file was created with the same permissions. Dnf5 sets 644 for all logs every time the log is rotated.

PR #2494, opened 2025-10-31
rotating logger: Preserve file mode and ACL

ppisar:
If a superuser set a custom access control list (ACL) on a /var/log/dnf5.log file and DNF5 was configured to rotate the logs, the ACL did not survive the rotation. Newly created log file was lacking the list:

    root@fedora-44:/tmp # setfacl -m user:test:rw /var/log/dnf5.log
    root@fedora-44:/tmp # getfacl -c /var/log/dnf5.log
    getfacl: Removing leading '/' from absolute path names
    user::rw-
    user:test:rw-
    group::r--
    mask::rw-
    other::r--

    root@fedora-44:/tmp # dnf5 --dump-main-config | grep log_rotate
    log_rotate = 4
    root@fedora-44:/tmp # getfacl -c /var/log/dnf5.log
    getfacl: Removing leading '/' from absolute path names
    user::rw-
    group::r--
    other::r--

The same problem was if the superuser changed the file mode.

This patch fixes it by copying the file mode and ACL from the old log file to the new log file on the rotation. A newly required dependency on libacl can be disabled at build-time with -DWITH_ACL=OFF CMake option.

Implementation detail: I used libacl for handling ACL. Alternative approach would be copying extended attributes with getxattr(2) etc. I chose libacl because I'm not sure that all file systems, especially XFS, implement ACL with extended attributes.

Implementation detail: This patch uses GNU-specific renameat2() function to atomically swap two log files. That enables DNF5 to rotate the logs without a gap when there is no /var/log/dnf5.log file present on the file system.

Resolve: #2487
Resolve: #1820

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1779


jrohel:
As I previously mentioned, I consider using `libacl` to be a good idea. I recommend sticking with it.

libdnf5/logger/rotating_file_logger.cpp
```diff
@@ -73,6 +83,31 @@ RotatingFileLogger::Impl::~Impl() {
...
+namespace {
+// Copy file mode and POSIX access control list from a file descriptor to
+// a file descriptor.
+// @return true on success, false otherwise.
+bool copy_mode(int from, int to) {
+    // File mode
+    struct stat stat;
+    if (fstat(from, &stat))
+        return false;
+    bool success = (0 == fchmod(to, 07777 & stat.st_mode));
```
dcantrell: I would prefer a more readable style here personally.  Define ```bool success = false``` at the beginning of the function and then just use an if block for this:

```
if (fchmod(to, 07777 & stat.st_mode) == 0)
    success = true;
```

libdnf5/logger/rotating_file_logger.cpp
```diff
@@ -73,6 +83,31 @@ RotatingFileLogger::Impl::~Impl() {
...
+#ifdef HAVE_LIBACL
+    // ACL
+    auto acl = acl_get_fd(from);
+    if (!acl) {
+        // Handle file systems which do not support ACL gracefully
+        success &= errno == ENOTSUP;
```
dcantrell: Same thing here, I would prefer to see the simpler:

```
if (errno == ENOTSUP)
    success = true;
```

libdnf5/logger/rotating_file_logger.cpp
```diff
@@ -73,6 +83,31 @@ RotatingFileLogger::Impl::~Impl() {
...
+        return success;
+    }
+    success &= (0 == acl_set_fd(to, acl));
```
dcantrell: And here:

```
if (acl_set_fd(to, acl) == 0)
    success = true;
```

ppisar:
I corrected handling a case when log_rotate option is 1 and dnf5.log.1 file already exists and is not empty.

libdnf5/logger/rotating_file_logger.cpp
ppisar: That wouldn't be equivalent. I want the function to report a failure when any operation fails: If fchmod() fails and ACL is unsupported, it should return false.

I chose the arithmetic notation over branching everywhere, because here and then it is in my opinion more legible then a maze of nested if-condition-assignment that would be otherwise required.

I can rewrite it, but are you sure it will be better?

ppisar: I had to invert to default `success` value. Then the branching is not so horrible.

dcantrell: Oh, I see now.  I think what you needed then was:

```
success &= ~(errno == ENOTSUP);
```

But that doesn't matter now, you've refactored it and I find it easier to read.  Thanks!

dcantrell (approved):

PR #2495, opened 2025-10-31
advisory: Add JSON documentation, standardize timestamps, and suppress error messages in JSON format

### LINKED ISSUE #2476: Document JSON output of `advisory` command
Document the JSON output of `dnf advisory` (both `list` and `info`). Opening this issue as a follow-up from the PR, since the `--json` option for `advisory` is not even mentioned in the manpage (except for the pre-generated help).

Also the “schema” of the JSON looks a lot different to the `repo` command.

Relevant parts of the sources:

https://github.com/rpm-software-management/dnf5/blob/6a8c404f4d8fd135540a99d3b28bba866f9a7348/libdnf5-cli/output/advisoryinfo.cpp#L99-L152

https://github.com/rpm-software-management/dnf5/blob/6a8c404f4d8fd135540a99d3b28bba866f9a7348/libdnf5-cli/output/advisorylist.cpp#L118-L169

https://github.com/rpm-software-management/dnf5/blob/6a8c404f4d8fd135540a99d3b28bba866f9a7348/libdnf5-cli/output/advisorylist.cpp#L210-L227

---

Related to #2428
Follow-up of #2470

### LINKED ISSUE #2493: Change timestamp in advisory JSON output from ISO to UNIX format
`dnf5 advisory list --json` prints time stamp in ISO format now:

~~~~
# dnf5 advisory list --json 
[
  {
    "name":"FEDORA-2025-793d1d7f22",
    "type":"bugfix",
    "severity":"Low",
    "nevra":"glibc-debuginfo-2.41-11.fc42.x86_64",
    "buildtime":"2025-08-28 00:50:46"
  },
[...]
~~~~

I request changing the format (`buildtime` keyword) to UNIX format (number of seconds since beginning of the UNIX epoch).

Reasons:
* We already have UNIX format on different places (`dnf5 repo info --json`, `dnf5 history info --json`).
* JSON is for machines and if a machine wants to understand the time, it needs to parse it (usually to UNIX time).
* The current format is actually not conforming to ISO. ISO requires `T` separator between the date and time.

Userfrom1995:
Add JSON format documentation for advisory commands, standardize timestamp
format, and implement enum-based error handling to maintain consistency
 with other JSON command outputs.

 Changes:
- Add JSON documentation for advisory list and info commands
- Change timestamps from ISO strings to UNIX integers for consistency
- Change advisory info JSON output from object to array format
- Replace boolean error handling flags with ErrorHandling enum
- Add determine_error_mode() helper function for centralized error mode selection
- Suppress error messages in JSON mode for clean programmatic output

Fixes #2476
Fixes #2493

### DIFF
--- a/dnf5/commands/advisory/advisory_info.cpp
+++ b/dnf5/commands/advisory/advisory_info.cpp
@@ -68,7 +68,6 @@ void AdvisoryInfoCommand::process_and_print_queries(
             advisory_info.add_advisory(cli_advisory);
         }
         advisory_info.print();
-        std::cout << std::endl;
     } else {
         for (auto advisory : advisories) {
             libdnf5::cli::output::AdvisoryInfo advisory_info;
--- a/dnf5/commands/advisory/advisory_subcommand.cpp
+++ b/dnf5/commands/advisory/advisory_subcommand.cpp
@@ -81,6 +81,7 @@ void AdvisorySubCommand::configure() {
 void AdvisorySubCommand::run() {
     auto & ctx = get_context();
 
+    ErrorHandling error_mode = determine_error_mode(ctx, false);
     auto advisories_opt = advisory_query_from_cli_input(
         ctx.get_base(),
         advisory_specs->get_value(),
@@ -91,7 +92,7 @@ void AdvisorySubCommand::run() {
         advisory_severity->get_value(),
         advisory_bz->get_value(),
         advisory_cve->get_value(),
-        false);
+        error_mode);
 
     auto advisories = advisories_opt.value_or(libdnf5::advisory::AdvisoryQuery(ctx.get_base()));
 
--- a/dnf5/commands/advisory_shared.hpp
+++ b/dnf5/commands/advisory_shared.hpp
@@ -24,6 +24,7 @@
 
 #include "utils/string.hpp"
 
+#include <dnf5/context.hpp>
 #include <libdnf5-cli/session.hpp>
 #include <libdnf5/advisory/advisory_query.hpp>
 #include <libdnf5/base/base_weak.hpp>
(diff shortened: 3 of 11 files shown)


Userfrom1995:
I have changed the output format of the advisory info from object to array because that's what other commands do to maintain consistency. If I should change it back, please let me know.

Userfrom1995:
I also noticed an inconsistency with error messages in JSON mode. Advisory commands print error messages before JSON output, while other commands don't:

**Advisory commands (inconsistent):**
```bash
dnf5 advisory list nonexistent --json
No advisory found matching the requested name: "nonexistent"
[]

dnf5 advisory info nonexistent --json  
No advisory found matching the requested name: "nonexistent"
[]

dnf5 history list 999 --json
[]

dnf5 history info 999 --json
[]

dnf5 repo list nonexistent --json
[]

dnf5 repo info nonexistent --json
[]
```
I just noticed this and I'm not sure if it's an issue or not with the API or with the output, but I think it was worth putting it here, so I'm pointing it out.

Userfrom1995:
I believe it is the change in the output of advisory info from object to array that is causing the test failures. Should I revert back? If that's the case, please let me know.

kontura:
I have created a CI tests update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1778

kontura:
> I believe it is the change in the output of advisory info from object to array that is causing the test failures. Should I revert back? If that's the case, please let me know.

Its also caused by the timestamp format, I have created a tests [update](https://github.com/rpm-software-management/ci-dnf-stack/pull/1778).

Since we are doing incompatible changes anyway I personally think it is a good consistency improvement.



> I also noticed an inconsistency with error messages in JSON mode. Advisory commands print error messages before JSON output, while other commands don't:

Oh right, this is a special advisory message: https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/advisory_shared.hpp#L75-L76. I believe it probably shouldn't be printed in the `--json` mode.

kontura:
> Should I also include the change to skip printing that message in --json mode in this PR?
> I can update the commit and PR description accordingly.

It would be great if you did. 

Briefly thinking about the implementation perhaps we could turn the `const bool strict` argument of `advisory_query_from_cli_input` into an enum?

Userfrom1995:
> > Should I also include the change to skip printing that message in --json mode in this PR?
> > I can update the commit and PR description accordingly.
> 
> It would be great if you did.
> 
No worries, I will do it.

> Briefly thinking about the implementation perhaps we could turn the `const bool strict` argument of `advisory_query_from_cli_input` into an enum?

I agree with the enum approach. Initially I thought about doing it with boolean flag, but yes, the enum sounds much better. I will try this approach and update the PR. Thank you for the suggestion and the idea. We can discuss the rest implementation after I push the changes.

kontura (approved):
Great, thank you!

Issue #2501, opened 2025-11-05
OptionStringContainer: Values containing delimiters are not properly escaped in `get_value_string` and `to_string` methods


jrohel:
When individual values within an `OptionStringContainer` contain the list delimiter itself (e.g., a comma `','`), the methods `to_string()` or `get_value_string()` do not escape these internal delimiters.

As a result, when the string is later parsed, the internal delimiter is misinterpreted as a separator between list items.

Issue is related to https://github.com/rpm-software-management/dnf5/issues/2483

Issue #2500, opened 2025-11-05
OptionStringContainer: Methods `to_string` and `get_value_string` use incorrect delimiter


jrohel:
The methods `to_string` and `get_value_string` in `OptionStringContainer` currently use a hardcoded delimiter `", "` instead of respecting default or custom delimiter settings.

1. **Steps to Reproduce Default Delimiter Issue**

When initializing with a list:
`OptionStringList option(std::vector<std::string>{"first", "second"});`

Expected Behavior:
`option.get_value_string()` should use the first default delimiter (`','`, `' '`, or `'\n'`) for consistency, resulting in `"first,second"`.

Actual Behavior:
The method returns `"first, second"`, using the hardcoded string `", "` as the delimiter.

2. **Steps to Reproduce Custom Delimiter Issue**

When initializing with a list and setting a custom delimiter (e.g., `:`):
`OptionStringList option(std::vector<std::string>{"first", "second"}, "", false, ":");`

Expected Behavior:
`option.get_value_string()` should return `"first:second"`.

Actual Behavior:
The method returns `"first, second"`. The custom delimiter setting is ignored, and the hardcoded string `", "` is used instead.

`OptionStringList` is in libdnf5 defined as `OptionStringContainer<std::vector<std::string>>`.

**The issue seems to be that the hardcoded string `", "` is always used, ignoring configured delimiters.**

PR #2502, opened 2025-11-05
Fix: `OptionStringContainer::to_string` delimiter and implement escaping

jrohel:
The methods `to_string` and `get_value_string` in `OptionStringContainer` were incorrectly using the hardcoded delimiter `", "`.

This commit fixes two issues:

1. Delimiter Usage - The methods now use the first defined delimiter instead of the hardcoded `", "`.

2. Escaping - Delimiters that are part of the value are now escaped with a backslash `'\'`.

Unit tests are included.

Closes: https://github.com/rpm-software-management/dnf5/issues/2500
Closes: https://github.com/rpm-software-management/dnf5/issues/2501

### DIFF
--- a/test/libdnf5/conf/test_option.cpp
+++ b/test/libdnf5/conf/test_option.cpp
@@ -494,6 +494,34 @@ void OptionTest::test_options_string_list_custom_delimiters() {
     CPPUNIT_ASSERT_EQUAL((std::vector<std::string>{"Dfirstx", "DsecondX"}), option3.get_value());
 }
 
+void OptionTest::test_options_string_list_to_string() {
+    OptionStringList option(std::vector<std::string>{"Dfirstx", "DsecondX"});
+    CPPUNIT_ASSERT_EQUAL(std::string{"Dfirstx,DsecondX"}, option.get_value_string());
+
+    option.set(Option::Priority::RUNTIME, std::vector<std::string>{" ", "DsecondX"});
+    CPPUNIT_ASSERT_EQUAL(std::string{"\\ ,DsecondX"}, option.get_value_string());
+
+    option.set(Option::Priority::RUNTIME, std::vector<std::string>{"asdasd", " ", "DsecondX"});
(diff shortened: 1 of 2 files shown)


libdnf5/conf/option_string_list.cpp
```diff
@@ -297,17 +297,28 @@ void OptionStringContainer<T, IsAppend>::add_item(Priority priority, const std::
...
+            if (!delimiters.empty()) {
+                result += delimiters[0];
+            }
         } else {
             next = true;
         }
-        oss << val;
+        for (const auto ch : item) {
+            if (delimiters.find(ch) != delimiters.npos) {
+                result += '\\';
```
m-blaha, line 315: Just thinking about this escaping of delimiters. It doesn't seem to be handled on the `from_string` side, so currently behavior is kind of asymmetric:

```
to_string(["first", " ", "second"]) => "first,\ ,second"
from_string("first,\ ,second") => ["first", "\", "second"]
```

jrohel: @m-blaha For from_string, I made a PR before this one: https://github.com/rpm-software-management/dnf5/pull/2485

m-blaha: Sorry, didn't know about it :)

libdnf5/conf/option_string_list.cpp
```diff
@@ -297,17 +297,28 @@ void OptionStringContainer<T, IsAppend>::add_item(Priority priority, const std::
...
-            oss << ", ";
+            if (!delimiters.empty()) {
+                result += delimiters[0];
+            }
         } else {
             next = true;
         }
-        oss << val;
+        for (const auto ch : item) {
+            if (delimiters.find(ch) != delimiters.npos) {
```
m-blaha: The backslash itself should also be escaped. Currently both ["first,second"] and ["first\","second"] are serialized as the same string "first\\,second".

jrohel: Thanks. The backslash character is now properly escaped.
I also updated the method docstring and modified unit tests to cover backslash escaping.

m-blaha (approved):
Thank you!

PR #2504, opened 2025-11-06
manifest: Add `--use-host-repos`

evan-goode:
This PR adds `dnf5 manifest --use-host-repos`:

```
``--use-host-repos``
    In addition to the repositories defined in the input file, load
    repositories from host configuration and use them for dependency
    resolution. Only applies to the ``resolve`` command.
```

This argument will be useful for keeping DNF repository configuration in base images while using an infile to list layered packages. The step of resolving an infile to a manifest becomes less pure with `--use-host-repos`, but this way we don't have to keep repository configuration in sync between two formats.

### DIFF
--- a/doc/dnf5_plugins/manifest.8.rst
+++ b/doc/dnf5_plugins/manifest.8.rst
@@ -91,6 +91,11 @@ Options
     Consider currently installed system packages for dependency resolution
     and exclude them from the resulting manifest file.
 
+``--use-host-repos``
+    In addition to the repositories defined in the input file, load
+    repositories from host configuration and use them for dependency
+    resolution. Only applies to the ``resolve`` command.
+
 ``--arch``
     Specify a comma-separated list of architectures to work with.
     By default, only packages corresponding to the current system architecture are processed.


kontura (commented):
The change looks good.

Though looking at it I have realized that since we don't apply the `setopts` to host repos doing something like: 
```bash
$ dnf5 manifest new --disable-repo=unwanted
```
or now also
```bash
$ dnf5 manifest resolve --use-host-repos --disable-repo=unwanted
```
will not work since `--disable-repo=` is implemented via `setopts`.
Of course any `--setopt=` manipulating the repos will not work either.

Sooner or later I think someone will run into this.

Perhaps we should follow up on this todo: https://github.com/rpm-software-management/dnf5/blob/main/dnf5/context.cpp#L282, move the logic to libdnf and use it here.
We could make this into a separate issue if you prefer.

dnf5-plugins/manifest_plugin/manifest.cpp
```diff
@@ -112,6 +113,19 @@ std::filesystem::path get_manifest_path(libdnf5::OptionPath & option, const std:
...
+void load_host_repos(dnf5::Context & ctx, libdnf5::Base & base) {
+    auto repo_sack = base.get_repo_sack();
+    repo_sack->create_repos_from_system_configuration();
+
+    std::vector<std::pair<std::string, std::string>> repos_from_path{ctx.get_repos_from_path()};
+    auto vars = base.get_vars();
+    for (auto & id_path_pair : repos_from_path) {
+        id_path_pair.first = vars->substitute(id_path_pair.first);
+        id_path_pair.second = vars->substitute(id_path_pair.second);
+    }
```
mfocko: I'd probably prefer

```suggestion
    for (auto & [id, path] : repos_from_path) {
        id = vars->substitute(id);
        path = vars->substitute(path);
    }
```

ngl, took me a moment to grep through the code to actually find a meaning of the pair…

evan-goode: Thanks, I forgot about destructuring! Done

kontura (approved):

PR #2508, opened 2025-11-10
`--json` for `check-upgrade`

### LINKED ISSUE #2481: Add JSON output to `dnf check-upgrade` command
Currently, `dnf check-update` produces output that is complex and difficult to parse programmatically. For example:

```bash
dnf check-update --refresh -yq | tail -n +2 | grep -E 'x86_64|i686|noarch|aarch64' | awk '{print $1, $2}'
```

This produces output like:

```
<package>.<arch> <new version>.<fedora version>
```

The output often includes extra information and requires multiple filtering and formatting steps to make it usable in scripts.

By comparison, `apt list --upgradable` produces structured and consistent output:

```
<package>/<repo> <new version> <arch> [upgradable from: <old version>]
```

## Proposed improvement

Add an option (e.g., `--simple-output` or `--upgradable`) for `dnf check-update` that outputs only:

* Package name
* Old version
* New version

For example:

```
<package name> <old version> => <new version>
```

This should be consistent and machine-readable, similar to Pacman’s output, making it much easier to parse for automation and scripting.

## Additional notes

* `dnf5` already supports JSON output for some commands. Extending this feature to `check-update` would be a natural fit.
* A simple flag like `--upgradable` or `--simple-output` would significantly improve usability for developers and maintainers of automated tools.

---

Also I'm using Arch Linux, I requesting this feature because I'm maintaining a project, see `zig-waybar-contrib`. This will be good if added.

mfocko:
For context of related changes, please see separate commits.

### TODO

- [x] Resolve the `std::move` remark
- [x] Maybe factor out the helper for `PackageSet` to sorted `std::vector`?
- [x] ~~How do I test?~~ Gotta rob already open PR…

---

Fixes #2481
Merge with rpm-software-management/ci-dnf-stack#1788

### DIFF
--- a/dnf5/commands/check-upgrade/check-upgrade.cpp
+++ b/dnf5/commands/check-upgrade/check-upgrade.cpp
@@ -208,7 +208,7 @@ void CheckUpgradeCommand::run() {
     auto sections = create_output();
 
     bool package_matched = false;
-    package_matched |= sections->add_section("", upgrades_query);
+    package_matched |= sections->add_section("Upgrades", upgrades_query);
     package_matched |= sections->add_section("Obsoleting packages", obsoletes_query, obsoletes);
 
     if (ctx.get_json_output_requested()) {
--- a/libdnf5-cli/output/package_list_sections.cpp
+++ b/libdnf5-cli/output/package_list_sections.cpp
@@ -162,9 +162,7 @@ void PackageListSections::print_json() {
             json_object_array_add(j_packages, j_pkg);
         }
 
-        // [NOTE](mfocko) Consider adding header to CLI “pretty” output too…
-        auto json_heading =
-            libdnf5::cli::utils::json::normalize_field(heading != "" ? heading : "Upgradeable packages");
+        auto json_heading = libdnf5::cli::utils::json::normalize_field(heading);
         json_object_object_add(j_output, json_heading.c_str(), j_packages);
     }
 


libdnf5-cli/output/package_list_sections.cpp
```diff
@@ -24,15 +24,48 @@
...
 
 #include <algorithm>
 
+namespace {
+
+template <class Compare>
+std::vector<libdnf5::rpm::Package> to_sorted_vector(const libdnf5::rpm::PackageSet & pkg_set, Compare cmp) {
+    std::vector<libdnf5::rpm::Package> packages;
+    for (const auto & pkg : pkg_set) {
+        // [FIXME](mfocko) Does this ‹std::move› actually move, since ‹pkg› is a const-ref?
```
mfocko: IMO for an actual _move_, it would require:

```diff
-std::vector<libdnf5::rpm::Package> to_sorted_vector(const libdnf5::rpm::PackageSet & pkg_set, Compare cmp) {
+std::vector<libdnf5::rpm::Package> to_sorted_vector(libdnf5::rpm::PackageSet && pkg_set, Compare cmp) {
```

and
```diff
-    for (const auto & pkg : pkg_set) {
+    for (auto && pkg : pkg_set) {
```

IMO this does a move from the copy.

---

Consuming the `PackageSet` feels quite… Rust-like. In CLI, where the `PackageSet` is iterated through once, it wouldn't be a problem, though not sure if it benefit is worth it.

dnf5/commands/check-upgrade/check-upgrade.cpp
```diff
@@ -209,6 +211,11 @@ void CheckUpgradeCommand::run() {
     package_matched |= sections->add_section("", upgrades_query);
     package_matched |= sections->add_section("Obsoleting packages", obsoletes_query, obsoletes);
 
+    if (package_matched && ctx.get_json_output_requested()) {
```
kontura: I see that if `package_matched` is `false` it can print the `No security updates needed, but...` line, personally I wouldn't expect that in the `--json` mode.
Would it make sense to rather still print json output but empty?

libdnf5-cli/output/package_list_sections.cpp
```diff
@@ -116,6 +133,43 @@ void PackageListSections::print(const std::unique_ptr<PkgColorizer> & colorizer)
     scols_unref_table(table);
 }
 
+// [NOTE] When editing JSON output format, do not forget to update the docs at doc/commands/check-upgrade.8.rst
+void PackageListSections::print_json() {
+    json_object * j_output = json_object_new_object();
+    for (const auto & [heading, pkg_set, obsoletes] : p_impl->sections) {
+        if (pkg_set.empty()) {
+            // skip empty sections
+            continue;
```
kontura, line 143: For parsing purposes I wonder if it would be better to also print the sections even if they are empty? 
I think it might also be more consistent with the rest of the commands' json outputs. 

Although I see the sections are not added at all in check-upgrade if their query is empty. :thinking:

include/libdnf5/rpm/package_set.hpp
```diff
@@ -153,6 +153,17 @@ class LIBDNF_API PackageSet {
...
 
+    /// @return An `std::vector` of packages sorted by the given comparator.
+    ///
+    /// @param cmp Comparator used for sorting.
+    /// @since 5.0 <-- [TODO] What goes here??? Do I guess?
+    template <class Compare>
+    std::vector<Package> to_sorted_vector(Compare cmp) const;
+
+    /// @return An `std::vector` of packages sorted by NEVRA.
+    /// @since 5.0 <-- [TODO] same as above
```
kontura: I would be fine with adding this to the `package_set`.

Here should be the version in which this is going to be released. It is a bit of a guessing game but I believe it will be 5.3.1.0 here.

include/libdnf5/rpm/package_set.hpp
```diff
@@ -153,6 +153,17 @@ class LIBDNF_API PackageSet {
 
     void swap(PackageSet & other) noexcept;
 
+    /// @return An `std::vector` of packages sorted by the given comparator.
+    ///
+    /// @param cmp Comparator used for sorting.
+    /// @since 5.0 <-- [TODO] What goes here??? Do I guess?
+    template <class Compare>
+    std::vector<Package> to_sorted_vector(Compare cmp) const;
```
kontura: Personally I wouldn't expose this overload, unless I am mistaken it is not used by anything from the outside.

libdnf5-cli/output/package_list_sections.cpp
kontura: I see this is outdated but the `PackageSet` doesn't really contain `Package` objects, it is only a `SolvMap` (bit map). The `Package` objects are constructed only by `PackageSet` [iterator](https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/rpm/package_set_iterator.cpp#L82-L84).

mfocko: Ah OK, makes sense then, cool!

dnf5/commands/check-upgrade/check-upgrade.cpp
mfocko: True, I missed that, thanks for noticing.

libdnf5-cli/output/package_list_sections.cpp
mfocko: There's also a smallish issue with the “headings”, I don't really like the idea of adding multiple edge cases because of this, so the keys in the JSON object look a bit weird… Maybe I could create some mapping for that, but it still feels meh…

include/libdnf5/rpm/package_set.hpp
mfocko: I don't see an issue with exposing it, but… can make it `private` and in case there's a use case, it can be made part of the interface.

libdnf5-cli/output/package_list_sections.cpp
mfocko: I could probably pull in `std::optional<>` to keep the sections, but not allocate anything for the “sets”.

kontura: > There's also a smallish issue with the “headings”, I don't really like the idea of adding multiple edge cases because of this, so the keys in the JSON object look a bit weird… Maybe I could create some mapping for that, but it still feels meh…

If I understand correctly you mean that the headings like "Obsoleting packages" are not great JSON keys, which I would agree with. Perhaps we could transform them during the JSON printing? Dynamically convert all letter to lower case and replace spaces with underscores?

I think this would work fairly well also for the section headings in the `list` command, I assume we will turn on the JSON output there as well since it uses this code and it will be very easy.

Regarding the special case of the missing upgrades heading in the `check-upgrade` output, personally I probably wouldn't mind adding it also to the "pretty" human readable output. However perhaps we could discuss this with others on our next developers meeting?

> I could probably pull in `std::optional<>` to keep the sections, but not allocate anything for the “sets”.

In my opinion we don't need to worry about this, the `PackageSet` (being a bit map) is quite light weight and I would say there will never be many many of these `sections`. Plus it seems the empty sets are [already skipped](https://github.com/rpm-software-management/dnf5/pull/2508/commits/2901ebbf0f8c325cfd22374f030afd0924c6bd00) in the human readable output.
Though the `std::optional<>` would work too. :slightly_smiling_face:

include/libdnf5-cli/utils/json.hpp
```diff
@@ -0,0 +1,43 @@
...
+// the Free Software Foundation, either version 2.1 of the License, or
+// (at your option) any later version.
+//
+// Libdnf is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
```
kontura: Can you shorten this license header?
For new files we have agreed to use just:
```
// Copyright Contributors to the DNF5 project
// SPDX-License-Identifier: LGPL-2.1-or-later
```
I believe it is described in https://github.com/rpm-software-management/dnf5/blob/main/CONTRIBUTING.md#license-header.

include/libdnf5/rpm/package_set.hpp
```diff
@@ -167,6 +171,13 @@ class LIBDNF_API PackageSet {
 
     class LIBDNF_LOCAL Impl;
     std::unique_ptr<Impl> p_impl;
+
+    /// @return An `std::vector` of packages sorted by the given comparator.
+    ///
+    /// @param cmp Comparator used for sorting.
+    /// @since 5.3.1.0
+    template <class Compare>
+    std::vector<Package> to_sorted_vector(Compare cmp) const;
```
kontura: Unfortunately having methods hidden using the `private` keyword is not great because I believe changing them to `public` later would break ABI.

It is mentioned for example [here](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B): 
> You cannot...
> ...
> Changing the access rights to some functions or data members, for example from private to public. With some compilers, this information may be part of the signature. If you need to make a private function protected or even public, you have to add a new function that calls the private one.

Ideally we would just move it to the `p_impl` however I realized that constructing the `Package` objects there is not so simple.

At this point I am not sure if it would be better to just inline it or keep it public like you originally had it.
I am sorry about being so annoying about this, I just want to be careful when adding things to the API.

include/libdnf5-cli/utils/json.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define LIBDNF5_CLI_UTILS_JSON_HPP
+
+#include <cctype>
+#include <string>
+
+namespace libdnf5::cli::utils::json {
+
+/// Normalizes the field name for a JSON field, i.e., converts to lower-case and
+/// replaces all whitespace with underscores.
+std::string normalize_field(std::string name) {
```
kontura: I think we should also mark this with the `LIBDNF_CLI_API` makro.

include/libdnf5/rpm/package_set.hpp
mfocko: _will check if there's any other occurrence of this pattern_

I've originally went with making it public, since it could be reused somewhere… tbf the most ideal way would be returning a range, so the consumer of the `PackageSet` can do whatever they want without involving the set itself 🤔 but I guess that's out of scope for this :D

mfocko: eh… I don't see it anywhere else, will drop the generic implementation, we can always redo it, if needed.

Issue #762, opened 2023-07-21
dnf [options] remove --oldinstallonly


j-mracek:
Removes old installonly packages, keeping only latest versions and version of running kernel.

myllynen:
Now that dnf5 is the default dnf on Fedora 41 there seems to be now easy way to remove e.g. old kernel versions.

Should this issue be prioritized higher to fill the feature gap in Fedora 41 or are there perhaps other plans to address this? Thanks.

evan-goode:
Yes, I think it'd be fair to revisit this. I'll make a note to discuss it with the team. There is already `installonly_limit`. By default it's `3`, but you can set it to `2` to only have 2 kernel versions installed concurrently (running kernel and just-installed kernel).

PR #2515, opened 2025-11-12
Add --oldinstallonly option to remove command

fhbash:
Adds `--oldinstallonly` flag to remove old installonly packages (e.g., kernels) that exceed the configured installonly_limit. 

Optional `--limit` parameter allows overriding the config value (must be >1 to keep newest installed version).

Usage:
```
  dnf5 remove --oldinstallonly              # Remove all old installonly pkgs
  dnf5 remove --oldinstallonly kernel       # Remove old kernels only
  dnf5 remove --oldinstallonly --limit=2    # Keep only 2 newest versions
```

ci-tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1782

Closes: #762

### DIFF
--- a/dnf5/commands/remove/remove.cpp
+++ b/dnf5/commands/remove/remove.cpp
@@ -20,6 +20,14 @@
 #include "remove.hpp"
 
 #include <dnf5/shared_options.hpp>
+#include <libdnf5-cli/argument_parser_errors.hpp>
+#include <libdnf5/conf/option_number.hpp>
+#include <libdnf5/rpm/package_query.hpp>
+#include <libdnf5/rpm/package_sack.hpp>
+#include <libdnf5/utils/bgettext/bgettext-mark.h>
+#include <libdnf5/utils/format.hpp>
+
+#include <map>
 
 namespace dnf5 {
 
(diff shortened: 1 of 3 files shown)


ppisar:
I guess it should depend on protect_running_kernel option value.

ppisar:
Yes, that's what I meant. I did not know how DNF4 behaves.

I described my expectation that protect_running_kernel works independently from install-only feature and that non-latest kernel is old not matter whether it is running. I think it would keep the code and a documentation simple.

Of course there is a matter of compatibility which I did not take into account. I'm not strongly against DNF4 behavior. I only find it more complex and less intuitive.

dnf5/commands/remove/remove.cpp
```diff
@@ -48,7 +56,27 @@ void RemoveCommand::set_argument_parser() {
...
+    cmd.register_named_arg(oldinstallonly_opt);
+
+    oldinstallonly_limit = dynamic_cast<libdnf5::OptionNumber<std::uint32_t> *>(parser.add_init_value(
+        std::unique_ptr<libdnf5::OptionNumber<std::uint32_t>>(new libdnf5::OptionNumber<std::uint32_t>(0))));
+    auto limit_opt = parser.add_new_named_arg("limit");
+    limit_opt->set_long_name("limit");
+    limit_opt->set_has_value(true);
+    limit_opt->set_arg_value_help("LIMIT");
+    limit_opt->set_description(
+        "Limit the number of installonly package versions to keep (must be >1, used with --oldinstallonly)");
```
m-blaha: It should say `must be >=1`.

dnf5/commands/remove/remove.cpp
```diff
@@ -71,21 +99,122 @@ void RemoveCommand::configure() {
...
+    auto goal = ctx.get_goal();
+
+    if (oldinstallonly->get_value()) {
+        auto & base = ctx.get_base();
+        auto & cfg_main = base.get_config();
+        std::uint32_t installonly_limit;
+
+        if (oldinstallonly_limit->get_value() > 0) {
+            installonly_limit = oldinstallonly_limit->get_value();
+            if (installonly_limit < 1) {
```
m-blaha, line 112: This condition can never be true - you are checking that installonly_limit is > 0 two lines above.

dnf5/commands/remove/remove.cpp
```diff
@@ -71,21 +99,122 @@ void RemoveCommand::configure() {
...
+            installonly_limit = cfg_main.get_installonly_limit_option().get_value();
+        }
+
+        libdnf5::rpm::PackageQuery installed_installonly(base);
+        installed_installonly.filter_installed();
+        installed_installonly.filter_installonly();
+
+        // If package specs provided, filter to only those packages
+        if (!pkg_specs.empty()) {
+            libdnf5::rpm::PackageQuery filtered_query(base);
```
m-blaha: This creates a PackageQuery object with all packages. You need to start with an empty query, otherwise union operations on line 131 would be effectively a no-op.
```suggestion
            libdnf5::rpm::PackageQuery filtered_query(base, libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES, true);
```

fhbash: nice catch.. thanks..

Issue #1868, opened 2024-11-17
I18N: The strings from userconfirm.hpp cannot be translated


sunwire:
https://github.com/rpm-software-management/dnf5/blob/708604c8ac6d1e9fa48dad97337e27423a880a5a/include/libdnf5-cli/utils/userconfirm.hpp#L43-L45

The strings cannot be translated even if I use the _() macro.
The `cmake --build build --target gettext-potfiles`  command seems to ignore them.
The same with "n", "N", "y" or "Y"

userconfirm() function is used by dnf5, dnf5daemon-client and copr plugin

sunwire:
Proposed changes
https://github.com/rpm-software-management/dnf5/compare/main...sunwire:dnf5:i18n-yesno

PS.
I don't like the multiple uses of #define but I can't think of any other solution.

```
$ sudo  ./dnf5 install s-tui
Aktualizowanie i ładowanie repozytoriów:
Załadowano repozytoria.
Pakiet                                                                                  Arch.              Wersja                                                                                  Repozytorium                                        Rozmiar
Instalowanie:
 s-tui                                                                                  noarch             1.1.6-3.fc41                                                                            fedora                                            293.4 KiB
Instalowanie zależności:
 python3-urwid                                                                          noarch             2.6.14-2.fc41                                                                           fedora                                              3.2 MiB
 python3-wcwidth                                                                        noarch             0.2.13-5.fc41                                                                           fedora                                            559.8 KiB

Podsumowanie transakcji:
 Instalowanie:         3 pakietów

Całkowity rozmiar pakietów przychodzących wynosi 1 MiB. Należy pobrać 1 MiB.
Po tej operacji zostanie użyte dodatkowo 4 MiB (zainstaluj 4 MiB, usuń 0 B).
Czy to jest ok [t/N]: t
```

ppisar:
Thanks for the report. I haven't looked at the code thoroughly, but I don't like two things on the proposed code: It multiplies the confirmation message and it breaks API and ABI of the public /usr/include/libdnf5-cli/utils/userconfirm.hpp header file. But it's possible that the fix will need changing API.

sunwire:
> I don't like two things on the proposed code

I don't like them either, which is why I didn't make a PR.
The problem is that one common message would have to be shared between different text domains. And I don't like that either.

geraldosimiao:
Hello. Any changes about this?

kontura:
[...]

PR #2518, opened 2025-11-17
Make userconfirm prompt strings translatable

Userfrom1995:
The confirmation prompt "Is this ok [Y/n]:" and response characters were hardcoded strings in the userconfirm.hpp template function, making them untranslatable.

Changes:
- Extract translatable strings to libdnf5-cli/utils/userconfirm.cpp
- Add helper functions: get_yes_no_prompt(), get_yes_response(), etc.
- Update libdnf5-cli.pot with new translatable strings

The userconfirm template now calls helper functions that return translated strings from the libdnf5-cli translation domain, allowing translators to localize the prompt and response characters.

Fixes: #1868

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ endif()
 
 if(WITH_TRANSLATIONS)
     # define common command to generate the pot file from sources
-    list(APPEND XGETTEXT_COMMAND xgettext -F --from-code=UTF-8 --keyword=_ --keyword=M_ --keyword=P_:1,2 --keyword=MP_:1,2 --keyword=C_:1c,2 --keyword=MC_:1c,2 --keyword=CP_:1c,2,3 --keyword=MCP_:1c,2,3 -c)
+    list(APPEND XGETTEXT_COMMAND xgettext -F --from-code=UTF-8 --keyword=_ --keyword=M_ --keyword=P_:1,2 --keyword=MP_:1,2 --keyword=C_:1c,2 --keyword=MC_:1c,2 --keyword=CP_:1c,2,3 --keyword=MCP_:1c,2,3 --keyword=dgettext:2 -c)
     add_custom_target(gettext-potfiles)
 endif()
 
--- a/cmake/Translations.cmake
+++ b/cmake/Translations.cmake
@@ -6,6 +6,15 @@ find_package(Gettext)
 set(COMPONENT_PATH ${CMAKE_CURRENT_SOURCE_DIR}/..)
 file(GLOB_RECURSE POT_SOURCES RELATIVE ${COMPONENT_PATH} ${COMPONENT_PATH}/*.cpp ${COMPONENT_PATH}/*.hpp)
 
+# For libdnf5-cli, also include the userconfirm header file from the global include directory
+if(GETTEXT_DOMAIN STREQUAL "libdnf5-cli")
+    set(USERCONFIRM_HEADER "${CMAKE_SOURCE_DIR}/include/libdnf5-cli/utils/userconfirm.hpp")
+    if(EXISTS ${USERCONFIRM_HEADER})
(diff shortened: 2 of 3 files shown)


Userfrom1995:
> Correct implementation should use nl_langinfo() to retrieve the translated Yes/No expressions. This manual reimplementation is a step in wrong direction.

Thanks for review. I was not aware of nl_langinfo().
I will update the implementation to use it.

evan-goode:
Thanks, this does seem to pick up the correct YESEXPR/NOEXPR. However, I still cannot see translations for the "Is this ok [Y/n]: " message, for example, if I add the following to `dnf5/po/es.po`:

```
msgid "Is this ok [Y/n]: "
msgstr "¿Está bien? [S/n]: "

msgid "Is this ok [y/N]: "
msgstr "¿Está bien? [s/N]: "
```

Am I just missing something, were you able to get it working?

Another concern is that the yesexpr/noexpr value and the translated message `[Y/n]` may be or may become out of sync. For example, we don't yet have a Swahili translation, so Swahili-speaking users would likely see the English fallback. Yet according to https://lh.2xlibre.net/values/yesexpr/, the `sw_KE` yesexpr is `^[+1nNyY]`. So I believe with this patch, it would print

```
Is this ok [y/N]:
```

and then if the user types `n`, it would be considered "yes" :upside_down_face:

Userfrom1995:
> > I ran `make libdnf5-cli-pot` to update the template. I updated the .po files using: `cd libdnf5-cli/po && for f in es.po cs.po de.po fr.po; do msgmerge --update "$f" libdnf5-cli.pot; done` I added the translations to the respective files. I committed the changes. I did an RPM build and installed the new binary, and then it worked.
> 
> Thanks! That worked. (Weblate usually takes care of this step for us)
> 
> > Yes, it was a major issue with my previous implementation and I have tried to solve it. Now, the `yesexpr` and `noexpr` from nl_langinfo are used only if a translation is available. If the prompt falls back to English, it strictly uses the "English regex".
> 
> That's a smart and reasonable solution. There is still the risk of a `yesexpr`/`noexpr` changing to become mismatched, but I don't see a way around that problem unless we could somehow dynamically parse the regex and display it in an intuitive way to the user. I am happy with this solution.
> 
We could print the regex pattern, but it would look quite bad, for example:
```Is this ok [^[+1sSyY]/^[-0nN]]:```
Parsing the regex pattern to display it in a user-friendly way would also be too complex for this solution, at least for now. 

> I have one more request: in DNF4 we have
> 
> ```
>             # If the English one letter names don't mix with the translated
>             # letters, allow them too:
>             if u'y' == choice and u'y' not in aui:
>                 choice = yui[0]
>                 break
>             if u'n' == choice and u'n' not in aui:
>                 choice = nui[0]
>                 break
> ```
> 
> could we do the same here? I think we'd just need to keep the old `if (choice == "y" || choice == "Y")` and `if (choice == "n" || choice == "N")` blocks in the loop after the regex tests.

Yes, I just added the conditions. However, there is one interesting thing I noticed on this page: https://lh.2xlibre.net/values/yesexpr/

All the yesexpr regex patterns contain `Yy`, and all the noexpr regex patterns contain `Nn`. So technically, the additional y/n conditions should never be reached unless:

- the regex pattern is missing or corrupted,

- we add a language that does not include `Yy` or `Nn` in its regex pattern, or

- one of the supported languages changes its regex pattern in the future.

So practically speaking, this is more of a safety measure than something that should trigger under normal circumstances.

evan-goode (approved):

evan-goode:
> Yes, I just added the conditions. However, there is one interesting thing I noticed on this page: https://lh.2xlibre.net/values/yesexpr/
> So practically speaking, this is more of a safety measure than something that should trigger under normal circumstances.

Good point.

I think we're ready to merge, thanks for the patch :)

PR #2519, opened 2025-11-17
Lock from the time system repository is loaded until exit

evan-goode:
Related: https://github.com/rpm-software-management/dnf5/issues/2129

This is another step towards proper locking so that multiple DNF5 processes don't interfere with each other.

Adds `Base::lock_system_repo` and `Base::unlock_system_repo`:

```
/// Acquire an advisory lock on the installroot's system repository.
/// The lock will be automatically released when Base goes out of scope, or manually when unlock_system_repo is called.
/// Can be called multiple times to upgrade or downgrade a READ lock to a WRITE lock or vice versa.
/// Should be called before the system repo is loaded, and the lock should be held until all transactions are
/// complete and other processes can safely re-read the RPMDB and resolve transactions.
/// @throw libdnf5::SystemError if an unexpected error occurs when locking
/// @return true if acquiring the lock succeeded, false otherwise
bool lock_system_repo(
    libdnf5::utils::LockAccessType access = libdnf5::utils::LockAccessType::WRITE,
    libdnf5::utils::LockBlockingType blocking = libdnf5::utils::LockBlockingType::NON_BLOCKING);

/// Release the lock obtained by lock_system_repo.
/// Idempotent. No-op if there is currently no lock.
/// @throw libdnf5::SystemError if an unexpected error occurs when unlocking
void unlock_system_repo();
```

And calls `Base::lock_system_repo` from `Context::load_repos` before the system repo is loaded.

The idea here is that we need to hold a lock the entire time from before the transaction is resolved until after the transaction is complete and the RPMDB is written. Otherwise, another DNF process could resolve a transaction using a soon-to-be-invalid version of the system repo and get an invalid transaction.

This approach is an improvement over DNF4 which blocks while another transaction is running but does not block before loading the system repo and resolving a new transaction.

A helpful message is printed listing the other processes DNF5 is waiting for:

```
$ sudo dnf5 install rsms-inter-fonts
Waiting for other processes to finish:
62409	dnf5 install blender
```

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -209,7 +209,8 @@ void Context::Impl::load_repos(bool load_system, bool load_available) {
 
     const auto skip_system_repo_lock = base.get_config().get_skip_system_repo_lock_option().get_value();
     if (load_system && !skip_system_repo_lock) {
-        const auto lock_access_type = libdnf5::utils::LockAccess::WRITE;
+        const auto lock_access_type =
+            cmd_requires_privileges() ? libdnf5::utils::LockAccess::WRITE : libdnf5::utils::LockAccess::READ;
         if (!base.lock_system_repo(lock_access_type, libdnf5::utils::LockBlocking::NON_BLOCKING)) {
             // A lock on the system repo could not immediately be acquired.
             // Gather and display information about other processes in line for the lock, then wait.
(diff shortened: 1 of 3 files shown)


dnf5/context.cpp
```diff
@@ -276,6 +280,7 @@ class Context::Impl {
 
     bool load_system_repo{false};
     LoadAvailableRepos load_available_repos{LoadAvailableRepos::NONE};
+    bool will_modify_system{true};
```
kontura: If I understand correctly this value is derived from (it duplicates) information already present in the `Context`. Is there a plan to use it somehow differently later on? 

Currently I would rather make `cmd_requires_privileges(dnf5::Context context)` a private method of `Context` and use it directly. It would require introducing a private header file for `Context`, where we would likely move `Context::Impl`, but I think that would be a good move regardless. 
What do you think?

dnf5/include/dnf5/context.hpp
```diff
@@ -77,7 +77,7 @@ class DNF_API Context : public libdnf5::cli::session::Session {
         const std::vector<std::string> & cves);
 
     /// Sets callbacks for repositories and loads them, updating metadata if necessary.
-    void load_repos(bool load_system);
+    void load_repos(bool load_system, bool load_available);
```
kontura: While we are doing changes here, I am wondering whether the `Context.load_repos(...)` API should be public at all. If I am not mistaken this API is for dnf5 plugins and those will already be using it through the dnf5 `main` function.
I see it is now used in `makecache` command but that feels more like an oversight to me because it results in calling it twice. I believe it could easily be removed.

This would need a private header for `Context` as welll though.

dnf5/context.cpp
evan-goode: It may be useful in the future for commands (especially those provided by plugins) to declare for themselves whether they will modify the system rather than doing the heuristics in `cmd_requires_privileges`. Also if we ever do something about https://issues.redhat.com/browse/RHEL-67915, there would be a distinction between "will modify the system repository" (inside the installroot) and "needs root privileges".

But for now your suggestion is maybe better, like you said, a `context_private.hpp` would be a good idea.

evan-goode:
> While I don't think these breaking changes will actually break anyone's code I am not thrilled about bumping the major version so soon again.
> 
> I think we could get around it by rather introducing new API but I am not sure if there is any real technical reason for this.

Most of the ABI/API changes are in `Locker`. So we could create a `LockerV2` or something, yes. IMO it's better to just do the ABI/API break. In my view, there is not much harm in a major version bump other than the ABI/API break that it signifies. Stable F43 is still on DNF 5.2 so we would not need to do backporting to 5.3.

dnf5/include/dnf5/context.hpp
evan-goode: I agree, it could be made private.

evan-goode: OK, back to working on this and I have a C++ question. How would you recommend making `load_repos` API-private exactly? Designating `Context::load_repos` with `DNF_LOCAL` doesn't seem to work by itself (that would be nice).

I don't think it's possible to split the definition of `Context` into two headers (One Definition Rule). Although adding `load_repos` shouldn't break ABI, so you could do something using `#ifdef DNF_BUILD_APPLICATION`.

We can't just use p_impl since `main` still has to run `ctx.load_repos`. Maybe a `ContextInternal` that inherits from `Context`?

There is also https://en.cppreference.com/w/cpp/language/modules.html.

kontura: > We can't just use p_impl since `main` still has to run `ctx.load_repos`.

This is the approach I was thinking about. I believe it will work if we make `main` a `friend` of `Context`. Initially I wasn't sure `main` can be a `friend` but it seems to work fine if it is forward declared. I tested here: https://github.com/kontura/dnf5/commits/private-context/.

If I recall correctly this setup (impl header + friends) is used a lot in `libdnf5`.

evan-goode: Alright, I moved it to the `Impl` and made `main` a friend. I went ahead and made `match_specs` a member of `Context` instead of making it another friend.

It does seem strange to me to make `main` a friend... would this have the side effect of making anyone's `main` function a friend of `Context`, not just dnf5's `main`? Probably not a big deal if so, more like a code smell. We can refactor further in the future if desired.

dnf5/context.cpp
evan-goode: Done.

dnf5/include/dnf5/context.hpp
kontura: Indeed I think anyone's `main` would be able to access it, though the definition of the `Impl` is not public.

I agree that it is weird.

PR #2522, opened 2025-11-20
transaction: sort packages when SOURCE_DATE_EPOCH set

jlebon:
For reproducible builds that involve RPMs, the exact order in which packages are installed matters. This affects some files that are easy to normalize in postprocessing (e.g. `/etc/passwd` line order), but it also affects the rpmdb, where e.g. row order may change. See this comment and following:

https://github.com/rpm-software-management/rpm/issues/2219#issuecomment-3220396780

This was fixed for the rpm-ostree use case by sorting the packages before adding them to the transaction and calling rpmtsOrder():

https://github.com/coreos/rpm-ostree/pull/5469

This is the dnf equivalent.

We only do this if SOURCE_DATE_EPOCH is defined for two reasons:
1. It reduces risk by limiting it to the group of people who actually care about these details. (Though ironically, this change itself will cause a rebuild of reproducible artifacts.)
2. It mitigates the concern raised in the linked discussions regarding well-defined reordering possibly masking bugs in RPMs that unknowingly rely on a specific order.

Ideally, this functionality would live on the rpm side instead. For the same reasons as above, this would also likely have to be an opt-in flag. Once we have that though, the logic here could be simplified.

Assisted-by: Claude Code / Sonnet 4.5

### DIFF
--- a/test/libdnf5/rpm/test_transaction.cpp
+++ b/test/libdnf5/rpm/test_transaction.cpp
@@ -78,6 +78,23 @@ public:
     int mirror_failure_cnt{0};
 };
 
+class OrderCapturingCallbacks : public libdnf5::rpm::TransactionCallbacks {
+public:
+    explicit OrderCapturingCallbacks(std::vector<std::string> expected) : expected_order(std::move(expected)) {}
+
+    void install_start(const libdnf5::base::TransactionPackage & item, [[maybe_unused]] uint64_t total) override {
+        auto nevra = item.get_package().get_nevra();
+        if (current_index >= expected_order.size() || nevra != expected_order[current_index]) {
+            order_correct = false;
+        }
+        current_index++;
+    }
+
+    std::vector<std::string> expected_order;
+    size_t current_index = 0;
+    bool order_correct = true;
+};
+
 }  // namespace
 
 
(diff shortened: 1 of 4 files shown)


test/libdnf5/rpm/test_transaction.cpp
```diff
@@ -161,3 +178,57 @@ void RpmTransactionTest::test_transaction_temp_files_cleanup() {
...
+        }
+
+        auto transaction = goal.resolve();
+        transaction.download();
+
+        auto callbacks = std::make_unique<OrderCapturingCallbacks>(expected_order);
+        auto callbacks_ptr = callbacks.get();
+        transaction.set_callbacks(std::move(callbacks));
+
+        // TODO(jkolarik): Temporarily disable the test to allow further investigation of issues on the RISC-V arch
```
cgwalters, line 219: This TODO looks misplaced?

test/libdnf5/rpm/test_transaction.cpp
```diff
@@ -161,3 +178,55 @@ void RpmTransactionTest::test_transaction_temp_files_cleanup() {
         CPPUNIT_ASSERT(!std::filesystem::exists(package_path));
     }
 }
+
+void RpmTransactionTest::test_source_date_epoch_sorting() {
+    add_repo_rpm("rpm-repo1", /* load */ false);
+    add_repo_rpm("rpm-repo2", /* load */ false);
+    add_repo_rpm("rpm-repo3", /* load */ true);
+
+    // setting a global var in a test is not great... no other tests right now
```
cgwalters, line 187: It's not threadsafe is the biggest problem, though very recent glibc versions paper over this.

ppisar (commented):
RPM library is free to reorder the packages and it actually does that (for handling pre and post script dependcies, to break dependency loops etc.).

This is simply wrong place to do it. You need to do it in RPM.

evan-goode:
> Is completely undocumented and not guaranteed feature.

I am fine with doing this in DNF if we can get RPM to promise+document that `rpmtsOrder` is deterministic. Current docstring is:

```
/** \ingroup rpmts
 * Determine package order in a transaction set according to dependencies.
 *
 * Order packages, returning error if circular dependencies cannot be
 * eliminated by removing Requires's from the loop(s). Only dependencies from
 * added or removed packages are used to determine ordering using a
 * topological sort (Knuth vol. 1, p. 262). Use rpmtsCheck() to verify
 * that all dependencies can be resolved.
 *
 * The final order ends up as installed packages followed by removed packages,
 * with packages removed for upgrades immediately following the new package
 * to be installed.
 *
 * @param ts		transaction set
 * @return		no. of (added) packages that could not be ordered
 */
int rpmtsOrder(rpmts ts);
```

test/libdnf5/rpm/test_transaction.cpp
jlebon: All `transaction.run()` calls in that file have the same temporary disablement. I suspect this logic could instead be conditional on `$arch == risv64` but I didn't want to detract from my patch. :)

jlebon:
Can I get a review on this?

evan-goode:
Sorry Jonathan, I'll take a look when I get back from PTO on Monday.

test/libdnf5/rpm/test_transaction.cpp
```diff
@@ -161,3 +178,55 @@ void RpmTransactionTest::test_transaction_temp_files_cleanup() {
...
+    // relate to SOURCE_DATE_EPOCH at least
+    setenv("SOURCE_DATE_EPOCH", "1234567890", 1);
+
+    std::vector<std::string> packages = {"one", "two", "three"};
+    // The exact final install order chosen by librpm doesn't matter. What
+    // matters is that it's consistent across all permutations. An alternative
+    // approach resilient to potential librpm changes is to capture the order on
+    // the first iteration and then compare future iterations against that.
+    std::vector<std::string> expected_order = {"two-2-2.noarch", "three-1-1.noarch", "one-2-1.noarch"};
```
evan-goode, line 196: I think doing it this way is fine for now.

Issue #948, opened 2023-10-05
`--debuginfo`


j-mracek:

derickdiaz:
I'll take this on

kontura:
We should make sure the `--debuginfo` flag respects `--resolve`, this is not working in the old dnf4 version: https://github.com/rpm-software-management/dnf-plugins-core/issues/535.

kontura:
@derickdiaz do you still plan to work on this?

TheHillBright:
In addition to @kontura comment regarding debuginfo debugsource, I think the new download command should also support downloading binary and debuginfo and debugsource rpms for the same set of packages with consistent versions in single transaction. This is not currently possible with dnf4.

PR #2529, opened 2025-11-26
Add --debuginfo option to download command

### LINKED ISSUE #494: Download command

fhbash:
Add a new `--debuginfo` flag to the download command that enables downloading debuginfo packages instead of regular packages.

ci test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1800

Fixes: #494 and  #948

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -75,6 +75,9 @@ void DownloadCommand::set_argument_parser() {
     srpm_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
+    debuginfo_option = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+
     auto resolve = parser.add_new_named_arg("resolve");
     resolve->set_long_name("resolve");
     resolve->set_description("Resolve and download needed dependencies");
@@ -93,6 +96,12 @@ void DownloadCommand::set_argument_parser() {
     srpm->set_description("Download the src.rpm instead");
     srpm->set_const_value("true");
     srpm->link_value(srpm_option);
+
+    auto debuginfo = parser.add_new_named_arg("debuginfo");
+    debuginfo->set_long_name("debuginfo");
+    debuginfo->set_description("Download the -debuginfo package instead");
+    debuginfo->set_const_value("true");
+    debuginfo->link_value(debuginfo_option);
 
     auto url = parser.add_new_named_arg("url");
     url->set_long_name("url");
@@ -162,6 +171,7 @@ void DownloadCommand::set_argument_parser() {
     create_destdir_option(*this);
     auto skip_unavailable = std::make_unique<SkipUnavailableOption>(*this);
     cmd.register_named_arg(srpm);
+    cmd.register_named_arg(debuginfo);
     cmd.register_named_arg(url);
     cmd.register_named_arg(urlprotocol);
     cmd.register_named_arg(allmirrors);
@@ -192,6 +202,10 @@ void DownloadCommand::configure() {
         context.get_base().get_repo_sack()->enable_source_repos();
     }
 
+    if (debuginfo_option->get_value()) {
+        context.get_base().get_repo_sack()->enable_debug_repos();
+    }
+
     context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
     // Default destination for downloaded rpms is the current directory
     context.get_base().get_config().get_destdir_option().set(libdnf5::Option::Priority::PLUGINDEFAULT, ".");
@@ -202,7 +216,10 @@ void DownloadCommand::run() {
 
     auto create_nevra_pkg_pair = [](const libdnf5::rpm::Package & pkg) { return std::make_pair(pkg.get_nevra(), pkg); };
 
-    std::map<std::string, libdnf5::rpm::Package> download_pkgs;
+    std::map<std::string, libdnf5::rpm::Package> original_binary_pkgs;
+    // Container to collect all packages to download
+    std::map<std::string, libdnf5::rpm::Package> packages_to_download;
+
     libdnf5::rpm::PackageQuery full_pkg_query(ctx.get_base(), libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK);
     for (auto & pattern : *patterns_to_download_options) {
         libdnf5::rpm::PackageQuery pkg_query(full_pkg_query);
@@ -231,7 +248,7 @@ void DownloadCommand::run() {
         }
 
         for (const auto & pkg : pkg_query) {
-            download_pkgs.insert(create_nevra_pkg_pair(pkg));
+            original_binary_pkgs.insert(create_nevra_pkg_pair(pkg));
 
             if (resolve_option->get_value()) {
                 auto goal = std::make_unique<libdnf5::Goal>(ctx.get_base());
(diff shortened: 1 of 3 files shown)


dnf5/commands/download/download.cpp
```diff
@@ -290,6 +309,37 @@ void DownloadCommand::run() {
...
+    if (debuginfo_option->get_value()) {
+        std::map<std::string, libdnf5::rpm::Package> debuginfo_pkgs;
+
+        libdnf5::rpm::PackageQuery debuginfo_pkg_query(ctx.get_base());
+        debuginfo_pkg_query.filter_available();
+
+        for (auto & [nevra, pkg] : download_pkgs) {
+            libdnf5::rpm::PackageQuery pkg_query(debuginfo_pkg_query);
+
+            std::string debuginfo_name = pkg.get_name() + "-debuginfo";
```
pkratoch: Please use the `get_debuginfo_name` method instead of appending the "-debuginfo" sufix. The method works correctly even for debuginfo packages (e.g. for "nano-debuginfo" it returns "nano-debuginfo" instead of creating "nano-debuginfo-debuginfo"). The `download --debuginfo` should work even if the debuginfo name is specified.

dnf5/commands/download/download.cpp
```diff
@@ -290,6 +309,37 @@ void DownloadCommand::run() {
...
+                for (const auto & dpkg : pkg_query) {
+                    debuginfo_pkgs.insert(create_nevra_pkg_pair(dpkg));
+                }
+            } else {
+                ctx.get_base().get_logger()->info("No debuginfo package found for: \"{}\"", pkg.get_name());
+            }
+        }
+
+        if (!debuginfo_pkgs.empty()) {
+            download_pkgs = debuginfo_pkgs;
```
pkratoch: The `download_pkgs` should be replaced by the `debuginfo_pkgs` in any case. If the debuginfo package is not found, then dnf should not download the base package. Right now it also masks a bug when a debuginfo package is not found when the "-debuginfo" sufix is specified. However it then results in a strange behavior when `dnf download --debuginfo nano-debuginfo htop` downloads only `htop-debuginfo`, but `dnf download --debuginfo nano-debuginfo` downloads `nano-debuginfo`.

pkratoch (changes requested):
Since `--debugsource` is not part of this PR, can you please create a separate issue for it?

Can you also write a few ci-dnf-stack tests for this feature? E.g. one that tests specifying both the base package name and the debuginfo name, or one for `dnf download --debuginfo --resolve`.

dnf5/commands/download/download.cpp
```diff
@@ -188,10 +198,19 @@ void DownloadCommand::configure() {
         context.set_load_system_repo(false);
     }
 
+    if (srpm_option->get_value() && debuginfo_option->get_value()) {
+        throw libdnf5::cli::ArgumentParserConflictingArgumentsError(
+            M_("Options \"--srpm\" and \"--debuginfo\" are mutually exclusive"));
+    }
+
```
m-blaha: Any particular reason for this? `dnf4 download --source --debuginfo` will happily download both - the source and debuginfo rpm packages.

fhbash: yeah.. you have a point here.. there's no reason to have this exclusive arg. I'll fix it.

dnf5/commands/download/download.cpp
```diff
@@ -258,14 +272,17 @@ void DownloadCommand::run() {
         return;
     }
 
-    if (srpm_option->get_value()) {
-        std::map<std::string, libdnf5::rpm::Package> source_pkgs;
+    std::map<std::string, libdnf5::rpm::Package> original_binary_pkgs = download_pkgs;
```
m-blaha: Is this copy needed? I think we can use `download_pkgs` in the for loops and drop `original_binary_pkgs` completely. Or maybe even better, rename `download_pkgs` -> `original_binary_pkgs`.

dnf5/commands/download/download.cpp
```diff
@@ -277,19 +294,52 @@ void DownloadCommand::run() {
...
+            ctx.get_base().get_logger()->warning("No debuginfo packages found for any of the specified packages");
+        }
     }
 
+    // If neither option is enabled, use original binary packages
+    if (!srpm_option->get_value() && !debuginfo_option->get_value()) {
+        packages_to_download = original_binary_pkgs;
+    }
+
+    download_pkgs = packages_to_download;
```
m-blaha: This copy can also be left out and we can use directly `packages_to_download` as source container in folowing `for` loops.

dnf5/commands/download/download.cpp
```diff
@@ -277,19 +294,52 @@ void DownloadCommand::run() {
...
+    if (debuginfo_option->get_value()) {
+        libdnf5::rpm::PackageQuery debuginfo_pkg_query(ctx.get_base());
+        debuginfo_pkg_query.filter_available();
+
+        bool found_any_debuginfo = false;
+        for (auto & [nevra, pkg] : original_binary_pkgs) {
+            libdnf5::rpm::PackageQuery pkg_query(debuginfo_pkg_query);
+
+            std::string debuginfo_name = pkg.get_debuginfo_name();
+            pkg_query.filter_name({debuginfo_name});
```
m-blaha, line 314: Please add also epoch filtering as mentioned in https://github.com/rpm-software-management/dnf5/pull/2547 review comment.

dnf5/commands/download/download.cpp
```diff
@@ -277,19 +294,52 @@ void DownloadCommand::run() {
...
 
-        download_pkgs = source_pkgs;
+        if (!found_any_debuginfo) {
+            ctx.get_base().get_logger()->warning("No debuginfo packages found for any of the specified packages");
+        }
     }
 
+    // If neither option is enabled, use original binary packages
+    if (!srpm_option->get_value() && !debuginfo_option->get_value()) {
+        packages_to_download = original_binary_pkgs;
```
m-blaha: Optimization: `packages_to_download = std::move(original_binary_pkgs);`

m-blaha (commented):
Thank you! This looks good. Please add epoch filtering and we are good to go.
Also consider optimizing performance by eliminating unnecessary copies of potentially large structures.

dnf5/commands/download/download.cpp
```diff
@@ -277,19 +294,52 @@ void DownloadCommand::run() {
...
+        debuginfo_pkg_query.filter_available();
+
+        bool found_any_debuginfo = false;
+        for (auto & [nevra, pkg] : original_binary_pkgs) {
+            libdnf5::rpm::PackageQuery pkg_query(debuginfo_pkg_query);
+
+            std::string debuginfo_name = pkg.get_debuginfo_name();
+            pkg_query.filter_name({debuginfo_name});
+            pkg_query.filter_version({pkg.get_version()});
+            pkg_query.filter_release({pkg.get_release()});
```
m-blaha, line 316: Just an idea - would `pkg_query.filter_evr(pkg.get_full_nevra())` be more effective than separate epoch/version/release filters? But I didn't do any tests and this can be definitely resolved as a follow up issue (if needed).

m-blaha: Oh, sorry, discard this comment. `get_full_nevra()` returns also name which needs to be altered. So the only option would be constructing a new `libdnf5::rpm::Nevra` object, setting correct name/epoch/version/release/arch there and then use `pkg_query.filter_nevra(nevra_object)`.

m-blaha (approved):

pkratoch (approved):

PR #2531, opened 2025-11-27
fix(pkg_list): add description of used colors

### LINKED ISSUE #2102: Document "dnf list --installed" output colors
Hello.

When I use `dnf list --installed *kernel*`, I get the following:
image

However, I cannot know what the colors of the output mean when there are multiple versions of the same package.
When I use `dnf list --installed`, almost all packages are bright green color.
Those of the kernel (used as example) are bright green, blue and white.

https://dnf.readthedocs.io/en/latest/command_ref.html#list-command
https://dnf5.readthedocs.io/en/latest/commands/list.8.html#list-command

I checked the above pages and didn't find an answer.

I would like to know what they mean, and I also want to make a suggestion to clear confusion for future users.

You could do it like so:
In the first line that says "Installed packages", acting as a title, you could put the meanings with colors in brackets.
Example:

Installed packages (**MeaningOne**, **MeaningTwo**, **MeaningThree**)

Where **MeaningOne** is colored with the appropriate color and says what it is. For example, "MeaningOne" can mean "Installed" with bright green color and "MeaningTwo" can mean "outdated package installed" with normal white color.

I believe this is a very simple addition to clear possible confusions for users.

Thanks.

mfocko:
### TODO

- [x] Also adjust the docs

### Summary

* Adjust colorizer
  * Add method that returns escaped description of used colors
  * Add a `reset` escape sequence to the list
* Use the description from the colorizer in `PackageListSections`' header

Screenshot: image

Fixes #2102

### DIFF
--- a/libdnf5-cli/output/pkg_colorizer.cpp
+++ b/libdnf5-cli/output/pkg_colorizer.cpp
@@ -46,6 +46,7 @@ const std::map<std::string_view, std::string_view> color_to_escape = {
     {"cyan", "\033[36m"},
     {"gray", "\033[37m"},
     {"white", "\033[1;37m"},
+    {"reset", "\033[0m"},
 };
 
 }
@@ -104,6 +105,34 @@ std::string PkgColorizer::to_escape(const std::string & color) {
         }
     }
     return output;
+}
+
+std::string PkgColorizer::get_coloring_description() {
+    std::ostringstream desc;
+
+    std::string separator("(");
+
+    // [NOTE](mfocko) Is there a possibility of any other meaning?
+    // Colorizer is written generically (with respect to the ordering against
+    // the “base set” of packages), but currently it is only used in the meaning
+    // as described below.
+    for (auto && [description, color] : {
+             std::make_pair("install", color_not_found),
+             std::make_pair("downgrade", color_lt),
+             std::make_pair("reinstall", color_eq),
+             std::make_pair("upgrade", color_gt),
+         }) {
+        if (color.empty()) {
+            // skip uncolored packages
+            continue;
+        }
+
+        desc << separator << "available for " << color << description << color_to_escape.at("reset");
+        separator = ", ";
+    }
+    desc << ")";
+
+    return desc.str();
 }
 
 }  // namespace libdnf5::cli::output
--- a/doc/commands/check-upgrade.8.rst
+++ b/doc/commands/check-upgrade.8.rst
@@ -39,6 +39,11 @@ is provided, it checks for updates for the entire system.
 ``DNF5`` will exit with code `100`` if updates are available and list them; `0` if no updates
 are available.
 
+If terminal is available, list of the packages is colored, packages available for reinstall are
+(by default) colored with bold green and packages available for upgrade with bold blue. This
+behavior can be adjusted in configuration via
+:ref:`color_list_available_upgrade <_color_list_available_upgrade_options-label>` and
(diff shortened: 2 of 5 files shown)


libdnf5-cli/output/pkg_colorizer.cpp
```diff
@@ -106,4 +107,30 @@ std::string PkgColorizer::to_escape(const std::string & color) {
     return output;
 }
 
+std::string PkgColorizer::get_coloring_description() {
+    std::ostringstream desc;
+    desc << "(";
+
+    // [NOTE](mfocko) Is there a possibility of any other meaning?
```
evan-goode, line 115: Even though PkgColorizer is only used with these meanings, we should keep it generic. Maybe `get_coloring_description` could take in the descriptions as 4 strings, similar to the constructor? Or we could add a new constructor and store the descriptions as member variables.

libdnf5-cli/output/pkg_colorizer.cpp
```diff
@@ -106,4 +107,30 @@ std::string PkgColorizer::to_escape(const std::string & color) {
...
+    // Colorizer is written generically (with respect to the ordering against
+    // the “base set” of packages), but currently it is only used in the meaning
+    // as described below.
+    for (auto && [description, color] : {
+             std::make_pair("install", color_not_found),
```
evan-goode, line 120: `reinstall` and `upgrade` are a little confusing on their own, IMO. Is there room to fit

```
Installed packages (available for reinstall, available for upgrade)
```

mfocko:
**Update**

I've updated the output to:
image

as for the suggestion to adjust the docs and update/add constructor for the colorizer…  I think that would involve multiple changes:
- I'm probably leaning more towards adjusting the defaults for coloring (from _ignored_ to _none by default_ for those 2 states that hard-coded to nothing in the code right now…)
- as for the possibility to pass custom description… there are multiple options of what I'd do, but none of them seems like something _clean & nice_? it would involve passing both the color and textual description of the meaning (that is to be displayed); they are 4 pairs of values, which are semantically tied together:
  - I don't think it makes sense to accept a `std::array` or even `std::initializer_list`, since you should be able to also omit the fields, this would imply that there needs to be a way to express _nothing_
  - as for the pairs themselves, `std::pair` would definitely work, but I don't like the fact that the fields don't carry any meaning (except for the one specified in the docs) and it's also more error-prone (e.g., swap color and description)
  - I'd probably introduce a simple struct holding these 2 values with named fields…
  - given the struct, it should be possible to wrap the arguments for coloring with `std::optional`; though I have a feeling it might not be necessary, as we can specify an empty default coloring (as we can right now…)

    also `std::optional` would require changes to parsing or how are the values in config handled…
    
    on top of that, I guess we could even make the descriptions customizable along the colors? though I'm not sure if there's a reasonable use-case that would be worth going so far :D

evan-goode:
> I'm probably leaning more towards adjusting the defaults for coloring (from ignored to none by default for those 2 states that hard-coded to nothing in the code right now…)

You mean the `color_not_found` and `color_lt` colors passed from check-upgrade and list commands?

> I'd probably introduce a simple struct holding these 2 values with named fields…

I like that.

> given the struct, it should be possible to wrap the arguments for coloring with std::optional

Maybe four `std::optional<SomeColorAndDescriptionStruct>`s?

Ordinarily I would say YAGNI but when building these APIs for maximum stability I think it's appropriate to try and make it general.

PR #2532, opened 2025-11-27
Musl support

aduskett:
Provide a few small changes to get musl working with dnf5. 

I've tested these on Yocto with DNF5 5.3.0.0 and it works perfectly!

### DIFF
--- a/libdnf5/common/sack/match_string.cpp
+++ b/libdnf5/common/sack/match_string.cpp
@@ -29,6 +29,10 @@
 #include <regex>
 #include <stdexcept>
 
+// TODO: Implement proper FNM_EXTMATCH support to fix MUSL errors.
+#if !defined(FNM_EXTMATCH)
+#define FNM_EXTMATCH (0)
+#endif
 
 namespace libdnf5::sack {
 


dnf5-plugins/config-manager_plugin/addrepo.cpp
```diff
@@ -168,7 +168,7 @@ void ConfigManagerAddRepoCommand::set_argument_parser() {
             throw cli::ArgumentParserError(
                 M_("{}: Badly formatted argument value \"{}\""), std::string{"set"}, std::string{value});
         }
-        std::string key{value, val};
+        std::string key{value, static_cast<const char *>(val)};
```
jrohel: How about instead of using `static_cast`, simply change the variable's declaration? That is, change the line `auto val = strchr(value + 1, '=');` to:
- `const auto * val = strchr(value + 1, '=');`
- or better `const auto * const val = strchr(value + 1, '=');`
- or `const char * const val = strchr(value + 1, '=');`

This also applies to other casts in this PR.

libdnf5/common/sack/match_string.cpp
```diff
@@ -29,6 +29,9 @@
 #include <regex>
 #include <stdexcept>
 
+#if !defined(FNM_EXTMATCH)
+#define FNM_EXTMATCH (0)
```
jrohel, line 34: This workaround ensures the code can be compiled with libraries that lack `FNM_EXTMATCH` support. However, this is not without consequence. The evaluation method for tested strings will change. But I agree with the decision. I prefer a functional DNF5 compiled with musl - even with a minor behavior deviation - over a non-compilable version.

Please add a comment along the lines: `// TODO: Implement proper FNM_EXTMATCH support`

dnf5-plugins/config-manager_plugin/addrepo.cpp
aduskett: Either way works for me! Fixed!

libdnf5/common/sack/match_string.cpp
aduskett: Added a comment as suggested!

jrohel (approved):
This PR is looking good. Thanks for your cooperation.

Issue #2540, opened 2016-07-19
Open Build Service repo management plugin (equivalent to copr plugin)


igor-raits:
Description of problem:
DNF doesn't include an equivalent of the copr command to manage Open Build Service repositories (either on private OBS instances or on build.opensuse.org). It's rather irritating to manually add repositories, and having them managed through a plugin makes it easier to work with.

Version-Release number of selected component (if applicable):
dnf-plugins-core-0.1.10-1.fc22
dnf-plugins-extras-0.0.9-1.fc22

Extra notes:
I don't mind if this is an extras plugin instead of a core one, but it would be pretty nice if it was easier to list/add/remove OBS repos from a manageability perspective. Obviously, since OBS serves a number of RPM-based distros (Fedora, RHEL, CentOS, OpenSUSE, SLE, and Mageia), it would make sense for this plugin to be somewhat configurable to be able to support whatever naming/versioning scheme is used by these distributions.

via @Conan-Kudo from https://bugzilla.redhat.com/show_bug.cgi?id=1251959

PR #2541, opened 2025-12-04
Add plugin for Open Build Service (OBS)

ddstreet:
This adds a 'obs' plugin to allow adding repos provided by the Open Build Service (OBS), i.e. https://build.opensuse.org/

The user interface is very similar to the COPR plugin; the same commands are provided (enable, disable, remove, list, debug). The main difference is the format of the repository "id":

- for COPR, the format is `[HUB/]OWNER/PROJECT [CHROOT]`
- for OBS the format is `[HUB/]PROJECT/REPONAME`

The `HUB` part is the same for both of them, i.e. the hostname (or shortname) of the webui/api of the service; if not specified the default (build.opensuse.org) is used, just like the COPR plugin.

The COPR repos will always have a `OWNER` that owns a specific `PROJECT`, and the OBS repos will also have some kind of `OWNER`, but the field doesn't have any fixed format; it can be "home:user" or "system:systemd" or other formats. The OBS repos also can have any number of levels of `PROJECT`, including none at all (e.g. "home:user" can have a repo, "home:user:subproject1" can have a repo also, "home:user:subproject1:subsubproject2" can also, etc.). It makes more sense, to me at least, to simply roll the `OWNER` and `PROJECT`(s) together as a single `PROJECT` (with each level delimited by colons; and this is how OBS presents it to users).

The OBS plugin also differs from COPR by including the actual repo name as the `REPONAME` in the id. The repo name roughly corresponds to the COPR "chroot" (e.g. "fedora-rawhide-ppc64le"), but the repo isn't arch-specific (technically it is, but OBS repos aren't separated by arch like COPR repos are, so most repos have multiple archs enabled), and the repo name isn't guessable based on the local system. Users can name their repos whatever they want.

One other note; when the plugin is provided a repo spec (i.e. `PROJECT/REPONAME`), it forms a URL to the OBS repo's "repository state" HTML web page, and downloads it, and then parses it (using libxml2 HTML parsing) to extract the HREF to the repo download URL. I don't think this is ideal, but it's required because of 2 reasons:

1. While the OBS does have an API, none of its methods are usable anonymously (at least, not in any of my testing using the standard api.opensuse.org). Even if this was changed, the plugin should still work with other "hubs" (i.e. other instances of OBS), which may not allow anonymous API access.
2. The URL to the repos is unfortunately not predictable, as it's up to the specific configuration of each OBS instance. The main OBS instance uses an entirely different hostname (https://download.opensuse.org) while other instances of OBS most likely just use the same hostname with a dedicated path to the download area.

Finally, this of course doesn't have any translations for the plugin, although I did add a .pot file.

Fixes: #2540

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -855,7 +855,7 @@ Provides:       dnf5-command(repomanage)
 Core DNF5 plugins that enhance dnf5 with builddep, changelog, config-manager,
 copr, obs, needs-restarting, repoclosure, repomanage, and reposync commands.
 
-%files -n dnf5-plugins -f dnf5-plugin-builddep.lang -f dnf5-plugin-changelog.lang -f dnf5-plugin-config-manager.lang -f dnf5-plugin-copr.lang -f dnf5-plugin-obs.lang -f dnf5-plugin-needs-restarting.lang -f dnf5-plugin-repoclosure.lang -f dnf5-plugin-reposync.lang
+%files -n dnf5-plugins -f dnf5-plugin-builddep.lang -f dnf5-plugin-changelog.lang -f dnf5-plugin-config-manager.lang -f dnf5-plugin-copr.lang -f dnf5-plugin-needs-restarting.lang -f dnf5-plugin-repoclosure.lang -f dnf5-plugin-reposync.lang
 %{_libdir}/dnf5/plugins/builddep_cmd_plugin.so
 %{_libdir}/dnf5/plugins/changelog_cmd_plugin.so
 %{_libdir}/dnf5/plugins/config-manager_cmd_plugin.so
@@ -1060,7 +1060,6 @@ mkdir -p %{buildroot}%{_libdir}/libdnf5/plugins
 %find_lang dnf5-plugin-changelog
 %find_lang dnf5-plugin-config-manager
 %find_lang dnf5-plugin-copr
-%find_lang dnf5-plugin-obs
 %find_lang dnf5-plugin-needs-restarting
 %find_lang dnf5-plugin-repoclosure
 %find_lang dnf5-plugin-reposync


dnf5-plugins/obs_plugin/obs_constants.hpp
```diff
@@ -0,0 +1,60 @@
...
+// along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef DNF5_COMMANDS_OBS_CONSTANTS_HPP
+#define DNF5_COMMANDS_OBS_CONSTANTS_HPP
+
+#include <libdnf5/utils/bgettext/bgettext-lib.h>
+#include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
+
+static const char * const OBS_COMMAND_DESCRIPTION =
+    _("Manage Obs repositories (add-ons provided by users/community/third-party)");
```
Conan-Kudo: Please consistently use `OBS` or `Open Build Service`, and not `Obs`.

dnf5-plugins/obs_plugin/obs_constants.hpp
```diff
@@ -0,0 +1,60 @@
...
+static const char * const OBS_THIRD_PARTY_WARNING =
+    _("Enabling a Obs repository. Please note that this repository is not part\n"
+      "of the main distribution, and quality may vary.\n"
+      "\n"
+      "The Fedora Project does not exercise any power over the contents of\n"
```
Conan-Kudo: Since the default is actually the openSUSE Build Service, this needs to be "The openSUSE Project"

dnf5-plugins/obs_plugin/obs_constants.hpp
```diff
@@ -0,0 +1,60 @@
...
+      "this repository, and packages are not held to any quality or security\n"
+      "level.\n"
+      "\n"
+      "Please do not file bug reports about these packages in Fedora\n"
```
Conan-Kudo: `Fedora`->`openSUSE`.

dnf5-plugins/obs_plugin/obs_constants.hpp
```diff
@@ -0,0 +1,60 @@
...
+      "Bugzilla. In case of problems, contact the owner of this repository.\n");
+
+static const char * const OBS_EXTERNAL_DEPS_WARNING =
+    _("Maintainer of the enabled Obs repository decided to make\n"
+      "it dependent on other repositories. Such repositories are\n"
+      "usually necessary for successful installation of RPMs from\n"
+      "the main Obs repository (they provide runtime dependencies).\n"
+      "\n"
+      "Be aware that the note about quality and bug-reporting\n"
+      "above applies here too, Fedora Project doesn't control the\n"
```
Conan-Kudo: `Fedora Project` -> `the openSUSE Project`

ddstreet: updated

dnf5-plugins/obs_plugin/obs_constants.hpp
ddstreet: updated

dnf5-plugins/obs_plugin/obs_constants.hpp
ddstreet: updated

dnf5-plugins/obs_plugin/obs_constants.hpp
ddstreet: updated, i think i got all of them but please let me know if i missed any

ddstreet:
per above comments, changed all 'Obs' -> 'OBS' in comments and strings, as well as 'Fedora' -> 'openSUSE' in strings

ddstreet:
another update to remove and adjust some code comments

Conan-Kudo:
Technically OBS repos could, but it's hard to identify them and it's better off not worrying about it.

Conan-Kudo:
Wouldn't it make sense to also add a key like `download_url_prefix` to the config where this can be set instead of scraping the web site for this?

It's largely formulaic after you get to the `https://download.opensuse.org/repositories/` part. This also makes it far less messy for third-party OBS instances.

Issue #2479, opened 2025-10-21
--no-gpgchecks does not override rpm enforcing policy


pmatilai:
Seems like the fix for #2229 needs to be extended a bit to cover --no-gpgchecks:

```
[root@lumikko ~]# dnf download telnet
Updating and loading repositories:
Repositories loaded.
Downloading Packages:
 telnet-1:0.17-94.fc42.x86_64           100% | 179.3 KiB/s |  61.9 KiB |  00m00s
[root@lumikko ~]# rpmsign --delsign telnet-0.17-94.fc42.x86_64.rpm
[root@lumikko ~]# dnf --quiet -y install --no-gpgchecks telnet-0.17-94.fc42.x86_64.rpm 
Package                 Arch   Version                  Repository          Size
Installing:
 telnet                 x86_64 1:0.17-94.fc42           @commandline   112.7 KiB

Transaction Summary:
 Installing:         1 package

Transaction failed: Rpm transaction failed.
Warning: skipped OpenPGP checks for 1 package from repository: @commandline
  - package telnet-1:0.17-94.fc42.x86_64 does not verify: no signature

```

ppisar:
I cannot reproduce it with developmental DNF5. Could you tell me what DNF5 do you use?

pmatilai:
Oh. Like #2229, this requires enforcing signature mode in rpm. Basically

```
echo "%_pkgverify_level all" > /etc/rpm/macros.verify
```

pmatilai:
This is with dnf5-5.2.16.0-1.fc42.x86_64, if there have been changes to this area since then I've missed that (entirely possible of course)

PR #2542, opened 2025-12-04
Move libdnf5/conf/config.h creation after feature detection

pmatilai:
Features need to be detected before they can be written down, doh. In particular this is needed for the rpmteSetVfyLevel() symbol detection from 4db642138c76474553213b0f9005a78f32fbf932, which previously only worked by chance in some incremental build cases.

Move the config header creation after all the detection is complete, next to the other configure_file() which is down there for similar reasons...

Fixes: #2479


pmatilai:
Thanks for the quick review!

PR #2546, opened 2025-12-05
spec: remove libpkgmanifest-devel dep from dnf5-plugin-manifest

yselkowitz:
`pkgconfig(libpkgmanifest)` is already declared as a `BuildRequires` earlier in the spec.  The plugin depends only on the runtime library, whose dependency is automatically generated, but not on the -devel package.


ppisar (commented):
It looks reasonable.

PR #2550, opened 2025-12-15
Convert `GoalProblem` enum to a 64 bit value (breaks ABI)

kontura:
Since we are [breaking the ABI](https://github.com/rpm-software-management/dnf5/pull/2519) again I though it would be good to do this change because I want to add another value to `GoalProblem` enum and its seems to be filling up.


m-blaha:
Yeah, this is a good idea.

PR #2558, opened 2026-01-06
Doc: Add plugin_conf_dir, fix pluginconfpath and pluginpath option descriptions 

jrohel:
- Add documentation for the `plugin_conf_dir` configuration option which supports a list of directories for loading libdnf5 plugin configuration files.
- Fix `pluginconfpath` and `pluginpath` option descriptions. The options support only a single directory, not a list of directories.
- Added details about `.conf` file extension requirement and plugin loading conditions.

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -433,22 +433,21 @@ repository configuration file should aside from repo ID consists of baseurl, met
 .. _pluginconfpath_options-label:
 
 ``pluginconfpath``
-    :ref:`list <list-label>`
+    :ref:`string <string-label>`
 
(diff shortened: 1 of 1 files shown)


m-blaha (approved):

m-blaha:
There seem to be problems with building dnf5 on rawhide. I guess swig 4.4 behaves differently?

ppisar:
If you mean the array offset warning-as-an-error in Ruby bindings, it could be triggered by GCC 16 which is now in Fedora 44.

m-blaha:
Or gcc, thanks! I did not look into it too closely since the failures are definitely not caused by documentation enhancements :)

kontura:
@jrohel  can you rebase?

PR #2561, opened 2026-01-11
copr: ignore case-sensitivity in the distrubution name

FrostyX:
Fix https://github.com/fedora-copr/copr/issues/3539

### DIFF
--- a/dnf5-plugins/copr_plugin/copr_config.cpp
+++ b/dnf5-plugins/copr_plugin/copr_config.cpp
@@ -98,6 +98,7 @@ void CoprConfig::load_all_configuration() {
 
     // Set the "name_version" for the later convenience
     std::string name_version = this->get_value("main", "distribution");
+    name_version = libdnf5::utils::string::tolower(name_version);
     name_version += "-" + this->get_value("main", "releasever");
     this->set_value("main", "name_version", name_version);
 
--- a/dnf5-plugins/copr_plugin/copr_config.hpp
+++ b/dnf5-plugins/copr_plugin/copr_config.hpp
@@ -22,6 +22,7 @@
 
 #include <libdnf5/base/base.hpp>
 #include <libdnf5/conf/config_parser.hpp>
+#include "utils/string.hpp"
 
 namespace dnf5 {
 


dnf5-plugins/copr_plugin/copr_config.cpp
```diff
@@ -98,6 +98,7 @@ void CoprConfig::load_all_configuration() {
 
     // Set the "name_version" for the later convenience
     std::string name_version = this->get_value("main", "distribution");
+    name_version = libdnf5::utils::string::tolower(name_version);
```
praiskup, line 101: Wouldn't it be better to fix the config so it avoids unnecessary capitalization? The config is only used by dnf copr enable, and these case conversions just add unnecessary complexity.

FrostyX: We could, and maybe the Asahi people already did so. I just thought it would be better to prevent similar problems for other distributions in the future, since this costs us basically nothing.

But if you are worried that someone somewhere might want mock configs that are not all lowercase, we can close this PR and I can just ping the Asahi people to lowercase it in their config.

dnf5-plugins/copr_plugin/copr_config.hpp
```diff
@@ -22,6 +22,7 @@
 
 #include <libdnf5/base/base.hpp>
 #include <libdnf5/conf/config_parser.hpp>
+#include "utils/string.hpp"
```
mfocko, line 25: I'd add it to the `.cpp`, since it's required only by the implementation.

I'd put it over (or under) the
```cpp
#include "libdnf5/utils/os_release.hpp"
```
that's already there

dnf5-plugins/copr_plugin/copr_config.cpp
mfocko: If I understand it correctly, it means you'd need to:
- accept the PR
- release to Fedora
- include the new mock release in the Copr deployment

From my perspective it would be better to fix just the config, if the mock is case-sensitive (which it most likely is, cause I got yelled at, the other day, for omitting just `.cfg` from the root :D).

> [!NOTE]
> I've also had a look at the mock configs, I expected openSUSE to have SUSE upper-case, but it's not the case.

OTOH I'm not sure if this really “hurts” per se…

praiskup: I believe all the tooling around is case-sensitive.

praiskup: This should be fixed in the broken config. Consider that one day, a distro might decide to add a random string like 'foo' to this config, and users will expect us to map it to 'bar' upstream.

PR #2564, opened 2026-01-13
Fix builds on rawhide

kontura:


ppisar:
I guess you need build everything from scratch. Then you will see that GCC 16 does not like how Swig generates Ruby bindings <https://bugzilla.redhat.com/show_bug.cgi?id=2427953>.

kontura:
I believe there was indeed an additional issue that wasn't showing in local non-mock builds due to the differences in configuration.

I have added one more commit `Replace all std::format with fmt::format` that seems to fix it however to be honest I am not sure why or what even is the original problem. I would be interested to know what is going on if someone understands it, perhaps @jrohel?

dnf5-plugins/copr_plugin/json.cpp
```diff
@@ -60,6 +60,8 @@ std::vector<std::string> Json::keys() {
     std::vector<std::string> retval;
     json_object_object_foreach(root, key, val) {
         retval.push_back(key);
+        // Silence error: variable ‘val’ set but not used
+        (void)val;
```
jrohel: Minor tip: `std::ignore = val;` would be more explicit. It requires `<utility>`, which might be included transitively, but it's cleaner to include it explicitly in this file.

kontura: Updated.

Issue #2544, opened 2025-12-05
RFE: Support restricting stickyvendor with {include,exclude}{src,}pkgs


michel-slm:
It would be much easier to onboard using sticky vendors if we can restrict the set of packages it applies to.

Making the syntax matches that used by DNF's repo configuration (includepkgs, excludepkgs) would make it easier - in fact we should probably reuse the same implementation.

But if we can extend this (both for sticky vendor and for repo configuration in general) to also allow restricting by *source RPM* - maybe includesrcpkgs and excludesrcpkgs - that would be really great, especially for source RPMs that generate hundreds if not more of binary RPMs.

(e.g. mesa, qemu)

ppisar:
I think this enhancement belongs to vendorpolicy configuration files rather than to main excludepkg.

jannau:
An alternative to this for the Fedora-Asahi-Remix use case might be using libsolv's vendor equivalence classes. See [pool_addvendorclass()](https://github.com/openSUSE/libsolv/blob/56f5ca0d86398ab68c38aa17883c4c9ebec95fe5/doc/libsolv-pool.txt?plain=1#L330).

The idea behind that would be to declare two equivalence relations (vendor names for illustration):
1. `copr-asahi` -> `copr-asahi-transitional`
2. `copr-asahi-transitional` -> `fedora`

The vendor name can be set freely in the spec file. So for every package we want to transition to the fedora build the copr spec file overrides the vendor to `copr-asahi-transitional` and eventually through normal updates the fedora package will be installed.

jannau:
A quick test on rawhide worked as expected. It's good enough to solve the immediate problem transitioning packages from vendored copr builds to Fedora builds. This is essentially what I was looking into to implement. Thanks.

For planning purposes: Is dnf5 5.3.0.x expected to land in Fedora 43 before Fedora 44 beta?

The request looks still useful to me. Fedora-Asahi-Remix uses `allow_vendor_change=false` globally only because it's necessary for a few packages since non-vendored packages either do not work or work poorly. There is no intention to diverge from Fedora's behaviour. Disallowing vendor changes causes usability issues for user which want to install a newer version of $RANDOM_PACKAGE from copr and expect that to work as documented for Fedora.

PR #2579, opened 2026-01-20
VendorChangeManager: Configuration version `1.1` with package-based filtering and enhanced flexibility

jrohel:
This PR introduces version `1.1` of the Vendor Change Manager configuration format, bringing improvements in flexibility and usability while maintaining full backward compatibility with version `1.0`.

Closes: https://github.com/rpm-software-management/dnf5/issues/2544
CI tests for new features: https://github.com/rpm-software-management/ci-dnf-stack/pull/1817

### 1. Simplified vendor array rules
- Combine `equivalent_vendors` with directional arrays:
  - Version `1.1` allows mixing `equivalent_vendors` with `outgoing_vendors`/`incoming_vendors` for more flexible policies  (Note: `equivalent_vendors` is automatically expanded to both `outgoing_vendors` and `incoming_vendors`)
  - Version `1.0` restricted this combination

- Independent vendor arrays: `outgoing_vendors` and `incoming_vendors` can now be specified independently
  - Missing array implies "any vendor is allowed"
  - Version `1.0` required both arrays to be present or both missing

## 2. Package-based filtering (new feature)

  - **Added `incoming_packages` and `outgoing_packages` arrays to restrict vendor changes based on package attributes**
  
  - **Supported package filters:**
    - `name`, `source_name`, `arch`, `repoid` (string-based)
    - `evr`, `epoch`, `version`, `release` (version-based with proper RPM comparison)
    - `cmdline_repo` (boolean)
  - String comparators: `EXACT`, `IEXACT`, `CONTAINS`, `STARTSWITH`, `ENDSWITH`, `REGEX`, `GLOB` (and negated variants)
  - Version comparators: `EXACT`, `NOT_EXACT`, `GT`, `GTE`, `LT`, `LTE` (with proper RPM version ordering)
  - Exclusion support: Rules can be marked with `exclude = true` for defining exclusion

###  3. Backward compatibility

  - Version `1.0` configurations continue to work unchanged
  - All version `1.0` restrictions remain enforced for `1.0` configs
  - Upgrade to `1.1` is opt-in via the version field

###  4. Improved validation

  - Vendor regex patterns are now validated during config file parsing
  - Errors are caught early rather than during goal resolution

###  5. Enhanced documentation

  - Documented improvements
  - Examples for new features
  - Documented config file locations and ordering rules (.conf extension requirement)

## Examples

Allow any vendor to be replaced by "My Trusted Vendor".
```toml
version = '1.1'

[[incoming_vendors]]
vendor = 'My Trusted Vendor'
```

Allow: Mutual changes between "First Vendor" and "Second Vendor" (equivalent vendors), change from "First Vendor"/"Second Vendor" to "Third Vendor"
```toml
version = '1.1'

[[equivalent_vendors]]
vendor = 'First Vendor'

[[equivalent_vendors]]
vendor = 'Second Vendor'

[[incoming_vendors]]
vendor = 'Third Vendor'
```

Allow use packages from the command-line repo from any vendor:
```toml
version = '1.1'

[[incoming_packages]]
filters = [
  { filter = 'cmdline_repo', value = 'true' }
]
```

Allow use packages from the command-line repo from any vendor, except for packages whose names start with "mypackage":
```toml
version = '1.1'

[[incoming_packages]]
filters = [
  { filter = 'name', value = 'mypackage', comparator = 'STARTSWITH' }
]
exclude = true

[[incoming_packages]]
filters = [
  { filter = 'cmdline_repo', value = 'true' }
]
```

This example allows a change from any vendor to "My Vendor", but only for incoming packages whose source package name is "mypackage" and version is greater than or equal to "2.0".
```toml
version = '1.1'

[[incoming_packages]]
filters = [
  { filter = 'source_name', value = 'mypackage' },
  { filter = 'version', value = '2.0', comparator = 'GTE' }
]

[[incoming_vendors]]
vendor = 'My Vendor'
```

### DIFF
--- a/doc/dnf5.conf-vendorpolicy-v1_1.5.rst
+++ b/doc/dnf5.conf-vendorpolicy-v1_1.5.rst
@@ -238,27 +238,6 @@ Each entry in ``[[outgoing_packages]]`` or ``[[incoming_packages]]`` can contain
 
     Default: ``false``
 
-Error Conditions
-================
-
-The following configurations are **invalid** and will cause an error:
-
-- Missing ``version`` field
-- Version value other than ``"1.1"``
-- Missing required ``vendor`` field in a vendor entry
-- Missing required ``filter`` or ``value`` field in a package filter
-- Missing ``filters`` array in a package entry
-- Empty ``filters`` array in a package entry
-- Unknown ``filter`` value (must be one of: "name", "source_name", "evr", "epoch", "version", "release", "arch", "repoid", "cmdline_repo")
-- Unknown ``comparator`` value
-- Invalid ``comparator`` for string-based filters (must be string comparison operators)
-- Invalid ``comparator`` for version-based filters (must be EXACT, NOT_EXACT, GT, GTE, LT, or LTE)
-- Invalid ``comparator`` for ``cmdline_repo`` filter (only EXACT is supported)
-- Invalid ``value`` for ``cmdline_repo`` filter (must be "true"/"1" or "false"/"0")
-- Invalid ``value`` for ``epoch`` filter (must be a valid number)
-- Invalid regex pattern in filter value
-- Unknown keys at the top level or inside entries
-
 Examples
 ========
 
--- a/doc/dnf5.conf-vendorpolicy-v1_0.5.rst
+++ b/doc/dnf5.conf-vendorpolicy-v1_0.5.rst
@@ -133,19 +133,6 @@ can contain the following fields:
 
     Default: ``false``
 
-Error Conditions
-================
-
-The following configurations are **invalid** and will cause an error:
-
-- Missing ``version`` field
-- Combination of ``equivalent_vendors`` with ``outgoing_vendors`` or ``incoming_vendors``
-- Only ``outgoing_vendors`` without ``incoming_vendors`` (or vice versa)
-- Missing required ``vendor`` field in a vendor entry
-- Unknown ``comparator`` value
-- Invalid regex pattern in vendor value
-- Unknown keys at the top level or inside entries
-
 Examples
 ========
 


doc/dnf5.conf-vendorpolicy.5.rst
```diff
@@ -40,7 +40,9 @@ Required Fields
 Vendor Mapping Definition
 --------------------------
 
-The file must contain **one** of the following definition methods:
+The file can use one or more of the following vendor definition methods.
```
evan-goode: I would rather not mix the documentation for version 1.0 and version 1.1 of the configuration file. Most users will just want to write a file in the latest version and won't care about limitations of the older spec, especially since 1.0 never made it to a stable Fedora. Could we keep the 1.0 documentation in a separate page?

libdnf5/solv/vendor_change_manager.cpp
```diff
@@ -119,6 +124,8 @@ const VendorChangeManager::VendorChangeMasks & VendorChangeManager::get_vendor_c
 void VendorChangeManager::load_vendor_change_policy(const std::filesystem::path & path) {
     VendorChangePolicy policy;
 
+    bool is_config_version_1_0;
```
evan-goode, line 546: Given the above, we may not even need to support 1.0 at all. We have a new major version with 5.4.0.0 after all. I'm fine either way.

libdnf5/solv/vendor_change_manager.cpp
```diff
@@ -219,6 +219,21 @@ void VendorChangeManager::load_vendor_change_policy(const std::filesystem::path
                         location_first_line_num(location));
                 }
 
+                // Validate regex pattern
```
evan-goode, line 726: Nice to have.

libdnf5/solv/vendor_change_manager.cpp
```diff
@@ -121,6 +187,202 @@ const VendorChangeManager::VendorChangeMasks & VendorChangeManager::get_vendor_c
 }
 
 
+bool VendorChangeManager::metches_package_defs(
```
evan-goode: typo: should be `matches`

libdnf5/solv/vendor_change_manager.cpp
```diff
@@ -70,6 +81,61 @@ class VendorChangePolicyConfigFileError : public Error {
...
+    }
+
+    const auto & outgoing_vendor_mask = get_vendor_change_masks(outgoing_vendor).outgoing_mask;
+    if (outgoing_vendor_mask.empty()) {
+        // The outgoing vendor is not involved in any valid policy change.
+        // Therefore, any change is illegal.
+        return false;
+    }
+
+    const auto & incomin_vendor_mask = get_vendor_change_masks(incoming_vendor).incoming_mask;
```
evan-goode: typo: `incoming`

evan-goode (commented):
I took a look but have yet to test this. It seems quite powerful!

@Conan-Kudo @michel-slm wdyt?

> Allow use packages from the command-line repo from any vendor:

We probably want this by default, will we ship this as a configuration file downstream in Fedora?

libdnf5/solv/vendor_change_manager.cpp
jrohel: I wanted to handle the configuration file extension properly, which is why I introduced version `1.1`. Version `1.1` is backward compatible with `1.0` - meaning that if you change `1.0` to `1.1` in an old config file, it will work correctly in the new libdnf. However, I intentionally implemented checks in the new libdnf to disallow version `1.1` extensions when version `1.0` is specified. This was done so the user can be sure that if they test a version `1.0` config in the new libdnf, it will still work in older versions of libdnf that lack `1.1` support. It's simply a clean solution, though it’s a question of whether it’s strictly necessary. libdnf with version 1.0 support was released in Fedora Rawhide - but Rawhide is not a stable release.

Conan-Kudo: It has been released into openSUSE Tumbleweed though, which _is_ a stable rolling release. So it does matter.

doc/dnf5.conf-vendorpolicy.5.rst
jrohel: @evan-goode OK. I added commit that splits documentation to version specific files.

libdnf5/solv/vendor_change_manager.cpp
```diff
@@ -114,28 +225,354 @@ const VendorChangeManager::VendorChangeMasks & VendorChangeManager::get_vendor_c
...
     VendorChangePolicy policy;
 
+    bool is_config_version_1_0;
+
     try {
         // Parse the TOML file
         auto config = toml::parse(path);
 
         // Check config file version
         const auto version = toml::find<std::string>(config, "version");
```
evan-goode: Perhaps not the fault of this PR, but this SIGABRTs if the file is missing a version. The error message is helpful at least:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  [error] toml::value::at: key "version" not found
 --> /etc/dnf/vendors.d/copr.conf
   |
 1 | # version = '1.0'
   | ^-- at the top-level table


Program received signal SIGABRT, Aborted.
```

doc/dnf5.conf-vendorpolicy-v1_1.5.rst
```diff
@@ -0,0 +1,421 @@
...
+
+    Optional field.
+
+    If ``true``, packages matching the filters are excluded from the rule. This is
+    useful for defining exceptions. Exclusion rules must appear before the rules
+    from which packages should be excluded.
+
+    Default: ``false``
+
+Error Conditions
```
evan-goode: This list of error conditions is a bit redundant The schema above already describes what's required.

doc/dnf5.conf-vendorpolicy-v1_1.5.rst
```diff
@@ -0,0 +1,421 @@
...
+-----------------------------
+
+This example shows vendors that are mutually equivalent, allowing changes
+in both directions.
+
+.. code-block:: toml
+
+    version = '1.1'
+
+    # All following vendors are mutually equivalent
```
evan-goode, line 283: Meaning the "CentOS" vendor would be considered equivalent to the "Red Hat" vendor, correct?

And if the user wanted to consider "CentOS*" equivalent, and consider "Red Hat*" equivalent, but not consider "CentOS*" equivalent to "Red Hat*", they would write two different configuration files, each with its own `[[equivalent_vendors]]` section, correct?

doc/dnf5.conf-vendorpolicy.5.rst
```diff
@@ -16,137 +16,26 @@ They specify which vendors (outgoing) are allowed to change to which other vendo
...
+The configuration files use **TOML format**.
 
-File Structure
-==============
+Supported Versions
+==================
 
-Required Fields
----------------
+The vendor change policy configuration format supports two versions:
```
evan-goode, line 24: Should we make it clearer that version 1.1 is preferred, and support for 1.0 is kept for compatibility?

evan-goode (commented):
LGTM otherwise. I tested manually, it works as expected. The docs are thorough and therefore long, but the large set of examples makes it easy enough to understand quickly. Thanks for making separate doc pages for 1.0 and 1.1.

CI tests with https://github.com/rpm-software-management/ci-dnf-stack/pull/1817 are passing locally.

doc/dnf5.conf-vendorpolicy.5.rst
Conan-Kudo: Yes.

doc/dnf5.conf-vendorpolicy-v1_1.5.rst
jrohel: @evan-goode 
> Meaning the "CentOS" vendor would be considered equivalent to the "Red Hat" vendor, correct?

Yes


> And if the user wanted to consider "CentOS*" equivalent, and consider "Red Hat*" equivalent, but not consider "CentOS*" equivalent to "Red Hat*", they would write two different configuration files, each with its own `[[equivalent_vendors]]` section, correct?

Yes

libdnf5/solv/vendor_change_manager.cpp
jrohel: Fixed. Handle missing `version` key to prevent unexpected crashes.

doc/dnf5.conf-vendorpolicy-v1_1.5.rst
jrohel: @evan-goode  OK. I removed section Error Conditions.

doc/dnf5.conf-vendorpolicy.5.rst
jrohel: OK. I've added this information to the documentation.

doc/dnf5.conf-vendorpolicy-v1_1.5.rst
evan-goode: OK, makes sense.

Issue #2326, opened 2025-06-27
history database doesn't use SOURCE_DATE_EPOCH


nalind:
Hi, I'm working on making container image builds one in [buildah](https://github.com/containers/buildah) easier to reproduce.
While a lot of this boils down to us improving our support for [`$SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) and leaning on RPM's built-in support for it (particularly https://github.com/rpm-software-management/rpm/pull/1803), setting the variable doesn't seem to affect the "dt_begin" and "dt_end" fields in the "trans" table in the history database, and that's giving me some trouble.
Could those values also be set to something other than the current time if $SOURCE_DATE_EPOCH is set in the environment?  For my case, they don't have to reflect that _exact_ value, as it's enough if they're merely consistent across invocations.

nalind:
> Just keep in mind that the `SOURCE_DATE_EPOCH` should be incremented between multiple `dnf install ...` calls during the container build. Otherwise all the transactions in the history db would get the same dt_begin (although dt_end might vary depending on how long did it take to complete the transaction).

Hmm, the dt_end value changing between runs would be something we'd need to control, too, since we're aiming for multiple runs to produce the same bit-for-bit history database files.

> Just keep in mind that the `SOURCE_DATE_EPOCH` should be incremented between multiple `dnf install ...` calls during the container build. Otherwise all the transactions in the history db would get the same dt_begin (although dt_end might vary depending on how long did it take to complete the transaction).

That could be difficult if, as https://reproducible-builds.org/docs/source-date-epoch/#dockerfile describes as its first option, the value is passed in from outside of the build.  We don't have syntax that let us do arithmetic evaluation to increment a previously-set numeric value in Dockerfiles like we could in a bash script.

> Meanwhile the direct sqlite manipulation hack mentioned in the first comment can be used.

I'll have to make a note of that one.  You definitely saved me some time with that snippet, thanks!

cgwalters:
I would suggest making the timestamps optional in the schema entirely if possible.

As far as I know the use case for having the dnf database in the container image is just to query which rpm-md repository the rpm came from. Which personally I think would make sense to just store in the rpm database and have at least some first class support for rpm for querying i.e. `rpm -q --queryformat='%{RPMMD_REPO}` foo`.

Then we likely wouldn't need the dnf database in server-built images at all.

ppisar:
[...]

PR #2584, opened 2026-01-23
transaction: honor SOURCE_DATE_EPOCH for history timestamps

jlebon:
When SOURCE_DATE_EPOCH is set, use it for transaction timestamps in the history database. This in turn makes it easier to achieve reproducible container image builds.

(Another option is to just not ship the history database, but at least Clair today needs this to be able to associate RPMs to yum repos. There are proposals to try to improve that in the future.)

Fixes: https://github.com/rpm-software-management/dnf5/issues/2326

Assisted-by: Claude Opus 4.5

### DIFF
--- a/doc/dnf5.8.rst
+++ b/doc/dnf5.8.rst
@@ -435,6 +435,12 @@ Environment
 ``LIBDNF_PLUGINS_CONFIG_DIR``
     Override a directory with libdnf5 plugin's configuration files.
 
+``SOURCE_DATE_EPOCH``
+    Specifies a timestamp given as number of seconds since UNIX epoch. When set,
+    transaction timestamps in the history database are set to the specified time
+    instead of the current time. Additionally, packages are ordered to ensure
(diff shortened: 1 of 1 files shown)


doc/dnf5.8.rst
```diff
@@ -435,6 +435,11 @@ Environment
 ``LIBDNF_PLUGINS_CONFIG_DIR``
     Override a directory with libdnf5 plugin's configuration files.
 
+``SOURCE_DATE_EPOCH``
+    When set, transaction timestamps in the history database are set to the
+    specified time instead of the current time. It also affects package ordering
```
ppisar: I'm missing documented format of the variable value. I think "…set to the specified UNIX time instead of the current time" could do it.

jlebon: Yes good point. I think I've been working in that area so long, it felt unnecessary. But for completeness, we should document it.

Issue #2586, opened 2026-01-27
repoclosure lacks any machine readable format for output


hrw:
In the issue #512 it was decided that JSON will be used as machine readable format.

`dnf repoclosure` still outputs mess and lacks `--json`:

```
dnf repoclosure --help
Użycie:
  dnf5 [GLOBAL OPTIONS] repoclosure [OPTIONS] [ARGUMENTS] 
                                
Options:                        
  --check=<REPO_ID>,...         Specify repo ids to check
  --arch=<ARCH>,...             Only check packages of specified architectures.
  --newest                      Only consider the latest version of a package from each repo.
                                
Arguments:                      
  specs                         List of package specs to check closure for
```

```
$ dnf --version
dnf5 wersja 5.2.17.0
dnf5 wersja API wtyczek 2.0
wersja libdnf5 5.2.17.0
wersja API wtyczek libdnf5 2.2
```

PR #2594, opened 2026-02-02
repoclosure: Add JSON output support and documentation

Userfrom1995:
Implement JSON machine-readable output for the repoclosure command, aligning it with the project standards for consistency with other commands like history and advisory.

Changes:
- Add --json option to repoclosure command
- Suppress resolution errors and progress messages in JSON mode
- Update repoclosure documentation (man page) with JSON output details

Fixes: #2586
Related to #512

### DIFF
--- a/dnf5-plugins/repoclosure_plugin/repoclosure.cpp
+++ b/dnf5-plugins/repoclosure_plugin/repoclosure.cpp
@@ -19,7 +19,9 @@
 
 #include "repoclosure.hpp"
 
+#include <json-c/json.h>
 #include <libdnf5-cli/argument_parser.hpp>
+#include <libdnf5-cli/exception.hpp>
 #include <libdnf5/conf/const.hpp>
 #include <libdnf5/rpm/package.hpp>
 #include <libdnf5/rpm/package_query.hpp>
@@ -94,6 +96,8 @@ void RepoclosureCommand::set_argument_parser() {
 
     newest = std::make_unique<libdnf5::cli::session::BoolOption>(
         *this, "newest", '\0', "Only consider the latest version of a package from each repo.", false);
+
+    create_json_option(*this);
 }
 
 void RepoclosureCommand::configure() {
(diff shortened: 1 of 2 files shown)


dnf5-plugins/repoclosure_plugin/repoclosure.cpp
```diff
@@ -152,12 +156,14 @@ void RepoclosureCommand::run() {
             auto nevra_pair = package_query.resolve_pkg_spec(spec, settings, true);
             if (!nevra_pair.first) {
                 specs_resolved = false;
-                std::cerr << libdnf5::utils::sformat(_("No match for argument \"{}\"."), spec) << std::endl;
+                if (!ctx.get_json_output_requested()) {
```
evan-goode: IMO `repoclosure --json` should behave the same as `repoclosure` when specs can't be resolved, rather than ignore the unmatched spec.

If we did want an option to ignore unmatched specs, we could add it separately.

For now, I would keep the error message on stderr, and keep `if (!specs_resolved) {` below. Maybe it would be better in the long run to also encode error messages in JSON as well, something like

```
{
    error_messages: [
        "No match for argument \"foobar\""
    ],
    packages: []
}
```

But I don't think we do this anywhere else yet.

doc/dnf5_plugins/repoclosure.8.rst
```diff
@@ -72,3 +75,23 @@ Examples
...
+
+The command returns a JSON array, each element describing one package with unresolved dependencies.
+Each package object contains the following fields:
+
+- ``package`` (string) - Package NEVRA.
+- ``repo`` (string) - Repository identifier.
+- ``unresolved_dependencies`` (array) - List of unsatisfied requirements (strings).
+
+For empty results:
+
```
evan-goode, line 96: This confused me at first, maybe this should be on one line:

```
For empty results, ``dnf5 repoclosure --json`` will return ``[]`` (empty array).
```

evan-goode (changes requested):
Thank you! Nice to have.

dnf5-plugins/repoclosure_plugin/repoclosure.cpp
kontura: It would be the only `--json` command exiting non-zero and with a message on stderr with unmatched argument (spec) though. 

I believe this is only a symptom of a general problem, the commands are inconsistent even without `--json`.
It is even already reported: https://github.com/rpm-software-management/dnf5/issues/1075, until we resolve it I guess we have to be inconsistent somewhere.

kontura (approved):

mfocko (approved):

Issue #2306, opened 2025-06-20
dnf5-debuginfo-install manual is not packaged


ppisar:
I cannot find dnf5-debuginfo-install(8) manual page dnf5 RPM package. It exists in sources as doc/commands/debuginfo-install.rst file.

PR #2602, opened 2026-02-07
Add dnf5-debuginfo-install(8) manual page and package provides

Userfrom1995:
Changes:
- Rename debuginfo-install.rst to debuginfo-install.8.rst
- Add debuginfo-install.8 to doc/commands/index.rst and conf.py.in
- Add install rule for debuginfo-install man page in doc/CMakeLists.txt
- Add Provides: dnf5-command(debuginfo-install) to dnf5.spec
- Update dnf5.spec %files to explicitly list install, upgrade man pages to avoid wildcard conflicts

Fixes: #2306


dnf5.spec
```diff
@@ -340,7 +342,10 @@ It supports RPM packages, modulemd modules, and comps groups & environments.
 %{_mandir}/man8/dnf*-group.8.*
 %{_mandir}/man8/dnf*-history.8.*
 %{_mandir}/man8/dnf*-info.8.*
-%{_mandir}/man8/dnf*-install.8.*
+%{_mandir}/man8/dnf5-install.8.*
+%if %{with dnf5_obsoletes_dnf}
```
m-blaha: This condition looks strange. It's here becouse of the wildcard conflict, right?
To be honest I'd rather drop all `%{_mandir}/manX/dnf*-XXX.8.*` lines and replace them with single `%if %{with dnf5_obsoletes_dnf}` block:

```
%{_mandir}/man8/dnf5.8.*
%{_mandir}/man8/dnf5-advisory.8.*
...
%{_mandir}/man8/dnf5-debuginfo-install.8.*
...

%if %{with dnf5_obsoletes_dnf}
# dnf compatibility symlinks
%{_mandir}/man8/dnf.8.*
%{_mandir}/man8/dnf-advisory.8.*
...
%endif
```

I understand that the issue existed even before your fix, but would you be willing to add a comit splitting the manpage files?

Userfrom1995: > This condition looks strange. It's here becouse of the wildcard conflict, right? To be honest I'd rather drop all `%{_mandir}/manX/dnf*-XXX.8.*` lines and replace them with single `%if %{with dnf5_obsoletes_dnf}` block:
> 
> ```
> %{_mandir}/man8/dnf5.8.*
> %{_mandir}/man8/dnf5-advisory.8.*
> ...
> %{_mandir}/man8/dnf5-debuginfo-install.8.*
> ...
> 
> %if %{with dnf5_obsoletes_dnf}
> # dnf compatibility symlinks
> %{_mandir}/man8/dnf.8.*
> %{_mandir}/man8/dnf-advisory.8.*
> ...
> %endif
> ```
> 
> I understand that the issue existed even before your fix, but would you be willing to add a comit splitting the manpage files?

I added the commit to split the man pages as you suggested. I also thought about doing it while working on the install and upgrade changes, but I only did the bare minimum to solve the problem and was a little hesitant to push such a big change in the spec.

Let me know what you think about the changes I just pushed and if there is anything else you would like to suggest or see changed.

Thank you for the review.

PR #2610, opened 2026-02-16
docs/broken man references

mfocko:
TODO:
- [x] What to do about:

  ```
    .. NOTE::
       This option in particular :ref:`can be set in your configuration file by
       your distribution <conf_distribution_specific_options-label>`.
  ```

  I'm unable to find what this references, the closest thing I've seen are distribution-specific overrides in `/usr/share/dnf5/libdnf.conf.d/`. Is that what it's supposed to reference?
  
  The note is twice, next to two options `best` and `skip_if_unavailable`

Fixes SWM-7819

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -96,8 +96,8 @@ repository configuration file should aside from repo ID consists of baseurl, met
     Default: ``True``.
 
     .. NOTE::
-       This option in particular :ref:`can be set in your configuration file by
-       your distribution <conf_distribution_specific_options-label>`.
+       This option in particular can be overriden by a distribution-specific
+       configuration. For more details see :ref:`drop-in configuration directories <drop_in_configuration_directories-label>`.
 
 .. _cachedir_options-label:
 
@@ -1221,8 +1221,8 @@ configuration.
     Default: ``False``.
 
     .. NOTE::
-       this option in particular :ref:`can be set in your configuration file
-       by your distribution <conf_distribution_specific_options-label>`.
+       This option in particular can be overriden by a distribution-specific
+       configuration. For more details see :ref:`drop-in configuration directories <drop_in_configuration_directories-label>`.
 
 .. _sslcacert_options-label:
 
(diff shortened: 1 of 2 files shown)


doc/misc/filtering.7.rst
```diff
@@ -50,7 +50,7 @@ Similar to global excludes, but this configuration is repository-specific and on
 User excludes filtering
 =======================
 
-API users have an additional option how to fine-tune excluded packages using ``*_user_excludes()`` and ``*_user_includes`` methods of the ``PackageSack`` object. See :ref:`PackageSack API <api_packagesack-label>`.
+API users have an additional option how to fine-tune excluded packages using ``*_user_excludes()`` and ``*_user_includes`` methods of the ``PackageSack`` object. See `libdnf5::rpm::PackageSack`_.
```
kontura: The `` `libdnf5::rpm::PackageSack`_`` doesn't seem to work in my testing.
By inspecting what ids are generated I see:

Image

So for some reason there is an id for `libdnf5::rpm::libdnf5::rpm::PackageSack` based on that I tried the link as: 
```
... See :cpp:class:`libdnf5::rpm::libdnf5::rpm::PackageSack`.
```
and looks like that works.

I have spent some more time investigating that and I believe its a `breathe` bug, I filled: https://github.com/breathe-doc/breathe/issues/1054

I am not sure how we want to deal with this now. Perhaps use the correct: ``:cpp:class:`libdnf5::rpm::PackageSack` `` and if it is ever fixed it would start working?

doc/dnf5.conf.5.rst
```diff
@@ -1380,7 +1380,7 @@ files<distro_configuration_files-label>`.
 The configuration files are processed following this order:
 
 1. Configuration files are alphabetically sorted in a list of names from the
-   :ref:`distribution configuration directory<distro_configuration_dir-label>.
+   :ref:`distribution configuration directory<distro_configuration_dir-label>`.
```
kontura: Can you also please remove the superfluous `.` before the `and the ...`?

mfocko:
> Regarding the distribution-specific overrides I believe those notes are left over from dnf4 version (originally added here: https://github.com/rpm-software-management/dnf/pull/1505)
As mentioned in the commit messages they exist due to https://bugzilla.redhat.com/show_bug.cgi?id=1738837: dnf man page says default is X but distro overrides it to Y.
On fedora specifically the overrides are now in /usr/share/dnf5/libdnf.conf.d/20-fedora-defaults.conf.
In my opinion the notes are still useful overall but yea we should reword them a bit and indeed point to the distribution-specific overrides.

I'd probably drop the reference, I opened the PR you linked, and it says basically verbatim the same thing twice.

kontura (approved):

PR #2613, opened 2026-02-17
dnf-automatic: use /bin/sh

aduskett:
/bin/sh is guaranteed by POSIX but /usr/bin/sh is not.


fhbash (approved):
Thanks for your contribution

Conan-Kudo:
Switching the shebangs to `#!/bin/sh` should be fine, as the brp script in Fedora will rewrite them back to `/usr` paths anyway. It'll make things moderately easier for other consumers of DNF5.

PR #2619, opened 2026-02-21
closed

Emayeah:
A couple of days ago I noticed that ^Z (SIGTSTP) doesn't make the cursor re-appear (set \e[?25 to high), I've implemented a handler that sets the cursor to high when SIGTSTP is caught, and then sets it back to low on resume (fg, SIGCONT)


Emayeah:
Whoops, my bad! I've forgot to branch!

PR #2633, opened 2026-02-26
Differentiate between optional and default groups in environments

pkratoch:
Requires: https://github.com/openSUSE/libsolv/pull/607

Related: http://bugzilla.redhat.com/show_bug.cgi?id=2437049

### DIFF
--- a/libdnf5/comps/environment/environment.cpp
+++ b/libdnf5/comps/environment/environment.cpp
@@ -95,6 +95,9 @@ Environment & Environment::operator=(Environment && src) noexcept = default;
 Environment & Environment::operator+=(const Environment & rhs) {
     p_impl->environment_ids.insert(
         p_impl->environment_ids.begin(), rhs.p_impl->environment_ids.begin(), rhs.p_impl->environment_ids.end());
+    p_impl->groups.clear();
+    p_impl->default_groups.clear();
+    p_impl->optional_groups.clear();
     return *this;
 }
 
@@ -355,6 +358,9 @@ void Environment::serialize(const std::string & path) {
 
 void Environment::add_environment_id(const EnvironmentId & environment_id) {
     p_impl->environment_ids.push_back(environment_id);
+    p_impl->groups.clear();
+    p_impl->default_groups.clear();
+    p_impl->optional_groups.clear();
 }
 
 }  // namespace libdnf5::comps
--- a/libdnf5/comps/group/group.cpp
+++ b/libdnf5/comps/group/group.cpp
@@ -93,6 +93,7 @@ Group & Group::operator=(Group && src) noexcept = default;
 
 Group & Group::operator+=(const Group & rhs) {
     p_impl->group_ids.insert(p_impl->group_ids.begin(), rhs.p_impl->group_ids.begin(), rhs.p_impl->group_ids.end());
+    p_impl->packages.clear();
     return *this;
 }
 
@@ -396,6 +397,7 @@ libdnf5::transaction::TransactionItemReason Group::get_reason() const {
 
 void Group::add_group_id(const GroupId & group_id) {
     p_impl->group_ids.push_back(group_id);
+    p_impl->packages.clear();
 }
 
 }  // namespace libdnf5::comps


dnf5.spec
```diff
@@ -132,7 +132,7 @@ Provides:       dnf5-command(versionlock)
 
 %global libmodulemd_version 2.5.0
 %global librepo_version 1.20.0
-%global libsolv_version 0.7.35
+%global libsolv_version 0.7.36
```
ppisar, line 135: I think you also want to update the version in libdnf5/CMakeLists.txt.

include/libdnf5/comps/environment/environment.hpp
```diff
@@ -102,6 +102,10 @@ class LIBDNF_API Environment {
     // @replaces dnf:dnf/comps.py:attribute:Environment.mandatory_groups
     std::vector<std::string> get_groups();
 
+    /// @return std::vector of default Group ids belonging to the Environment.
+    /// @since 5.4.0.1
+    std::vector<std::string> get_default_groups();
```
ppisar, line 107: This adds a new pubic method. That means we need to increase minor version number. from 5.4.0.0 to 5.4.1.0 in VERSION.cmake and dnf5.spec.

Also correct the version in a comment at this method declaration.

libdnf5/comps/environment/environment.cpp
```diff
@@ -171,20 +173,27 @@ int Environment::get_order_int() const {
 
 
 std::vector<std::string> load_groups_from_pool(
-    libdnf5::solv::CompsPool & pool, const std::vector<EnvironmentId> & environment_ids, bool required = true) {
+    libdnf5::solv::CompsPool & pool,
+    const std::vector<EnvironmentId> & environment_ids,
+    GroupType type = GroupType::MANDATORY) {
```
ppisar: Shouldn't the `type` argument be called `types`? Your implementation supports a disjunction of multiple group types (e.g. GroupType::MANDATORY|GroupType::DEFAULT), so the meaning of the argument is not a discrete GroupType.

(If this was not our intention, then I don't understand the hassle with adding all the GroupType operators.)

libdnf5/comps/environment/environment.cpp
```diff
@@ -171,20 +173,27 @@ int Environment::get_order_int() const {
...
+        }
+        if (any(type & GroupType::DEFAULT)) {
+            offsets.push_back(solvable->dep_recommends);
+        }
+        if (any(type & GroupType::OPTIONAL)) {
+            offsets.push_back(solvable->dep_suggests);
         }
 
-        if (offset) {
+        for (const auto offset : offsets) {
```
ppisar, line 199: Original function skipped this block if `offset` (solvable->dep_requires or solvable->dep_suggests) was false. The new implementation does not do the check. Is that intentional? Are all the three new offsets (dep_requires, etc.) guaranteed to be always populated?

libdnf5/comps/environment/environment.cpp
```diff
@@ -310,13 +329,18 @@ void Environment::serialize(const std::string & path) {
...
-        // Create an XML node for this package
+        // Create an XML node for this group
         node = utils::xml::add_subnode_with_text(node_grouplist, "groupid", group);
     }
     xmlNodePtr node_optionlist = xmlNewNode(NULL, BAD_CAST "optionlist");
     xmlAddChild(node_environment, node_optionlist);
+    for (const auto & group : get_default_groups()) {
+        // Create an XML node for this group
+        node = utils::xml::add_subnode_with_text(node_optionlist, "groupid", group);
+        xmlNewProp(node, BAD_CAST "default", BAD_CAST "true");
```
ppisar, line 343: Just a note: xmlNewProp() can return NULL (in case memory allocation fails and thus no attribute has been added) and well-writen code should handle it. Though I can see that no libxml2 call in this function handles errors. This is would be great to fix it in a separate pull request.

libdnf5/comps/environment/environment.cpp
pkratoch: I added the check before adding each offset into the vector. Thanks!

libdnf5/comps/environment/environment.cpp
pkratoch: I'll make a separate PR.

PR #2634, opened 2026-02-27
dnf5: Reuse cached packages for offline transactions

m-blaha:
Before downloading packages for an offline transaction, reuse packages already downloaded to the cache (e.g., by dnf-automatic).

I'm not sure this is the best solution, but it doesn't affect the library in any way and resolves my use case: I have dnf-automatic configured to download available updates. However, sometimes (e.g., when a kernel or desktop environment update is available) I prefer to apply the updates offline. This scenario caused all the upgrades to be redownloaded to the separate offline cache. With this patch, the already downloaded packages are reused.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -39,6 +39,7 @@
 #include <libdnf5/rpm/package_set.hpp>
 #include <libdnf5/rpm/rpm_signature.hpp>
 #include <libdnf5/transaction/offline.hpp>
+#include <libdnf5/transaction/transaction_item_action.hpp>
 #include <libdnf5/utils/bgettext/bgettext-lib.h>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 #include <libdnf5/utils/fs/file.hpp>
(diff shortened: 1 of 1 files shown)


dnf5/context.cpp
```diff
@@ -337,6 +338,33 @@ void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
     state.write();
 }
 
+/// Prepopulate the offline destdir with already-cached packages from the repo
+/// cache to avoid re-downloading them.  Must be called before setting destdir,
+/// so that get_package_path() / is_available_locally() still resolve to the
+/// repo cache.
+static void prepopulate_offline_cache(
+    libdnf5::base::Transaction & transaction, const std::filesystem::path & packages_dir) {
+    std::filesystem::create_directories(packages_dir);
```
kontura: This always creates the `packages` directory even if it ends up empty, for example when the stored transaction only removes packages.
It is not really a problem but the tests currently don't expect it.

Do we want to adapt the tests or create the directory only when needed?
Probably it would be good for both the `comps` and `packages` directories to behave the same.

m-blaha: Oh, I'll take a look!

m-blaha: I deferred the directory creation until there is a package to copy.

kontura: Great, thank you!

PR #2635, opened 2026-02-27
dnf5: Reuse cached packages for offline and store transactions v2

m-blaha:
Download packages to the repo cache first (where already-cached packages are found and skipped), then move them to the final destination directory. This avoids re-downloading packages that are already present in the cache.

This is alternative solution to https://github.com/rpm-software-management/dnf5/pull/2634 that should handle also transaction store/replay use  cases.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -39,6 +39,7 @@
 #include <libdnf5/rpm/package_set.hpp>
 #include <libdnf5/rpm/rpm_signature.hpp>
 #include <libdnf5/transaction/offline.hpp>
+#include <libdnf5/transaction/transaction_item_action.hpp>
 #include <libdnf5/utils/bgettext/bgettext-lib.h>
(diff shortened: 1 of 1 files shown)


kontura:
I like the optimization, simplification and that it works for both `--offline` (`system-upgrade`) and `--store` options.

It seems the tests are failing because it always creates the `packages` directory, even if there are none there.
I have put together some additional tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1837

I do worry that it is always "stealing" packages from the cache, even if user has `keepcache=true` in the configuration. I have included a test for that.
I am also wondering what is the cost of moving all the packages from cache to the `packages` directory across filesystems, particularly for big `system-upgrades`.

Personally I would probably prefer the first approach https://github.com/rpm-software-management/dnf5/pull/2634 if it did a copy instead of a move and was extended to the `--store` option?

Issue #2271, opened 2025-05-29
Throw read-only image-mode system error for install command before refreshing metadata


ppisar:
Not sure if there is a way, but it would be nice if we could do the check before we download many MiB of repo metadata:

```
[core@cosa-devsh ~]$ sudo dnf install htop 
Updating and loading repositories:
 Fedora 42 - x86_64 - Updates Archive                                                                                                                                                               100% |  12.2 MiB/s |  10.4 MiB |  00m01s
 Fedora 42 openh264 (From Cisco) - x86_64                                                                                                                                                           100% |   8.9 KiB/s |   6.0 KiB |  00m01s
 Fedora 42 - x86_64 - Updates                                                                                                                                                                       100% |   3.2 MiB/s |   5.7 MiB |  00m02s
 Fedora 42 - x86_64                                                                                                                                                                                 100% |  15.2 MiB/s |  35.4 MiB |  00m02s
Repositories loaded.
Package                                                                          Arch              Version                                                                          Repository                                          Size
Installing:
 htop                                                                            x86_64            3.4.1-1.fc42                                                                     updates                                        456.4 KiB
Installing dependencies:
 hwloc-libs                                                                      x86_64            2.12.0-1.fc42                                                                    fedora                                           2.9 MiB

Transaction Summary:
 Installing:         2 packages

Total size of inbound packages is 2 MiB. Need to download 2 MiB.
After this operation, 3 MiB extra will be used (install 3 MiB, remove 0 B).
Error: this bootc system is configured to be read-only. For more information, run `bootc --help`.
```

_Originally posted by @dustymabe in https://github.com/rpm-software-management/dnf5/issues/2133#issuecomment-2917447723_

ppisar:
This optimization should affect any command potentially executing an RPM transaction (install, remove, upgrade, swap, distro-sync, ...) and should not affect query commands (repoquery).

Issue #2358, opened 2025-07-21
Add --transient support to DNF5 for bootc systems


mrguitar:
This is issue is to track the work to implement `--transient` support for bootc systems using DNF 5. This functionality has been incredibly useful in the CentOS-stream and RHEL space w/ DNF v4, but is currently lacking on modern Fedora systems.

evan-goode:
Resolved by https://github.com/rpm-software-management/dnf5/pull/2727.

PR #2637, opened 2026-02-28
Improve detection and handling of bootc systems with read-only /usr

evan-goode:
Details in commit messages.

Resolves https://github.com/rpm-software-management/dnf5/issues/2271.

This is largely in preparation for https://github.com/rpm-software-management/dnf5/issues/2358.

### DIFF
--- a/libdnf5/utils/subprocess.cpp
+++ b/libdnf5/utils/subprocess.cpp
@@ -96,8 +96,8 @@ namespace libdnf5::utils::subprocess {
 CompletedProcess run(const std::string & command, const std::vector<std::string> & args) {
     // Struct is used to pass a possible error from a child process before starting a new program.
     struct ErrorMessage {
-        enum { BIND_STDOUT, BIND_STDERR, EXEC } error;  // what failed
-        int err_code;                                   // errno
+        enum { BIND_STDIN, BIND_STDOUT, BIND_STDERR, EXEC } error;  // what failed
+        int err_code;                                               // errno
     };
 
     Pipe pipe_error_msg_from_child;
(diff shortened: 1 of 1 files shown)


libdnf5/utils/subprocess.cpp
```diff
@@ -0,0 +1,267 @@
...
+        // Check for errors or hangup on any descriptor
+        if (fds[STDOUT_FD].revents & (POLLERR | POLLHUP)) {
+            fds[STDOUT_FD].fd = -1;
+        }
+        if (fds[STDERR_FD].revents & (POLLERR | POLLHUP)) {
+            fds[STDERR_FD].fd = -1;
+        }
+    }
+
+    waitpid(child_pid, &child_exit_status, 0);
```
jrohel: In the original version (in my code in "actions.cpp"), `waitpid` was called automatically in the `OsScopeExit finish` destructor. The execution of the destructor at that point was guaranteed by the end of the `else` block. There is no need to write an explicit `else`; simply defining a block with `{ }` is enough.

The `waitpid` call should remain in the destructor to ensure it is called even if the code terminates prematurely due to an exception. We have several places here that throw a `SystemError`. If we don't handle `waitpid` and the caller catches the exception to continue (e.g., is_bootc_system ENOENT), a zombie process will remain at least until the end of the dnf5 run. If the `run` function becomes part of a public API in the future, it could become a source of zombie processes when used in daemon applications.

jrohel:
The `run` function does not redirect the standard input for the executed process. The question is whether we want the spawned process to be able to 'steal' data from our standard input. We likely don't want this unless the caller of the `run` function explicitly allows it via an argument. Alternatively, we could pass a file descriptor with input data to the `run` function (using -1 as an invalid file descriptor when we want to close the input). These are just ideas, but in any case, we probably don't want an unredirected standard input as the default.

libdnf5/utils/subprocess.cpp
evan-goode: Good catch, done.

PR #2640, opened 2026-03-03
needs-restarting: Add JSON output support and documentation

Userfrom1995:
Implement JSON machine-readable output for the needs-restarting command.

Changes:
- Add --json option to needs-restarting command for all three modes (reboot hint, services, processes).
- Extract JSON output into dedicated helper functions (print_reboot_json, print_services_json, print_processes_json).
- Add unit tests for --json option registration.
- Update needs-restarting documentation (man page) with JSON output details and field descriptions.

Fixes: #2341

### DIFF
--- a/dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
+++ b/dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
@@ -656,7 +656,6 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
 
     // Now check which processes need restarting
     std::set<ProcessResult> processes_needing_restart;
-    libdnf5::rpm::PackageSet updated_packages{ctx.get_base()};
 
     for (const auto & [exe_path, process_info] : running_processes) {
         // Recursively get all dependencies of the package that
@@ -669,8 +668,6 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
             if (install_time > process_info.start_time) {
                 processes_needing_restart.insert(
                     {process_info.pid, process_info.cmdline, process_info.package.get_name()});
-                // Track this package as updated so we can check reverse dependencies
-                updated_packages.add(dep);
                 break;
             }
         }


dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -213,6 +217,19 @@ time_t NeedsRestartingCommand::get_boot_time(Context & ctx) {
     return boot_time;
 }
 
+struct ProcessResult {
```
mfocko, line 221: I would probably expand this type with:

```cpp
json_object * ProcessResult::json() const;
```

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -237,6 +254,62 @@ libdnf5::rpm::PackageSet recursive_dependencies(
...
+    json_object_object_add(
+        j_output, "documentation", json_object_new_string("https://access.redhat.com/solutions/27943"));
+
+    json_object_array_add(j_output_array, j_output);
+    std::cout << json_object_to_json_string_ext(j_output_array, JSON_FLAGS) << std::endl;
+    json_object_put(j_output_array);
+}
+
+static void print_services_json(const std::vector<std::string> & service_names) {
+    json_object * j_output = json_object_new_array();
```
mfocko, line 297: I don't really like the repetitiveness of these arrays for each of the available switches, but looking at the code below where it's called from, I don't think we have much way for improvement here :/

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -374,11 +456,20 @@ void NeedsRestartingCommand::services_need_restarting(Context & ctx) {
...
+        }
+    } else {
+        if (!service_names.empty()) {
+            for (const auto & service_name : service_names) {
+                std::cout << service_name << std::endl;
+            }
+            throw libdnf5::cli::SilentCommandExitError(1);
         }
-        throw libdnf5::cli::SilentCommandExitError(1);
     }
```
mfocko:
```suggestion
    } else if (!service_names.empty()) {
        for (const auto & service_name : service_names) {
            std::cout << service_name << std::endl;
        }
        throw libdnf5::cli::SilentCommandExitError(1);
    }
```

mfocko: this diff doesn't make any sense… github pls, how can you screw THIS up? :clawing_my_eyes_out:

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -457,7 +548,7 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
     // Map of executable paths to their process start times and packages
     struct ProcessInfo {
```
mfocko, line 551: Same here, I'd move the JSON conversion into the struct as suggested above.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -556,15 +645,15 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
...
+                running_processes.insert({exe_path_str, ProcessInfo{pid, cmdline_args, process_start_time, package}});
             }
         }
     }
 
     closedir(proc_dir);
 
     // Now check which processes need restarting
-    std::unordered_set<std::string> processes_needing_restart;
+    std::vector<ProcessResult> processes_needing_restart;
```
mfocko: Based on what you're doing below, wouldn't it be easier to use `std::set`? I'm afraid I don't have metrics on the performance downsides of `unordered_set` -> `set`

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -576,21 +665,37 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
...
+                        cmdline_str += " ";
+                    }
+                }
+
+                std::cout << process.pid << "\t" << cmdline_str << std::endl;
+            }
+            throw libdnf5::cli::SilentCommandExitError(1);
         }
-        throw libdnf5::cli::SilentCommandExitError(1);
     }
```
mfocko:
```suggestion
    } else if (!processes_needing_restart.empty()) {
        for (const auto & process : processes_needing_restart) {
            std::string cmdline_str;
            for (size_t i = 0; i < process.cmdline.size(); ++i) {
                cmdline_str += process.cmdline[i];
                if (i < process.cmdline.size() - 1) {
                    cmdline_str += " ";
                }
            }

            std::cout << process.pid << "\t" << cmdline_str << std::endl;
        }
        throw libdnf5::cli::SilentCommandExitError(1);
    }
```

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
Userfrom1995: I don't think we need it here, because the data from `ProcessInfo` flows into `ProcessResult`, which I've already added the JSON conversion to. 
Not all data from `ProcessInfo` goes to `ProcessResult` either so even if I add a `json()` method there, I don't see where it would actually be used. Constructing and passing the JSON format from `ProcessInfo` to `ProcessResult` would be unnecessarily complicated. But if you still want it, let me know.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
Userfrom1995: Changed to `std::set`.
Personally, I still prefer `std::vector` here: we insert everything in one loop, sort once, and then extract in another loop. With std::set we avoid the explicit sort, but iteration over a vector benefits from better cache locality since elements are contiguous in memory, whereas std::set requires pointer chasing through tree nodes. 
That said, the number of entries is unlikely to exceed a few thousand, so the performance difference is negligible either way, mostly a matter of taste.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -576,19 +667,30 @@ void NeedsRestartingCommand::processes_need_restarting(Context & ctx, bool exclu
             // updated since the process started, the process needs restart
             const auto install_time = static_cast<time_t>(dep.get_install_time());
             if (install_time > process_info.start_time) {
-                processes_needing_restart.insert(std::string(process_info.pid) + '\t' + process_info.cmdline);
+                processes_needing_restart.insert(
+                    {process_info.pid, process_info.cmdline, process_info.package.get_name()});
                 // Track this package as updated so we can check reverse dependencies
                 updated_packages.add(dep);
```
Userfrom1995: Not related to this PR, but I noticed `updated_packages` is never used after being declared and populated. From the comment, it looks like it was intended for some reverse dependency check, but that logic was never implemented. Should this be removed, or are there future plans for it?

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
Userfrom1995: Yes, even I can't think of anything better for this one, at least not yet.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
mfocko: I guess it's OK to remove, I'll reach out to the colleague that edited that part to ask about it.

dcantrell: Yes, it's ok to remove this.  It was overly ambitious at the time and then I never got back around to it.

Userfrom1995: > Yes, it's ok to remove this. It was overly ambitious at the time and then I never got back around to it.

Thanks.

Should I add a second commit in this PR to remove it, or will it require a completely separate PR?

mfocko: I'd include it in this one.

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
mfocko: It looks OK 👀 I guess I saw a pattern, so I went for it without making sure it fits the context…

mfocko (approved):
LGTM, thanks a lot for the contribution 🚀

Issue #712, opened 2023-07-12
Use sticky vendors by default


Conan-Kudo:
### Description of request

Over the years, we've had a number of complaints of unintended package switching between providers. While users can use versionlock and excludes to manually constrain things, it has become pretty obvious that the intent is to "lock on" to particular providers of packages when initially installed or explicitly switched.

In Fedora, the following cases exist today:

* `-freeworld` packages from third party sources
* Alternative builds of packages from COPR/OBS
* Personal packages replacing distro packages

While those cases may exist for RHEL/CentOS, there is also the case of CentOS SIG packages replacing distro packages.

### Benefit to Fedora/CentOS/RHEL

By switching this on by default (`allow_vendor_change=False`), there will be less surprises with upgrades, and the solver can make more educated choices based on the expected desires of the user.

### Benefit to (open)SUSE

This is already the default behavior in SUSE distributions, and DNF v4 in openSUSE is configured to have sticky vendors on by default. This would just make it the upstream default for DNF v5 and make it easier in SUSE distributions.

kontura:
I believe this has already been done: https://github.com/rpm-software-management/dnf5/pull/623

We have already had a number of reports why are packages not updating.
I like the suggestion from: https://bugzilla.redhat.com/show_bug.cgi?id=2219624 to also add a warning like:
`Warning: 1 vendor change prevented, consider allow_vendor_change=true`

Conan-Kudo:
At least in DNF 4, sticky vendors can be overridden if you use either `dnf repository-packages` to install or if you specify the full NEVRA to install/upgrade to. If we port over `dnf repository-packages` functionality to DNF 5, that could be used to suggest how to switch vendors for a package.

praiskup:
As this got reverted by https://github.com/rpm-software-management/dnf5/pull/745 and https://github.com/rpm-software-management/dnf5/issues/750 was created, I think we should reopen this one?

pemensik:
I think vendor changes should be highlighted somehow. Either put then on top or bottom of the list of upgraded packages. Make the vendor change bold if possible. I think vendor updates should get higher visibility to ensure this is wanted and intended, before I press Y to start upgrading. Explicitly showing change of vendor repository used for the package would be good.

While I like this idea in general, it has broken too many workflows on Fedora rawhide. It *needs* to communicate better what happens or why it does not.

Conan-Kudo:
[...]

PR #2641, opened 2026-03-03
Suggest allow_vendor_change=true when solver blocks vendor updates

### LINKED ISSUE #750: Add descriptive output for sticky vendor options

fhbash:
- When the solver blocks an update due to a vendor change restriction, print a hint suggesting `--setopt=allow_vendor_change=true` (#750)

A previous attempt to change this default was reverted because solver errors gave no indication that vendor locking was the cause. The new hint in `print_resolve_hints()` addresses that by detecting `RULE_UPDATE` problems and telling the user how to override the restriction.

Closes: #750 SWMBZBUGSM-153 BZ#2219624

CI Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1839

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -1107,6 +1107,7 @@ static void print_resolve_hints(dnf5::Context & context) {
         bool conflict = false;
         bool broken_file_dep = false;
         bool best = false;
+        bool vendor_change = false;
         // walk through all solver problem to detect a conflict, missing file dependency and best
         for (const auto & resolve_log : context.get_transaction()->get_resolve_logs()) {
             if (resolve_log.get_problem() == libdnf5::GoalProblem::SOLVER_ERROR) {
@@ -1129,6 +1130,9 @@ static void print_resolve_hints(dnf5::Context & context) {
                             case libdnf5::ProblemRules::RULE_BEST_2:
                                 best = true;
                                 break;
+                            case libdnf5::ProblemRules::RULE_UPDATE:
+                                vendor_change = true;
+                                break;
                             default:
                                 break;
                         }
@@ -1149,6 +1153,11 @@ static void print_resolve_hints(dnf5::Context & context) {
                 hints.emplace_back(
                     libdnf5::utils::sformat(_("{} to allow removing of installed packages to resolve problems"), arg));
             }
+        }
+
+        if (!conf.get_allow_vendor_change_option().get_value() && vendor_change) {
+            const std::string_view arg{"--setopt=allow_vendor_change=true"};
+            hints.emplace_back(libdnf5::utils::sformat(_("{} to allow changing package vendors"), arg));
         }
 
         if (broken_file_dep) {


doc/dnf5.conf.5.rst
```diff
@@ -60,7 +60,7 @@ repository configuration file should aside from repo ID consists of baseurl, met
     If disabled, DNF5 will stick to the original vendor during RPM upgrades or downgrades.
     This preference, however, can be fine-tuned using :ref:`Vendor change policies <dnf5_vendor_change_policy-label>`.
 
-    Default: ``True``.
+    Default: ``False``.
```
evan-goode: Let's change this back to `True` (until we have an accepted CP) and then this LGTM. We can add the `--{,no-}allow-vendor-change` in a follow-up PR. Personally I would prefer having the argument vs just `setopt` since it'll be frequently used and shouldn't feel "off the beaten path".

evan-goode (changes requested):

doc/dnf5.conf.5.rst
fhbash: Sorry.. this was completely vanish from my radar after the PTO.. 

Fixed.

Conan-Kudo:
The commit message needs fixing now.

dnf5/main.cpp
```diff
@@ -1129,6 +1130,9 @@ static void print_resolve_hints(dnf5::Context & context) {
                             case libdnf5::ProblemRules::RULE_BEST_2:
                                 best = true;
                                 break;
+                            case libdnf5::ProblemRules::RULE_UPDATE:
```
m-blaha, line 1133: I'm a bit worried that `libdnf5::ProblemRules::RULE_UPDATE` might be too broad and there might be causes other than vendor change firing it. However I haven't been able to craft an example demonstrating it.

m-blaha (changes requested):
Please update the commit message - since the PR does not actually change the default, and remove #712 issue reference from "Closes: "
Other than that, I have only one concern regarding possible false positives when showing the hint, but I would not block merge on it.

dnf5/main.cpp
fhbash: Yeah.. I couldn't make an test to demonstrating this as well.

fhbash:
> Please update the commit message - since the PR does not actually change the default, and remove #712 issue reference from "Closes: " Other than that, I have only one concern regarding possible false positives when showing the hint, but I would not block merge on it.

I updated the commit message.

Conan-Kudo:
It doesn't look like that happened? The message indicates changes that the commit hasn't done.

fhbash:
> It doesn't look like that happened? The message indicates changes that the commit hasn't done.

Ow.. I made the change only on commit message, forgot on PR message.. doing now..

Conan-Kudo:
No, you didn't change the commit message either.

m-blaha (approved):
Thanks!

Conan-Kudo (approved):

Conan-Kudo:
I assume the tests need fixing before we can land this?

fhbash:
> I assume the tests need fixing before we can land this?

Yeah, I made this test https://github.com/rpm-software-management/ci-dnf-stack/pull/1839

evan-goode (approved):
Thanks, LGTM!

PR #2642, opened 2026-03-09
Add '--noplugins' tsflag config value

jermy:
Hi - I pushed a change to dnf recently as https://github.com/rpm-software-management/dnf/pull/2299 and @pkratoch suggested the same option should be available in dnf5, so this is basically that. Again, I don't know how much this is useful, but for dnf it allowed me to install some packages in a docker container where there isn't sufficient permission to run the unshare RPM plugin.

Similar to the dnf change, this will allow running `dnf5 --setopt=tsflags=noplugins` to disable all RPM plugins when installing/reinstalling/removing packages.

I thought it would be a fairly trivial change, but unfortunately had to do a bit of digging to find out why the flags weren't being applied, which is due to the `rpmtsAdd*` functions in turn calling `rpmtsSetupTransactionPlugins`. The latter needs to know about this flag before the transaction is set up, otherwise the plugins will be initialised and cannot be unloaded.

Also - and correct me if I'm wrong - since this will be compiled against librpm, then the more recently added flags `RPMTRANS_FLAG_NOPLUGINS` and `RPMTRANS_FLAG_NOCAPS` must be available at compile time and there is no such support for just reporting flags as invalid (indeed, an exception will be thrown and the transaction aborted if an invalid flag is given). I've therefore updated the documentation to remove those references.

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -594,6 +594,7 @@ repository configuration file should aside from repo ID consists of baseurl, met
     ``nocaps``        ``RPMTRANS_FLAG_NOCAPS``
     ``nocrypto``      ``RPMTRANS_FLAG_NOFILEDIGEST``
     ``deploops``      ``RPMTRANS_FLAG_DEPLOOPS``
+    ``noplugins``     ``RPMTRANS_FLAG_NOPLUGINS``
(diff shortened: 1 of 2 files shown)


pkratoch (approved):
LGTM, thanks!

PR #2643, opened 2026-03-09
dnf5daemon-server: Fix daemon crash for invalid locale

jrohel:
Add exception handling for set_thread_locale() to prevent daemon crash when a client sends an invalid locale string. Previously, an unhandled exception would crash the daemon, and this could be triggered by any unauthenticated client.

When an exception occurs during locale setting execution continues with the default locale.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2445771


m-blaha (approved):
Thank you!

PR #2647, opened 2026-03-10
Shell completion: use space for all options

### LINKED ISSUE #2609: CLI autocompletion should not print equals sign `=`
This is an issue of style.

Steps to reproduce:
```
$ dnf repoquery --whatp<TAB><TAB>
...
$ dnf repoquery --whatprovides=
```
I don't know any other CLI tool that autocompletes equals signs instead of spaces. Moreover, `rpm` itself autocompletes spaces as does `dnf4`:
```
$ rpm -q --whatp<TAB><TAB>
...
$ rpm -q --whatprovides<SPACE>
```

Userfrom1995:
Shell completion now uses a space for long options that require an argument instead of '='. 
Space is now used for both long and short options, regardless of whether they require an argument.

Fixes: #2609

### DIFF
--- a/libdnf5-cli/argument_parser.cpp
+++ b/libdnf5-cli/argument_parser.cpp
@@ -925,30 +925,22 @@ void ArgumentParser::Command::print_complete(
                 } else {
                     suggestions.emplace_back(name);
                 }
-                last = name;
-                if (!opt->get_has_value()) {
-                    last += ' ';
-                }
+                last = name + ' ';
             }
             if (!opt->get_long_name().empty()) {
                 std::string name = "--" + opt->get_long_name();
                 if (name.compare(0, strlen(arg), arg) == 0) {
-                    if (opt->get_has_value()) {
-                        name += '=';
-                    }
                     if (add_description) {
                         std::string extended_name = name;
                         if (opt->get_has_value()) {
+                            extended_name += ' ';
                             extended_name += opt->get_arg_value_help().empty() ? "VALUE" : opt->get_arg_value_help();
                         }
                         help.add_line(extended_name, '(' + opt->get_description() + ')', nullptr);
                     } else {
                         suggestions.emplace_back(name);
                     }
-                    last = name;
-                    if (!opt->get_has_value()) {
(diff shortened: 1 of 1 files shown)


Conan-Kudo (dismissed):

libdnf5-cli/argument_parser.cpp
```diff
@@ -925,30 +925,22 @@ void ArgumentParser::Command::print_complete(
...
             if (!opt->get_long_name().empty()) {
                 std::string name = "--" + opt->get_long_name();
                 if (name.compare(0, strlen(arg), arg) == 0) {
-                    if (opt->get_has_value()) {
-                        name += '=';
-                    }
                     if (add_description) {
                         std::string extended_name = name;
                         if (opt->get_has_value()) {
+                            extended_name += '=';
```
kontura: If we no longer complete the `=` perhaps it shouldn't also be shown in the description/help? So that it matches with the form the user will use.

Conan-Kudo: It's still supposed to be valid and supported. This is just changing the style of export for the tab completion.

Userfrom1995: > If we no longer complete the `=` perhaps it shouldn't also be shown in the description/help? So that it matches with the form the user will use.

Yes, I agree. I have changed it to show a space instead of `=`. 

I’ve also updated the man pages. Now that we are pushing the space-based auto-completion, I thought it would be good to update the man pages to reflect the same format.

If you want me to revert the documentation changes, I can do that. 

Please have a look and let me know if there are any other changes you want.

Sorry for the delay.

kontura: It is hard with these stylistic things.

Personally I would drop the man pages changes and keep `=` there.

Userfrom1995: > It is hard with these stylistic things.
> 
> Personally I would drop the man pages changes and keep `=` there.

Done.
Please take a look and let me know if any other change is required.

kontura (approved):

PR #2650, opened 2026-03-12
Fix UB in repo sorting

### LINKED ISSUE #2644: Signing key not found when importing multiple signing keys
### To Reproduce

```
git clone https://github.com/Ultramarine-Linux/builder --revision c597d47f3bc38f1c2b335ca28588cceea41b0d2d --depth 1
cd builder
podman build .
```

### Example Output

```
 Terra rawhide                          100% | 139.0   B/s | 374.0   B |  00m03s
>>> repomd.xml GPG signature verification error: Signing key not found
 Terra rawhide (Extras)                 100% | 139.0   B/s | 374.0   B |  00m03s
>>> repomd.xml GPG signature verification error: Signing key not found
 Ultramarine Linux rawhide              100% | 139.0   B/s | 374.0   B |  00m03s
>>> repomd.xml GPG signature verification error: Signing key not found
 Fedora - Rawhide - Developmental packa 100% |   2.0 MiB/s |  21.9 MiB |  00m11s
Importing OpenPGP key 0x68F437CA:
 UserID     : "Terra Rawhide Distribution <security@fyralabs.com>"
 Fingerprint: 6AB2516D559868D141F8E0600E2C394268F437CA
 From       : file:///etc/pki/rpm-gpg/RPM-GPG-KEY-terrarawhide
The key was successfully imported.
Importing OpenPGP key 0x3CD99338:
 UserID     : "Terra Rawhide Extras Distribution <security@fyralabs.com>"
 Fingerprint: C05A925CFCF7C583482FF8CC022D9E9B3CD99338
 From       : file:///etc/pki/rpm-gpg/RPM-GPG-KEY-terrarawhide-extras
The key was successfully imported.
Importing OpenPGP key 0xE9185619:
 UserID     : "Ultramarine Linux Cheetah <security@fyralabs.com>"
 Fingerprint: 351BB6D4633A21F344F35D69716FE667E9185619
 From       : file:///etc/pki/rpm-gpg/RPM-GPG-KEY-umrawhide
The key was successfully imported.
Failed to download metadata (baseurl: "https://repos.fyralabs.com/umrawhide") for repository "ultramarine": repomd.xml GPG signature verification error: Signing key not found
```

### Context

- The issue seems to involve simultaneously importing 2 or more signing keys from different repos. If you change `dnf.conf` to enable only one of the repos, it works. 2 or more, it starts to fail.
- We cannot reproduce this consistently outside of mock. Above is already the cleanest repro we've found without involving mock, but if you add more dnf5 commands into the Dockerfile, it probably will start working magically, which makes it extremely difficult to bisect the issue.
- This only happens on rawhide.
- Doesn't happen on `dnf4`, so probably an issue with 5.4.0.0, not a regression from rpm.

lleyton:
See title. The compare function needs to return a boolean if a < b. However, .compare returns -int, 0, and +int, which then gets cast into bool, which leads to UB.

Closes #2644.

### DIFF
--- a/doc/dnf5.conf-vendorpolicy.5.rst
+++ b/doc/dnf5.conf-vendorpolicy.5.rst
@@ -60,6 +60,6 @@ by creating a file with the same name in the ``/etc/dnf/vendors.d/`` directory.
 See Also
 ========
 
-* :ref:`DNF5 Vendor Change Policy File Reference - v1.1 <dnf5_vendor_change_policy_v1_1-label>`
-* :ref:`DNF5 Vendor Change Policy File Reference - v1.0 <dnf5_vendor_change_policy_v1_0-label>`
+* :manpage:`dnf5.conf-vendorpolicy-v1_1(5)`, :ref:`DNF5 Vendor Change Policy File Reference - v1.1 <dnf5_vendor_change_policy_v1_1-label>`
+* :manpage:`dnf5.conf-vendorpolicy-v1_0(5)`, :ref:`DNF5 Vendor Change Policy File Reference - v1.0 <dnf5_vendor_change_policy_v1_0-label>`
 * :manpage:`dnf5.conf(5)`, :ref:`DNF5 Configuration Reference <dnf5_conf-label>`


ppisar:
Next time, please write the pull request body into the commit message because git log is closer to the code than a pull request management system.

Also the commit title is wrong. Coerction from int to boolean is well defined.  What was wrong was that std::set comparator returned is-not-equal instead is less-than.

PR #2651, opened 2026-03-13
Parallelize solv cache building during repo loading

aruiz:
Speed up makecache / repo loading by parallelizing several serial steps in update_and_load_repos. Cold-solv makecache with Fedora + updates repos goes from ~4.1s to ~2.7s wall time (1.5x faster), same CPU time. Note that these changes have been created with assistance by Cursor/Claude and this is my first time contributing. Any feedback to make the contribution useful is welcome.

### Changes:

- Add a generic ThreadPool utility (libdnf5/utils/thread_pool.hpp) with a submit()/std::future interface; explicitly link pthreads in CMake
- Pre-build .solv caches on worker threads in isolated libsolv pools — no shared mutable state, no locking needed on the main pool
- Parallelize read_metadata_cache() calls during cache validation (independent local I/O)
- Dispatch pre-built repos to the sack loader asynchronously, so pre-build of repo B overlaps with the loader consuming repo A
- Defer chmod + rename of written .solv files to background tasks (file already loaded via fd dup)
- Populate metadata paths directly from the librepo download result, skipping a redundant load_local() call for freshly downloaded repos
- All parallel paths are best-effort with silent fallback to existing serial code — no new failure modes, no public API changes
- Tests for parallel loading correctness and ThreadPool

### DIFF
--- a/libdnf5/repo/solv_repo.cpp
+++ b/libdnf5/repo/solv_repo.cpp
@@ -59,7 +59,7 @@ static std::array<char, SOLV_USERDATA_SOLV_TOOLVERSION_SIZE> get_padded_solv_too
     return padded_solv_toolversion;
 }
 
-void SolvRepo::userdata_fill(SolvUserdata * userdata) {
+static void userdata_fill(SolvUserdata * userdata, const unsigned char * chksum) {
     if (strlen(solv_toolversion) > SOLV_USERDATA_SOLV_TOOLVERSION_SIZE) {
         libdnf_throw_assertion(
             "Libsolv's solv_toolvesion is: {} long but we expect max of: {}",
@@ -70,7 +70,12 @@ void SolvRepo::userdata_fill(SolvUserdata * userdata) {
     memcpy(userdata->dnf_magic, SOLV_USERDATA_MAGIC.data(), SOLV_USERDATA_MAGIC.size());
     memcpy(userdata->dnf_version, SOLV_USERDATA_DNF_VERSION.data(), SOLV_USERDATA_DNF_VERSION.size());
     memcpy(userdata->libsolv_version, get_padded_solv_toolversion().data(), SOLV_USERDATA_SOLV_TOOLVERSION_SIZE);
-    memcpy(userdata->checksum, checksum, CHKSUM_BYTES);
+    memcpy(userdata->checksum, chksum, CHKSUM_BYTES);
+}
+
+
+void SolvRepo::userdata_fill(SolvUserdata * userdata) {
+    ::libdnf5::repo::userdata_fill(userdata, checksum);
 }
 
 static bool validate_solv_cache(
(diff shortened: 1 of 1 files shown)


libdnf5/repo/solv_repo.cpp
```diff
@@ -821,4 +846,157 @@ void SolvRepo::create_environment_solvable(
...
+    std::filesystem::create_directories(parent_dir);
+
+    auto tmp_file = fs::TempFile(parent_dir, target_path.filename());
+    auto & cache_file = tmp_file.open_as_file("w+");
+
+    SolvUserdata solv_userdata{};
+    memcpy(solv_userdata.dnf_magic, SOLV_USERDATA_MAGIC.data(), SOLV_USERDATA_MAGIC.size());
+    memcpy(solv_userdata.dnf_version, SOLV_USERDATA_DNF_VERSION.data(), SOLV_USERDATA_DNF_VERSION.size());
+    memcpy(solv_userdata.libsolv_version, get_padded_solv_toolversion().data(), SOLV_USERDATA_SOLV_TOOLVERSION_SIZE);
+    memcpy(solv_userdata.checksum, chksum, CHKSUM_BYTES);
```
kontura: This should not be duplicated, I think it would be better to use `userdata_fill`.
In general the whole `write_solv_file` shares a lot with `write_main` de-duplicating them would also be nice.

libdnf5/repo/solv_repo.cpp
```diff
@@ -821,4 +846,157 @@ void SolvRepo::create_environment_solvable(
...
+/// Computes the solv cache file path for a given repo config and optional type.
+static std::filesystem::path compute_solv_file_path(const ConfigRepo & config, const char * type) {
+    std::string filename;
+    if (type != nullptr) {
+        filename = fmt::format("{}-{}.solvx", config.get_id(), type);
+    } else {
+        filename = config.get_id() + ".solv";
+    }
+    return std::filesystem::path(config.get_cachedir()) / CACHE_SOLV_FILES_DIR / filename;
+}
```
kontura: This duplicates the logic from `solv_file_name` and `solv_file_path`.

libdnf5/repo/solv_repo.cpp
```diff
@@ -821,4 +846,157 @@ void SolvRepo::create_environment_solvable(
...
+            return false;
+        }
+        if (memcmp(ud->checksum, chksum, CHKSUM_BYTES) != 0) {
+            return false;
+        }
+        return true;
+    } catch (...) {
+        return false;
+    }
+}
```
kontura, line 909: There already is `can_use_solvfile_cache`.

We should have the logic just once I understand you might need a different interface to it since this is a static function but it should be possible to unify them and only transform the arguments in each call site.

libdnf5/repo/solv_repo.cpp
```diff
@@ -700,12 +718,19 @@ void SolvRepo::write_ext(Id repodata_id, RepodataType type, const std::string &
...
+        try {
+            std::filesystem::permissions(
+                tmp_path,
+                std::filesystem::perms::group_read | std::filesystem::perms::others_read,
+                std::filesystem::perm_options::add);
+            std::filesystem::rename(tmp_path, solvfile_path);
+        } catch (...) {
+        }
+    });
+    deferred_solv_writes.push_back(std::move(finalize));
```
kontura: I wonder if this is really worth it, do I understand correctly it creates a new thread just to set permissions and rename the file?

aruiz: > I wonder if this is really worth it, do I understand correctly it creates a new thread just to set permissions and rename the file?

I avoided this and tested performance, it did not make much of a difference. I can drop it if you want.

libdnf5/repo/solv_repo.cpp
aruiz: Noted, will update this branch with a fix.

libdnf5/repo/solv_repo.cpp
aruiz: I addressed this on the updated branch

libdnf5/repo/solv_repo.cpp
aruiz: Dropped it in the new branch.

libdnf5/repo/solv_repo.cpp
aruiz: Fixed in the new branch.

aruiz:
I have updated the branch with all the changes with additional commits. I can squash changes into a more sensible set of commits (or a single commit) before we merge.

PR #2653, opened 2026-03-16
docs(builddep): add warning about enabling repos

### LINKED ISSUE #2574: Can't handle when repo set enable=0
[docker-ce-stable-source]
name=Docker CE Stable - Sources
baseurl=https://download.docker.com/linux/fedora/$releasever/source/stable
enabled=0
gpgcheck=1
gpgkey=https://download.docker.com/linux/fedora/gpg

 Docker CE Stable - Sources                                                                                   100% | 705.0   B/s |   1.5 KiB |  00m02s
>>> Status code: 404 for https://download.docker.com/linux/fedora/43/source/stable/repodata/repomd.xml (IP: 13.32.121.111) - https://download.docker.c
>>> Status code: 404 for https://download.docker.com/linux/fedora/43/source/stable/repodata/repomd.xml (IP: 13.32.121.111) - https://download.docker.c
>>> Status code: 404 for https://download.docker.com/linux/fedora/43/source/stable/repodata/repomd.xml (IP: 13.32.121.111) - https://download.docker.c
>>> Status code: 404 for https://download.docker.com/linux/fedora/43/source/stable/repodata/repomd.xml (IP: 13.32.121.111) - https://download.docker.c
>>> Usable URL not found                                                                                                                              
Repositories loaded.

mfocko:
Fixes #2574

### DIFF
--- a/doc/dnf5_plugins/builddep.8.rst
+++ b/doc/dnf5_plugins/builddep.8.rst
@@ -39,6 +39,10 @@ Install missing dependencies for building an RPM package.
              than you would expect because dependencies were evaluated according macros
              set on the package build host.
 
+.. warning:: Because build dependency information is only available in source
+             packages, the ``dnf5 builddep`` command ignores `enabled=0` on the
+             source repositories matching :ref:`the naming convention <repo_names-label>`.
+
 
 Options
 =======
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -817,6 +817,8 @@ Repo Options
     Default: empty.
 
 
+.. _repo_names-label:
+
 Source and debuginfo repository names
 =====================================
 


doc/dnf5_plugins/builddep.8.rst
```diff
@@ -39,6 +39,12 @@ Install missing dependencies for building an RPM package.
              than you would expect because dependencies were evaluated according macros
              set on the package build host.
 
+.. warning:: When running ``dnf5 builddep`` the source repositories are enabled
```
evan-goode: What we really want to make clear is that `builddep` (via calling `enable_source_repos`) may enable source repos that are `enabled=0` in repo configuration files. It uses the heuristic documented in `man 5 dnf5.conf` under "SOURCE AND DEBUGINFO REPOSITORY NAMES" to identify source repositories to enable.

doc/dnf5_plugins/builddep.8.rst
```diff
@@ -39,6 +39,12 @@ Install missing dependencies for building an RPM package.
...
+             as the information about build dependencies is available only
+             in the source packages (i.e. src.rpm). Because of this warnings may
+             be issued (e.g. for defined, but non-reachable source repositories)
```
evan-goode: I don't think we need to mention these warnings, just explain that it may try to load source repos that are not usually enabled. In some cases, the warnings will absolutely be relevant to the user (unreachable source RPM repo = no builddep info about SRPMs in that repo)

doc/dnf5_plugins/builddep.8.rst
mfocko: Does it sound better? 🤔 I still feel like running in circles trying to explain it.

dcantrell: We can revise the text over and over again and will never be satisfied.  I think about messages like this as just stating a fact, not describing a reaction to user behavior (even though it is).  The phrasing should tell the user what dnf will do, not in a way that comes across as a negative reaction to the user behavior.

My attempt:

```
. warning:: Because build dependency information is only available in source packages, the ``dnf5 builddep`` command ignores `enabled=0` on source repositories matching :ref:`naming heuristic <repo_names-label>`.
```

dcantrell (changes requested):
I am fine with the warning, but think it needs some rephrasing.  I have offered a suggestion, but feel free to change it however you want.

doc/dnf5_plugins/builddep.8.rst
mfocko: Applied. I'll wait for @evan-goode's review and maybe one more suggestion.

evan-goode: Looks good, thanks! I'd change the link text to `the naming heuristic` or `the naming convention`.

dcantrell (approved):

PR #2654, opened 2026-03-16
`dnf5 config-manager enable` (+ `disable`)

mfocko:
#### TODO

- [x] Test manually
- [x] Tests
  rpm-software-management/ci-dnf-stack#1850
- [x] Docs
- [x] Resolve the licenses…

### DIFF
--- a/doc/dnf5_plugins/config-manager.8.rst
+++ b/doc/dnf5_plugins/config-manager.8.rst
@@ -84,7 +84,7 @@ Subcommands
 
     The ``setopt`` command writes the repositories configuration overrides to a file named ``99-config_manager.repo``
     located in the system repository override directory. Repository settings are written only to this override file.
-    The original repository configuration file is not changed. Reposiory ID may contain globs.
+    The original repository configuration file is not changed. Repository ID may contain globs.
     Override files also support globs. But the ``setopt`` command resolves the ``repoid`` pattern, and overrides are set
     for each matching repository independently. This means repositories added later will not be affected by these overrides.
 


dnf5-plugins/config-manager_plugin/repos.hpp
```diff
@@ -0,0 +1,126 @@
+// Copyright Contributors to the DNF5 project
+// SPDX-License-Identifier: LGPL-2.1-or-later
```
m-blaha, line 2: All other sources in the config-manager plugin use `GPL-2.0-or-later` license. Is LGPL here really intended?

dnf5-plugins/config-manager_plugin/repos.hpp
```diff
@@ -0,0 +1,126 @@
...
+        create_missing_dirs_opt->set_long_name("create-missing-dir");
+        create_missing_dirs_opt->set_description("Allow to create missing directories");
+        create_missing_dirs_opt->set_has_value(false);
+        create_missing_dirs_opt->set_parse_hook_func(
+            [this](cli::ArgumentParser::NamedArg *, const char *, const char *) {
+                create_missing_dirs = true;
+                return true;
+            });
+        cmd.register_named_arg(create_missing_dirs_opt);
+    };
```
m-blaha: Is there a reason to implement the methods directly in the header file? Since the methods are neither templates nor trivial one-liners, it would be more consistent with the rest of the code base to move the implementation to the repos.cpp file.

dnf5-plugins/config-manager_plugin/repos.hpp
mfocko: I remember checking the `CONTRIBUTING.md` and `COPYING.md`, though… now that I'm looking at it, I'm finding discrepancies anyways 👀 

_edit 1º_: it's a mess… `dnf_automatic`:
```
λ rg -INo '// SPDX-License-Identifier: .*' dnf5-plugins/automatic_plugin/*.{c,h}pp | sort | uniq
// SPDX-License-Identifier: GPL-2.0-or-later
// SPDX-License-Identifier: LGPL-2.0-or-later
// SPDX-License-Identifier: LGPL-2.1-or-later
```

dnf5-plugins/config-manager_plugin/repos.hpp
mfocko: I initially tried to hide it away in an anonymous namespace, but that failed in the end… I'll move it out to source.

dnf5-plugins/config-manager_plugin/repos.hpp
mfocko: Or is that because some of those sources are from dnf4, libdnf, or dnf-plugins-core (they all seem to use GPL)?

mfocko: Regardless of `automatic_plugin`:

```
λ rg -INo '// SPDX-License-Identifier: .*' dnf5-plugins/*/*.hpp | sort | uniq -c
  17 // SPDX-License-Identifier: GPL-2.0-or-later
   1 // SPDX-License-Identifier: LGPL-2.0-or-later
  11 // SPDX-License-Identifier: LGPL-2.1-or-later

λ rg -INo '// SPDX-License-Identifier: .*' dnf5-plugins/*/*.cpp | sort | uniq -c
  24 // SPDX-License-Identifier: GPL-2.0-or-later
   3 // SPDX-License-Identifier: LGPL-2.0-or-later
  13 // SPDX-License-Identifier: LGPL-2.1-or-later
```

I've also had a look at the specfile and `dnf5-plugins` have:
```spec
```

which is not correct?

```diff
-License:        LGPL-2.1-or-later
+License:        GPL-2.0-or-later AND LGPL-2.0-or-later AND LGPL-2.1-or-later
```

dnf5-plugins/config-manager_plugin/repos.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto create_missing_dirs_opt = parser.add_new_named_arg("create-missing-dir");
+    create_missing_dirs_opt->set_long_name("create-missing-dir");
+    create_missing_dirs_opt->set_description("Allow to create missing directories");
+    create_missing_dirs_opt->set_has_value(false);
+    create_missing_dirs_opt->set_parse_hook_func([this](cli::ArgumentParser::NamedArg *, const char *, const char *) {
+        create_missing_dirs = true;
+        return true;
+    });
+    cmd.register_named_arg(create_missing_dirs_opt);
```
mfocko, line 39: To be fair I'm not sure if this makes sense to have configurable… Since the `enable` / `disable` always works only with the override directory.

When testing in the container I didn't hit any issues even when not setting it.

dnf5-plugins/config-manager_plugin/repos.hpp
mfocko: Checked the plugins (included in `dnf5-plugins`) one-by-one: there's only Copr plugin that introduces `LGPL-2.1-or-later`, otherwise they're all `GPL-2.0-or-later`. In this case I'm not sure if it really matters, since the Copr pulls in LGPL anyways…

ppisar: It matters. All licenses must be enumerated.

mfocko: > It matters. All licenses must be enumerated.

I meant the choice for the 2 new files doesn't matter that much, cause the spec will have that license anyways.

ppisar: Last time the team discussed licenses the decision was to make fresh new files LGPL. Even files for dnf5 executable. The reason was that in the future we might want to move a piece of code from the tool to a libdnf5 library and then the GPL license would prevent that.

m-blaha: @ppisar Thanks for clarification!

PR #2667, opened 2026-03-30
feat(daemon): add History.list method

### LINKED ISSUE #2538: dnf5daemon-server: provide a history list API
In Cockpit our software updates page has a history card with old transactions. The PackageKit API we use is quite straightforward and returns a timestamp of the transaction and the "action" with related packages. For reference the current implementation looks as following:

Image

So having something like `org.rpm.dnf.v0.History.List` which returns a list of transactions:

```
[{
  "start": <timestamp>,
  "end": <timestamp>,
 "removed": [],
 "upgraded": [],
 "installed": [],
  "downgraded": [],
}]
```

As Cockpit is a webui being able to pass a `limit` of maximum transactions returned would be welcome. (PackageKit only accepts a limit, unlike the recent_changes API which only has a `since`.)

mfocko:
Fixes #2538

### DIFF
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
@@ -83,6 +83,44 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         <arg name="options" type="a{sv}" direction="in" />
         <arg name="changeset" type="a{saa{sv}}" direction="out" />
     </method>
+
+    <!--
+        list:
+        @options: an array of key/value pairs
+        @transactions: array of transactions with their packages
+
+        Get a list of transactions from the history database. Each transaction
+        includes metadata and categorized package changes.
+
+        The packages in each category are represented as dictionaries containing
+        the requested set of attributes. For upgraded and downgraded packages,
+        an "original_evr" attribute is included with the original package EVR.
+
+        Following options are supported:
+
+            - limit: int64
+                Maximum number of transactions to return.
+            - since: int64, unix timestamp
+                Only return transactions completed after this time.
+            - reverse: boolean, default "false"
+                If true, return oldest transactions first. Default is newest first.
+            - contains_pkgs: list of strings
+                Filter to transactions containing packages with these names.
+            - transaction_attrs: list of strings, default ["id", "start", "end", "user_id", "description", "status"]
+                List of transaction attributes to include. Supported: id, start, end,
+                user_id, description, status, releasever, comment.
+            - package_attrs: list of strings, default ["name", "arch", "evr"]
+                List of package attributes to include. Supported: name, epoch,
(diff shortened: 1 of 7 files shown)


dnf5daemon-server/services/history/history.cpp
```diff
@@ -272,3 +287,193 @@ sdbus::MethodReply History::recent_changes(sdbus::MethodCall & call) {
...
+    // Get transactions
+    std::vector<libdnf5::transaction::Transaction> transactions;
+    if (options.contains("since")) {
+        int64_t timestamp = dnfdaemon::key_value_map_get<int64_t>(options, "since");
+        auto all_transactions = history.list_all_transactions();
+        for (auto & trans : all_transactions) {
+            if (trans.get_dt_end() > timestamp) {
+                transactions.emplace_back(std::move(trans));
+            }
+        }
```
kontura: Should this also be updated to use the new `list_transactions_since`?

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
```diff
@@ -81,6 +81,44 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
...
+        list:
+        @options: an array of key/value pairs
+        @transactions: array of transactions with their packages
+
+        Get a list of transactions from the history database. Each transaction
+        includes metadata and categorized package changes.
+
+        The packages in each category are represented as dictionaries containing
+        the requested set of attributes. For upgraded and downgraded packages,
+        a "from_evr" attribute is included with the original package EVR.
```
kontura: Was the `from_evr` renamed to `original_evr`?

I have also noticed the presence of `original_evr` is not controlled by `package_attrs`, is that intended?

kontura (commented):
Sorry for the delay, this looks good.
I have created a few CI tests to cover this functionality: https://github.com/rpm-software-management/ci-dnf-stack/pull/1865

mfocko:
> Sorry for the delay, this looks good.

@kontura no problem, the initial iteration went pretty well, though I've been putting off the deduplication of constructing the JSONs of packages for `history` and `list`… so it's still kinda not finished :)

dnf5daemon-server/services/history/history.cpp
mfocko: I think it should, must've missed that 👀

dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.History.xml
mfocko: > Was the `from_evr` renamed to `original_evr`?

Yes, it's for the consistency with other interfaces.

> I have also noticed the presence `original_evr` is not controlled by `package_attrs`, is that intended?

I'd say so? @m-blaha might know more, I tried to be as close as possible to already implemented behavior

kontura (approved):
Thank you!

PR #2669, opened 2026-03-30
CI: Build and test on ELN

ppisar:
Because ELN is configured as RHEL, i.e. differently from Fedora.

I left out building en explicit non-modular configuration on ELN because ELN will disable modularity implicitly in near future. There is no need to build the same thing twice.

Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1898
Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1901

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -64,6 +64,7 @@ jobs:
     trigger: pull_request
     targets:
       - fedora-all
+      - fedora-eln
     packages: [dnf5]
     additional_repos:
       - "copr://rpmsoftwaremanagement/dnf-nightly"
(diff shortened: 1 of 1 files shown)


kontura (commented):
It seems now some of the tests fail due to missing `useradd` should we add `shadow-utils` as an explicit require to `ci-dnf-stack` since we use `useradd`? 

I am guessing that on normal fedora it is pulled in by something automatically.

ppisar:
Indeed. I will review run_in_context() calls and add the dependencies.

ppisar:
This <https://github.com/rpm-software-management/ci-dnf-stack/pull/1901> PR should fix the dependencies.

kontura (approved):
Everything passed, thanks.

Issue #1610, opened 2024-08-05
Prepare mechanism to regenerate the system state *.toml files


m-blaha:
We should have a supported mechanism how to recover from incorrect system state *.toml files. Something like `dnf5 system-state regenerate` command that would regenerate as much of system state as possible using rpmdb and dnf5 transaction history database.

See https://github.com/rpm-software-management/dnf5/issues/1550 for more context.

ppisar:
Regarding the name I would propose "dnf5 system-state rebuild" to be similar to "rpmbuild --rebuilddb".

evan-goode:
Do we need this now that the TOML files are written more atomically (https://github.com/rpm-software-management/dnf5/pull/2289)?

kontura:
Personally I still think this would be nice to have.

The TOML writing can still end in a non ideal state, I have even added a TODO that references this issue: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/system/state.cpp#L676-L678

PR #2675, opened 2026-04-07
state: Don't delete .new files in State::load()

### LINKED ISSUE #2601: Intermittent 'cannot copy' errors for /usr/lib/sysimage/libdnf5/packages.toml/new in dnf5-5.3.0.0-7.fc44
In recent F44 / Rawhide I'm periodically seeing openQA tests fail like this:

Image

There doesn't seem to be any particular trigger, it just...sometimes happens.

m-blaha:
When multiple libdnf5 processes run concurrently (e.g. dnf5 transaction + PackageKit/GNOME Software loading the system repo), the .new file recovery code in State::load() can race with State::save() in another process. The recovery code would delete or rename .new files that the concurrent save() just wrote, causing save()'s subsequent rename() to fail with "cannot copy/rename" errors.

Fix by making load() purely read-only with respect to .new files. In case there are any .new files present, just log a warning and keep using the non-.new state files from the last successful save().

An alternative approach using Locker to synchronize State::load() and State::save() was considered. This would preserve crash recovery for the case where save() was interrupted during the rename phase (all .new files fully written). However, a crash during the write phase still leaves state inconsistent with rpmdb, requiring a full state rebuild (see #1610). Also, Locker requires write access to create the lock file in the state directory, which would break non-root read-only operations (e.g. dnf5 repoquery) unless fallback logic was added. The added complexity was not justified given these limitations.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2601

### DIFF
--- a/libdnf5/system/state.cpp
+++ b/libdnf5/system/state.cpp
@@ -666,29 +666,20 @@ void State::load() {
     try {
         auto new_suffix_files = gather_suffix_new_files(this->path);
         if (!new_suffix_files.empty()) {
+            // The .new files are either stale leftovers from a previously
+            // crashed save(), or they belong to a concurrent save() in another
+            // process. We cannot safely delete or rename them - doing so would
+            // race with a concurrent save() and cause errors (
+            // https://github.com/rpm-software-management/dnf5/issues/2601).
+            // Load from the non-.new files which are always in a consistent
+            // state from the last successful save(). Stale .new
+            // files will be overwritten by the next save().
+            // TODO(amatej): Once https://github.com/rpm-software-management/dnf5/issues/1610
+            //               is done we should suggest to rebuild the system state.
             auto logger = base->get_logger();
-            logger->warning("System state: unfinished update found");
-
-            try {
-                // package_state new file is renamed first, if it exists the .new system state
-                // files are likely incomplete and we cannot use them
-                if (std::filesystem::exists(suffix_new(get_package_state_path()))) {
-                    // TODO(amatej): Once https://github.com/rpm-software-management/dnf5/issues/1610 is done we should
-                    //               suggest to rebuild the system state, some information is likely missing because
-                    //               system state update happens after the transaction is finished.
-                    logger->error("System state: cannot use partially written system state files");
-                    for (const auto & f : new_suffix_files) {
-                        std::filesystem::remove(f);
-                    }
-                } else {
-                    logger->warning(
-                        "System state: update interruption happened during renaming which means all the new system "
-                        "state files were written, using them.");
-                    rename_new_system_state_files(true);
-                }
-            } catch (const std::filesystem::filesystem_error & ex) {
-                logger->error("System state: cannot recover: ", ex.what());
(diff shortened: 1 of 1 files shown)


evan-goode (approved):
Thanks for the investigation and welcome back from the Packit exchange :)

The fix looks correct to me.

evan-goode:
> An alternative approach using Locker to synchronize State::load() and State::save() was considered. This would preserve crash recovery for the case where save() was interrupted during the rename phase (all .new files fully written). However, a crash during the write phase still leaves state inconsistent with rpmdb, requiring a full state rebuild (see https://github.com/rpm-software-management/dnf5/issues/1610).

IMO ideally we would have locking in addition to this change.

> Also, Locker requires write access to create the lock file in the state directory, which would break non-root read-only operations (e.g. dnf5 repoquery) unless fallback logic was added. The added complexity was not justified given these limitations.

For the "system repo" lock (https://github.com/rpm-software-management/dnf5/pull/2519), we worked around that by having the lock file (`/usr/lib/sysimage/libdnf5/system-repo.lock`) be persistent and owned by root with 0664. Then unprivileged users can obtain read locks but not write locks.

Maybe standard practice should be to use the system repo lock for these state files too? The system repo lock is obtained in `Context::load_repos`; it's not automatically obtained for libdnf5 API users. The libdnf5 tutorial [was updated to recommend obtaining the lock](https://dnf5.readthedocs.io/en/latest/tutorial/repos.html#load-repositories-from-system-configuration-files). I guess dnf5daemon consumers and the new PackageKit backend are not using it (yet).

I would hesitate to somehow enforce obtaining the system repo lock in libdnf5, since there are use cases where it's better to read a soon-to-be-invalid state than to wait for a long DNF5 process to finish and release a write lock. But maybe it could be opt-out instead of opt-in.

kontura:
I also think we need the locking and I agree we should treat the system state the same as the system repo. While it is fine if a read only operation reads stale system state it can just as well be a different write operation in the associated reports in https://github.com/rpm-software-management/dnf5/issues/2601.
We use the system state to resolve a goal correctly (reasons / comps info), in my opinion we cannot allow using outdated system state for transactions (just like with system repo).


> I would hesitate to somehow enforce obtaining the system repo lock in libdnf5, since there are use cases where it's better to read a soon-to-be-invalid state than to wait for a long DNF5 process to finish and release a write lock. But maybe it could be opt-out instead of opt-in.

Personally I would enforce it for write operations. Could we somehow `assert` that `system_repo_lock` is `write` locked when running a transaction? I believe it might help libdnf5 API clients to use it correctly.

m-blaha:
The system repo write lock assertion for transactions seems reasonable to me. This would ensure a consistent and current system state without requiring an additional system state lock. For read-only operations, the lock is not strictly required, which might result in slightly stale or inconsistent system state (there is a race condition between system state load() and save() operations), but I believe that's acceptable.
Let me file a new issue to track this. I guess some enhancement of the Locker API would also be needed so it's out of scope for this fix.

PR #2683, opened 2026-04-11
refactor(fs): simplify and correct logic of `File::read`

vil02:
This PR consist of two parts:
- it replaces the _pair_ `length_to_end` and `length_detected` with an optional variable. This simplifies the code - it is easier to _prove_ that the `length_to_end` is initialized,
- the standard says that [`std::fseak`](https://en.cppreference.com/w/cpp/io/c/fseek.html) returns 0​ upon success, nonzero value otherwise.


m-blaha (approved):
Thank you!

PR #2684, opened 2026-04-12
style(dnf5daemon-server): add missing `override`

vil02:
cf. [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final).

One could consider marking these functions as `final`.


jrohel:
Thanks for the PR.

Regarding the `final` keyword: it's more about considering whether we ever plan to inherit from these classes. If not, we can mark the entire classes as `final` instead of just specific methods.

jrohel (approved):

PR #2687, opened 2026-04-13
spec: Require rpm-libs >= 5.99.90 because of rpmteSetVfyLevel()

ppisar:
Commit 4db642138c76474553213b0f9005a78f32fbf932 (Honor per-repo pgp_gpgcheck=0 even in rpm's enforcing signature mode) started to call rpmteSetVfyLevel() if it is available in RPM library on build time. The function emerged in RPM 5.99.90.

Users who update libdnf5 built against RPM >= 5.99.90 before updating RPM library on their system first will get a linking error on a subsequent execution of DNF5:

    # dnf -y update --nogpgcheck --refresh --no-best --allowerasing
    dnf: symbol lookup error: /lib64/libdnf5.so.2: undefined symbol: rpmteSetVfyLevel

This can happen e.g. when partially upgrading (libdnf5 without rpm-libs) from Fedora 43 to 44.

This patch prevents from upgrading DNF5 before RPM.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -441,6 +441,9 @@ License:        LGPL-2.1-or-later
 #Requires:       libmodulemd{?_isa} >= {libmodulemd_version}
 Requires:       libsolv%{?_isa} >= %{libsolv_version}
 Requires:       librepo%{?_isa} >= %{librepo_version}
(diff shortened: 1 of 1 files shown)


m-blaha (approved):
Thank you!

PR #2688, opened 2026-04-13
Introduce gpgcheck_policy to control gpgcheck

### LINKED ISSUE #727: Introduce `gpgcheck_policy` to control `gpgcheck` defaults and add support for `pkg_gpgcheck`

fhbash:
Introducing `gpgcheck_policy` option controlling what gpgcheck expands into: legacy (pkg only), full (pkg+repo, default), all (pkg+repo+localpkg).

CI test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1859

Close: https://github.com/rpm-software-management/dnf5/issues/727

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -243,6 +243,31 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     Default: ``[]``.
 
+.. _gpgcheck_policy_options-label:
+
+``gpgcheck_policy``
+    ``legacy``, ``full``, or ``all``
+
+    Controls how the ``gpgcheck`` option expands when set in a configuration section.
+
+    ``legacy``
+        ``gpgcheck=1`` sets only ``pkg_gpgcheck=1``.
+        This preserves the original DNF behavior.
+
+    ``full``
+        ``gpgcheck=1`` sets both ``pkg_gpgcheck=1`` and ``repo_gpgcheck=1``.
+
+    ``all``
+        ``gpgcheck=1`` sets ``pkg_gpgcheck=1``, ``repo_gpgcheck=1``, and ``localpkg_gpgcheck=1``.
+
+    Explicitly setting ``pkg_gpgcheck``, ``repo_gpgcheck``, or ``localpkg_gpgcheck`` in the
+    same configuration section overrides the policy for that specific option.
+
+    The policy is a global setting defined in ``[main]`` and affects all repository
+    sections that use ``gpgcheck``.
+
+    Default: ``full``.
+
 .. _group_package_types_options-label:
 
 ``group_package_types``
@@ -1004,7 +1029,9 @@ configuration.
     Doesn't apply for packages passed directly as arguments, as they are not in any repository,
     see :ref:`localpkg_gpgcheck <localpkg_gpgcheck_options-label>`.
 
-    Due to compatibility `gpgcheck` option is supported as well but `pkg_gpgcheck` is preferred.
+    Due to compatibility ``gpgcheck`` option is supported as well but ``pkg_gpgcheck`` is preferred.
+    The behavior of ``gpgcheck`` is controlled by the
+    :ref:`gpgcheck_policy <gpgcheck_policy_options-label>` option.
 
 .. _includepkgs_options-label:
 
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -243,6 +243,42 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     Default: ``[]``.
 
+.. _gpgcheck_options-label:
+
+``gpgcheck``
+
+    Due to compatibility ``gpgcheck`` option is supported as well but ``pkg_gpgcheck`` is preferred.
+    The meaning of ``gpgcheck`` is controlled by the
+    :ref:`gpgcheck_policy <gpgcheck_policy_options-label>` option.
+
+    Default: ``pkg_gpgcheck`` value.
+.. :
+
+.. _gpgcheck_policy_options-label:
+
+``gpgcheck_policy``
+    ``legacy``, ``full``, or ``all``
+
+    Controls how the ``gpgcheck`` option expands when explicitly set in a configuration section.
+
+    ``legacy``
+        ``gpgcheck=1`` sets only ``pkg_gpgcheck=1``.
+        This preserves the original DNF behavior.
+
+    ``full``
+        ``gpgcheck=1`` sets both ``pkg_gpgcheck=1`` and ``repo_gpgcheck=1``.
+
+    ``all``
+        ``gpgcheck=1`` sets ``pkg_gpgcheck=1``, ``repo_gpgcheck=1``, and ``localpkg_gpgcheck=1``.
+
+    Explicitly setting ``pkg_gpgcheck``, ``repo_gpgcheck``, or ``localpkg_gpgcheck`` in the
+    same configuration section overrides the policy for that specific option.
+
+    The policy is a global setting defined in ``[main]`` and affects all repository
+    sections that use ``gpgcheck``.
+
+    Default: ``legacy``.
+
 .. _group_package_types_options-label:
 
 ``group_package_types``
@@ -1003,8 +1039,6 @@ configuration.
 
     Doesn't apply for packages passed directly as arguments, as they are not in any repository,
     see :ref:`localpkg_gpgcheck <localpkg_gpgcheck_options-label>`.
-
-    Due to compatibility `gpgcheck` option is supported as well but `pkg_gpgcheck` is preferred.
 
 .. _includepkgs_options-label:
 
--- a/libdnf5/repo/config_repo.cpp
+++ b/libdnf5/repo/config_repo.cpp
@@ -59,6 +59,7 @@ class ConfigRepo::Impl {
     OptionChild<OptionString> username{main_config.get_username_option()};
     OptionChild<OptionString> password{main_config.get_password_option()};
     OptionChild<OptionStringAppendList> protected_packages{main_config.get_protected_packages_option()};
+    OptionBool gpgcheck{false};
     OptionChild<OptionBool> pkg_gpgcheck{main_config.get_pkg_gpgcheck_option()};
     OptionChild<OptionBool> repo_gpgcheck{main_config.get_repo_gpgcheck_option()};
     OptionChild<OptionBool> enablegroups{main_config.get_enablegroups_option()};
@@ -152,9 +153,8 @@ ConfigRepo::Impl::Impl(Config & owner, ConfigMain & main_config, const std::stri
     owner.opt_binds().add("username", username);
     owner.opt_binds().add("password", password);
     owner.opt_binds().add("protected_packages", protected_packages);
+    owner.opt_binds().add("gpgcheck", gpgcheck);
     owner.opt_binds().add("pkg_gpgcheck", pkg_gpgcheck);
-    // Compatibility alias for pkg_gpgcheck
-    owner.opt_binds().add("gpgcheck", pkg_gpgcheck);
     owner.opt_binds().add("repo_gpgcheck", repo_gpgcheck);
     owner.opt_binds().add("enablegroups", enablegroups);
     owner.opt_binds().add("retries", retries);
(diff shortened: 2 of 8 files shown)


evan-goode:
cc @Conan-Kudo

Conan-Kudo:
Why is everything failing?

evan-goode:
Transient infra failure? Let's see

doc/dnf5.conf.5.rst
```diff
@@ -243,6 +243,31 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     Default: ``[]``.
 
+.. _gpgcheck_policy_options-label:
+
+``gpgcheck_policy``
+    ``legacy``, ``full``, or ``all``
+
+    Controls how the ``gpgcheck`` option expands when set in a configuration section.
```
ppisar, line 251: There is no documentation entry for `gpgcheck` option (because it's deprecated). Only a notice at `pkg_gpgcheck`:


>    Due to compatibility ``gpgcheck`` option is supported as well but ``pkg_gpgcheck`` is preferred.
    The behavior of ``gpgcheck`` is controlled by the `gpgcheck_policy` option.

Reading the documentation made me wonder whether `gpgcheck` and `pkg_gpgcheck` are aliases. Reading the source code revealed they are independent now. To make the documentation clear, I'd like to see the documentation worded differently.

Probably the "Controls how the ``gpgcheck`` option expands" sentence should be "Controls how the deprecated ``gpgcheck`` option expands". Maybe the whole `gpgcheck_policy` option should be marked as deprecated since it depends on a deprecated option.

Then maybe an entry for `gpgcheck` could be added and the notice "Due to compatibility..." moved there.

libdnf5/repo/config_repo.cpp
```diff
@@ -616,6 +616,33 @@ void ConfigRepo::load_from_parser(
     Logger & logger,
     Option::Priority priority) {
     Config::load_from_parser(parser, section, vars, logger, priority);
+
+    // Apply gpgcheck_policy: if gpgcheck was set in this repo section, expand it
+    if (p_impl->gpgcheck.get_priority() >= priority) {
```
ppisar, line 621: Ideally code in this block should be deduplicacated with code in ConfigMain::Impl::apply_gpgcheck_policy(). I know the classes do not inherit from each other, so it's difficult to write one function which work both with ConfigRepo->p_impl and ConfigMain::Impl object. But if it would be possible and concise, it would be great.

ppisar (commented):
Otherwise, it looks good. I will have yet to test this implementation.

Because upstream default policy will break Fedora, we need to override the policy to `legacy` in fedora-release-common package before DNF5 with this change reaches Fedora.

Conan-Kudo:
> Otherwise, it looks good. I will have yet to test this implementation.
> 
> Because upstream default policy will break Fedora, we need to override the policy to `legacy` in fedora-release-common package before DNF5 with this change reaches Fedora.

Do we? We could also update fedora-repos so that `repo_gpgcheck=0` is set in the repo files.

ppisar:
> > Otherwise, it looks good. I will have yet to test this implementation.
> > Because upstream default policy will break Fedora, we need to override the policy to `legacy` in fedora-release-common package before DNF5 with this change reaches Fedora.
> 
> Do we? We could also update fedora-repos so that `repo_gpgcheck=0` is set in the repo files.

That's would be only a partial solution because it would leave non-Fedora repositories affected. Either Fedora needs to stick to legacy, or FESCo to approve the change.

Conan-Kudo:
Oh, hah all the fedora repos already set `repo_gpgcheck=0` because I added it back when I attempted to make Zypper work on Fedora.

Conan-Kudo:
> 
> That's would be only a partial solution because it would leave non-Fedora repositories affected. Either Fedora needs to stick to legacy, or FESCo to approve the change.

Yeah, keeping it set to legacy in Fedora packaging makes sense.

ppisar:
I merged the https://github.com/rpm-software-management/ci-dnf-stack/pull/1862. Could you please address improve the documentation as I suggested in-line and rebase this pull request to current main HEAD?

libdnf5/repo/config_repo.cpp
fhbash: I made a change trying to deduplicating the code on my last commit

doc/dnf5.conf.5.rst
```diff
@@ -243,6 +243,41 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     Default: ``[]``.
 
+.. _gpgcheck_options-label:
+
+``gpgcheck``
+
+    Due to compatibility ``gpgcheck`` option is supported as well but ``pkg_gpgcheck`` is preferred.
+    The behavior of ``gpgcheck`` is controlled by the
```
ppisar: I would replace "behavior" with "meaning" in this sentence.

libdnf5/repo/config_repo.cpp
ppisar: The ci-dnf-stack tests still fail. They shouldn't. I recommend rebasing this pull request to the current main branch HEAD. It could resolve some of the failures.

ppisar:
Could you squash the documentation commit "Applying requested changes on documentation" into previous "Introduce gpgcheck_policy to control gpgcheck" commit?

libdnf5/repo/config_repo.cpp
fhbash: @ppisar  Done, rebased on top of main.

ppisar:
The current ci-dnf-stack tests, like dnf/gpg.feature:137, still fail.

fhbash:
Hi @ppisar  with my last commit and the code rebase, I got all tests passed using the ci-test I created: https://github.com/rpm-software-management/ci-dnf-stack/pull/1859

```1 feature passed, 0 failed, 0 skipped
22 scenarios passed, 0 failed, 3 skipped
182 steps passed, 0 failed, 36 skipped, 0 undefined
Took 0m2.677s
```

ppisar:
That <https://github.com/rpm-software-management/ci-dnf-stack/pull/1859> result shows failures.

But we can disregard that ci-dnf-stack pull request now, as that only adds new tests to exhibit the new option and I don't understand why it globally sets "gpgcheck_policy" to "legacy" (It shouldn't be needed if we now set repo_gpgcheck=0 everywhere.)

Most importantly current tests should pass. And they don't. I just rerun locally dnf/gpg.feature and it still fails at dnf/gpg.feature:137 and dnf/gpg.feature:138. The same failure which is listed disply in results in this pull request.

Also thinking more about this change, we should bump major DNF5 version because it breaks compatibility (gpgcheck=1 behaves differently).

Conan-Kudo:
It doesn't break ABI though, we can at least bump the feature version.

ppisar:
It does not break ABI. It breaks API. See few comments above at <https://github.com/rpm-software-management/dnf5/pull/2688#issuecomment-4267265996>.

Conan-Kudo:
Yeah, given that we can ship an override in the dnf5 packaging in Fedora, I don't think that's a big enough deal to worry about.

ppisar:
https://github.com/rpm-software-management/ci-dnf-stack/pull/1859 tests pass now.

Tests without that ci-dnf-stack pull request still don't pass. I guess some tests do not utilize the repo_gpgcheck. If you don't want to make fix this issue I'm fine with it. Though it shows some deficiency in the tests which can bite us later.

Nevertheless, this code change should bump major version and properly highlight the compatibility breakage in the commit message.

fhbash:
We had an team meeting to discuss this change, and we decided to move forward with `legacy` as default value for now, so we do not break anything after this release.  I'll create a new issue to change the default value to `full` for the next major version change.

ppisar:
Now that we don't change the behavior, the tests should really past. They don't.

doc/dnf5.conf.5.rst
```diff
@@ -243,6 +243,41 @@ repository configuration file should aside from repo ID consists of baseurl, met
...
     Default: ``[]``.
 
+.. _gpgcheck_options-label:
+
+``gpgcheck``
+
+    Due to compatibility ``gpgcheck`` option is supported as well but ``pkg_gpgcheck`` is preferred.
+    The meaning of ``gpgcheck`` is controlled by the
+    :ref:`gpgcheck_policy <gpgcheck_policy_options-label>` option.
+    :ref:`pkg_gpgcheck <_gpgcheck_options-label>` option.
```
ppisar: This line `:ref:`pkg_gpgcheck <_gpgcheck_options-label>` option.` looks like an omission. I think it shouldn't be here.

doc/dnf5.conf.5.rst
```diff
@@ -243,6 +243,41 @@ repository configuration file should aside from repo ID consists of baseurl, met
...
+.. :
+
+.. _gpgcheck_policy_options-label:
+
+``gpgcheck_policy``
+    ``legacy``, ``full``, or ``all``
+
+    Controls how the deprecated``gpgcheck`` option expands when set in a configuration section.
```
ppisar: Missing a space between _depreacated_ and _\`\`gpgcheck\`\`_.

ppisar (commented):
Besides the nitpicks in the documentation, I have a problem with your current implementation:

gpgcheck and pkg_gpgcheck works identically, as an alias. That's a problem with this configuration:

~~~
[main]
gpgcheck_policy=all

[test]
pkg_gpgcheck=1
~~~

Your current code sets repo_gpgcheck to 1, while I would expect it to leave it to the default 0. I think the previous implementation was better, because it stopped aliasing gpgcheck to pkg_gpgcheck.

doc/dnf5.conf.5.rst
fhbash: Yeah.. I missed that... thanks!

doc/dnf5.conf.5.rst
```diff
@@ -243,6 +243,40 @@ repository configuration file should aside from repo ID consists of baseurl, met
...
+    The meaning of ``gpgcheck`` is controlled by the
+    :ref:`gpgcheck_policy <gpgcheck_policy_options-label>` option.
+.. :
+
+.. _gpgcheck_policy_options-label:
+
+``gpgcheck_policy``
+    ``legacy``, ``full``, or ``all``
+
+    Controls how the deprecated ``gpgcheck`` option expands when set in a configuration section.
```
Conan-Kudo: It is effectively undeprecated with this PR.

Conan-Kudo (changes requested):

ppisar:
The documentation is better, but what about the aliasing of gpgcheck and pkg_gpgcheck I pointed before? Is it really intended? I think the aliasing completely destroys the meaning of pkg_gpgcheck as an option which SHOULD NOT control repo_gpgcheck and localpkg_gpgcheck.

ppisar (dismissed):
Now it behaves well. Thanks.

Regarding documentation, default `gpgcheck` value is not documented. Current behavior, probably because of compatibility, is that it defaults to `pkg_gpgcheck` value. Maybe that could be mentioned in the documentation ("Default: `pkg_gpgcheck` value."). Also maybe `gpgcheck_policy` documentation should emphasize that the expansion does not apply to the default, implicit `gpgcheck` setting. E.g. I would add _explicitly_ adverb to: "Controls how the `gpgcheck` option expands when explicitly set in a configuration section."

Otherwise, this patch is good.

Current ci-dnf-stack tests pass.

I rerun the related ci-dnf-stack PR and it passes except that on Rawhide it wanted to install the nonmodular builds leading to an installation conflict. I think that's fine.

I approve this pull request. If you want ammend the documention and I will rereview it again. Otherwise, feel free to merge this code.

Conan-Kudo (approved):

ppisar (approved):
Good.

PR #2691, opened 2026-04-14
local plugin: Use separate repo for packages from nogpgcheck repos

### LINKED ISSUE #2580: local plugin does not work with unsigned packages
If local libdnf5 plugin is installed and then an unsisnged package is installed from a repository with disabled pkg_gpgcheck, subsequent installation of that package retrieves the unsisnged to the _dnf_local repository which has hard-coded pkg_gpgcheck=true and that causes an error:

~~~~
# dnf reinstall perl-Sub-HandlesVia
 Copr repo for rpminspect                                      100% |   3.2 KiB/s |   1.8 KiB |  00m01s
 Copr repo for dnfnightly owned by rpmsoftwaremanagement       100% |   3.2 KiB/s |   1.8 KiB |  00m01s
 Local libdnf5 plugin repo                                     100% |   2.6 KiB/s |   1.5 KiB |  00m01s
Repositories loaded.
Package                            Arch     Version                          Repository            Size
Reinstalling:
 perl-Sub-HandlesVia               noarch   0:0.052000-1.fc44                _dnf_local       495.5 KiB
   replacing perl-Sub-HandlesVia   noarch   0:0.052000-1.fc44                f44-build        495.5 KiB

Transaction Summary:
 Reinstalling:       1 package
 Replacing:          1 package

Total size of inbound packages is 145 KiB. Need to download 0 B.
After this operation, 0 B extra will be used (install 495 KiB, remove 495 KiB).
Is this ok [y/N]: y
[1/1] perl-Sub-HandlesVia-0:0.052000-1.fc44.noarch             100% |   0.0   B/s |   0.0   B |  00m00s
>>> Already downloaded
-------------------------------------------------------------------------------------------------------
[1/1] Total                                                    100% |   0.0   B/s |   0.0   B |  00m00s
Running transaction
Transaction failed: Signature verification failed.
OpenPGP check for package "perl-Sub-HandlesVia-0.052000-1.fc44.noarch" (/var/cache/libdnf5/_dnf_local-71c913707df56d1b/packages/perl-Sub-HandlesVia-0.052000-1.fc44.noarch.rpm) from repo "_dnf_local" has failed: The package is not signed.
~~~~

This is documented behavior:

> Note that the repository has ``pkg_gpgcheck`` verification enabled by default but doesn't specify any ``gpgkey``, it assumes all required keys were already imported.

But I'd like to make pkg_gpgcheck configurable so that this plugin can be reconfigured to be helpful also for people who installs unsigned packages.

Note that the current behavior breaks Fedora CI <https://pagure.io/fedora-ci/general/issue/538>.

### LINKED ISSUE #2581: First invocation after installing a local plugin warns: Could not open file /var/lib/dnf/plugins/local/repodata/repomd.xml
After installing the local plugin, until another package is installed and thus the default /var/lib/dnf/plugins/local repository populated, the directory does not exist and DNF5 warns about it:

~~~~
# dnf reinstall perl-Sub-HandlesVia
 Copr repo for rpminspect                                      100% |  11.0 KiB/s |   1.8 KiB |  00m00s
 Copr repo for dnfnightly owned by rpmsoftwaremanagement       100% |  11.1 KiB/s |   1.8 KiB |  00m00s
 Local libdnf5 plugin repo                                     ???% |   0.0   B/s |  -1.0   B |  ?     
>>> Curl error (37): Could not read a file:// file for file:///var/lib/dnf/plugins/local/repodata/repomd.xml [Could not open file /var/lib/dnf/plugins/local/repodata/repomd.xml] - file:///var/lib/dnf/plugins/local/repodata/repomd.xml
>>> Usable URL not found
Repositories loaded.
~~~~

Would it be possible to enhance the plugin to auto-disable to repository if the (default) directory does not exist? Or to install the plugin with a prepopulated, empty repository?

m-blaha:
Create a second repository `_dnf_local_nogpgcheck` with pkg_gpgcheck disabled
for packages originally from repositories with pkg_gpgcheck disabled. This
avoids signature verification failures when reinstalling or upgrading such
packages.

The `_dnf_local_nogpgcheck` repository is stored in a separate directory
and is only created when there are packages to cache from repositories with
pkg_gpgcheck disabled.

The `_dnf_local` repository continues to use its existing directory
for backward compatibility.

Also fix metadata_expire being set globally instead of per-repo, which was
causing unnecessary metadata downloads for all repositories.

Also only create the repo directories when there are packages to put in
them, avoiding warnings about missing repodata on first use.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2580 
Resolves: https://github.com/rpm-software-management/dnf5/issues/2581

### DIFF
--- a/doc/libdnf5_plugins/local.8.rst
+++ b/doc/libdnf5_plugins/local.8.rst
@@ -11,20 +11,29 @@ Local Plugin
 Description
 ===========
 
-After each libdnf5 transaction copy all downloaded packages to a ``_dnf_local`` repository on the local filesystem and generate repository metadata.
+After each libdnf5 transaction copy all downloaded packages to local repositories on the filesystem and generate repository metadata.
 
-The repository is automatically added by the plugin with the following options::
+The plugin creates two repositories:
 
-    [_dnf_local]
-    name = Local libdnf5 plugin repo
-    baseurl = <repodir>
+* ``_dnf_local`` for packages from repositories with ``pkg_gpgcheck`` enabled.
+  The ``pkg_gpgcheck`` setting is inherited from the main configuration.
+  The repository doesn't specify any ``gpgkey``, it assumes all required keys
+  were already imported.
+* ``_dnf_local_nogpgcheck`` for packages from repositories with ``pkg_gpgcheck``
+  disabled. This repository has ``pkg_gpgcheck`` set to ``false``.
+
+Each repository is only created when there are cached packages in it.
+
+Both repositories are configured with the following options::
+
     skip_if_unavailable = true
-    cost = 500
     metadata_expire = 0
 
-Note that the repository has ``pkg_gpgcheck`` verification enabled by default but doesn't specify any ``gpgkey``, it assumes all required keys were already imported.
+The ``_dnf_local`` repository has ``cost = 500`` and the ``_dnf_local_nogpgcheck``
+repository has ``cost = 501``, so the gpgcheck-enabled repository is preferred
+when both contain the same package.
 
-To generate the metedata `createrepo_c` is required.
+To generate the metadata `createrepo_c` is required.
 
 Configuration
 =============
@@ -52,6 +61,13 @@ It can also contain:
 
     Path where the local repository is located.
     By default it is in :ref:`persistdir <persistdir_options-label>` in ``plugins/local`` subdirectory.
+
+``repodir_nogpgcheck``
+    :ref:`string <string-label>`
+
+    Path where the local repository for packages from repos with ``pkg_gpgcheck``
+    disabled is located.
+    By default it is in :ref:`persistdir <persistdir_options-label>` in ``plugins/local-nogpgcheck`` subdirectory.
 
 
 The ``createrepo`` section requires:
--- a/libdnf5-plugins/local/local.conf
+++ b/libdnf5-plugins/local/local.conf
@@ -2,8 +2,11 @@
 name = local
 enabled = host-only
 
-# Path to the local repository.
+# Path to the local repository for packages from repos with pkg_gpgcheck enabled.
 # repodir = /var/lib/dnf/plugins/local
+
+# Path to the local repository for packages from repos with pkg_gpgcheck disabled.
+# repodir_nogpgcheck = /var/lib/dnf/plugins/local-nogpgcheck
 
 # Createrepo options. See man createrepo_c
 [createrepo]
--- a/include/libdnf5/rpm/package.hpp
+++ b/include/libdnf5/rpm/package.hpp
@@ -503,6 +503,11 @@ public:
     // @replaces libdnf:libdnf/dnf-package.h:function:dnf_package_get_repo(DnfPackage * pkg)
     libdnf5::repo::RepoWeakPtr get_repo() const;
 
+    /// Whether OpenPGP signature verification is required for this package
+    /// based on its repository configuration. Checks localpkg_gpgcheck for
+    /// command-line packages and pkg_gpgcheck for repository packages.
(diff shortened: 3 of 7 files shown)


fhbash (dismissed):
This is nice! Thanks!

One question do we'll need aditional CI tests for this change?

m-blaha:
Yeah, if we decide to merge this, the CI tests are definitely needed. I'd like to get feedback also from @ppisar (who had different preferred solution in mind) and @evan-goode . I'll ask them for review.

ppisar:
This pull request is good. I only don't understand why a location of the signed `_dnf_local` repository is configurable with `repodir` plugin option, but a location of the unsigned `_dnf_local_nogpgcheck` repository is not configurable.

m-blaha:
> This pull request is good. I only don't understand why a location of the signed `_dnf_local` repository is configurable with `repodir` plugin option, but a location of the unsigned `_dnf_local_nogpgcheck` repository is not configurable.

Oh, good catch! Thank you.

m-blaha:
I've added `repodir_nogpgcheck` config option to make _dnf_local_nogpgcheck repo location configurable.

doc/libdnf5_plugins/local.8.rst
```diff
@@ -11,20 +11,28 @@ Local Plugin
...
-The repository is automatically added by the plugin with the following options::
+The plugin creates two repositories:
+
+* ``_dnf_local`` for packages from repositories with ``pkg_gpgcheck`` enabled.
+  The ``pkg_gpgcheck`` setting is inherited from the main configuration.
+  The repository doesn't specify any ``gpgkey``, it assumes all required keys
+  were already imported.
+* ``_dnf_local_nogpgcheck`` for packages from repositories with ``pkg_gpgcheck``
+  disabled. This repository has ``pkg_gpgcheck`` set to ``false`` and is stored
+  in a separate directory (``plugins/local-nogpgcheck`` under :ref:`persistdir
```
ppisar: I would remove a text in the parentheses, maybe "and is stored in a separate directory" clause too. The location is not detailed in the above item for _dnf_local either. The location is well explained later `repodir_nogpgcheck` option.

libdnf5-plugins/local/local.conf
```diff
@@ -5,6 +5,9 @@ enabled = host-only
 # Path to the local repository.
```
ppisar: I would append "for packages from repos with pkg_gpgcheck enabled" to make clear how it differs from repodir_nogpgcheck.

ppisar (commented):
Besides the documentation and comments, it looks good and works for me. But I have two suggestions:

Both repositories have the same cost 500. Wouldn't be better to have the unsigned repository with slightly higher cost (501) than the singed one so that it's clear what happens when both repositories contain the same package? Preferably to prefer the signed one? According to my tests, the plugin prefers a repository which was not updated the last one, making the repository selection nondeterministic with respect to the currently being installed package.

Would it be possible to rename old identifiers related to the signed repository in the plugin code to bear "signed" word. E.g. to have LOCAL_YESGPGCHECK_REPO_NAME instead of LOCAL_REPO_NAME? To be on par with LOCAL_NOGPGCHECK_REPO_NAME identifier and enjoy the symmetry also in the code?

m-blaha:
@ppisar I've addressed your comments.

ppisar:
Could you also rename private member variable `repodir` at line 147?

m-blaha:
Well, I was thinking about it, but it's closely connected with `repodir` config option, which cannot be renamed without breaking compatibility.

ppisar:
I know. I meant keeping the configuration option intact, while renaming the variable in the code. But it's not important. I will approve it as it is.

ppisar (approved):
Good.

PR #2699, opened 2026-04-20
daemon: Add schedule_for_next_boot method to Offline interface

### LINKED ISSUE #2693: dnf5daemon: Add a method to prepare offline update for the next machine start
It's possible to prepare offline update or upgrade just by downloading the packages, similar to `dnf update --offline`, which does not schedule the update for the next machine start (no `/system-update` file created), that can be done as a separate step with `dnf offline reboot`. I miss the option in the dnf5daemon.

The `org.rpm.dnf.v0.Offline` -> `get_status` returns whether the update is pending (which also means the `/system-update` file exists, if I'm not mistaken), and the actual transaction status. With the download only operation the `status` member says `download-complete`, while when the transaction was there, but cancelled, the `status` says `ready`. I do not know how much other work is needed to turn from `download-complete` to `ready` state, but maybe the download also checked the transaction and a similar `dnf offline reboot` procedure is possible. When I do `dnf update --offline` is also ends the transaction in the `download-complete` status.

What I really want here is to create the symlink file `/system-update`, the restart will be done by other means, when the prepared offline update is in the correct state.

You can already cancel it with `org.rpm.dnf.v0.Offline` -> `cancel_with_options`, but you cannot "prepare_for_restart".

Why I need this: the prepared offline update can be a simple package update or a distribution upgrade, depending whether it was done by `org.rpm.dnf.v0.rpm.Rpm::upgrade()` or by `org.rpm.dnf.v0.rpm.Rpm::system_upgrade()`. I can examine the `transaction_status` from the `get_status` and decide what to do, but re-running these commands is not good, it can take time to refresh and everything, while the offline transaction is already prepared, not talking if you change anything in the `transaction_status` content, being it a key name or its value, the apps would need to be updated for that change. The suggested "only create the symlink when the offline update is in an installable state" is much simpler for all interested parties.

Alternatively, if you'd want to not change the API, which is perfectly fine by me, you can create the symlink (if possible to run the prepared offline update) as part of the `org.rpm.dnf.v0.Offline`->`set_finish_action_with_options`/`set_finish_action`. That would make sense, because you won't call this method without wanting to apply the updates the next start. You call `cancel_with_options`/`cancel` when you want to skip the update the next start.

m-blaha:
Add a new D-Bus method to the org.rpm.dnf.v0.Offline interface that schedules a prepared offline transaction for the next boot without rebooting. This creates the /system-update magic symlink and sets the transaction status to "ready", equivalent to `dnf5 offline reboot` without the actual reboot.

This is needed by GUI applications (e.g. GNOME Software) that download offline updates via the daemon and handle the reboot themselves.

Also fixes transaction status inconsistency after cancelling the offline transaction - the status is reverted to "download-complete" in such situations.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2693

### DIFF
--- a/dnf5daemon-server/services/offline/offline.cpp
+++ b/dnf5daemon-server/services/offline/offline.cpp
@@ -30,6 +30,26 @@
 
 const char * const ERR_ANOTHER_TOOL = "Offline transaction was initiated by another tool.";
 
+std::optional<libdnf5::offline::OfflineTransactionState> Offline::read_transaction_state(std::string & error_msg) {
+    const std::filesystem::path state_path{get_datadir() / libdnf5::offline::TRANSACTION_STATE_FILENAME};
+    std::error_code ec;
+    if (!std::filesystem::exists(state_path, ec)) {
+        error_msg = "No offline transaction is stored.";
+        return std::nullopt;
+    }
+    libdnf5::offline::OfflineTransactionState state{state_path};
+    const auto & read_exception = state.get_read_exception();
+    if (read_exception != nullptr) {
+        try {
+            std::rethrow_exception(read_exception);
+        } catch (const std::exception & ex) {
+            error_msg = ex.what();
+        }
+        return std::nullopt;
+    }
+    return state;
+}
+
 std::filesystem::path Offline::get_datadir() {
     auto base = session.get_base();
     const auto & installroot = base->get_config().get_installroot_option().get_value();
@@ -135,6 +155,17 @@ void Offline::dbus_register() {
                     session.get_threads_manager().handle_method(
                         *this, &Offline::set_finish_action, call, session.session_locale);
                 },
+                {}},
+            sdbus::MethodVTableItem{
+                sdbus::MethodName{"schedule_for_next_boot"},
+                sdbus::Signature{"a{sv}"},
+                {"options"},
+                sdbus::Signature{"bs"},
+                {"success", "error_msg"},
+                [this](sdbus::MethodCall call) -> void {
+                    session.get_threads_manager().handle_method(
+                        *this, &Offline::schedule_for_next_boot, call, session.session_locale);
+                },
                 {}})
         .forInterface(dnfdaemon::INTERFACE_OFFLINE);
 #else
@@ -211,6 +242,17 @@ void Offline::dbus_register() {
         [this](sdbus::MethodCall call) -> void {
             session.get_threads_manager().handle_method(
                 *this, &Offline::set_finish_action, call, session.session_locale);
+        });
+    dbus_object->registerMethod(
+        dnfdaemon::INTERFACE_OFFLINE,
+        "schedule_for_next_boot",
+        {"a{sv}"},
+        {"options"},
+        "bs",
+        {"success", "error_msg"},
+        [this](sdbus::MethodCall call) -> void {
+            session.get_threads_manager().handle_method(
+                *this, &Offline::schedule_for_next_boot, call, session.session_locale);
         });
 #endif
 }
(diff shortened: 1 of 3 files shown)


m-blaha:
@mcrha I cannot assign you as a reviewer, but I'd like to get your opinion before merging this.

mcrha:
I gave it some testing (I cannot really review the code itself), and it does what I expect for it. Thank you.

It uncovered a little regression on the gnome-shell side in the 50.x series too, which I'm going to deal with there.

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -253,6 +294,15 @@ sdbus::MethodReply Offline::impl_cancel(sdbus::MethodCall & call, const dnfdaemo
...
                 success = false;
                 error_msg = ec.message();
+            } else {
+                // Reset the status back to download-complete since the transaction
+                // is no longer scheduled for the next boot.
+                std::string state_error;
+                auto state = read_transaction_state(state_error);
+                if (state && state->get_data().get_status() == libdnf5::offline::STATUS_READY) {
+                    state->get_data().set_status(libdnf5::offline::STATUS_DOWNLOAD_COMPLETE);
+                    state->write();
```
ppisar, line 304: It's not probably good to fix it in this pull request, but I cannot see any file locking inside the state->write() call. What a D-bus clients cancels an offline update with this code, while another libdnf5 application reads the state? Will there be half-written state file?

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -388,3 +423,48 @@ sdbus::MethodReply Offline::set_finish_action(sdbus::MethodCall & call) {
...
+                if (ec) {
+                    error_msg = ec.message();
+                } else {
+                    state->get_data().set_status(libdnf5::offline::STATUS_READY);
+                    state->write();
+                    success = true;
+                }
+            }
+        }
+    }
```
ppisar, line 465: Ideally this should go to libdnf5 to deduplicate code we have in dnf5/commands/offline/offline.cpp. I leave it up to you whether you want to deduplicate it now, or later, or never.

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -30,6 +30,26 @@
 
 const char * const ERR_ANOTHER_TOOL = "Offline transaction was initiated by another tool.";
 
+std::optional<libdnf5::offline::OfflineTransactionState> Offline::read_transaction_state(std::string & error_msg) {
+    const std::filesystem::path state_path{get_datadir() / libdnf5::offline::TRANSACTION_STATE_FILENAME};
+    std::error_code ec;
+    if (!std::filesystem::exists(state_path, ec)) {
+        error_msg = "No offline transaction is stored.";
```
ppisar, line 37: How is this message internationalized?

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -388,3 +423,48 @@ sdbus::MethodReply Offline::set_finish_action(sdbus::MethodCall & call) {
...
+    }
+    bool success{false};
+    std::string error_msg{};
+
+    auto state = read_transaction_state(error_msg);
+    if (state) {
+        const auto & status = state->get_data().get_status();
+        if (status != libdnf5::offline::STATUS_DOWNLOAD_COMPLETE && status != libdnf5::offline::STATUS_READY) {
+            error_msg =
+                fmt::format("Cannot schedule offline transaction for next boot. Transaction status is \"{}\".", status);
```
ppisar, line 444: Internationalization is missing.

ppisar (approved):
Overall good. I have only small nitpicks with internationalizing error messages, and idealistic opinion that much of the code should be shared with dnf5 executable, probably in libdnf5.

I will leave it up to you whether you want to follow the ideas, or not. Feel free to merge it as it is.

Conan-Kudo:
FYI @anaselli (for dnfdragora offline-update support)

dnf5daemon-server/services/offline/offline.cpp
m-blaha: Thanks, the problem is real and I filed a follow-up issue for this: https://github.com/rpm-software-management/dnf5/issues/2702

dnf5daemon-server/services/offline/offline.cpp
m-blaha: Since the daemon currently barely uses gettext, I filed a follow-up issue for internationalization of all user-facing strings: https://github.com/rpm-software-management/dnf5/issues/2703

dnf5daemon-server/services/offline/offline.cpp
m-blaha: Thanks, thats a valid point and code could most likely be shared with CLI `offline reboot` command. TBH, I'd prefer to do this later in a follow-up PR.

anaselli:
> FYI @anaselli (for dnfdragora offline-update support)

I was thinking to look at that after some other task i've in my todo list, will this change  be into incoming mageia and fedora stable?

evan-goode:
Yes for Fedora stable. No breaking API/ABI changes here. This is motivated by https://bugzilla.redhat.com/show_bug.cgi?id=2457884.

PR #2706, opened 2026-04-22
Fix subprocess issues found with dash shell

### LINKED ISSUE #2704: test_binary_data fails since 5.4.1.0
test_binary_data started failing since 5.4.1.0:

```
test_subprocess.cpp:100:Assertion
Test name: UtilsSubprocessTest::test_binary_data
equality assertion failed
- Expected: 5
- Actual  : 20
Failures !!!
Run: 467   Failure total: 1   Failures: 1   Errors: 0
```

https://salsa.debian.org/bluca/dnf5/-/jobs/9450318

Is this a known issue?

m-blaha:
The \xNN hex escapes in printf are not supported by POSIX shells like dash. Replace with \NNN octal escapes so the test works regardless of which shell is used.

Also a race condition in child process output polling was discovered while running the tests with dash shell.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2704

### DIFF
--- a/libdnf5/utils/subprocess.cpp
+++ b/libdnf5/utils/subprocess.cpp
@@ -231,7 +231,7 @@ CompletedProcess run(const std::string & command, const std::vector<std::string>
                 throw SystemError(errno, M_("poll failed during subprocess I/O"));
             }
 
-            if (fds[STDOUT_FD].revents & POLLIN) {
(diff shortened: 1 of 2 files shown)


kontura (approved):
Looks good to me!

PR #2701, opened 2026-04-20
refactor: pass by const reference

vil02:
Continuation of #2698.

Sorry for splitting this into such small pieces - in the mentioned PR I got some _strange_ build errors and I am trying to understand where do the failures originate from.

PR #2708, opened 2026-04-22
refactor: pass by const reference

vil02:
Similar to #2698, #2701 and #2705.

So I guess in the first PR, I did not update one of the header files. Sorry for that.

### DIFF
--- a/dnf5daemon-server/services/rpm/rpm.cpp
+++ b/dnf5daemon-server/services/rpm/rpm.cpp
@@ -383,7 +383,8 @@ std::vector<std::string> get_filter_patterns(dnfdaemon::KeyValueMap options, con
     return filter_patterns;
 }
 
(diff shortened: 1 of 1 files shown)


include/libdnf5/comps/group/package_type.hpp
```diff
@@ -59,7 +59,7 @@ inline constexpr bool any(PackageType flags) {
 }
 
 LIBDNF_API PackageType package_type_from_string(const std::string & type);
-LIBDNF_API PackageType package_type_from_string(const std::vector<std::string> types);
+LIBDNF_API PackageType package_type_from_string(const std::vector<std::string> & types);
```
m-blaha: This is ABI breaking change on public API.

m-blaha (approved):
Thank you!

Issue #1985, opened 2025-01-04
dnf5 install --downloadonly does not check package signatures


marmarek:
It looks like dnf5 checks signatures only just before installing packages, but not after downloading them. dnf4 verified signatures just after downloading. This has significant security implication for `--downloadonly` option! Thankfully I have a CI check for this case...

Reproducer:
```console
[user@testbuilder dd]$ cat /etc/yum.repos.d/unknown-key.repo 
[unknown-key]
name=unknown-key
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-41-primary
baseurl=https://ftp.qubes-os.org/~marmarek/repo-verify-unknown-key
gpgcheck=1

[user@testbuilder dd]$ sudo dnf4 install --repoid=unknown-key -y --downloadonly tmux
unknown-key                                     2.6 kB/s | 580  B     00:00    
Dependencies resolved.
================================================================================
 Package      Architecture   Version                  Repository           Size
================================================================================
Installing:
 tmux         x86_64         1000:500-1.fc30          unknown-key         5.8 k

Transaction Summary
================================================================================
Install  1 Package

Total download size: 5.8 k
Installed size: 0  
DNF will only download packages for the transaction.
Downloading Packages:
tmux-500-1.fc30.x86_64.rpm                       53 kB/s | 5.8 kB     00:00    
--------------------------------------------------------------------------------
Total                                            51 kB/s | 5.8 kB     00:00     
unknown-key                                     1.6 MB/s | 1.6 kB     00:00    
GPG key at file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-41-primary (0xE99D6AD1) is already installed
The GPG keys listed for the "unknown-key" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: tmux-1000:500-1.fc30.x86_64
 GPG Keys are configured as: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-41-primary
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: GPG check FAILED
[user@testbuilder dd]$ echo $?
1
```

```console
[user@testbuilder dd]$ sudo dnf5 install --repoid=unknown-key -y --downloadonly tmux
Updating and loading repositories:
Repositories loaded.
Package                 Arch    Version                  Repository         Size
Installing:
[...]

PR #2710, opened 2026-04-22
Verify RPM signatures per package in dnf download

Tomurisk:
With all due respect, I'm asking to implement this warning, because as an user, I had no idea that `dnf download` downloads RPMs without any GPG verification whatsoever. Then when you end up using those packages in your personal projects, there's supply chain risk, as well as when you install a RPM with `sudo dnf install` – no verification in that case as well, because:
```
keepassxc              x86_64 2.7.5-1.mga9             @commandline    31.1 MiB
Warning: skipped OpenPGP checks for 1 package from repository: @commandline
```
Now consider this: one downloads with `dnf download`, then does `sudo dnf install [whatever RPM]`, all security checks get bypassed because `@commandline` "repository", which is a major self-reinforcing security oversight. APT, on other hand, screams loudly and proudly that the repository has no public key, therefore downloads don't even succeed. I understand that there are fundamental philosophical differences between APT and DNF, however, we can take a glance:
```
The following signatures couldn't be verified because the public key is not available
W: GPG error: https://ports.ubuntu.com/ubuntu-ports xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 40976EAF437D05B5 NO_PUBKEY 3B4FE6ACC0B21F32
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
```
Those are **_three warnings_**, that happen during `apt update`, and the user needs to provide extra flags/repo configuration for the download to succeed. `apt-get download` blocks DEBs from unsigned repos and reports them as:
```
E: Version '[version]' for '[package]' was not found
```
I strongly believe we can do better than whatever we have now. It doesn't help that the overwhelming majority share of DNF5 comes from Fedora, which only reinforces the cognitive bias that "Fedora has us covered by default", when in this instance, this is not the case.

I'm thankful to the DNF community for the work that was made over the years, but this issue needs to be solved.

PS. This warning does not change any existing behavior or break any workflows, it simply informs users of a security property they currently have no way of knowing.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -27,8 +27,10 @@
 #include <libdnf5/rpm/package.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/rpm/package_set.hpp>
+#include <libdnf5/rpm/rpm_signature.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
 
+#include <filesystem>
 #include <algorithm>
 #include <iostream>
 #include <map>
@@ -170,7 +172,6 @@ void DownloadCommand::set_argument_parser() {
             arch_option.emplace(value);
             return true;
         });
-
 
     cmd.register_named_arg(arch);
     cmd.register_named_arg(resolve);
@@ -415,6 +416,36 @@ void DownloadCommand::run() {
         std::cout << "Downloading Packages:" << std::endl;
     }
     downloader.download();
+
+    // gpg verification
+    bool all_ok = true;
+    std::error_code ec;
+    libdnf5::rpm::RpmSignature rpm_signature(ctx.get_base());
+
+    for (const auto & [nevra, pkg] : packages_to_download) {
+        if (!pkg.is_pkg_gpgcheck_enabled()) {
+            continue;
+        }
+
+        auto pkg_path = pkg.get_package_path();
+        if (std::filesystem::exists(pkg_path, ec)) {
+            auto check_result = rpm_signature.check_package_signature(pkg_path);
+            if (check_result != libdnf5::rpm::RpmSignature::CheckResult::OK) {
+                std::cerr << libdnf5::utils::sformat(
+                    _("OpenPGP verification FAILED for '{}': {}"),
+                    std::filesystem::path(pkg_path).filename().c_str(),
+                    rpm_signature.check_result_to_string(check_result))
+                          << std::endl;
+                std::filesystem::remove(pkg_path, ec);
+                all_ok = false;
+            }
+        }
+    }
+
+    if (!all_ok) {
+        throw libdnf5::cli::CommandExitError(
+            1, M_("OpenPGP signature verification failed for some packages."));
+    }
 }
 
 


m-blaha:
Hi @Tomurisk!

Thanks for bringing this up! The issue is real and definitely worth addressing.

I'm concerned that unconditionally adding another warning message may contribute to "warning fatigue" - users tend to start ignoring warnings they see repeatedly, which defeats the purpose of having them.

Given that we already have all the necessary infrastructure, I'd like to encourage you to consider implementing **actual GPG verification** for the download command instead of the warning. Similar to what we already have in place for the `reposync` plugin. The implementation could even take the existing `--no-gpgchecks` flag into account to skip verification and allow downloads from repositories without the pkg_gpgcheck config option set.

What do you think? Would you be interested in this? The dnf team is happy to provide any help or guidance needed.

Thanks again for caring about dnf5's security!

Tomurisk:
@m-blaha I think you have a good point, however I'm not quite advanced with DNF tooling, conventions, or C++, so I used Perplexity to implement this feature. It seems to compile well on its own, however, when I used tito to create RPMs, I was met with SWIG errors, I'm unsure how the entire build pipeline works, so I will test out an artifact when a COPR build finishes.

Tomurisk:
Tested it, but unfortunately the GPG checking never fires off... Any help, please?

That said, I'd much rather not vibe-code the implementation you've suggested, because it will end up polluting the codebase. Otherwise, it takes around 20 minutes to compile DNF5 on my own machine, and evidently on COPR, so realistically I can do only 15 compilations in 5 hours, whether it works, not sure at all. I believe the code as-is might be a good starting point, but somebody with proper expertise should look into it. While I could theoretically be learning the codebase by building-waiting-rebuilding, following some sort of design specification and guidelines, sharpening C++ skills, we might have a nuclear war before I'm done with it. Alas, it just happens so that the warning I made is likely the best quality coding I can realistically do, otherwise we'd have code that is not memory safe and such.

I'm not entirely sure whether the upcoming release of Fedora will use the latest version of DNF5, but perhaps it'd be logical to add the warning as in my previous (now detached) commit as a quick-and-dirty stub, then move upon to the actual verification within `dnf download`?

ppisar:
This is #1985.

m-blaha:
Sorry, I overlooked the #1985 issue.
                                                                                         
I can see two paths forward here:
                                                                                     
1. Close this PR and link it to 1985. Your analysis of the problem in the description adds valuable context for resolving the issue. Someone from the team can pick it up from there. To be honest, I don't think the warning-only approach adds enough value to justify merging it - users who don't know about the verification gap won't understand what to do with the warning, and those who do know don't need it.

2. Get this PR to a mergeable state. Issue 1985 is broader, it covers both `dnf download` and `dnf install --downloadonly`. But these two most likely would not share code and can be implemented independently, so resolving `dnf download` on its own is fine. Your current PR has a few bugs and some design adjustments are needed, but the core approach is sound. I understand that the learning curve for contributing to a C++ project you're not familiar with is steep. It will most likely take several rounds of code and review back and forth. If you're still interested in continuing, I'm happy to work on it with you, but no pressure at all.

Either way, this will get addressed. Let me know what works best for you!

m-blaha:
> Tested it, but unfortunately the GPG checking never fires off... Any help, please?

I think the issue is here https://github.com/rpm-software-management/dnf5/pull/2710/changes#diff-9c153c08a2655daec485cc9768c7576ed9924c510c712486dd676688d21ab75aR435-R437

```
        auto dest_dir = ctx.get_base().get_config().get_destdir_option().get_value();
        auto pkg_path = std::filesystem::path(dest_dir) / pkg.get_location();
        downloaded_paths.push_back(pkg_path);
```

Getting the right download location is a bit messy in dnf5. What you used will return something like `./Packages/a/acpi-1.8-2.fc43.x86_64.rpm` (the destdir and relative location in the repository joined together). But the real path is `./acpi-1.8-2.fc43.x86_64.rpm` - the package downloader uses only `pkg.location().filaname()` and costructs it this way:
https://github.com/rpm-software-management/dnf5/blob/0bc7deeb3dd29974096e1cff0e9c0c6c789c9d89/libdnf5/repo/package_downloader.cpp#L274-L275

Thus your code does not find the file and the gpg check is skipped.

m-blaha:
Actually, you don't need to keep a separate `downloaded_paths` vector. I think you can simply do this for the verification loop:
```
for (const auto & [nevra, pkg] : packages_to_download) {
      auto pkg_path = pkg.get_package_path();
      // verify...                                                                       
  }
```

dnf5/commands/download/download.cpp
```diff
@@ -171,6 +176,14 @@ void DownloadCommand::set_argument_parser() {
...
         });
 
+    auto * gpgcheck = gpgcheck_option;
+
+    auto nogpgchecks = parser.add_new_named_arg("no-gpgchecks");
+    nogpgchecks->set_long_name("no-gpgchecks");
+    nogpgchecks->set_description("Skip GPG signature verification of downloaded packages");
+    nogpgchecks->set_const_value("false");
+    nogpgchecks->link_value(gpgcheck);
+    cmd.register_named_arg(nogpgchecks);
```
m-blaha: We can drop this command specific no-gpgchecks option, since there already is a global one: https://github.com/rpm-software-management/dnf5/blob/23eff3731cef4d048b6e4eb75677a8f467e349a3/dnf5/main.cpp#L429-L444

dnf5/commands/download/download.cpp
```diff
@@ -415,6 +428,36 @@ void DownloadCommand::run() {
         std::cout << "Downloading Packages:" << std::endl;
     }
     downloader.download();
+
+    // gpg verification
+    if (gpgcheck_option->get_value() &&
```
m-blaha: I would move this test to validation loop and check it per package using `pkg.is_pkg_gpgcheck_enabled()`. This method returns true, if `pkg_gpgcheck` option for repository the package is downloaded from is set to `true`. It doesn't make sense to try to verify package comming from repo with gpgcheck disabled (which most likely even doesn't have a key).

dnf5/commands/download/download.cpp
```diff
@@ -415,6 +428,36 @@ void DownloadCommand::run() {
...
+        std::error_code ec;
+        libdnf5::rpm::RpmSignature rpm_signature(ctx.get_base());
+        
+        for (const auto & [nevra, pkg] : packages_to_download) {
+            auto pkg_path = pkg.get_package_path();
+            if (std::filesystem::exists(pkg_path, ec)) {
+                auto check_result = rpm_signature.check_package_signature(pkg_path);
+                if (check_result != libdnf5::rpm::RpmSignature::CheckResult::OK) {
+                    std::cerr << libdnf5::utils::sformat(
+                        _("GPG verification FAILED for '{}': {}"),
```
m-blaha: The user facing string across dnf5 use `OpenPGP` instead of `GPG`.

Tomurisk:
@m-blaha Looks like it finally works properly, I implemented those fixes you commented on my code

Fundamentally, it's working, however as far as I understand there's a test workflow for Rawhide that has:
```
Command 'set -eo pipefail; dnf copr enable rpmsoftwaremanagement/dnf-nightly fedora-rawhide-x86_64 -y && dnf download libdnf5 libdnf5-cli --repo copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly --destdir=/var/share/test-artifacts/nightly' returned 1.
```
Which naturally causes:
```
Repositories loaded.
libdnf5-cli-0:5.4.2.0^2026042402102430 100% |  40.5 MiB/s | 332.0 KiB |  00m00s
libdnf5-0:5.4.2.0^20260424021024309548 100% |  76.4 MiB/s |   1.3 MiB |  00m00s
OpenPGP verification FAILED for 'libdnf5-5.4.2.0^20260424021024309548.main.3.g0bc7deeb-1.fc45.x86_64.rpm': No corresponding key was found.
OpenPGP verification FAILED for 'libdnf5-cli-5.4.2.0^20260424021024309548.main.3.g0bc7deeb-1.fc45.x86_64.rpm': No corresponding key was found.
OpenPGP signature verification failed for some packages.
```

I assume this breaks `dnf download` workflows that need to be updated with `dnf download --no-gpgchecks`, a simple sed might work for this, but I can't speak for the infrastructure.

m-blaha:
Thank you, it looks good! I'll investigate the failing test scenarios and prepare a fix for them. The issue with the ABI check is that although packages in Copr repositories are signed, the key has not been imported to rpmdb yet. The download command, unlike install/update/... commands, does not import keys to the RPM database. It uses only already imported keys.

One remaining question I'll raise at the next DNF team meeting is what should happen when verification fails. There are basically two options:

1. What you've currently implemented: print an error message, delete the failed RPM file, and return a non-zero exit code. Only keep the failing files if `--no-gpgchecks` option was used. This is the strictest approach.
2. A more relaxed behavior: print a warning, keep the downloaded file for user inspection, and return a non-zero exit code. This is closer to your original implementation, but prints a warning only when a package actually fails verification.

Depending on the decision, the man pages will also need to be updated accordingly.

dnf5/commands/download/download.cpp
```diff
@@ -81,6 +83,9 @@ void DownloadCommand::set_argument_parser() {
     debugsource_option = dynamic_cast<libdnf5::OptionBool *>(
         parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
 
+    gpgcheck_option = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(true))));  // default ON
+
```
m-blaha: The gpgcheck_option is not used any more, this can be removed.

dnf5/commands/download/download.hpp
```diff
@@ -51,6 +51,7 @@ class DownloadCommand : public Command {
     libdnf5::OptionBool * srpm_option{nullptr};
     libdnf5::OptionBool * debuginfo_option{nullptr};
     libdnf5::OptionBool * debugsource_option{nullptr};
+    libdnf5::OptionBool * gpgcheck_option{nullptr};
```
m-blaha: The gpgcheck_option attribute is not used, can be removed.

Tomurisk:
> 2\. A more relaxed behavior: print a warning, keep the downloaded file for user inspection, and return a non-zero exit code. This is closer to your original implementation, but prints a warning only when a package actually fails verification.

I suggest pushing in the meeting for this one, because as you mentioned before, users tend to skip warnings. When one needs to explicitly use `--no-gpgchecks`, it naturally evokes (at least should) the question "Why do I need to use this extra flag?". Therefore, removing the RPM is a more proactive action, for the advantage of user's security.

Also, an example from my personal use case, from DNF4, which is legacy at this point, but relevant nonetheless:
I maintain an AppImage, which uses `dnf download` to fetch the RPMs, then bundle them up. It requires this extra step:
```
    if rpm --checksig "$rpm" >/dev/null 2>&1; then
        echo "RPM signature OK"
    else
        echo "ERROR: Signature verification failed!"
        exit 1
    fi
```
I believe it'd be easier to delete the file, I guess with this new approach I'd need a new condition to check whether it's needed to download the RPM again, but it will overall make the code cleaner. Yet we're looking couple years into downstream, when Fedora, major RHEL derivatives adopt my implementation.

pmatilai:
> Now consider this: one downloads with dnf download, then does sudo dnf install [whatever RPM], all security checks get bypassed because @commandline "repository", which is a major self-reinforcing security oversight. 

Yes, this was always an unfortunate choice, I believe it was basically done to match with the long-standing rpm behavior where you can install whatever locally, signatures or no.

In RPM >= 6.0 signatures are getting enforced everywhere on RPM level by default. Fedora 43-44 still default to the old behavior but it's changing to the upstream default in 45: https://fedoraproject.org/wiki/Changes/Enforcing_signature_checking_by_default

Of course there are still RPM 4.x distros out there so doing something about this in dnf5 is not a bad idea.

Tomurisk:
@m-blaha Hi again, what's the progress on this PR?

m-blaha:
Hi @Tomurisk, apologies for the delay, and thank you for following up!
  
We discussed this at the team meeting and really appreciate the work you've put into this. After discussion, the team agreed that enabling GPG verification unconditionally by default introduces some problems we'd like to avoid:

- It's a significant behavior change for existing `dnf download` workflows.
- Packages from repositories whose GPG keys haven't been imported into the RPM database yet (e.g. a newly enabled Copr repo) would fail verification even though there's nothing wrong with them - users would be forced to add `--no-gpgchecks` just to get a working download.

We see two possible paths forward:

1. Make it opt-in with a `--gpgcheck` flag (similar to what reposync already does). Users who care about verification would explicitly pass the flag. This keeps existing workflows intact and avoids confusion with repos that don't have keys imported yet.
2. Print a warning to stderr for packages that fail the check, without deleting them or exiting with a non-zero code.

We'd recommend option 1 - it gives security-conscious users a clear, intentional tool, and it's consistent with how reposync handles this.

Either way, the core verification logic you wrote is solid and would carry over. The main change would be gating it behind the flag (or switching from error to warning).

PR #2715, opened 2026-04-28
manifest: Split downloaded packages by basearch with --per-arch option

pkratoch:
Related: https://redhat.atlassian.net/browse/SWM-5555


fhbash (approved):
LGTM, thanks!

PR #2717, opened 2026-04-29
Avoid 5 minutes timeout on repo key import for non-interactive sessions

m-blaha:
When dnf5daemon-server needs to import a repository GPG key, it would emit
a D-Bus signal and wait up to 5 minutes for client confirmation. This
causes a hang when the client doesn't handle the signal (e.g., GNOME
Software) or the call is non-interactive.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2458182


kparal:
With this PR, gnome-software works just fine for me, even with Cursor repo added, resolving https://bugzilla.redhat.com/show_bug.cgi?id=2463519

kontura (commented):
If I understand correctly this adds the `interactive` option to several interfaces, should we also document they accept it?

Also would it make sense to add the `interactive` option to [`Base::read_all_repos`](https://dnf5.readthedocs.io/en/latest/dnf_daemon/dnf5daemon_dbus_api.8.html#org.rpm.dnf.v0.Base.read_all_repos) as well?

m-blaha:
Correct. I've added documentation to methods that now accept the option.

Regarding `read_all_repos` - yes, it would make sense, but doing so would break the API. Currently the method does not accept `options` parameter. We can deal with it by adding a new `Base::read_all_repos_with_options` (as we've already done for several other methods), but that would require also change on the caller's side.

kontura (approved):
Thank you!

PR #2721, opened 2026-04-30
add dup alias

sfalken:
Adds dup alias to distro-sync

openSUSE Tumbleweed and MicroOS users are used to this command alias, via zypper dup, and with the forthcoming availability of transactional updates via libdnf-plugin-txnupd making transactional systems available to fedora, and the use of the same plugin on openSUSE systems using dnf it provides a familiar command

### DIFF
--- a/dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
+++ b/dnf5/config/usr/share/dnf5/aliases.d/compatibility.conf
@@ -110,6 +110,13 @@ descr = "Alias for 'distro-sync'"
 group_id = 'commands-compatibility-aliases'
 complete = false
 
+['dup']
+type = 'command'
+attached_command = 'distro-sync'
+descr = "Alias for 'distro-sync'"
+group_id = 'commands-compatibility-aliases'
+complete = false
+
 ['grp']
 type = 'command'
 attached_command = 'group'


ppisar:
Downstream distributions are recommended to ship and maintain their aliases in their downstream distributions.

sfalken:
> Downstream distributions are recommended to ship and maintain their aliases in their downstream distributions.

Is Fedora considered *downstream* because I'm working on bringing this functionality into Fedora, this is more for that than for openSUSE.

ppisar:
Yes, Fedora is downstream either. Upstream defines only a minimal set of aliases to preserve a basic compatibility with DNF4 as a predecessor of DNF5.

Distributions can install a file with the alias definition into /usr/share/dnf5/aliases.d/.

As far as I know Fedora has fedora-release-common package with distribution-wide overrides, also for DNF5.

Conan-Kudo:
Yeah, no. This existed in DNF4 and it's perfectly reasonable to have it in DNF5.

As it is, I want to kick out global fedora dnf5 configuration from fedora-release because it's not a branding package thing and it needs to go somewhere that will exist regardless of branding.

Conan-Kudo (approved):

PR #2723, opened 2026-05-02
libdnf5: fix inappropriate backslash consumption in config files

### LINKED ISSUE #2503: libdnf5: Backslashes '\' are inappropriately consumed/de-escaped when reading configuration files
**Examples with repository name**
- Example with `--setopt=myrepo.name='M\y ug\\ly name'` is correct:
```
# dnf --repo=myrepo --setopt=myrepo.name='M\y ug\\ly name' repo list
repo id                           repo name                                                                                     
myrepo                          M\y ug\\ly name
```

- Example with `name='M\y ug\\ly name'` in the repository configuration file. Backslashes are de-escaped (incorrectly): 
```
# dnf --repo=myrepo repo list
repo id                           repo name                                                                                     
myrepo                          My ug\ly name
```

Userfrom1995:
The variable substitution engine in libdnf5 was overly aggressive, unconditionally removing backslashes even when they preceded normal characters (e.g., 'M\y' became 'My').

This change restricts backslash de-escaping to only special characters:
- '$' (to allow literal dollar signs)
- '}' (to allow literal braces within subexpressions)

This ensures that backslashes in repository names and paths are preserved literally, matching the behavior of --setopt and restoring compatibility with user expectations from DNF4.

Added unit tests to verify the fix and prevent regressions.

Fixes: #2503

### DIFF
--- a/libdnf5/conf/vars.cpp
+++ b/libdnf5/conf/vars.cpp
@@ -173,14 +173,17 @@ std::pair<std::string, size_t> Vars::substitute_expression(std::string_view text
         }
 
         if (res[pos] == '\\') {
-            // Escape the next character (if there is one)
+            // Escape the next character only if it is special ($ or } in subexpressions)
             if (pos + 1 >= res.length()) {
                 break;
             }
-            res.erase(pos, 1);
-            total_scanned += 2;
-            pos += 1;
-            continue;
+            char next = res[pos + 1];
+            if (next == '$' || (next == '}' && depth > 0)) {
+                res.erase(pos, 1);
+                total_scanned += 2;
+                pos += 1;
+                continue;
+            }
         }
         if (res[pos] == '$') {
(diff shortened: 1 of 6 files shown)


Userfrom1995:
The CI failures are expected, as now we don't de-escape the double backslashes and they're passed as-is to printf, which is causing the issue. So the test stack needs updating, but before doing that I need confirmation from one of the reviewers that this is the way you guys want to go. Or maybe we can skip the vars.substitute() function for the command_format? That even sounds like a bad idea to me.

ppisar:
@jrohel, what is your opinion on this change?

Userfrom1995:
Could I get a review or any thoughts on this one?

Userfrom1995:
> Could I get a review or any thoughts on this one?

Another friendly ping for a review.

PR #2724, opened 2026-05-04
Move system-repo.lock to /var/lib/dnf/system-repo.lock

### LINKED ISSUE #2709: dnf info and search commands fail with lock error on Fedora Atomic
On atomic systems like Fedora Kinoite, Silverblue, etc., it's expected that dnf can't install or remove packages, because `/usr` is read-only. (rpm-ostree is used for package management instead.) However, commands that just query package repositories, like `dnf info` and `dnf search`, are still useful and in fact used to work.

As of a recent version—I first observed this after I switched over to the Fedora 44 beta (specifically, the Kinoite and Silverblue atomic desktop images)—these commands are no longer working on systems where `/usr` is read-only: `dnf info` and `dnf search` (and possibly other dnf commands) fail with the following error message:

> Failed to open lock file "/usr/lib/sysimage/libdnf5/system-repo.lock": (30) - Read-only file system

A few notes:

* `/usr/lib/sysimage/libdnf5` is empty on these images in addition to `/usr` being read-only.
* I can also reproduce this on a Fedora 44 container image (from `quay.io/fedora/fedora:44`) by deleting the contents of `/usr/lib/sysimage/libdnf5` and then mounting a read-only overlay over `/usr`.
* Passing the option `--skip-file-locks` is a workaround and makes the commands behave as expected. I think it would make sense for the behavior given by `--skip-file-locks` to be the default if `/usr` is read-only, but I'm not all that familiar with the internals of dnf5, so maybe there's some subtlety here.

evan-goode:
The lockfile should not be `%ghost` since DNF5 expects it to always exist.

/var is a better place for the lockfile, in part since on bootc systems, we can open it for writing before unlocking /usr.

Resolves https://github.com/rpm-software-management/dnf5/issues/2709.

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -488,10 +488,10 @@ Package management library.
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/packages.toml
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/system.toml
 %verify(not md5 size mtime) %attr(0644, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/transaction_history.sqlite{,-shm,-wal}
-%verify(not md5 size mtime) %attr(0664, root, root) %ghost %{_prefix}/lib/sysimage/libdnf5/system-repo.lock
 %license lgpl-2.1.txt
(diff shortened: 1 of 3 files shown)


m-blaha (approved):
Thank you! The rawhide testing failures are unrelated (see https://github.com/rpm-software-management/ci-dnf-stack/pull/1875).

kontura (approved):

PR #2713, opened 2026-04-26
refactor(download): preallocate memory in `DownloadCommand::configure`

vil02:

PR #2725, opened 2026-05-04
refactor(search): optimise `search_processor`

vil02:
Similar to #2713.

### DIFF
--- a/dnf5/commands/search/search_processor.cpp
+++ b/dnf5/commands/search/search_processor.cpp
@@ -176,6 +176,7 @@ SearchResults SearchProcessor::get_results() {
 
     // In the end just push everything into the final result structure.
     SearchResults results;
+    results.group_results.reserve(priority_matches.size());
     for (auto const & [priority, packages] : priority_matches) {
         results.group_results.push_back(
             {.matched_keys = get_matched_keys(priority), .matched_packages = std::move(packages)});
--- a/dnf5/commands/search/search_processor.cpp
+++ b/dnf5/commands/search/search_processor.cpp
@@ -177,7 +177,7 @@ SearchResults SearchProcessor::get_results() {
     // In the end just push everything into the final result structure.
     SearchResults results;
     results.group_results.reserve(priority_matches.size());
-    for (auto const & [priority, packages] : priority_matches) {
+    for (auto & [priority, packages] : priority_matches) {
         results.group_results.push_back(
             {.matched_keys = get_matched_keys(priority), .matched_packages = std::move(packages)});
     }


m-blaha (dismissed):
Thank you! The fix is correct, although the gain is most likely negligible.

jrohel:
@vil02 
Thank you.

This is really just a minor optimization. Here are tips for further optimizations in the code - if you want, you can add:

If we want `std::move` to actually have an effect in:
`{.matched_keys = get_matched_keys(priority), .matched_packages = std::move(packages)});`
we need to drop the const in:
`for (auto const & [priority, packages] : priority_matches) {`

Similarly, for `std::move` to work in:
`priority_matches[priority].packages.insert(std::move(package));`
we should change the loop to pass by value:
`for (auto package : all_matches) {`

If you prefer to leave the code as it is, just let me know - I have no objections to merging this.

vil02:
@jrohel: thanks for the suggestions. I added a commit with the suggested change.

EDIT (cf. [log](https://download.copr.fedorainfracloud.org/results/packit/rpm-software-management-dnf5-2725/fedora-rawhide-x86_64/10476462-dnf5/builder-live.log.gz)):
```
/builddir/build/BUILD/dnf5-5.4.3.0-build/dnf5-5.4.3.0/dnf5/commands/search/search_processor.cpp: In member function ‘libdnf5::cli::output::SearchResults dnf5::SearchProcessor::get_results()’:
/builddir/build/BUILD/dnf5-5.4.3.0-build/dnf5-5.4.3.0/dnf5/commands/search/search_processor.cpp:172:27: error: cannot bind non-const lvalue reference of type ‘libdnf5::rpm::Package&’ to an rvalue of type ‘libdnf5::rpm::Package’
  172 |     for (auto & package : all_matches) {
      |                           ^~~~~~~~~~~
```

jrohel:
@vil02 
```
  172 |     for (auto & package : all_matches) {
      |                           ^~~~~~~~~~~
```
`PackageSetIterator::operator*()` returns by value. It looks like you removed `const` but kept the reference. Since it returns a temporary object, you need to remove the reference as well and loop by value:
`for (auto package : all_matches) {`
...just as I suggested in my first comment.

vil02:
With
```cpp
for (auto package : all_matches) {
```
a copy is made anyway, so I decided to leave the original code.

jrohel:
@vil02 
The point of the proposed change lies elsewhere. `PackageSetIterator::operator*()` returns by value, so keeping a `const` reference to the returned value doesn't help. On the contrary, it makes the situation worse. The subsequent `insert(std::move(package))` still creates a copy of package because of the `const`. If we use `for (auto package : all_matches) {`, `package` becomes a `mutable` variable, and the subsequent `insert(std::move(package))` will actually perform a move (saving a copy). But those are just details. I will merge it.

jrohel (approved):
Thank you for the PR.

PR #2727, opened 2026-05-05
Port bootc functionality from DNF4

evan-goode:
Adds the following bootc features from DNF4:

- `persistence` configuration option with values `transient`, `auto`, and `persist` (https://github.com/rpm-software-management/dnf/pull/2155)
- `--transient` flag to set `persistence=transient` (https://github.com/rpm-software-management/dnf/pull/2155)
- In transient mode, DNF calls `bootc usr-overlay --read-only` to mount a read-only overlayfs on /usr, and via `unshare` mounts this overlayfs as read/write in a private mount namespace. DNF can write to the overlay but it appears read-only to everything else on the system. (https://github.com/rpm-software-management/dnf/pull/2155)
- ~Whether a transaction is persistent or transient is stored in the history DB and listed by `history info` (https://github.com/rpm-software-management/dnf/pull/2240)~
- Check whether unsafe paths will be modified in transient transactions, as defined by the `usr_drift_protected_paths` option (https://github.com/rpm-software-management/dnf/pull/2244)

Additionally:

- ~history list --json` lists transaction persistence~
- ~Adds a migration mechanism to the transaction history DB. Using a database with a newer minor version than the latest version is allowed, but using a database with a new major version throws an error.~
- Changes `cmd_requires_privileges` to return `false` when storing a transaction via `--store`, which allows `--store` without root privileges or obtaining a write lock on the system repo, and allows it on bootc systems without unlocking `/usr`

~Marking as draft for now; I may try to move some functionality out of main.cpp and into libdnf5 or Context.~

Some commits were `Assisted-by Claude Opus 4.6`.

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -240,6 +240,12 @@ void Context::Impl::update_repo_metadata_from_advisory_options(
 }
 
 void Context::Impl::load_repos(bool load_system, bool load_available) {
+    if (persistence == libdnf5::base::TransactionPersistence::TRANSIENT &&
+        !base.get_config().get_usr_drift_protected_paths_option().get_value().empty()) {
+        base.get_config().get_optional_metadata_types_option().add_item(
+            libdnf5::Option::Priority::RUNTIME, libdnf5::METADATA_TYPE_FILELISTS);
+    }
+
     if (load_available) {
         libdnf5::repo::RepoQuery repos(base);
         repos.filter_enabled(true);
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -622,6 +622,11 @@ bool Context::Impl::cmd_requires_privileges() const {
         return false;
     }
 
+    // when storing a transaction, system should not be modified
+    if (!get_transaction_store_path().empty()) {
+        return false;
+    }
+
     // otherwise, transactional cmds with store option defined are expected to modify the system
     auto it_store = std::find_if(
         all_cmd_args.begin(), all_cmd_args.end(), [](auto arg) { return arg->get_long_name() == "store"; });
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -508,6 +508,8 @@ void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
         transaction.set_comment(comment);
     }
 
+    transaction.set_persistence(persistence);
+
     auto result = transaction.run();
     if (result != libdnf5::base::Transaction::TransactionRunResult::SUCCESS) {
         print_error(libdnf5::utils::sformat(
@@ -662,6 +664,14 @@ const char * Context::get_comment() const noexcept {
 
 void Context::set_comment(const char * comment) noexcept {
     p_impl->set_comment(comment);
+}
+
+libdnf5::base::TransactionPersistence Context::get_persistence() const noexcept {
+    return p_impl->get_persistence();
+}
+
+void Context::set_persistence(libdnf5::base::TransactionPersistence persistence) noexcept {
+    p_impl->set_persistence(persistence);
 }
 
 std::string Context::get_cmdline() {
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -524,12 +524,12 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     They are protected via Obsoletes as well as user/plugin removals.
 
+    An entry prefixed with ``glob:``, such as ``glob:/path/to/dir/*.conf``,
+    expands to all matching files and reads one package per line from each.
+
     Default: ``dnf5,glob:/etc/dnf/protected.d/*.conf``.
 
     .. NOTE::
-       Any packages which should be protected can do so by including a file in ``/etc/dnf/protected.d``
-       with their  package name in it.
-
        DNF5 will protect also the package corresponding to the running version of the kernel. See also
        :ref:`protect_running_kernel <protect_running_kernel_options-label>` option.
 
@@ -680,9 +680,6 @@ repository configuration file should aside from repo ID consists of baseurl, met
     List of paths that are likely to cause problems when their contents drift
     with respect to ``/usr``, e.g. ``/etc/pam.d/*``. If a transient transaction
     would modify these paths, DNF5 aborts the operation and prints an error.
-    Supports globs. A list of paths can be protected by creating a ``.conf``
-    file in ``/etc/dnf/usr-drift-protected-paths.d/`` containing one path (or
-    glob pattern) per line.
 
     When using ``persistence=transient`` on bootc systems, a transient overlay
     is created on ``/usr``, and any changes DNF5 makes to ``/usr`` will be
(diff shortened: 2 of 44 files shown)


dnf5/context.cpp
```diff
@@ -240,6 +240,12 @@ void Context::Impl::update_repo_metadata_from_advisory_options(
 }
 
 void Context::Impl::load_repos(bool load_system, bool load_available) {
+    if (persistence == libdnf5::transaction::TransactionPersistence::TRANSIENT &&
+        !base.get_config().get_usr_drift_protected_paths_option().get_value().empty()) {
+        base.get_config().get_optional_metadata_types_option().add_item(
+            libdnf5::Option::Priority::RUNTIME, libdnf5::METADATA_TYPE_FILELISTS);
+    }
+
```
ppisar: This (automatic download of file lists in case of transient transaction and non-empty usr_drift_protected_paths) should be documented. I expect that people will complain that DNF5 is slow (read downloads file lists) on bootc systems. That way we can react "We've told you."

doc/dnf5.conf.5.rst
```diff
@@ -618,6 +636,31 @@ repository configuration file should aside from repo ID consists of baseurl, met
...
+.. _usr_drift_protected_paths_options-label:
+
+``usr_drift_protected_paths``
+    :ref:`list <list-label>`
+
+    List of paths that are likely to cause problems when their contents drift
+    with respect to ``/usr``, e.g. ``/etc/pam.d/*``. If a transient transaction
+    would modify these paths, DNF5 aborts the operation and prints an error.
+    Supports globs. Defaults to
+    ``glob:/etc/dnf/usr-drift-protected-paths.d/*.conf``, so a list of paths can
```
ppisar: This duplicates information in the last Default: paragraph. Can you simplify it?

libdnf5/transaction/transaction.cpp
```diff
@@ -80,6 +81,18 @@ std::string transaction_state_to_string(TransactionState state) {
...
 
 
+std::string transaction_persistence_to_string(TransactionPersistence persistence) {
+    switch (persistence) {
+        case TransactionPersistence::PERSIST:
+            return "Persist";
+        case TransactionPersistence::TRANSIENT:
+            return "Transient";
+        default:
+            return "Unknown";
```
ppisar: The literal strings needs to be internationalized. I can see you use them in JSON as well as in human-readable output. That needs to be sorted out.

ppisar (commented):
Overall good, I left few questions in-line. I did not try to undetstand the database migration. You should ask somebody else to review that piece of code.

doc/dnf5.conf.5.rst
evan-goode: Better?

dnf5/context.cpp
evan-goode: True, I added a sentence to the `usr_drift_protected_path` option documentation.

libdnf5/transaction/transaction.cpp
evan-goode: Good catch, the whole `history list` and `history info` outputs are untranslated. We might as well do that now. Done in latest commit.

doc/dnf5.conf.5.rst
ppisar: Not much. Now it reads that regardless what you have in the option, "A list of paths can be protected by creating a ``.conf``  file in ``/etc/dnf/usr-drift-protected-paths.d/``". That's not true.

If you want to to hint the users how use the globs, write something like:

~~~~
Supports globs. A list of paths can be protected by creating a file matching
the glob and writing one path (or glob pattern) per line in the file.
~~~~

Or better, since this is not specific to this option, document in general how `glob:` value works at list-type options, probably in a paragraph between DESCRIPTION and [MAIN] OPTIONS headings in this manual.

evan-goode: True. I updated it again.

> Or better, since this is not specific to this option, document in general how glob: value works at list-type options, probably in a paragraph between DESCRIPTION and [MAIN] OPTIONS headings in this manual.

Since it's currently only this and the `protected_paths` option, I left the note duplicated in both places.

evan-goode:
Rebased and marked ready for review.

libdnf5/transaction/db/sql/create_tables.sql
```diff
@@ -19,6 +19,7 @@ R"**(
         "description" TEXT,                               /* A description of the transaction (e.g. the CLI command being executed) */
         "comment" TEXT,                                   /* An arbitrary comment */
         "state_id" INTEGER,                               /* (enum) */
+        "persistence" INTEGER,                            /* (enum) 0 = Unknown, 1 = Persist, 2 = Transient */
```
m-blaha: Since this column maps to `TransactionPersistence` enum with possible values 0,1,2, shouldn't it be defined as `NOT NULL DEFAULT 0`? Currently there are two values - `0` and `NULL` that (kind of coincidentally, due to sqlite3_column_int behavior) map to `TransactionPersistence::UNKNOWN`.

libdnf5/transaction/db/sql/migrate_1_1_to_1_2.sql
```diff
@@ -0,0 +1,9 @@
+R"**(
+    BEGIN TRANSACTION;
+
+    ALTER TABLE "trans" ADD COLUMN "persistence" INTEGER;
```
jrohel: @m-blaha made a good point about `NOT NULL DEFAULT 0`. It should also be handled in the migration code:
`ALTER TABLE "trans" ADD COLUMN "persistence" INTEGER NOT NULL DEFAULT 0;`

libdnf5/transaction/db/sql/create_tables.sql
evan-goode: Yes that's better, thanks!

libdnf5/transaction/db/sql/migrate_1_1_to_1_2.sql
evan-goode: Done!

Conan-Kudo:
Why isn't this externalized as plugins? This shouldn't be part of DNF itself. Some of the changes that are going on here don't make that much sense to me either. What's the point in recording transient transactions, for example? They aren't permanent, so they might as well not be recorded at all.

Conan-Kudo:
Having some of this functionality built-in will also conflict with the txnupd plugin, which does this for btrfs-based atomic systems. It seems to be very flawed to integrate it into the core like this.

evan-goode:
> Why isn't this externalized as plugins? This shouldn't be part of DNF itself. Some of the changes that are going on here don't make that much sense to me either. What's the point in recording transient transactions, for example? They aren't permanent, so they might as well not be recorded at all.

To paraphrase from last month's DNF community meeting, I'm sympathetic with this take. Indeed there may come other technologies similar to bootc that require a similar but different setup process to get DNF to work with them, and having these interop bits be pluggable would be nice. However, the plugin API is missing a lot of necessary things today, a plugin couldn't for example add a `--transient` option to all the transaction-running commands. So with this PR I would rather just try to not introduce too much new API surface so it's easier to move this functionality to a plugin in the future if we want.

I ripped out the "persistence" entry in the history database (along with the DB migration logic, though we'll surely need some of that later for something else). The history DB is under `/usr` so entries made by transient transactions would disappear on reboot anyway.

The helper functions in `libdnf5::utils::bootc` are OK IMO, two of them already existed.

The new `persistence` and `usr_drift_protected_paths` options are generically named, don't mention bootc, so they could conceivably be used by other plugins/"transience providers" in the future.

ppisar (approved):
Good. The Rawhide test failure is unrelated and the ELN test failures are an infrastructure failure.

PR #2728, opened 2026-05-06
Update system state when installing a package already installed via rpm

### LINKED ISSUE #1787: Update system state when installing a package that was already installed via rpm
There is already a test scenario "Install a package that was already installed via rpm" in install-remove.feature

When a package is installed via rpm, and then again via dnf, the system state should contain the information about the package and it should have reason "user".

m-blaha:
When a package is installed via rpm (outside dnf) and then the user runs
`dnf install <package>`, the system state was not updated. The package
remained with EXTERNAL_USER reason.

Now when an ALREADY_INSTALLED package has EXTERNAL_USER reason, a
REASON_CHANGE action is added to set the reason to USER.

Closes: https://github.com/rpm-software-management/dnf5/issues/1787

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/libdnf5/base/goal.cpp
+++ b/libdnf5/base/goal.cpp
@@ -1527,6 +1527,13 @@ std::pair<GoalProblem, libdnf5::solv::IdQueue> Goal::Impl::add_install_to_goal(
             spec,
             {pool.get_nevra(package_id)},
             libdnf5::Logger::Level::WARNING);
+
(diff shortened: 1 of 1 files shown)


evan-goode (approved):
LGTM, thanks!

I considered whether reinstall should also change the reason, but I decided probably not, since reinstall currently never changes the reason. Anyway we already have `dnf5 mark`.

PR #2731, opened 2026-05-11
Repo info: add unique NEVRAs count

kontura:
Port https://github.com/rpm-software-management/dnf/pull/2324 to dnf5.

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1880

### DIFF
--- a/include/libdnf5-cli/output/interfaces/repo.hpp
+++ b/include/libdnf5-cli/output/interfaces/repo.hpp
@@ -69,6 +69,7 @@ public:
     virtual uint64_t get_size() const = 0;
     virtual uint64_t get_pkgs() const = 0;
     virtual uint64_t get_available_pkgs() const = 0;
+    virtual uint64_t get_unique_nevras() const = 0;
     virtual std::vector<std::string> get_mirrors() const = 0;
 };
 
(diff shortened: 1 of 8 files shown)


include/libdnf5-cli/output/interfaces/repo.hpp
```diff
@@ -69,6 +69,7 @@ class IRepoInfo {
     virtual uint64_t get_size() const = 0;
     virtual uint64_t get_pkgs() const = 0;
     virtual uint64_t get_available_pkgs() const = 0;
+    virtual uint64_t get_unique_nevras() const = 0;
```
jrohel, line 72: Adding a virtual method to a public interface class causes a public ABI break.

jrohel (commented):
The code is simple and looks OK functionally. However, I spot an ABI break at first glance. Two approaches come to mind:

1. We agree on a point when we will introduce breaking changes. But generally speaking, we should minimize breaking the library's ABI.

2. We create a new interface, for example `IRepoInfo2_1`, which will inherit from `IRepoInfo` and contain the newly added method. This would introduce interface versioning. Usage of the passed interface object (e.g., in libdnf5-cli/output/repo_info.cpp) would then require checking which class it inherits from to determine which methods are available.

This way, we would remain backward compatible. Applications compiled with an older version of the library (using the original `IRepoInfo`) would continue to work with the new library. (Of course, in this case, the new library wouldn't report the number of unique NEVRs, because it wouldn't receive them from the old application using the old interface).

Issue #2200, opened 2025-04-16
option to use local time timestamps


realsobek:
dnf-4.23.0-1.fc40.noarch uses local time timestamps (at the moment +0200 due to summer time) in `dnf history` and `/var/log/dnf.log`.
dnf5-5.2.12.0-2.fc42.x86_64 uses UTC time timestamps in `dnf history` and `/var/log/dnf5.log`. In example below you see the 2 hour difference. The difference makes it harder to trace actions, because all other apps' logs and timestamps use local time.
Because RPM scriptlet action's output gets truncated I `grep /var/log/dnf.log` after automated system installation is done.

I cannot find an option to force dnf5 to use local time. Is there no such option? Something like `dnf history list --uselocaltime` would be handy. 

I found an issue was created for upstream tests https://github.com/rpm-software-management/dnf5/issues/900 reply:
> It's normal that time formatting functions and thus applications honor a localtime zone.

In https://github.com/rpm-software-management/ci-dnf-stack/issues/1386 the same issue got reply:
> This got changed in https://github.com/rpm-software-management/dnf5/pull/844 when the `Issued` time was unified with the rest of dates printed by dnf5 (like in `repo info` or `history list`). The times are now always in UTC.

I do not know the big picture: brainstorming: Add dnf5 option to force UTC time output? Update the upstream tests to cope with local time? 
If no change is intended, please, add a note to https://dnf5.readthedocs.io/en/stable/changes_from_dnf4.7.html that dnf5 output is unified on UTC.

Example with Fedora 42 and dnf5:
```
# timedatectl
               Local time: Wed 2025-04-16 13:40:20 CEST
           Universal time: Wed 2025-04-16 11:40:20 UTC
                 RTC time: Wed 2025-04-16 11:40:20
                Time zone: Europe/Berlin (CEST, +0200)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no
# date
Wed Apr 16 01:40:18 PM CEST 2025
# dnf install -y leafpad
...
# dnf history list
 ID Command line                        Date and time       Action(s) Altered
255 dnf install -y leafpad              2025-04-16 11:40:28                 1
...
# vim /var/log/dnf5.log
...
2025-04-16T11:40:25+0000 [8393] INFO --- DNF5 launched with arguments: "dnf install -y leafpad" ---
...
```

Kudos for the timestamp formatting:
2025-04-16T11:40:25
I like it much better than `journalctl`'s output:
[...]

PR #2733, opened 2026-05-12
reposync: add `--min-buildtime <YYYY-MM-DD>` option

kontura:
dnf5 counterpart to https://github.com/rpm-software-management/dnf-plugins-core/pull/615

test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1881


include/libdnf5-cli/session.hpp
```diff
@@ -207,6 +207,28 @@ class LIBDNF_CLI_API BoolOption : public Option {
...
+    explicit DateOption(
+        libdnf5::cli::session::Command & command,
+        const std::string & long_name,
+        const std::string & desc,
+        const std::string & help);
+
+    ~DateOption();
+
+    /// @return Parsed value.
+    /// @since 5.0
```
ppisar: Update "since" comment to the current version. DateOption is added now.

include/libdnf5-cli/session.hpp
```diff
@@ -207,6 +207,28 @@ class LIBDNF_CLI_API BoolOption : public Option {
 };
 
 
+class LIBDNF_CLI_API DateOption : public Option {
+public:
+    explicit DateOption(
+        libdnf5::cli::session::Command & command,
+        const std::string & long_name,
```
ppisar, line 217: Can we add `char short_name` agument after the long_name argument to support a short option in case somebody needs it? This is public API so we should make it perfect from the beginning.

include/libdnf5-cli/session.hpp
```diff
@@ -207,6 +207,28 @@ class LIBDNF_CLI_API BoolOption : public Option {
 };
 
 
+class LIBDNF_CLI_API DateOption : public Option {
```
ppisar, line 213: This enhances public API. I think it would be good to:

* increase minor version number of DNF5
* add a documentation for this class. Like what's the purpose, what date format it expects, that argument parser will throw an exception if it's invalid etc.

dnf5-plugins/reposync_plugin/reposync.cpp
```diff
@@ -236,6 +239,10 @@ ReposyncCommand::download_list_type ReposyncCommand::get_packages_list(const lib
         limit_to_latest(query);
     }
 
+    if (min_buildtime_option->get_value() > 0) {
```
ppisar: get_value() returns number of seconds since UNIX epoch. 0 What if the user supplies 1970-01-01, i.e. 0 seconds? Or 1969-12-31, i.e. -(24*60*60) seconds?

We should use a different way how the recognizes that min-buildtime option was given.

doc/dnf5_plugins/reposync.8.rst
```diff
@@ -72,6 +72,10 @@ Options
     Specifies the root path where downloaded metadata files are stored. If not
     specified, it defaults to the value of ``--destdir``.
 
+``--min-buildtime <YYYY-MM-DD>``
```
ppisar, line 75: In my opinion, the name of the option does not match how it behaves: When I see that name, I think it means the package build time is greater or equal of the given value. But the code and the documentation does sharp, noninclusive is greater than.

Can we change the name (e.g. `--build-after`), or the meaning?

include/libdnf5-cli/session.hpp
kontura: Added.

dnf5-plugins/reposync_plugin/reposync.cpp
kontura: I have updated it to explicitly check if the argument was used.

doc/dnf5_plugins/reposync.8.rst
kontura: Good point, I have updated the docs but I think the code was already effectively doing greater or equal.
The `PackageQuery::filter_recent` doesn't include the element if its buildtime is smaller than provided timestamp.

It is also further complicated by the fact the we provide only a date so the time from the option is 00:00. I believe that regardless of used operator `<`/`<=` if the package was built any other second except the first (time 00:00) it will be included, based on that I am not sure `--build-after` is good.

Anyway unless I got confused now it should behave as expected and according to the docs.

ppisar: You are right with PackageQuery::filter_recent(). When I saw at for the first time, I did not notice that the filter removes packages. So the implementation is fine.

Regarding the 00:00 time, it would be great to notice it in the documentation, like:

~~~~
Download only packages with buildtime newer or equal to YYYY-MM-DD 00:00:00 UTC.
~~~~

But according to my tests the option value is parsed with respect to a local time zone. That means 
--min-buildtime 1970-01-02 in Europe/Prague zone is interpreted as 1970-01-01 23:00:00 UTC. If this is a feature, the documentation should read:

~~~~
Download only packages with buildtime newer or equal to YYYY-MM-DD 00:00:00 in your local time zone.
~~~~

I also discovered that --min-buildtime 1970-01-01 in Europe/Prague misbehaves. It's correctly parsed as -3600. But when PackageQuery::filter_recent() is called, there is this type cast:

~~~~
    const unsigned long long time_long = static_cast<unsigned long long>(timestamp);
~~~~

which misinterprets timestamp=-3600  as time_long=18446744073709548016. This bug could be addressed in a separate pull request.

kontura: I see, I am not a fan of it using local time, I have added an extra commit that switches the `DateOption` to UTC and enhances the docs.

However if you prefer the local time I can keep it.

> there is this type cast

I have also noticed that, I would prefer to fix that separately.

ppisar: The best approach is parsing and formatting times in the same time zone. Now we format everything in UTC, so parsing in UTC is good.

Later when implementing #2200 (formatting in local zone), we will also change the parsing.

ppisar (approved):
Good.

PR #2735, opened 2026-05-13
 libdnf5 Plugin-to-App Communication

### LINKED ISSUE #2587: libdnf5 plugin is missing API for sending error/warning/info messages to client applications
We (CSI  team: subscription-manager and RHSM) are working on libdnf5 plugins for future major version of RHEL. We implement our libdnf5 and dnf5 plugins in this repository:

https://github.com/candlepin/rhsm-dnf5-plugins

We will have to implement replacement of current `subscription-manager` dnf4 plugin, which prints error/warning/info messages to terminal, when e.g. system is not registered, entitlement certificate is not present or expired, etc. We need to message to users the reason, why they do not have access to content on RHEL.

We already write error/warning/info/log messages to dnf log using provided API, but such critical information like system is nor registered has to be printed to console or displayed in the GUI to have a functionality parity with existing dnf4 plugin. It is something must have in a next major version of RHEL. Something like this:

```console
[root@rhel10 ~]# dnf -y install zsh
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use "rhc" or "subscription-manager" to register.

No match for argument: zsh
Error: Unable to find a match: zsh
```

for the case, when system is not registered.

We could use some quick and dirty workaround and write such messages directly to `stderr`, but there is is a high risk that we will corrupt output of client application like `dnf` or the GUI application will not be able to display such message.  Writing anything to terminal using e.g. `std::cerr << "Error: some error message << std::endl;" would not look great, when `dnf` was started with `--json` CLI option.

jrohel:
Closes: https://github.com/rpm-software-management/dnf5/issues/2587

### DIFF
--- a/include/libdnf5/base/interaction_callbacks.hpp
+++ b/include/libdnf5/base/interaction_callbacks.hpp
@@ -1,0 +1,126 @@
+// Copyright Contributors to the DNF5 project.
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#ifndef LIBDNF5_BASE_INTERACTION_CALLBACKS_HPP
+#define LIBDNF5_BASE_INTERACTION_CALLBACKS_HPP
+
+#include "base_weak.hpp"
+
+#include "libdnf5/base/text_validator.hpp"
+#include "libdnf5/common/message.hpp"
+#include "libdnf5/defs.h"
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace libdnf5::base {
+
+/// Return value constants for InteractionCallbacks confirm(), choice(), and input_text() methods.
+#ifdef SWIG
+/* clang-format off */
+#define constexpr %constant
+/* clang-format on */
+#endif
+constexpr int32_t ANSWER_YES = -1;      ///< User confirmed / a string entered
+constexpr int32_t ANSWER_NO = -2;       ///< User denied / undefined string
+constexpr int32_t ANSWER_DEFAULT = -3;  ///< Use the caller-supplied default
+#ifdef SWIG
+#undef constexpr
+#endif
+
+/// Base class for interaction callbacks.
+/// To implement callbacks, inherit from this class and override virtual methods.
+///
+/// These callbacks allow libdnf5 plugins to interact with the user through the application.
+/// The application implements these methods to provide user interface
+/// for messages, confirmations, choices, text input, and progress.
+class LIBDNF_API InteractionCallbacks {
+public:
+    enum class MessageLevel : unsigned int { ERROR = 1, WARNING, NOTICE, INFO };
+    enum class ProgressState { NEW, UPDATE, END_OK, END_ERROR, INFO, WARNING, ERROR };
+
+    explicit InteractionCallbacks();
+    InteractionCallbacks(const InteractionCallbacks &) = delete;
+    InteractionCallbacks(InteractionCallbacks &&) = delete;
+    virtual ~InteractionCallbacks();
+
+    InteractionCallbacks & operator=(const InteractionCallbacks &) = delete;
+    InteractionCallbacks & operator=(InteractionCallbacks &&) = delete;
+
+protected:
+    /// @return The `Base` object to which this object belongs.
+    libdnf5::BaseWeakPtr get_base() const;
+
+    /// Override to display a message to the user. Called via Base::message().
+    /// @param level The message level (ERROR, WARNING, NOTICE, INFO)
+    /// @param message The message to display
+    virtual void message(MessageLevel level, const libdnf5::Message & msg);
+
+    /// Override to ask the user to confirm an action. Called via Base::confirm().
+    /// The default implementation returns `ANSWER_NO` if `assumeno` is set, `ANSWER_YES` if `assumeyes`
+    /// is set, and `ANSWER_DEFAULT` otherwise.
+    /// @param message The confirmation message describing the action
+    /// @param default_answer The answer preferred by the caller, used when neither `assumeyes`
+    ///        nor `assumeno` configuration option is active
+    /// @return `ANSWER_YES` (-1) to confirm, `ANSWER_NO` (-2) to deny, `ANSWER_DEFAULT` (-3) to use caller default
+    virtual int32_t confirm(const libdnf5::Message & msg, bool default_answer);
+
+    /// Override to ask the user to select an option. Called via Base::choice().
+    /// @param msg The message describing the choice
+    /// @param options Vector of pointers to available option messages (for translation support)
+    /// @param default_option Index of the option preferred by the caller (0-based)
+    /// @return Index of the selected choice (0-based), or `ANSWER_DEFAULT` (-3) to use caller default
+    virtual int32_t choice(
+        const libdnf5::Message & msg, const std::vector<libdnf5::Message *> & options, int32_t default_option);
+
+    /// Override to ask the user for text input. Called via Base::input_text().
+    /// The implementation should repeatedly prompt the user until valid input is provided
+    /// or the user cancels. If a validator is provided, call validator->validate() on each
+    /// input and display any error message before prompting again.
+    /// @param out_text Output parameter - filled with user input when returning `ANSWER_YES` or `ANSWER_NO`
+    /// @param msg The message describing what input is needed
+    /// @param default_text Optional pre-filled text offered to the user, or nullptr (default is undefined text).
+    /// @param validator Optional validator for checking input validity, or nullptr to accept any input.
+    /// @return `ANSWER_YES` (-1) if a string was entered (out_text filled; may be empty/zero-length),
+    ///         `ANSWER_NO` (-2) if no string was returned (out_text unchanged; analogous to NULL char*),
+    ///         `ANSWER_DEFAULT` (-3) if the caller should use default_text (out_text unchanged),
+    ///         `ANSWER_ABORT` (-4) to abort the operation (out_text unchanged)
+    virtual int32_t input_text(
+        std::string & out_text,
+        const libdnf5::Message & msg,
+        const char * default_text,
+        libdnf5::base::TextValidator * validator);
+
+    /// Override to report progress. Called via Base::progress().
+    /// On NEW the implementation allocates a handle and returns it; the caller
+    /// passes the same handle back in all subsequent calls for that entry.
+    /// The handle is an opaque integer token whose meaning is defined solely
+    /// by the implementation; the caller treats it as an integer number.
+    /// @param handle  For NEW: ignored. For other states: the handle returned
+    ///                by the corresponding NEW call.
+    /// @param state   The progress state
+    /// @param msg     Optional description of current activity, or nullptr
+    /// @param count   Number of completed items
+    /// @param total   Total number of items
+    /// @return For NEW: the handle of the new progress entry.
+    ///         For other states: >= 0 to continue, < 0 to request cancellation.
+    /// The default implementation returns 0 for NEW and returns the handle
+    /// unchanged for all other states without performing any action.
+    virtual int progress(int handle, ProgressState state, const libdnf5::Message * msg, int64_t count, int64_t total);
+
+private:
+    friend class ::libdnf5::Base;
+
+    /// Called by Base::set_interaction_callbacks() to associate this object with a Base instance.
+    void register_base(Base & base);
+
+    /// Weak pointer to the Base object that owns this callbacks instance.
+    libdnf5::BaseWeakPtr base;
+};
+
+}  // namespace libdnf5::base
+
+#endif  // LIBDNF5_BASE_INTERACTION_CALLBACKS_HPP
--- a/include/libdnf5/base/interaction_callbacks.hpp
+++ b/include/libdnf5/base/interaction_callbacks.hpp
@@ -27,6 +27,7 @@ namespace libdnf5::base {
 constexpr int32_t ANSWER_YES = -1;      ///< User confirmed / a string entered
 constexpr int32_t ANSWER_NO = -2;       ///< User denied / undefined string
 constexpr int32_t ANSWER_DEFAULT = -3;  ///< Use the caller-supplied default
+constexpr int32_t ANSWER_ABORT = -4;    ///< Abort the operation (user request or I/O error)
 #ifdef SWIG
 #undef constexpr
 #endif
@@ -65,14 +66,16 @@ protected:
     /// @param message The confirmation message describing the action
     /// @param default_answer The answer preferred by the caller, used when neither `assumeyes`
     ///        nor `assumeno` configuration option is active
-    /// @return `ANSWER_YES` (-1) to confirm, `ANSWER_NO` (-2) to deny, `ANSWER_DEFAULT` (-3) to use caller default
+    /// @return `ANSWER_YES` (-1) to confirm, `ANSWER_NO` (-2) to deny, `ANSWER_DEFAULT` (-3) to use caller default,
+    ///         `ANSWER_ABORT` (-4) to abort the operation
     virtual int32_t confirm(const libdnf5::Message & msg, bool default_answer);
 
     /// Override to ask the user to select an option. Called via Base::choice().
     /// @param msg The message describing the choice
     /// @param options Vector of pointers to available option messages (for translation support)
     /// @param default_option Index of the option preferred by the caller (0-based)
-    /// @return Index of the selected choice (0-based), or `ANSWER_DEFAULT` (-3) to use caller default
+    /// @return Index of the selected choice (0-based), `ANSWER_DEFAULT` (-3) to use caller default,
+    ///         or `ANSWER_ABORT` (-4) to abort the operation
     virtual int32_t choice(
         const libdnf5::Message & msg, const std::vector<libdnf5::Message *> & options, int32_t default_option);
 
--- a/libdnf5/base/interaction_callbacks.cpp
+++ b/libdnf5/base/interaction_callbacks.cpp
@@ -1,0 +1,72 @@
+// Copyright Contributors to the DNF5 project.
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include "libdnf5/base/interaction_callbacks.hpp"
+
+#include "libdnf5/base/base.hpp"
+
+namespace libdnf5::base {
+
+InteractionCallbacks::InteractionCallbacks() = default;
+
+
+InteractionCallbacks::~InteractionCallbacks() = default;
+
+
+BaseWeakPtr InteractionCallbacks::get_base() const {
+    return base;
+}
+
+
+void InteractionCallbacks::register_base(Base & base) {
+    this->base = base.get_weak_ptr();
+}
+
+
+void InteractionCallbacks::message([[maybe_unused]] MessageLevel level, [[maybe_unused]] const libdnf5::Message & msg) {
+}
+
+
+int32_t InteractionCallbacks::confirm(
+    [[maybe_unused]] const libdnf5::Message & msg, [[maybe_unused]] bool default_answer) {
+    auto & config = base->get_config();
+    if (config.get_assumeno_option().get_value()) {
+        return ANSWER_NO;
+    }
+    if (config.get_assumeyes_option().get_value()) {
+        return ANSWER_YES;
+    }
+    return ANSWER_DEFAULT;
+}
+
+
+int32_t InteractionCallbacks::choice(
+    [[maybe_unused]] const libdnf5::Message & msg,
+    [[maybe_unused]] const std::vector<libdnf5::Message *> & options,
+    [[maybe_unused]] int32_t default_option) {
+    return ANSWER_DEFAULT;
+}
+
+
+int InteractionCallbacks::progress(
+    int handle,
+    [[maybe_unused]] ProgressState state,
+    [[maybe_unused]] const libdnf5::Message * msg,
+    [[maybe_unused]] int64_t count,
+    [[maybe_unused]] int64_t total) {
+    if (state == ProgressState::NEW) {
+        return 0;
+    }
+    return handle;
+}
+
+
+int32_t InteractionCallbacks::input_text(
+    [[maybe_unused]] std::string & out_text,
+    [[maybe_unused]] const libdnf5::Message & msg,
+    [[maybe_unused]] const char * default_text,
+    [[maybe_unused]] libdnf5::base::TextValidator * validator) {
+    return ANSWER_DEFAULT;
+}
+
+}  // namespace libdnf5::base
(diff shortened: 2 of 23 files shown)


libdnf5/base/interaction_callbacks.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+InteractionCallbacks::~InteractionCallbacks() = default;
+
+
+void InteractionCallbacks::message(
+    [[maybe_unused]] Level level, [[maybe_unused]] const libdnf5::Message & removal_info) {}
+
+
+bool InteractionCallbacks::confirm([[maybe_unused]] const libdnf5::Message & removal_info) {
+    return true;
```
m-blaha: Wouldn't `return false;` be a safer option here? This "autoconfirm" behavior kind of contradicts the dnf5 CLI default behavior, where `assumeno` takes precedence over `assumeyes` and also `defaultyes` config option defaults to false resulting in the default answer for the user confirmation being `no`.

libdnf5/base/interaction_callbacks.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace libdnf5::base {
+
+InteractionCallbacks::InteractionCallbacks() = default;
+
+InteractionCallbacks::~InteractionCallbacks() = default;
+
+
+void InteractionCallbacks::message(
+    [[maybe_unused]] Level level, [[maybe_unused]] const libdnf5::Message & removal_info) {}
```
m-blaha:
```suggestion
    [[maybe_unused]] Level level, [[maybe_unused]] const libdnf5::Message & message) {}
```
Also in `InteractionCallbacks::confirm()`.

include/libdnf5/base/interaction_callbacks.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+/// Base class for interaction callbacks.
+/// To implement callbacks, inherit from this class and override virtual methods.
+///
+/// These callbacks allow libdnf5 plugins to interact with the user through the
+/// application. The application implements these methods to provide user interface
+/// for messages, confirmations, choices, and text input.
+class LIBDNF_API InteractionCallbacks {
+public:
+    enum class Level : unsigned int { ERROR = 1, WARNING, NOTICE, INFO };
```
m-blaha: Should we re-use already existing `Logger::Level` enum?

libdnf5/base/interaction_callbacks.cpp
jrohel: I was thinking about that too. This code is used in case the library user (the derived class) doesn't implement the `confirm` method. I figured that in such a case, the user probably doesn't want to handle prompts. However, if the default is set to false, the program might silently halt on some prompts. But truth be told, I don't have a crystal ball. Instead of guessing whether `true` or `false` is a better default, it might be better to **force the library user to state their intent (enforce its implementation -> make the method pure virtual).**

The question is whether to let `choice` and `input_text` return default values, or to also make them pure virtual methods and thus force the library user to implement them.

libdnf5/base/interaction_callbacks.cpp
jrohel: hanks for noticing. That copy-paste mistake slipped in while I was testing.

include/libdnf5/base/interaction_callbacks.hpp
jrohel: Originally, I was only considering `ERROR`, `WARNING`, and `INFO`. Then I added `NOTICE`, but I'm not sure if that's overkill. Using `Logger::Level` would introduce even more levels. Anyway, I'll think about it some more, thanks for the feedback.

libdnf5/base/interaction_callbacks.cpp
ppisar: Virtual methods makes interfaces inextensible with respect to preserving ABI. I'd rather keep libdnf5 to provide a default implementation with a default answer false. The application can override the method if it has better default.

I don't understand the comment that default false could make a program to halt silently. Did you mean to block indefinitely, or to exit without printing the prompt?

I guess first user of the application would notice and report it to the application. So I don't it's a big problem.

Another approach would be using good old pointer (or you can use fancy std::optional<> if C++ supports it on function pointers) and if the the pointer remains default NULL (i.e. an application did not registered a callback), libdnf5 can provide a fallback implementation which will print the prompt and then the default answer as messages (e.g. using an application callback for printing messages).

jrohel: @ppisar 
> Virtual methods makes interfaces inextensible with respect to preserving ABI. I'd rather keep libdnf5 to provide a default implementation with a default answer false. The application can override the method if it has better default.

Yes, adding a virtual method to a class in C++ breaks the ABI. However, whether the method is pure virtual or provides a default implementation makes no difference - the ABI issue remains exactly the same.

I have previously considered alternative approaches, such as using plain C callback functions or, ideally, `std::function` instead of interface classes with virtual methods. However, SWIG support has always been the bottleneck. SWIG natively supports interfaces with virtual methods via its "directors" feature. Supporting C callback functions is more complicated, and `std::function` is even worse.

At one point, I even considered having a dedicated class with a single virtual method for each callback. Each class would represent exactly one callback function, making it very easy to expose to other languages using SWIG directors. Adding a new callback would simply mean introducing a new class, leaving the original classes untouched and their ABI intact.

However, since we already had callback classes with multiple methods, I continued down that path. To address the ABI issue when adding a new virtual method, our plan is to introduce a new "versioned" class that inherits from the original one and appends the new methods. This way, legacy code will continue to work with the old class, while new code can inherit from the new, extended version.

---
> I don't understand the comment that default false could make a program to halt silently. Did you mean to block indefinitely, or to exit without printing the prompt?

Take this example from the expired-pgp-keys plugin:
```
if (callbacks && !callbacks->confirm(KeyExpiryInfoMessage(key_info, key_timestamp))) {
    // User declined removing this key.
    continue;
}
```
Here, the desired default behavior is `true`. If the library user does not register a callback class at all, the plugin will proceed to remove the expired keys. (In the original code without InteractionCallback, the same situation occurs if no enabled repository is found).

However, if the library user does register their callback class but leaves the confirm method unimplemented, and confirm has a default implementation returning `false`, the plugin's behavior silently changes and it stops removing expired keys.

For other plugins, this kind of silent change could easily cause the application to stop doing what it is supposed to do, leaving the user with no idea why.

Therefore, keeping the method pure virtual has a distinct advantage: it forces the library user to explicitly acknowledge that since they are installing their own custom handler, they must provide their own implementation for `confirm`.

---
> Another approach would be using good old pointer (or you can use fancy std::optional<> if C++ supports it on function pointers) and if the the pointer remains default NULL (i.e. an application did not registered a callback)

C++ is capable of many things. However, we must not forget what I mentioned in my response to the first paragraph. Both sides - the callback implementation (the library user/application) and the invocation (the libdnf5 plugin) - can be written in languages other than C++.

The requirement to project the C++ interface into other languages using SWIG (for both implementation and invocation) severely limits the features we can reasonably use. SWIG cannot do everything, and furthermore, its level of support varies across different languages (Python, Perl, Ruby).

But perhaps we just misunderstood each other. The library user is not required to register their own `InteractionCallbacks` handler. As seen in the previous expired-pgp-keys plugin example, if they don't, the plugin's native default behavior is used instead.

The case I am addressing is specifically when a user does choose to register an `InteractionCallbacks` handler - in that scenario, they should be forced to implement their own confirm method.

---
>, libdnf5 can provide a fallback implementation which will print the prompt and then the default answer as messages (e.g. using an application callback for printing messages).

As a library, libdnf5 cannot print anything directly (at most, it can log messages). It doesn't have a console; it's a library.

Furthermore, the suggestion to "e.g. using an application callback for printing messages" creates a circular argument. I am designing `InteractionCallbacks` precisely so that library plugins can send messages they want to print (via the `message` method) and ask yes/no questions (via the `confirm` method).

If a user registers their custom `InteractionCallbacks` but fails to implement `confirm` and `message` (because they weren't forced to by a pure virtual declaration), the plugin is simply out of luck. Sure, even with a pure virtual method, a library user can still write a dummy implementation that always returns `true` or `false` - but at that point, they made an explicit design choice in their code; it didn't just happen silently by default.

Otherwise, what is the alternative? Should we introduce an `InteractionCallbacksBackup` interface? And if the library user fails to define the methods in the backup interface as well, should we introduce a backup for the backup? And if they don't implement that either, then what...?

jrohel:
I had a discussion (and a brainstorming session) with @m-blaha. We came up with questions.

Going back to the beginning: while writing the text below, I formed an opinion on all questions except for point 2 regarding the default behavior of `confirm`.

Regarding point 1, I would keep the implementation of everything I have already done (`message`, `confirm`, `choice`, `text_input`). For point 3, `choice` should return that nothing was selected (cancel) by default, and for point 4, `input_text`, the default implementation should return that no input was provided (cancel).

Even so, I am still interested in your feedback on each point, and most importantly, your thoughts on point 2 `confirm` default behavior.

1. Adding virtual methods to a class breaks the ABI (currently, we handle this by introducing a new version of the class that inherits from the original one and appends the new methods). This is why, from the very beginning, I wanted to have all reasonable callback methods covered to avoid having to create multiple versions of the interface over time. I have implemented `message`, `confirm`, `choice`, and `input_text`, and I was also planning to add `progress`.

   **Right now, we already need `message` and `confirm`. However, the question arises: is it OK to implement the other methods as well, or should we stay minimalist and include the other callback methods - `choice`, `input_text` (and potentially `progress`) - only when someone actually requests them?** The advantage is that we wouldn't have to deal with questions 3. and 4. listed below right now and simpler PR. The disadvantage is that if someone does come up with a request for these callbacks later (for example, @jirihnidek wrote: "but ability to display some spinner could be useful." https://github.com/rpm-software-management/dnf5/issues/2587#issuecomment-4023681803), these questions will be back, and on top of that, we will have to deal with ABI compatibility - most likely by creating `InteractionCallbacks2_1` to follow the established style in libdnf5.

2. **Default behavior of the `confirm` callback method.** If a library user (application developer) decides to implement the `InteractionCallbacks` handler, do we want to provide a default implementation for the `confirm` method, or do we want to force them to always implement it (by making it a pure virtual method)?

    a) Default implementation (first implementation returns `true`). There were suggestions to return `false` - see the previous communication. Ultimately, during the discussion with @m-blaha, I thought of creating a default implementation for `confirm` that would respect the `assumeyes` and `assumeno` configuration. If neither is set (which is a very common case), it would either follow `defaultyes` or we could add an argument to the `confirm` method allowing the plugin to specify its preferred default (for instance, the expired-pgp-keys plugin assumes a yes response if no callback handler is found).

    b) Pure virtual confirm method. This represents the approach: "Since you decided to implement your own custom callbacks (which is recommended), you must implement `confirm`, as it is a critical callback that directly affects the resulting behavior of the library plugin."

    To me, option a) makes more sense: a default that respects `assumeyes` and `assumeno`, with an argument where the plugin passes what it prefers as a fallback default when no assume option is configured.

3. **If we include the implementation of `choice`, a question regarding its default behavior** arises, similar to `confirm`.

   a) Currently, the default implementation states that the first option from the list is selected, and if the list is empty, it returns that nothing was selected.

   b) I am thinking to always return that nothing was selected by default. We don't know what the plugin wants, so let the plugin deal with it.

4. **If we include the implementation of `input_text`, a question regarding its default behavior** arises as well.

   a) Currently, the default implementation returns information indicating that the user provided no input.

   b) It could alternatively return that the user entered an empty line, but that seems like a poor choice to me.

libdnf5/base/interaction_callbacks.cpp
ppisar: Thanks for the answers. As usually I forgot the Swig layer.

> As a library, libdnf5 cannot print anything directly (at most, it can log messages). It doesn't have a console; it's a library.
> Otherwise, what is the alternative? Should we introduce an InteractionCallbacksBackup interface? And if the library user fails to define the methods in the backup interface as well, should we introduce a backup for the backup? And if they don't implement that either, then what...?

My problem was this situation: At the beginning the library can only register a callback for `message`. So the application registers it and everybody is happy. Later version of the library offers registering a callback for `confirm` and starts using it. But then what happens to the application executed against the new library when the library attempts to issue `confirm`? Obviously the library sees that no callback for `confirm` is registered, the library will use a default answer and the program will continue. But the user never notice there was a prompt attempted by the library.

To remedy it, I suggested that the library should try to approximate the `confirm` with a `message`. That way the user will have a chance no notice that his application does not implement `confirm` and will report a bug against the application. That is based on my observation that application authors study the application interface and implements it once and then stop following new features added to the library.

Now that you explained you are going use C++ interfaces and adding new callbacks only as a new interface, it cannot happen that the application implements only half of the interface. But the problem that the application only implements old interface while the library('s plugin) wants to use new interface remains.

Of course this problem happens only when we introduce the new interface. So nothing we have to do now. But we should have a clear idea now how we will handle it when it (addition of new interface) happens.

ppisar:
From my experience with using readline, openssl, and gpgme, the application should always be able to return `no-answer` and `cancel`, besides the expected choices (like `yes` and `no` for `confirm`).

The `no-answer` should be interpreted by the library as if the user selected default. This is common when a user runs a terminal application without a standard input or with an input connected to /dev/null.

The library should signal to the application which answer is default. If the library always does that, then theoretically we don't need to support `no-answer` answer. But then the responsibility to pick and return the default answer will be on the application. That might be bothersome to the application writers.

The `cancel` should be interpreted by the library as if the user wants to abort the process. This is pretty common in GUI when a user terminates the dialogue with closing the modal window by pressing a cross icon on the window decoration instead of clicking the offered option buttons. Or in a terminal when the user presses ^D or ^C.

I wouldn't signal default for `choice` to the application as the first option. Imagine you want to have options sorted, default being the middle one:

~~~~
Refresh repository metadata on:
  (1) Monday
  (2) Tuesday
* (3) Wednesday
  (4) Thursday
  (5) Friday
  (6) Saturday
  (7) Sunday
? _
~~~~

We should instead document that the order matters and signal default to the application explicitly.

Regarding who defines default, I'm confused. I though that `asssumeyes`, `assumeno`, and `defaultyes` are DNF5 program options. Not libdnf5 options. If they are DNF5 program options, then I wouldn't harness them to the libdnf5 at all. Instead I would have libdnf5 plugins mandatory to specify the default value and the options would be consulted only in callbacks which DNF5 (or libdnf5-cli) implements. If my assumption was wrong and the options are already integral part of libdnf5, then I would first use plugin's default, then `defaultyes` as the default value. And if `asssumeyes` or `assumeno` is defined, then I would intercept the `confirm` callback before reaching to the application, and substituted the answer based on them.

include/libdnf5/base/interaction_callbacks.hpp
jirihnidek: I think that `ERROR`, `WARNING`, `NOTICE` and `INFO` would be enough for our purpose. It would be good to explain severity of each level in API documentation. It is not clear for me, what is difference between `NOTICE` and `INFO`.

jirihnidek: Oh, I can see it now in your source file. INFO messages goes to `stdout` and `NOTICE` messages goes to `stderr` and moreover `NOTICE` messages could be suppressed with `quiet`. It should be emphasized in the API doc.

jirihnidek:
Hi @jrohel,
Thank for opening this PR. It is great that you are working on this. The ability to display some spinner and progress would be nice to have, but we do not have any hard requirements for it.

I can see that the new API is used in `gpg-expired-keys` (introduced in this commit: https://github.com/rpm-software-management/dnf5/pull/2735/changes/2c44adba9a22bbb426ec17663735bbc42abc985a). Thus, it could be used as inspiration for other plugin authors. However, It would be good to extend example of libdnf5 plugin to show how to print message in plugins.

jirihnidek:
One question related to bash completion: Any message printed by libdnf5 plugin could "poison" bash completion script of `dnf` CLI and the bash completion script could stop working as expected.

Should it be duty for authors of libdnf5 plugins or will new API handle this? If it should be duty for authors of libdnf5 plugins, how should we suppress printing anything to console, when `dnf` CLI tool is triggered in bash completion script? How could we detect that `dnf` CLI is used for generating output for bash completion script?

Maybe, there some other uses cases (not only bash completion script), when plugins should not print anything to console.

ppisar:
> One question related to bash completion: Any message printed by libdnf5 plugin could "poison" bash completion script of dnf CLI and the bash completion script could stop working as expected.

Plugin does not print anything. It's the dnf5 application. When `dnf5 --complete=...` is executed for the completion purpose, the dnf5 application should register callbacks that discards any messages. So it's not a plugin's job to do it.

jrohel:
@jirihnidek 
- I simplified the interface. In the new libdnf5 version, the InteractionCallbacks API is exposed through `Base` so that libdnf5 plugins can request user interaction without holding a direct pointer to the callbacks object.

- Based on the requirements that emerged from the discussions, I have extended the interface further. Callbacks returning a query response can now return `ANSWER_ABORT`. This informs the plugin that the user wants to abort the execution of the plugin (or the application wants to, because something went wrong, e.g., displaying the query). This usually means terminating the application or the transaction in the case of a daemon. The usage of `message` and `confirm` can again be seen in the expired-pgp-keys plugin.

jirihnidek:
> Plugin does not print anything. It's the dnf5 application. When `dnf5 --complete=...` is executed for the completion purpose, the dnf5 application should register callbacks that discards any messages. So it's not a plugin's job to do it.

Yes, I know that plugin itself will not print anything, but plugin will only instruct client application to print messages somehow. It is purpose of this PR. When I wrote: "Any message printed by libdnf5 plugin ..." I only simplified my expression too much.

It is very good message that it will not be our job.

Thanks

libdnf5/base/interaction_callbacks.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+
+void InteractionCallbacks::message([[maybe_unused]] MessageLevel level, [[maybe_unused]] const libdnf5::Message & msg) {
+}
+
+
+int32_t InteractionCallbacks::confirm(
+    [[maybe_unused]] const libdnf5::Message & msg, [[maybe_unused]] bool default_answer) {
+    auto & config = base->get_config();
```
m-blaha, line 32: What happens if the main application the libdnf plugin runs inside did not set interaction_callbacks? In case the libdnf plugin called `base.confirm()`, the default instance of InteractionCallbacks is created, but without registering the base instance. Then `confirm()` method of this InteractionCallbacks instace is called (see  https://github.com/rpm-software-management/dnf5/pull/2735/changes#diff-bdbb71b590628ac0c4559721f0e5ae1028f8730a2a62d7a96f8c2d8594dc39a9R485-R490) and here it tries to dereference base, which null.

include/libdnf5/base/interaction_callbacks.hpp
```diff
@@ -0,0 +1,129 @@
...
+    InteractionCallbacks & operator=(InteractionCallbacks &&) = delete;
+
+protected:
+    /// @return The `Base` object to which this object belongs.
+    libdnf5::BaseWeakPtr get_base() const;
+
+    /// Override to display a message to the user. Called via Base::message().
+    /// @param level The message level (ERROR, WARNING, NOTICE, INFO)
+    /// @param message The message to display
+    virtual void message(MessageLevel level, const libdnf5::Message & msg);
```
jirihnidek, line 61: @mcrha had some comment to original issue here: https://github.com/rpm-software-management/dnf5/issues/2587#issuecomment-3834091433 Is this API enough for  GUI applications? Wouldn't be beneficial to have another optional argument with some constant representing message, because GUI application would like to alter some messages in some circumstances. Something like

```c++
virtual void message(MessageLevel level, const libdnf5::Message & msg, const msg_id);
```

Thus GUI applications could alter error/warn/info message based on the `msq_id` and provide some link for running another application e.g. for registration the system, verification GPG, etc.

Doing something like this based on the `msg` is impossible, because it will land in the GUI application translated.

mcrha: I'm still not sure what it should look like specifically on the gnome-software side, but I agree a `msg_id` + `msg_str` is a good thing to do for the reasons you wrote (with some caveats, see below). The `level` is needed as well. The app is not meant to filter out some messages, it should show all levels, right?

One problem can be when multiple plugins use this API. Should the "listener" know whom the origin is? As the plugins are independent, the `msg_id`-s can clash unintentionally.

It's probably no concern for the "listener" app whom the origin is, and considering the clashing problem the apps should not know of the specific plugins (to override the message one would need to know not only the code, but also the dnf5 plugin, which is not great). I mean, it seems like the `msg_id` would be good to have, but a generic "show something somehow to the user" API the listener should rather trust the invoker and show the message always verbatim.

jirihnidek: The `msg_id` was just basic "starting" idea. In real time it would be hard to use as it was proposed, because there had to be some central database of these IDs. Thus, some combination of `plugin_id` + `msg_id` would be probably needed. Not sure, what `plugin_id` should be.

I can imagine that any client application will be able to filter messages based on severity of message. E.g. a client application can display only INFO, WARN, ERROR messages.

If you will want to show messages only as plugins will send these messages, then I do not have any problem with that.

libdnf5/base/interaction_callbacks.cpp
jrohel: Good catch. The problem is with the lazy initialization in the facade methods. Specifically, the issue is with the line:
`p_impl->interaction_callbacks = std::make_unique<base::InteractionCallbacks>();`
It should be:
`set_interaction_callbacks(std::make_unique<base::InteractionCallbacks>());`

I have fixed it. Thanks!

include/libdnf5/base/interaction_callbacks.hpp
jrohel: Generally speaking, anyone (any third party) can write their own libdnf5 plugin. So we cannot have a complete database of all existing libdnf5 plugins. This also creates a problem with the database and `msg_id` collisions. That is why I didn't complicate the API any further and left out the `msg_id`.

jirihnidek: > That is why I didn't complicate the API any further and left out the msg_id.

I see, I only recall some discussion with desktop team. Thus, I started this discussion.

I think that we have a conclusion and only message should be sent without anything like `msg_id`.

mcrha: Yes, I agree too. The msg_id looked like a good thing on the paper, but once it gets into the practice it turns into a "nonsense". I'm sorry about that.

m-blaha (approved):
The code looks good, thank you!

The only thing that concerns me is that the parts introducing most of the complexity currently have no consumer: progress, choice, input_text with their validators. I understand your point that adding them later would require an ABI break, or versioned subclasses (as we use in libdnf5 plugins callbacks). However, this still creates a maintenance burden for components that are not actively used.

Conan-Kudo:
Can we please get a sample plugin using these new APIs? Otherwise it's hard to tell how to use them effectively.

PR #2736, opened 2026-05-14
Invalidate offline transaction after successful online transaction

m-blaha:
- Add `OfflineTransactionState(Base&)` constructor that derives the state file path from installroot. This is just convenience layer to deduplicate some code.
- After a successful online RPM transaction (both CLI and daemon), check for a pending offline transaction and invalidate it. Removes the state file, transaction.json, and the /system-update magic symlink while preserving downloaded packages for reuse

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2467655

### DIFF
--- a/dnf5/commands/offline/offline.cpp
+++ b/dnf5/commands/offline/offline.cpp
@@ -255,6 +255,22 @@ void check_state(const libdnf5::offline::OfflineTransactionState & state) {
     }
 }
 
+void invalidate_if_not_valid(libdnf5::offline::OfflineTransactionState & state, libdnf5::Base & base) {
+    if (state.is_pending() && !state.check_rpmdb_cookie(base)) {
+        auto cmd_line = state.get_data().get_cmd_line();
+        state.invalidate();
+        if (cmd_line.empty()) {
+            throw libdnf5::cli::CommandExitError(
+                1, M_("The system has been modified since the offline transaction was prepared."));
+        }
+        throw libdnf5::cli::CommandExitError(
+            1,
+            M_("The system has been modified since the offline transaction was prepared. "
+               "To reschedule, run: {}"),
+            cmd_line);
+    }
+}
+
 void reboot(bool poweroff = false) {
     if (std::getenv("DNF_SYSTEM_UPGRADE_NO_REBOOT")) {
         if (poweroff) {
@@ -328,6 +344,7 @@ void OfflineRebootCommand::run() {
     }
 
     check_state(*state);
+    invalidate_if_not_valid(*state, ctx.get_base());
 
     auto & offline_data = state->get_data();
     if (offline_data.get_status() != libdnf5::offline::STATUS_DOWNLOAD_COMPLETE &&
@@ -465,6 +482,9 @@ void OfflineExecuteCommand::configure() {
 
 void OfflineExecuteCommand::run() {
     auto & ctx = get_context();
+
+    invalidate_if_not_valid(*state, ctx.get_base());
+
     auto & offline_data = state->get_data();
 
     const auto & system_releasever = offline_data.get_system_releasever();
@@ -723,8 +743,16 @@ void OfflineStatusCommand::run() {
     }
     check_state(*state);
 
+    const bool rpmdb_changed = state->is_pending() && !state->check_rpmdb_cookie(get_context().get_base());
+
     const auto & status = state->get_data().get_status();
-    if (status == libdnf5::offline::STATUS_DOWNLOAD_INCOMPLETE) {
+    if (rpmdb_changed) {
+        std::cout << _("The system has been modified since the offline transaction was prepared. "
+                       "The offline transaction initiated by the following command is no longer valid:")
+                  << std::endl
+                  << "  " << state->get_data().get_cmd_line() << std::endl
+                  << _("To reschedule, run the command above. To clean up, run `dnf5 offline clean`.") << std::endl;
+    } else if (status == libdnf5::offline::STATUS_DOWNLOAD_INCOMPLETE) {
         std::cout << no_transaction_message << std::endl;
     } else if (status == libdnf5::offline::STATUS_DOWNLOAD_COMPLETE || status == libdnf5::offline::STATUS_READY) {
         std::cout << _("An offline transaction was initiated by the following command:") << std::endl
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -347,8 +347,7 @@ void Context::Impl::store_offline(libdnf5::base::Transaction & transaction) {
     const auto & packages_location = offline_destdir / "packages";
     const auto & comps_location = offline_destdir / "comps";
 
-    const std::filesystem::path state_path{offline_datadir / libdnf5::offline::TRANSACTION_STATE_FILENAME};
-    libdnf5::offline::OfflineTransactionState state{state_path};
+    auto state = libdnf5::offline::OfflineTransactionState::from_base(base);
 
     auto & offline_data = state.get_data();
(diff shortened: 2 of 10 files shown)


kontura:
This would work for dnf5 and our daemon but not for anything else (other libdnf API users, rpm...).

What about if we stored the rpmdb cookie in the offline transaction state and then if we want to perform it (be it by `dnf5 offline reboot` or anything else) we check the cookie still matches?
Possibly that would also be a good place for message saying the transaction was invalidated.

m-blaha:
@kontura Great point! I was experimenting with an rpmdb cookie check to cover the gap between transaction resolution and storing the offline transaction (which mostly involves downloading packages), but had to remove it because importing a GPG key into the rpmdb during a transaction test changes the cookie. And also the system repo lock covers this time window. I didn't realize it could still be useful to store the cookie with the state file and check it before performing the reboot.

Let me experiment with that a bit.

That said, I still think this current "early warning" about an online transaction invalidating the stored one is useful - it allows the user to immediately re-run the command to reschedule it.

kontura:
> That said, I still think this current "early warning" about an online transaction invalidating the stored one is useful - it allows the user to immediately re-run the command to reschedule it.

Yea, I also like the warning.

dnf5/commands/offline/offline.cpp
```diff
@@ -723,8 +744,14 @@ void OfflineStatusCommand::run() {
     }
     check_state(*state);
 
+    const bool rpmdb_changed = state->is_pending() && !state->check_rpmdb_cookie(get_context().get_base());
+
     const auto & status = state->get_data().get_status();
-    if (status == libdnf5::offline::STATUS_DOWNLOAD_INCOMPLETE) {
+    if (rpmdb_changed || status == libdnf5::offline::STATUS_INVALIDATED) {
+        std::cout << _("The offline transaction initiated by the following command has been invalidated:") << std::endl
```
kontura: This is really a nitpick but I like how the message in the `reboot` command says: `The system has been modified since the offline transaction was prepared.` Could we somehow mention it here as well as the reason why the transaction has been invalidated?

dnf5/commands/offline/offline.cpp
```diff
@@ -465,6 +475,17 @@ void OfflineExecuteCommand::configure() {
...
 void OfflineExecuteCommand::run() {
     auto & ctx = get_context();
+
+    if (state->is_pending() && !state->check_rpmdb_cookie(ctx.get_base())) {
+        auto cmd_line = state->get_data().get_cmd_line();
+        state->invalidate();
+        throw libdnf5::cli::CommandExitError(
+            1,
+            M_("The system has been modified since the offline transaction was prepared. "
+               "To reschedule, run: {}"),
```
kontura: This is also very minor: I see this message is present twice exactly the same, would it make sense to de-duplicate it?

dnf5daemon-server/services/offline/offline.cpp
```diff
@@ -262,6 +262,13 @@ sdbus::MethodReply Offline::get_status(sdbus::MethodCall & call) {
 
     std::string state_error;
     auto state = read_transaction_state(state_error);
+    bool rpmdb_changed = false;
+    if (state && state->is_pending()) {
+        try {
+            rpmdb_changed = !state->check_rpmdb_cookie(*session.get_base());
+        } catch (const std::exception &) {
```
kontura: Would it make sense to at least log the exception here?

dnf5/commands/offline/offline.cpp
```diff
@@ -329,6 +329,16 @@ void OfflineRebootCommand::run() {
 
     check_state(*state);
 
+    if (state->is_pending() && !state->check_rpmdb_cookie(ctx.get_base())) {
```
kontura: The condition here and in `execute` command doesn't trigger if the offline transaction was invalidated by dnf5 or dnf5daemon and so it prints just: `System is not ready for offline transaction.`.
Is this intended? 

I think I would prefer the message from this condition, similarly to the `status` command.

dnf5/context.cpp
```diff
@@ -534,6 +532,17 @@ void Context::Impl::download_and_run(libdnf5::base::Transaction & transaction) {
     for (auto const & entry : transaction.get_gpg_signature_problems()) {
         print_error(entry);
     }
+
+    libdnf5::offline::OfflineTransactionState state(base);
+    if (state.is_pending()) {
+        auto cmd_line = state.get_data().get_cmd_line();
+        state.invalidate();
+        print_error(_("Warning: Pending offline transaction has been invalidated."));
```
kontura, line 540: Would it be even more user friendly to move this warning before the user confirms the transaction?
That way they could decide if they want to proceed and invalidate the offline transaction or abort.

kontura (commented):
I am thinking about the new INVALIDATED status and to be honest it feels a little bit redundant to me now.
I am not 100% sure but the way I understand it it basically means: for this offline transaction rpmdb has changed and the user was already notified/warned? 

Practically it only ensures dnf5 prints the warning just once? Personally I would be fine with dropping it and printing the warning repeatedly (maybe with slightly modified wording). 
What do you think?

Sorry for being so picky here, I do like the new functionality.

m-blaha:
@kontura Hopefully your review comments should be resolved now.

kontura:
> @kontura Hopefully your review comments should be resolved now.

Thank you!

I think the only thing left is to resolve the failing CI test. It hard-codes an offline state toml without `rpmdb_cookie` and it fails to parse.

m-blaha:
> I think the only thing left is to resolve the failing CI test. It hard-codes an offline state toml without `rpmdb_cookie` and it fails to parse.

Thanks! I added the fix to https://github.com/rpm-software-management/ci-dnf-stack/pull/1882

kontura (dismissed):

include/libdnf5/transaction/offline.hpp
```diff
@@ -120,6 +128,10 @@ class LIBDNF_API OfflineTransactionState {
     /// @param path Path to the state file (default location is /usr/lib/sysimage/libdnf5/offline/offline-transaction-state.toml).
     OfflineTransactionState(std::filesystem::path path);
 
+    /// Constructs a new OfflineTransactionState deriving the state file path
+    /// from the Base installroot configuration.
+    OfflineTransactionState(const libdnf5::Base & base);
```
kontura: I just realized this technically breaks API, doesn't it?

Before doing `libdnf5::offline::OfflineTransactionState state2{{}};` was fine but now it fails to build because its ambiguous.

m-blaha: Oh, right. I can add `explicit` to both one parameter constructors, but it still will not resolve the `libdnf5::offline::OfflineTransactionState state2{{}};` ambiguity.

m-blaha: Or if we want to keep the API, I can convert the Base-based constructor to a static factory method. That should resolve the ambiguity completely.

kontura: I don't think we would break any code by the API change but we probably shouldn't try to sneak it in.

To be honest I don't really have a preference here.

m-blaha: OK, thanks for catching! Even though the real impact would be small, it was API change anyway. So I replaced the newly introduced constructor with a factory method. This way we are only adding to the current API and no ambiguity is introduced.

kontura (approved):

include/libdnf5/transaction/offline.hpp
kontura: Thank you!

PR #2737, opened 2026-05-14
dnfdaemon: Add install_time package attribute

### LINKED ISSUE #2734: dnf5daemon: Provide install/change timestamp
Coming from:  
https://bugzilla.redhat.com/show_bug.cgi?id=2432617#c20

When looking for changes since some time, you can limit by the `since` option in the `org.rpm.dnf.v0.History` interface, which is great, but it's not known when the change was done to the caller. I'm mainly interested in the install/update time, the removals are not that interesting to me, but the information is not propagated to the caller, thus I use `buildtime` from the package metainfo, which breaks things like in the downstream bug report.

I'd like to know when the package was installed, or when the transaction(s) the history response contains ended, like `trans.get_dt_end()`, for each returned package. With that, I can use more precise times to consider whether a new update notification should or should not be shown.

_Side improvement, with the "since" tag, you can filter changelog entries before that time, leaving up to the top change log entry (in case that would filter out all changelogs). It's completely unrelated, I only noticed a very long list of the changelog entries is returned for each package, which is a waste of (D-Bus)  bandwidth from my point of view. It's an option candidate for sure._

m-blaha:
Expose the install_time attribute through the D-Bus package attribute system. This allows callers of the History.recent_changes() and Rpm.list() methods to query when a package was actually installed on the system, instead of having to use buildtime (when the RPM was built by the packager) as an inaccurate proxy.

The underlying libdnf5::rpm::Package::get_install_time() API already existed but was not wired into the daemon's attribute serialization.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2734

### DIFF
--- a/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
+++ b/dnf5daemon-server/dbus/interfaces/org.rpm.dnf.v0.rpm.Rpm.xml
@@ -42,7 +42,7 @@ along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
         Following options and filters are supported:
 
             - package_attrs: list of strings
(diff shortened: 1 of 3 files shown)


fhbash (approved):
lgtm, thanks!

PR #2738, opened 2026-05-14
filter_recent: drop time_t timestamp cast

kontura:
POSIX claims: `time_t` shall be an integer type with a width of at least 64 bits. We can compare it directly.

This fixes the comparison in case `timestamp` is negative.

### DIFF
--- a/libdnf5/rpm/package_query.cpp
+++ b/libdnf5/rpm/package_query.cpp
@@ -40,6 +40,7 @@ extern "C" {
 #include <fnmatch.h>
 
 #include <filesystem>
+#include <utility>
 
 namespace libdnf5::rpm {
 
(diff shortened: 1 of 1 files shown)


m-blaha (approved):
The CI failures seem to be caused by collision with https://github.com/rpm-software-management/dnf5/pull/2733 .

kontura:
Rebased.

ppisar (approved):
`dnf5 reposync --min-buildtime 1960-01-01` works as execpted now.

ppisar:
Something is spooky in the CI: dnf5-tests passed everywhere except of F43 (the new --min-buildtime option test failed with `Unknown argument "--min-buildtime" for command "reposync"`). Yet tmt-verbose-log files for all Fedoras are consistent with installing dnf5-plugins-5.4.3.0^20260514124419362424.pr2738.10.gc970e87d-1 package.

Maybe we should add a dummy behave test which will print installed dnf5 NEVRA to be really sure what we test. Maybe CI mistook a container image.

PR #2739, opened 2026-05-14
dnfdaemon: Include transaction_time in History.recent_changes() response

m-blaha:
Add a transaction_time attribute (int64, unix timestamp) to every
package returned by the recent_changes() method. This tells callers
when the change actually happened, which is especially important for
removed packages where install_time is not available.

A two-pass approach is used to ensure the timestamp reflects the latest
transaction that touched each package, not the oldest one.

Related: https://github.com/rpm-software-management/dnf5/issues/2734

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


mcrha:
There is a mismatch between the type of the values, because both `buildtime` and `install_time` are `uint64`, but this new `transaction_time` is `int64`. It'll be better to have the times of the same type.

mcrha:
Looks good to me, code-running wise.

ppisar (approved):
It looks good.

PR #2745, opened 2026-05-19
dnfdaemon: Use Goal resolution for Rpm.list upgrades/upgradable

### LINKED ISSUE #2744: dnf5daemon incorrectly calculates list of upgradable packages
The `Rpm.list(scope="upgrades")` used by Gnome Software dnf5 backend relies solely on queries in its implementation (similar to `dnf repoquery --upgrades`) and does not resolve an actual upgrade transaction (as `dnf upgrade` does).
This leads to differing results, as obsoletes, or repository priorities are not taken into account.

Particularly repo priorities are causing firefox issues from the issue in https://discussion.fedoraproject.org/t/firefox-already-updated-but-still-appears-on-the-software-manager/191502

m-blaha:
The previous implementation used PackageQuery::filter_upgrades() / filter_upgradable() which are simple version-comparison filters that don't match what an actual `dnf upgrade` transaction would produce.
This caused GNOME Software (and other D-Bus clients) to show updates that would never be performed (e.g. from low-priority repos or with uninstallable dependencies) and miss updates that would (e.g.  obsoletes).

Replace the query-based approach with actual Goal-based transaction resolution.  This correctly accounts for repo priorities, excludes, versionlock, obsoletes, vendor changes, dependency constraints, and uninstallable packages.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2744

### DIFF
--- a/dnf5daemon-server/services/rpm/rpm.cpp
+++ b/dnf5daemon-server/services/rpm/rpm.cpp
@@ -25,8 +25,11 @@
 #include "libdnf5/comps/environment/query.hpp"
 #include "libdnf5/comps/group/query.hpp"
 
+#include <libdnf5/base/goal.hpp>
+#include <libdnf5/base/transaction_package.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/rpm/package_set.hpp>
+#include <libdnf5/transaction/transaction_item_action.hpp>
 #include <sdbus-c++/sdbus-c++.h>
 
 #include <iostream>
@@ -411,10 +414,8 @@ libdnf5::rpm::PackageQuery Rpm::filter_packages(const dnfdaemon::KeyValueMap & o
 
     std::string scope = dnfdaemon::key_value_map_get<std::string>(options, "scope", "all");
     // start with all packages
-    libdnf5::sack::ExcludeFlags flags = libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES;
-    if (scope != "upgrades" && scope != "upgradable") {
-        flags = flags | libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK;
-    }
+    libdnf5::sack::ExcludeFlags flags =
+        libdnf5::sack::ExcludeFlags::APPLY_EXCLUDES | libdnf5::sack::ExcludeFlags::IGNORE_VERSIONLOCK;
     libdnf5::rpm::PackageQuery query(*base, flags);
 
     // toplevel filtering - the scope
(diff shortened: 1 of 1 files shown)


kontura:
I think the current implementation of the filters provides different and useful information.

Also will this match `dnf upgrade` in all cases? I believe for it to match 100% we would have to basically reimplement it exactly. Would it make sense to rather guide the users to use their normal upgrade workflow without performing the transaction (essentially the equivalent of `upgrade --assumeno`)?

m-blaha:
This is valid point. Maybe instead of introducing this inconsistency between D-Bus `Rpm.list()` and CLI `repoquery` we could have added a new call to return upgrade candidates. @mcrha what do you think? Would switching Gnome SW to a different D-Bus call be viable? I was considering it, but eventually came up with this implementation, basically to avoid requiring changes on Gnome SW side. If the equivalent of `dnf repoquery --upgrades` is needed for D-Bus clients in the future, it can be added later (e.g. by adding a different `scope`, or adding a new option like `resolve=false`).

What are the cases where this would not match `dnf upgrade` that you have in mind? I can think of `installonly_limit` which is handled outside of libsolv, and possibly passing non-default options like `best` or repos limiting to the actual goal that Gnome SW uses for upgrade.

m-blaha:
Also to be honest, the Firefox situation from the related Fedora discussion (https://discussion.fedoraproject.org/t/firefox-already-updated-but-still-appears-on-the-software-manager/191502) is kind of non-standard: the `mozilla` repo provides a version of firefox that compares lower than the one in the fedora repos. And it's resolved by increasing the mozilla repo priority.

kontura:
> And it's resolved by increasing the mozilla repo priority.

Oh, interesting.

> What are the cases where this would not match `dnf upgrade` that you have in mind? I can think of `installonly_limit` which is handled outside of libsolv, and possibly passing non-default options like `best` or repos limiting to the actual goal that Gnome SW uses for upgrade.

Yea, I was mostly thinking about the options and repos but the `installonly_limit` is also a very good point. I worry that any difference will eventually come back to bite us.

I would really prefer to keep the CLI and D-Bus consistent.
If the new call would be just a convenience shortcut with potential differences I am not sure its that helpful.
Honestly in my opinion we should just enhance the docs.

mcrha:
> Would switching Gnome SW to a different D-Bus call be viable?

I do not mind changing the call. I just more carefully re-read the gnome-software's dnf5 plugin code which returns a list of the packages for update, and it calls `list` with `scope=upgrades` first, then, if any package is returned, then it runs `upgrade` transaction, but only up to `resolve()`, it does not run the transaction further. Then it fiddles other things with the set of the returned packages.

It feels like the `list` with `scope=upgrades` is unneeded here.

Issue #2700, opened 2026-04-20
Document that positional argument separator "--" is no longer recognized


evan-goode:
See https://github.com/rpm-software-management/dnf5/issues/700. In the meantime, we should document the difference from DNF4 in changes_from_dnf4.

PR #2751, opened 2026-05-21
libdnf5-cli: Stop option parsing after "--"

### LINKED ISSUE #700: Positional argument separator "--" is no longer recognized
Prior DNF versions accepted the common "`--`" "stop-parsing" indicator to demarcate where all remaining command arguments should be treated as positional (I believe this "came for free" by using Python `argparse`).

dnf5 rejects the option as an unknown argument:

```
[root@b0e5aba396a0 /]# dnf5 install -y -- tar
Unknown argument "--" for command "install". Add "--help" for more information about the arguments.
```

This syntax is especially useful when passing-thru strings in scripts:

```sh
dnf install -y -- "$@"
```

Userfrom1995:
Treat "--" as the positional argument separator. Once it is
encountered, all remaining command line arguments are handled as
positional values, even if they start with a hyphen or match a
subcommand name.

Update completion handling so option and subcommand suggestions are not
offered after the separator, and add parser tests for the new behavior.

Fixes: #700 
Relates to: #2700

### DIFF
--- a/dnf5/context.cpp
+++ b/dnf5/context.cpp
@@ -1237,6 +1237,16 @@ std::vector<std::string> Context::match_specs(
         settings.set_with_provides(false);
         settings.set_with_filenames(false);
         settings.set_with_binaries(false);
+        // Patterns starting with '-' are not valid RPM package names.
+        // Without this, resolve_pkg_spec would parse '-' as a NEVRA
+        // name-version delimiter, causing '-v*' to match packages by
+        // version starting with 'v' instead of name. Safe to disable
+        // NEVRA because match_specs is only called from completion hooks
+        // the '-' prefixed arg only reaches here after "--" (the argument
+        // separator that stops named-arg/option parsing).
+        if (!pattern.empty() && pattern[0] == '-') {
+            settings.set_with_nevra(false);
+        }
         matched_pkgs_query.resolve_pkg_spec(pattern + '*', settings, true);
 
         for (const auto & package : matched_pkgs_query) {
--- a/doc/dnf5.8.rst
+++ b/doc/dnf5.8.rst
@@ -353,6 +353,12 @@ Following options are applicable in the general context for any ``dnf5`` command
     | Exclude packages specified in ``PACKAGE-SPEC-N`` arguments from the transaction.
     | This is a list option.
 
+Special Arguments
+=================
+
+``--``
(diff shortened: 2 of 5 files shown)


pkratoch:
Thanks for the patch!

I think there is still a small bug in the completion handling. When I try to auto-complete e.g. `dnf5 install -- -v`, it doesn't offer packages that start with "-v" (which would be an empty set), but instead prints a seemingly nonsensical result. It appears that it matches packages whose versions start with "v", because it tries to match nevra and the "-" is the delimiter between the name and version. I believe this happens because the `print_complete` is bypassed after the `--` and it goes through a different completion path. It could be fixed by passing the `stop_at_named_args` to `print_complete` and handle it there.

Userfrom1995:
> Thanks for the patch!
> 
> I think there is still a small bug in the completion handling. When I try to auto-complete e.g. `dnf5 install -- -v`, it doesn't offer packages that start with "-v" (which would be an empty set), but instead prints a seemingly nonsensical result. It appears that it matches packages whose versions start with "v", because it tries to match nevra and the "-" is the delimiter between the name and version. I believe this happens because the `print_complete` is bypassed after the `--` and it goes through a different completion path. It could be fixed by passing the `stop_at_named_args` to `print_complete` and handle it there.

The -v NEVRA issue happens because both completion paths (through `print_complete` and through `PositionalArg::parse`) end up calling the same `complete_hook` → `match_specs` → `resolve_pkg_spec`, which interprets - as a `NEVRA` version delimiter. 

The issue is not in the routing, it's in how `match_specs` handles - prefixed patterns.
Without `--`, `-v` is caught by the short-option guard and never reaches `match_specs`. 
After `--`, it becomes a positional arg and flows to `match_specs`, where NEVRA parsing misinterprets `-v*` as name="" version="v*".

I've changed `match_specs (context.cpp)` to disable NEVRA matching when the pattern starts with -. 
I beleieve this is safe because RPM package names cannot start with `-`, and `match_specs` is only used for completion hooks. 

I also implemented the routing change you suggested (passing stop_at_named_args to print_complete()). This ensures that completions after -- skip named arguments and commands, routing directly to the positional complete_hook. However, the actual fix for the -v issue is disabling NEVRA matching for --prefixed patterns.

Please take another look and let me know what you think of the changes, the fix, and whether you think any additional changes are required.

pkratoch (approved):

PR #2757, opened 2026-05-28
Fix GoalPrivate::operator= not resetting resolve-time members

### LINKED ISSUE #2756: Goal::resolve() crashes with "Id is out of bitmap range" when resolving multiple times in a session
I was using gnome-software to install multiple RPMs during the same session. After three successful installs, my fourth install failed with "Id is out of bitmap range". The package installs fine when using `rpm` to install it.

Distro: Fedora 44
Package: libdnf5-5.4.2.1

## Repro steps

1. Download 4-5 RPMs, I had downloaded Discord, 1password, VS Code, Microsoft Edge, and Google Chrome.
2. Open each RPM in gnome-software and install it.
3. After three or so, observe that the install fails.

I was able to reproduce with a script:

```python
#!/usr/bin/env python3
"""Reproduce bitmap range bug by padding pool to byte boundary.
Uses a single session - each Rpm::list adds a new solvable, then tries resolve."""

import dbus
import dbus.mainloop.glib
import sys

DEST = "org.rpm.dnf.v0"
OBJ = "/org/rpm/dnf/v0"

def main():
    if len(sys.argv) < 2:
        print(f"Usage: {sys.argv[0]} <path-to-local-rpm>")
        sys.exit(2)

    rpm_path = sys.argv[1]
    rpm_uri = f"file://{rpm_path}"

    dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
    bus = dbus.SystemBus()

    # Open single session
    mgr = bus.get_object(DEST, OBJ)
    session_path = mgr.open_session(
        dbus.Dictionary({}, signature='sv'),
        dbus_interface="org.rpm.dnf.v0.SessionManager"
    )
    print(f"Session: {session_path}")
    session = bus.get_object(DEST, session_path)
    rpm_iface = dbus.Interface(session, "org.rpm.dnf.v0.rpm.Rpm")
    goal_iface = dbus.Interface(session, "org.rpm.dnf.v0.Goal")

    list_options = dbus.Dictionary({
        'patterns': dbus.Array([rpm_path], signature='s', variant_level=1),
        'with_provides': dbus.Boolean(False, variant_level=1),
        'scope': dbus.String('all', variant_level=1),
    }, signature='sv')

    # First list call loads repos + adds the package
    print("Loading repos (first Rpm::list call)...")
    try:
        rpm_iface.list(list_options, timeout=600)
        print("Repos loaded, package added to pool (solvable #1)")
    except dbus.exceptions.DBusException as e:
        print(f"First list failed: {e}")
        return 1

    # Now iterate: after each Rpm::list (which adds the package again),
    # try Goal::reset -> Rpm::install(URI) -> Goal::resolve
    # Each iteration:
    #   - Rpm::list adds 1 solvable
    #   - Goal::resolve's add_paths_to_goal adds 1 more solvable
    # So pool grows by 2 per cycle (after the first list which was just 1)

    for iteration in range(16):  # 16 iterations covers all mod-8 values
        print(f"\nIteration {iteration}: adding package via Rpm::list...")
        try:
            rpm_iface.list(list_options, timeout=60)
        except dbus.exceptions.DBusException as e:
            print(f"  List failed: {e}")
            break

        # Try the install flow
        try:
            goal_iface.reset(timeout=30)
        except dbus.exceptions.DBusException as e:
            print(f"  Reset failed: {e}")
            break

        try:
            rpm_iface.install(
                dbus.Array([rpm_uri], signature='s'),
                dbus.Dictionary({}, signature='sv'),
                timeout=30
            )
        except dbus.exceptions.DBusException as e:
            print(f"  Install failed: {e}")
            break

        try:
            result = goal_iface.resolve(
                dbus.Dictionary({}, signature='sv'),
                timeout=600
            )
            print(f"  Resolve: OK")
        except dbus.exceptions.DBusException as e:
            msg = str(e)
            if "bitmap range" in msg:
                print(f"  *** BUG REPRODUCED at iteration {iteration}! ***")
                print(f"  Error: {msg}")
                print(f"  Pool additions from base: {2*(iteration+1) + 1}")
                return 0
            else:
                print(f"  Resolve error: {msg}")
                break

    print("\nNo crash in 16 iterations. Bug may require specific base pool size.")

    # Close session
    try:
        mgr.close_session(session_path, dbus_interface="org.rpm.dnf.v0.SessionManager")
    except:
        pass

    return 1

if __name__ == "__main__":
    sys.exit(main())
```

Run it with the path to any RPM:

```bash
sudo python3 reproduce_bug.py any_package.rpm
```

## Root cause

This appears to be caused by the `transaction_user_installed` bitmap retaining its original size. During `Goal::resolve()` the `p_impl->rpm_goal` member is reassigned via `operator=`, which doesn't reset the bitmap. On subsequent resolves, new solvables are added to the pool with IDs that exceed the bitmap's allocated capacity, causing `SolvMap::add()` to throw. This was likely missed because in the case of the copy constructor the field stays `nullptr` until the first use.

## Fix

The fix is to reset the fields:

```cpp
transaction_group_reason.reset()
transaction_user_installed.reset()
```

Based on comments and copy constructor behavior, these fields are not intended to be persisted. Comparing further, there are other fields that likely should be copied and reset, but that's not necessary to resolve the issue here.

kfarnung:
GoalPrivate::operator= was missing resets for several members that were
added after the original implementation. Most critically,
transaction_user_installed and transaction_group_reason bitmaps were
never reset, causing 'Id is out of bitmap range' crashes when
Goal::resolve() was called multiple times in the same session (e.g. via
dnf5daemon-server when gnome-software installs multiple local RPMs).

The fix aligns operator= with the copy constructor:
- Reset resolve-time artifacts (transaction_group_reason,
  transaction_user_installed)
- Copy input pointers (exclude_from_weak, user_installed_packages)
- Reset resolve-time state (clean_deps_present, groups, environments,
  reason_changes)

Fixes: https://github.com/rpm-software-management/dnf5/issues/2756
Assisted-by: GitHub Copilot

### DIFF
--- a/libdnf5/rpm/solv/goal_private.hpp
+++ b/libdnf5/rpm/solv/goal_private.hpp
@@ -263,10 +263,20 @@ inline GoalPrivate::~GoalPrivate() {
 
 inline GoalPrivate & GoalPrivate::operator=(const GoalPrivate & src) {
     if (this != &src) {
+        // Copy inputs
         base = src.base;
         staging = src.staging;
         installonly = src.installonly;
         installonly_limit = src.installonly_limit;
+
+        // Reset resolve-time artifacts
+        transaction_group_reason.reset();
+        transaction_user_installed.reset();
+
+        // Copy inputs (pointers)
+        exclude_from_weak.reset(src.exclude_from_weak ? new libdnf5::solv::SolvMap(*src.exclude_from_weak) : nullptr);
+
+        // Reset solver state
         if (libsolv_solver.is_initialized()) {
             libsolv_solver.reset();
         }
@@ -274,15 +284,27 @@ inline GoalPrivate & GoalPrivate::operator=(const GoalPrivate & src) {
             transaction_free(libsolv_transaction);
             libsolv_transaction = nullptr;
         }
+
+        // Copy inputs (pointers)
         protected_packages.reset(
             src.protected_packages ? new libdnf5::solv::SolvMap(*src.protected_packages) : nullptr);
         removal_of_protected.reset();
         protected_running_kernel = src.protected_running_kernel;
+        user_installed_packages.reset(
+            src.user_installed_packages ? new libdnf5::solv::IdQueue(*src.user_installed_packages) : nullptr);
+
+        // Copy inputs (flags)
         allow_downgrade = src.allow_downgrade;
         allow_erasing = src.allow_erasing;
         allow_vendor_change = src.allow_vendor_change;
         install_weak_deps = src.install_weak_deps;
         run_in_strict_mode = src.run_in_strict_mode;
+
+        // Reset resolve-time state
+        clean_deps_present = false;
+        groups.clear();
+        environments.clear();
+        reason_changes.clear();
     }
     return *this;
 }
--- a/test/dnf5daemon-server/test_install.py
+++ b/test/dnf5daemon-server/test_install.py
@@ -161,3 +161,18 @@ class InstallTest(support.InstallrootCase):
         )
 
         self.iface_goal.do_transaction(dbus.Dictionary({}, signature='sv'))
+
+    def test_install_file_multiple_resolves(self):
+        '''
+        Test that resolving a file install multiple times in the same session
+        does not crash with "Id is out of bitmap range".
+        '''
+        rpm_path = self.path_to_repo_rpm("rpm-repo1", "one-1-1.noarch.rpm")
+
+        for i in range(16):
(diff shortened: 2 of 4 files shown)


m-blaha (approved):
Thank you, this looks good! The test failures are due to libsolv-0.7.38 bug https://github.com/openSUSE/libsolv/pull/623. Locally with libsolv 0.7.39 the tests passed.

PR #2763, opened 2026-06-02
spec: Remove unused dependency on "check" pkg-config file

ppisar:
The Check library is not used anywhere now. It seems the build-time dependency was copied from libdnf project, without actually being used at that time either.


m-blaha (approved):
Thanks, it's unused leftover from the libdnf days...

PR #2768, opened 2026-06-04
Add distro-sync support to dnf-automatic

sfalken:
Adds distro-sync capability to the automatic plugin, and a D-bus emitter for desktop notifications, this is in preparation for the addition of a transactional style system in Fedora that's being worked on (similar to MicroOS)

### DIFF
--- a/dnf5-plugins/automatic_plugin/CMakeLists.txt
+++ b/dnf5-plugins/automatic_plugin/CMakeLists.txt
@@ -2,14 +2,18 @@
 set(GETTEXT_DOMAIN dnf5-plugin-automatic)
 add_definitions(-DGETTEXT_DOMAIN=\"${GETTEXT_DOMAIN}\")
 
+include(sdbus_cpp)
+
 file(GLOB_RECURSE AUTOMATIC_SOURCES *.cpp)
 add_library(automatic_cmd_plugin MODULE ${AUTOMATIC_SOURCES})
 
 # disable the 'lib' prefix in order to create automatic_cmd_plugin.so
 set_target_properties(automatic_cmd_plugin PROPERTIES PREFIX "")
 
+target_include_directories(automatic_cmd_plugin PRIVATE ${SDBUS_CPP_INCLUDE_DIRS})
 target_link_libraries(automatic_cmd_plugin PRIVATE libdnf5 libdnf5-cli)
 target_link_libraries(automatic_cmd_plugin PRIVATE dnf5)
+target_link_libraries(automatic_cmd_plugin PRIVATE ${SDBUS_CPP_LIBRARIES})
 
 install(TARGETS automatic_cmd_plugin LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/dnf5/plugins/)
 install(DIRECTORY "config/usr/share" DESTINATION "${CMAKE_INSTALL_PREFIX}")
--- a/dnf5-plugins/automatic_plugin/emitters.cpp
+++ b/dnf5-plugins/automatic_plugin/emitters.cpp
@@ -25,11 +25,13 @@
 #include <libdnf5/base/transaction_package.hpp>
 #include <libdnf5/utils/bgettext/bgettext-lib.h>
 #include <libdnf5/utils/format.hpp>
+#include <sdbus-c++/sdbus-c++.h>
 
 #include <cstdio>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <map>
 
 namespace dnf5 {
 
@@ -218,6 +220,17 @@ void EmitterEmail::notify() {
             curl_slist_free_all(recipients);
             curl_easy_cleanup(curl);
         }
+    }
+}
+
+void EmitterDBus::notify() {
+    try {
+        auto conn = sdbus::createSystemBusConnection(sdbus::ServiceName{"org.rpm.dnf.v0.Automatic"});
+        auto obj = sdbus::createObject(*conn, sdbus::ObjectPath{"/org/rpm/dnf/v0/Automatic"});
+        std::map<std::string, sdbus::Variant> result{{"success", sdbus::Variant{success}}};
+        obj->emitSignal("UpdateResult").onInterface("org.rpm.dnf.v0.Automatic").withArguments(result);
+    } catch (const sdbus::Error & e) {
+        std::cerr << "dnf5-automatic: D-Bus error: " << e.what() << std::endl;
     }
 }
 
--- a/dnf5-plugins/automatic_plugin/automatic.cpp
+++ b/dnf5-plugins/automatic_plugin/automatic.cpp
@@ -471,6 +471,8 @@ void AutomaticCommand::run() {
                 emitter = std::make_unique<EmitterCommandEmail>(config_automatic, transaction, output_stream, success);
             } else if (emitter_name == "email") {
                 emitter = std::make_unique<EmitterEmail>(config_automatic, transaction, output_stream, success);
+            } else if (emitter_name == "dbus") {
+                emitter = std::make_unique<EmitterDBus>(config_automatic, transaction, output_stream, success);
             } else {
                 auto & logger = *base.get_logger();
                 logger.warning(_("Unknown report emitter for dnf5 automatic: \"{}\"."), emitter_name);
--- a/dnf5-plugins/automatic_plugin/automatic.cpp
+++ b/dnf5-plugins/automatic_plugin/automatic.cpp
@@ -365,15 +365,20 @@ void AutomaticCommand::run() {
     // setup upgrade transaction goal
     auto settings = libdnf5::GoalJobSettings();
 
+    libdnf5::Goal goal(base);
+    const auto upgrade_type = config_automatic.config_commands.upgrade_type.get_value();
     // TODO(mblaha): Use advisory_query_from_cli_input from dnf5/commands/advisory_shared.hpp?
-    if (config_automatic.config_commands.upgrade_type.get_value() == "security") {
+    if (upgrade_type == "security") {
         auto advisories = libdnf5::advisory::AdvisoryQuery(base);
         advisories.filter_type("security");
         settings.set_advisory_filter(advisories);
         // TODO(mblaha): set also minimal=true?
+        goal.add_rpm_upgrade(settings);
+    } else if (upgrade_type == "distro-sync") {
+        goal.add_rpm_distro_sync(settings);
+    } else {
+        goal.add_rpm_upgrade(settings);
     }
-    libdnf5::Goal goal(base);
-    goal.add_rpm_upgrade(settings);
 
     auto transaction = goal.resolve();
 
--- a/dnf5-plugins/automatic_plugin/config/usr/share/dbus-1/interfaces/org.rpm.dnf.v0.Automatic.xml
+++ b/dnf5-plugins/automatic_plugin/config/usr/share/dbus-1/interfaces/org.rpm.dnf.v0.Automatic.xml
@@ -1,0 +1,49 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+
+<!--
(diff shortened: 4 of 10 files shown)


dnf5-plugins/automatic_plugin/CMakeLists.txt
```diff
@@ -2,14 +2,18 @@
 set(GETTEXT_DOMAIN dnf5-plugin-automatic)
 add_definitions(-DGETTEXT_DOMAIN=\"${GETTEXT_DOMAIN}\")
 
+pkg_check_modules(SDBUS REQUIRED sdbus-c++)
```
Conan-Kudo: You should probably rework this to use our common code for finding sdbus-cpp. 

```suggestion
include(sdbus_cpp)
```

The definition is in: https://github.com/rpm-software-management/dnf5/blob/main/cmake/sdbus_cpp.cmake

sfalken: I can do that.

sfalken: done in 6bd3953

dnf5-plugins/automatic_plugin/emitters.cpp
```diff
@@ -221,4 +222,14 @@ void EmitterEmail::notify() {
...
 
+void EmitterDBus::notify() {
+    try {
+        auto conn = sdbus::createSystemBusConnection(sdbus::ServiceName{"org.rpm.dnf.v0.Automatic"});
+        auto obj = sdbus::createObject(*conn, sdbus::ObjectPath{"/org/rpm/dnf/v0/Automatic"});
+        obj->emitSignal("UpdateResult").onInterface("org.rpm.dnf.v0.Automatic").withArguments(success);
+    } catch (const sdbus::Error & e) {
+        std::cerr << "dnf5-automatic: D-Bus error: " << e.what() << std::endl;
+    }
+}
```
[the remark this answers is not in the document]
sfalken: I'm aware a dbus policy configuration is required, I'm just unsure where it needs to be in the codebase.

dnf5-plugins/automatic_plugin/CMakeLists.txt
```diff
@@ -2,14 +2,18 @@
...
 file(GLOB_RECURSE AUTOMATIC_SOURCES *.cpp)
 add_library(automatic_cmd_plugin MODULE ${AUTOMATIC_SOURCES})
 
 # disable the 'lib' prefix in order to create automatic_cmd_plugin.so
 set_target_properties(automatic_cmd_plugin PROPERTIES PREFIX "")
 
+target_include_directories(automatic_cmd_plugin PRIVATE ${SDBUS_INCLUDE_DIRS})
 target_link_libraries(automatic_cmd_plugin PRIVATE libdnf5 libdnf5-cli)
 target_link_libraries(automatic_cmd_plugin PRIVATE dnf5)
+target_link_libraries(automatic_cmd_plugin PRIVATE ${SDBUS_LIBRARIES})
```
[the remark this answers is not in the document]
sfalken: This has been changed, to conform better with dnf5 policy, by using the internal sdbus_cpp mechanism

dnf5-plugins/automatic_plugin/emitters.cpp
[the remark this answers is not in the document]
m-blaha: It should be shipped with the dnf5-plugin-automatic subpackage, the file itself I would store somewhere under dnf5-plugins/automatic_plugin/config. I would keep requesting the well-known name (as it currently is) just to enable consumers to filter the signals by sender to protect against spoofed signals sent by regular users.

m-blaha:
Regarding the new "distro-sync" upgrade type, please add it also to the `automatic_plugin/config/usr/share/dnf5/dnf5-plugins/automatic.conf` and `doc/dnf5_plugins/automatic.8.rst` files.

dnf5-plugins/automatic_plugin/emitters.cpp
```diff
@@ -221,4 +222,14 @@ void EmitterEmail::notify() {
     }
 }
 
+void EmitterDBus::notify() {
+    try {
+        auto conn = sdbus::createSystemBusConnection(sdbus::ServiceName{"org.rpm.dnf.v0.Automatic"});
+        auto obj = sdbus::createObject(*conn, sdbus::ObjectPath{"/org/rpm/dnf/v0/Automatic"});
+        obj->emitSignal("UpdateResult").onInterface("org.rpm.dnf.v0.Automatic").withArguments(success);
```
m-blaha: We should carefully think about signature of the signal, since adding extra data later would be API breaking change. Is having only one `bool` enough? Using `a{sv}` signature instead of `b` could help making the payload extensible.

m-blaha:
One more think - there should be an introspection XML file for the Automatic interface (similar to what we already have in dnf5daemon-server/dbus/interfaces/). It should contain the signal description, signature and arguments.

dnf5-plugins/automatic_plugin/emitters.cpp
sfalken: I don't personally have any use for any signal other than "success" or "failure", what other possible values are you thinking of?

m-blaha: Nothing in particular - maybe error messages from failed transactions, a list of upgraded packages... I've just been burned several times by APIs not being designed flexible enough, resulting in the need to create a new signal/method with a new signature :(

ppisar: I have no idea what the foreseen transactional style means, but what does make the automatic plugin so special that it deserves its own D-Bus service? Shouldn't manual "dnf5 upgrade" command also emit the same D-Bus signal?

sfalken: > Nothing in particular - maybe error messages from failed transactions, a list of upgraded packages... I've just been burned several times by APIs not being designed flexible enough, resulting in the need to create a new signal/method with a new signature :(

Ok, I'm good with flexible, let me poke at it a bit.  Making it more useful for other usecases is fine by me.

sfalken: > I have no idea what the foreseen transactional style means, but what does make the automatic plugin so special that it deserves its own D-Bus service? Shouldn't manual "dnf5 upgrade" command also emit the same D-Bus signal?

Probably?   I'm trying to keep the scope down to what I'll actually be using for, rather than pushing for a major change to dnf5 itself.

Transactional style is using [dnf5-plugin-txnupd,](https://gitlab.com/VelocityLimitless/Projects/libdnf-plugin-txnupd) to recreate [MicroOS](https://microos.opensuse.org) style systems (read-only base system, based on btrfs snapshotting), using dnf5 as the frontend, rather than the bespoke shell scripting of [transactional-update](https://github.com/openSUSE/transactional-update) as is currently done on openSUSE based transactional systems.

I'm planning on making dnf5-txnupd the update mechanism in [Kalpa Desktop](https://kalpadesktop.org), as well as implement the possibility of transactional systems within Fedora, and it's downstreams (if folks are interested)

m-blaha: > I have no idea what the foreseen transactional style means, but what does make the automatic plugin so special that it deserves its own D-Bus service? Shouldn't manual "dnf5 upgrade" command also emit the same D-Bus signal?

That's a good point - @sfalken would creating libdnf5 plugin similar to `rpm-plugin-dbus-announce` help with your use case?

sfalken: > > I have no idea what the foreseen transactional style means, but what does make the automatic plugin so special that it deserves its own D-Bus service? Shouldn't manual "dnf5 upgrade" command also emit the same D-Bus signal?
> 
> That's a good point - @sfalken would creating libdnf5 plugin similar to `rpm-plugin-dbus-announce` help with your use case?

I potentially could?  @Conan-Kudo your thoughts?

ppisar: Thanks for the links, especially <https://kubic.opensuse.org/documentation/transactional-update-guide/transactional-update.html> was useful for me.

> what does make the automatic plugin so special that it deserves its own D-Bus service?

After reading them, I think it's the asynchronous nature of DNF automatic. Users who have the automatic service/timer enabled, they want to know when an update was installed to do the reboot. While other invocations of DNF5 (or other libdnf5 applications, like Gnome Software) are synchronous. I guess the this D-Bus notification only targets people because the Btrfs snapshot will have been completed by the existing txnupd plugin before receiving the notification. Also a user is not interested anytime he installs a package to get a pop-up window "You have installed a package". I would immediately tried to disable the notifications. With this use case on their mind, a notification specific to the automatic plugin is probably good enough.

Conan-Kudo: Yes. In this case, it runs in the background on a schedule, stacks updates, and then the notifier would let the user know a reboot is needed.

sfalken: > Thanks for the links, especially https://kubic.opensuse.org/documentation/transactional-update-guide/transactional-update.html was useful for me.
> 
> > what does make the automatic plugin so special that it deserves its own D-Bus service?
> 
> After reading them, I think it's the asynchronous nature of DNF automatic. Users who have the automatic service/timer enabled, they want to know when an update was installed to do the reboot. While other invocations of DNF5 (or other libdnf5 applications, like Gnome Software) are synchronous. I guess the this D-Bus notification only targets people because the Btrfs snapshot will have been completed by the existing txnupd plugin before receiving the notification. Also a user is not interested anytime he installs a package to get a pop-up window "You have installed a package". I would immediately tried to disable the notifications. With this use case on their mind, a notification specific to the automatic plugin is probably good enough.

Right, this emitter only fires, when a new snapshot, with a successful update is created during the maintenance window, I'm working on making the signal extensible, as was requested in https://github.com/rpm-software-management/dnf5/pull/2768#discussion_r3381772401 as that's not a huge deal, but only defining the boolean key "success" for my needs at the moment.

Folks could extend it to do other things, if they wanted to.

dnf5-plugins/automatic_plugin/emitters.cpp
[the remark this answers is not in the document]
sfalken: should be in the latest commit

sfalken:
> Regarding the new "distro-sync" upgrade type, please add it also to the `automatic_plugin/config/usr/share/dnf5/dnf5-plugins/automatic.conf` and `doc/dnf5_plugins/automatic.8.rst` files.

done in the latest push

sfalken:
> One more think - there should be an introspection XML file for the Automatic interface (similar to what we already have in dnf5daemon-server/dbus/interfaces/). It should contain the signal description, signature and arguments.

added

dnf5-plugins/automatic_plugin/emitters.cpp
sfalken: I changed this to a{sv}, and just defined the boolean I needed

m-blaha (dismissed):
rpm builds are failing:
```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /builddir/build/BUILD/dnf5-5.4.3.0_20260610165555687381.pr2768.33.g4da8e843-build/BUILDROOT
error: Installed (but unpackaged) file(s) found:
   /usr/share/dbus-1/system.d/org.rpm.dnf.v0.Automatic.conf
    Installed (but unpackaged) file(s) found:
   /usr/share/dbus-1/system.d/org.rpm.dnf.v0.Automatic.conf
```

sfalken:
> rpm builds are failing:
> 
> ```
> Checking for unpackaged file(s): /usr/lib/rpm/check-files /builddir/build/BUILD/dnf5-5.4.3.0_20260610165555687381.pr2768.33.g4da8e843-build/BUILDROOT
> error: Installed (but unpackaged) file(s) found:
>    /usr/share/dbus-1/system.d/org.rpm.dnf.v0.Automatic.conf
>     Installed (but unpackaged) file(s) found:
>    /usr/share/dbus-1/system.d/org.rpm.dnf.v0.Automatic.conf
> ```

Fixed.   Sorry, forgot to add it the spec file.

Conan-Kudo (approved):

PR #2769, opened 2026-06-04
needs-restarting: Reuse system bus connection everywhere

### LINKED ISSUE #2562: Inconsistencies with dnf needs-restarting
dnf5 version 5.2.17.0
dnf5 plugin API version 2.0
libdnf5 version 5.2.17.0
libdnf5 plugin API version 2.2

name: needs_restarting
version: 1.0.0
API version: 2.0

This was done reverting and re-doing upgrades to kernel-0:6.18.4-200.fc43.x86_64

Without sudo this has worked for me with: 
dnf needs-restarting
dnf needs-restarting -r
dnf needs-restarting --reboothint
dnf needs-restarting -s 

When I run with sudo
Running with -r or --reboothint sometimes indicates a reboot is needed other times it does not. 
Running with sudo after running as the current user exhibits the same behavior.  
When the failure is observed with sudo, running as the user still works. 
 
Thanks

evan-goode:
Memoize sdbus::createSystemBusConnection() to ensure the same D-Bus connection is reused for the lifetime of the NeedsRestartingCommand.

Also, always pass this connection explicitly to sdbus::createProxy; if no connection is specified, sdbus::createProxy will create its own internal D-Bus connection "to either the session bus or the system bus, depending on the context, see man sd_bus_open)." [1]

Resolves https://github.com/rpm-software-management/dnf5/issues/2562

[1] https://kistler-group.github.io/sdbus-cpp/docs/using-sdbus-c++.html

### DIFF
--- a/dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
+++ b/dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
@@ -161,6 +161,22 @@ time_t NeedsRestartingCommand::get_kernel_boot_time(Context & ctx) {
     return btime;
 }
 
+// Memoized sdbus::createSystemBusConnection
+sdbus::IConnection & NeedsRestartingCommand::get_system_bus_connection() {
+    if (system_bus_connection_exception) {
+        std::rethrow_exception(system_bus_connection_exception);
+    }
+    if (system_bus_connection == nullptr) {
+        try {
+            system_bus_connection = sdbus::createSystemBusConnection();
+        } catch (...) {
+            system_bus_connection_exception = std::current_exception();
+            std::rethrow_exception(system_bus_connection_exception);
+        }
+    }
+    return *system_bus_connection;
+}
+
 time_t NeedsRestartingCommand::get_boot_time(Context & ctx) {
     // We have three sources from which to derive the boot time. These values
     // vary depending on containerization, existing of a Real Time Clock, etc:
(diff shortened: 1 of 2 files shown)


dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
```diff
@@ -161,6 +161,23 @@ time_t NeedsRestartingCommand::get_kernel_boot_time(Context & ctx) {
     return btime;
 }
 
+// Memoized sdbus::createSystemBusConnection
+sdbus::IConnection & NeedsRestartingCommand::get_system_bus_connection() {
+    if (system_bus_connection_exception) {
+        std::rethrow_exception(system_bus_connection_exception);
+    }
+    if (system_bus_connection != nullptr) {
+        return *system_bus_connection;
```
ppisar: Could you move this return statement and the same statement in the try block to the end of this method and move the try-catch blocks into this if block while inverting the condition?

ppisar (changes requested):
The sdbus::createProxy() protype you chose does not start an event loop for delivering asynchronous calls and signals. I believe that no D-Bus interface the needs-restarting plugin uses use those features, so it's not a problem here.

I also verified the plugin works (again.)

dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp
evan-goode: Sure, done.

ppisar (approved):
Good.

PR #2774, opened 2026-06-10
reposync: exit non-zero when a package fails to download

### LINKED ISSUE #1926: API: PackageDownloader::download() method does not indicate all download errors
The `PackageDownloader::download()` method throws an exception only in the case of "irrecoverable" download errors. When the `fail_fast` flag is set to false, the method does not throw an error even if some targets fail to download. Errors related to individual downloads are reported exclusively via the corresponding librepo targets. Since these librepo targets are stored as local variables within the method, the caller of download() cannot check the status of individual downloads.

This behavior impacts `reposync`, which is expected to:

- Avoid failing on the first error, downloading as many packages as possible (hence, fail_fast is set to false).
- Return an exit code of 1 if some packages fail to download (now the API does not provide way how to check it)

liulinC:
reposync runs PackageDownloader with fail_fast=false and never checks per-package results, so it exits 0 even when a package fails its size/checksum verification and is dropped, silently producing an incomplete mirror.

Expose per-package download failures from libdnf5 PackageDownloader via a new get_failed_packages() method, and make reposync print the failed packages and exit 1 when any package failed (download as much as possible, then fail).

Resolves #1926

### DIFF
--- a/dnf5-plugins/reposync_plugin/reposync.cpp
+++ b/dnf5-plugins/reposync_plugin/reposync.cpp
@@ -279,10 +279,14 @@ void ReposyncCommand::download_packages(const ReposyncCommand::download_list_typ
         downloader.add(pkg, pth.parent_path());
     }
     downloader.download();
-    // TODO(mblaha): Return exit code 1 if any of packages was not downloaded.
-    // In case of fail_fast set to false, the download() method does
-    // not throw an exception if particular package could not be downloaded.
-    // See https://github.com/rpm-software-management/dnf5/issues/1926 for details
+    const auto failed_packages = downloader.get_failed_packages();
+    if (!failed_packages.empty()) {
+        for (const auto & pkg : failed_packages) {
+            std::cerr << libdnf5::utils::sformat(_("Failed to download package: {}"), pkg.get_full_nevra())
+                      << std::endl;
+        }
+        throw libdnf5::cli::CommandExitError(1, M_("Failed to download one or more packages"));
+    }
 }
 
 void ReposyncCommand::delete_old_local_packages(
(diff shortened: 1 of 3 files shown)


liulinC:
This is referring https://github.com/rpm-software-management/yum-utils/blob/master/reposync.py#L340
```
probs = my.downloadPkgs(download_list)   # downloads ALL pkgs; returns {pkg: [errors]}, never raises
if probs:
    exit_code = 1                         # any per-package failure -> non-zero
    for key in probs:
        for error in probs[key]:
            my.logger.error('%s: %s', key, error)
...
sys.exit(exit_code)
```

m-blaha:
I believe that changing `Addresses #1926` to `Resolves https://github.com/rpm-software-management/dnf5/issues/1926` in the commit message/PR description will automatically close the issue once this is merged.

m-blaha:
Could you split this into two commits?
  
1. API commit with PackageDownloader enhancements
2. Reposync changes leveraging the new API (this one should have Resolves: <issue> in description)

m-blaha:
The rawhide tests failure is unrelated - see https://github.com/rpm-software-management/ci-dnf-stack/pull/1891

liulinC:
> The rawhide tests failure is unrelated - see [rpm-software-management/ci-dnf-stack#1891](https://github.com/rpm-software-management/ci-dnf-stack/pull/1891)

Thanks for confirming this.

m-blaha (approved):
I filed a follow-up issue https://github.com/rpm-software-management/dnf5/issues/2775. The issue is pre-existing and I would like to get some feedback from the team whether we even want to change this.

liulinC:
Is there any review update about this ?

PR #2776, opened 2026-06-10
Add remove --duplicates option to remove older duplicate packages

### LINKED ISSUE #760: dnf [options] remove --duplicates

fhbash:
Removes older versions of duplicate packages (same name and arch, different version) and reinstalls the newest version to ensure system integrity. Installonly packages are excluded. Available repos are loaded when needed for the reinstall step.

Resolves: https://github.com/rpm-software-management/dnf5/issues/760

CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1895

### DIFF
--- a/dnf5/commands/remove/remove.cpp
+++ b/dnf5/commands/remove/remove.cpp
@@ -65,6 +65,17 @@ void RemoveCommand::set_argument_parser() {
     oldinstallonly_opt->link_value(oldinstallonly);
     cmd.register_named_arg(oldinstallonly_opt);
 
+    duplicates = dynamic_cast<libdnf5::OptionBool *>(
+        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
+    auto duplicates_opt = parser.add_new_named_arg("duplicates");
+    duplicates_opt->set_long_name("duplicates");
+    duplicates_opt->set_description("Remove older versions of duplicate packages and reinstall the newest version");
+    duplicates_opt->set_const_value("true");
+    duplicates_opt->link_value(duplicates);
+    cmd.register_named_arg(duplicates_opt);
+
+    oldinstallonly_opt->add_conflict_argument(*duplicates_opt);
+
     oldinstallonly_limit = dynamic_cast<libdnf5::OptionNumber<std::int32_t> *>(parser.add_init_value(
         std::unique_ptr<libdnf5::OptionNumber<std::int32_t>>(new libdnf5::OptionNumber<std::int32_t>(0))));
     auto limit_opt = parser.add_new_named_arg("limit");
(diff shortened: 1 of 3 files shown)


dnf5/commands/remove/remove.cpp
```diff
@@ -175,6 +187,49 @@ void RemoveCommand::run() {
                 }
             }
         }
+    } else if (duplicates->get_value()) {
+        if (oldinstallonly_limit->get_priority() >= libdnf5::Option::Priority::COMMANDLINE) {
+            throw libdnf5::cli::ArgumentParserMissingDependentArgumentError(
+                M_("Option '--limit' can only be used with '--oldinstallonly'."));
```
m-blaha: For consistency - wouldn't it be better to handle --duplicates/--limit conflict the same way as the --duplicates/--oldinstallonly one? This way the user get the same errors in both situations:
```
limit_opt->add_conflict_argument(*duplicates_opt);
```

dnf5/commands/remove/remove.cpp
```diff
@@ -175,6 +187,49 @@ void RemoveCommand::run() {
...
+        for (const auto & pkg : duplicate_query) {
+            packages_by_na[pkg.get_na()].push_back(pkg);
+        }
+
+        for (auto & [na, packages] : packages_by_na) {
+            std::sort(packages.begin(), packages.end(), [](const auto & a, const auto & b) {
+                return libdnf5::rpm::cmp_nevra(b, a);
+            });
+
+            goal->add_rpm_reinstall(packages[0]);
```
m-blaha: There's a bug in `Goal::Impl::add_rpms_to_goal()`: when `add_rpm_reinstall(const Package&)` is called with an installed package, the `REINSTALL` case verifies the solvable id is installed and then passes that same **installed** solvable id to `rpm_goal.add_install()`. This results in libsolv job like `job install oneof acpi-1.8-2.fc43.x86_64@@System [setevr,setarch]` which is no-op - the package is already installed.

The string-based `add_reinstall_to_goal()` (called by goal->add_rpm_reinstall(<nevra spec>)) handles this correctly: it resolves the spec, splits installed from available, finds the available package with matching NEVRA, and feeds the **available** solvable id to the solver. Now the libsolv job looks like `job install oneof acpi-1.8-2.fc43.x86_64@fedora [setevr,setarch]` which correctly works.
  
This is a pre-existing bug in libdnf5, not introduced by this PR. I'd file a separate issue for it.

As a workaround, you can use the string-based overload instead:

```suggestion
            goal->add_rpm_reinstall(packages[0].get_full_nevra());
```

m-blaha: Filed as issue https://github.com/rpm-software-management/dnf5/issues/2777

dnf5/commands/remove/remove.cpp
```diff
@@ -175,6 +187,49 @@ void RemoveCommand::run() {
...
+
+            for (size_t i = 1; i < packages.size(); ++i) {
+                goal->add_rpm_remove(packages[i]);
+            }
+        }
```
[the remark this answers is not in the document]
m-blaha: actually, there is a simpler solution - use `skip_unavailable` option for add_rpm_reinstall(). You don't need to manually check availability and get a warning for free:

```suggestion
        libdnf5::GoalJobSettings settings;
        settings.set_skip_unavailable(libdnf5::GoalSetting::SET_TRUE);
        for (auto & [na, packages] : packages_by_na) {
            std::sort(packages.begin(), packages.end(), [](const auto & a, const auto & b) {
                return libdnf5::rpm::cmp_nevra(b, a);
            });

            goal->add_rpm_reinstall(packages[0].get_full_nevra(), settings);

            for (size_t i = 1; i < packages.size(); ++i) {
                goal->add_rpm_remove(packages[i]);
            }
        }
```

m-blaha (approved):
Looks good, thank you!

PR #2785, opened 2026-06-19
Assert system repo is locked for writing when performing transactions

### LINKED ISSUE #2676: Assert system repo write lock is held before running a transaction
The system state files (`packages.toml`, `nevras.toml`, etc.) in `/usr/lib/sysimage/libdnf5/` are used during goal resolution to determine package reasons and comps information. These files are read by `State::load()` (triggered via `get_system_state()` during system repo loading) and written by `State::save()` at the end of a transaction.

Without proper synchronization, a reader process can observe inconsistent system state during a concurrent transaction's `State::save()`. The save method writes .new files and then renames them sequentially during the rename phase, a reader can see a mix of old and new state files (e.g., updated packages.toml but stale nevras.toml).

Add an assertion in `Transaction::Impl::_run()` that the system repo write lock is held before a transaction is executed. This ensures that

- Transactions always run under the write lock, protecting both rpmdb and system state from concurrent access.
- Read-only operations (e.g., repoquery) do not strictly require the lock. They may read slightly stale or inconsistent system state during a concurrent transaction, which is I think acceptable.
- libdnf5 API consumers (PackageKit, dnf5daemon, custom integrations) that forget to acquire the lock before running a transaction get a clear assertion failure instead of silent data corruption.

Also the `Locker` class API enhancement are needed - currently, `Locker` does not expose whether a lock is held or whether it's a READ or WRITE lock.

Related:
- #2601 (fixed by #2675)
- #2519
- #1610

kontura:
Closes: https://github.com/rpm-software-management/dnf5/issues/2676

### DIFF
--- a/bindings/libdnf5/base.i
+++ b/bindings/libdnf5/base.i
@@ -23,6 +23,7 @@
 %import "repo.i"
 %import "rpm.i"
 %import "transaction.i"
+%import "utils.i"
 
 %{
     #include "bindings/libdnf5/exception.hpp"
--- a/bindings/libdnf5/base.i
+++ b/bindings/libdnf5/base.i
@@ -24,6 +24,7 @@
 %import "repo.i"
 %import "rpm.i"
 %import "transaction.i"
+%import "utils.i"
 
 %{
     #include "bindings/libdnf5/exception.hpp"
--- a/dnf5daemon-server/threads_manager.hpp
+++ b/dnf5daemon-server/threads_manager.hpp
@@ -23,6 +23,8 @@
 #include "dbus.hpp"
 #include "utils.hpp"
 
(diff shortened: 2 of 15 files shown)


kontura:
The builds currently fail due to too many open file descriptors, the libdnf5 tests now additionally open the system repo lock, I think this is expected.

I have discovered librepo leaks a lot of `fd`s, fix: https://github.com/rpm-software-management/librepo/pull/377
With the librepo change the builds here should pass.

kontura:
Test update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1902

Further requires: https://github.com/rpm-software-management/dnf5/pull/2788

PR #2804, opened 2026-06-29
Reflink local packages into the cache when supported

daandemeyer:
When copying locally-available packages into their download destination, attempt a copy-on-write clone (FICLONE) first and fall back to a plain byte copy when the filesystem doesn't support it (cross-device, non-CoW fs, older kernel). On CoW filesystems this makes the copy near-instant and avoids duplicating storage.

### DIFF
--- a/libdnf5/repo/package_downloader.cpp
+++ b/libdnf5/repo/package_downloader.cpp
@@ -21,6 +21,7 @@
 
 #include "repo_downloader.hpp"
 #include "temp_files_memory.hpp"
+#include "utils/fs/utils.hpp"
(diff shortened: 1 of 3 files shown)


m-blaha:
Cool, thank you! There are even more places across dnf5 code base that could leverage this, once it is merged.
Can you please resolve the clang formatting issue (https://github.com/rpm-software-management/dnf5/actions/runs/28377046700/job/84248444226?pr=2804)?

m-blaha (approved):
Thank you!

daandemeyer:
Thanks!

PR #2809, opened 2026-07-03
More copy->reflink replacements across the code base

m-blaha:
This is a follow-up to https://github.com/rpm-software-management/dnf5/pull/2804

The code now leverages `reflink_or_copy` in several additional places:

- prepopulating the cache for offline and stored transactions
- the local plugin
- cloning the user cache from the root one

### DIFF
--- a/libdnf5/repo/repo.cpp
+++ b/libdnf5/repo/repo.cpp
@@ -30,6 +30,7 @@ constexpr const char * REPOID_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmno
 #include "libdnf5/common/exception.hpp"
 #include "libdnf5/conf/const.hpp"
 #include "libdnf5/utils/bgettext/bgettext-mark-domain.h"
+#include "libdnf5/utils/fs/utils.hpp"
 
 extern "C" {
 #include <solv/repo_rpmdb.h>
@@ -544,6 +545,27 @@ void Repo::internalize() {
 }
 
 
+namespace {
+
+// Reflink-or-copy every entry from a flat cache directory into `dest_dir`.
+// Regular files are reflinked when possible; other entry types fall back to std::filesystem::copy.
+void reflink_or_copy_cache(
+    const std::filesystem::path & src_dir, const std::filesystem::path & dest_dir, std::error_code & ec) {
+    for (const auto & entry : std::filesystem::directory_iterator(src_dir, ec)) {
+        auto dest = dest_dir / entry.path().filename();
+        if (entry.is_regular_file()) {
+            utils::fs::reflink_or_copy(entry.path(), dest, ec);
+        } else {
+            std::filesystem::copy(entry.path(), dest, ec);
+        }
+        if (ec) {
+            return;
+        }
+    }
+}
+
+}  // namespace
+
 bool Repo::clone_root_metadata() {
     auto & logger = *p_impl->base->get_logger();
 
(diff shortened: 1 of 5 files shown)


libdnf5/utils/fs/utils.cpp
```diff
@@ -134,6 +134,28 @@ void reflink_or_copy(const stdfs::path & src, const stdfs::path & dest, std::err
...
+    for (auto end = stdfs::directory_iterator(); it != end; it.increment(ec)) {
+        if (ec) {
+            return;
+        }
+        auto dest = dest_dir / it->path().filename();
+        if (it->is_regular_file(ec)) {
+            reflink_or_copy(it->path(), dest, ec);
+        } else {
+            stdfs::copy(it->path(), dest, ec);
+        }
```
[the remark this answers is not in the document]
m-blaha: I'm not sure this is worth fixing. We still resort to `stdfs::copy` which might either succeed, fail with the same reason or fail differently. We can get slightly misleading error message, but data integrity is not compromised in any case.

libdnf5/utils/fs/utils.cpp
```diff
@@ -134,6 +134,28 @@ void reflink_or_copy(const stdfs::path & src, const stdfs::path & dest, std::err
     stdfs::copy(src, dest, stdfs::copy_options::overwrite_existing, ec);
 }
 
+void reflink_or_copy_dir(const stdfs::path & src_dir, const stdfs::path & dest_dir, std::error_code & ec) noexcept {
```
jrohel: Very interesting function. **It copies the contents of the nested directory as well - but only the files.** It doesn't go any deeper than that. Is this intentional? If so, it definitely needs a prominent warning in the documentation.

Additionally, the reflink feature is not applied to files within the nested directory.

Personally, I don't really like this behavior.

m-blaha: The regular file type check is intentional, any other type falls back to `std::filesystem::copy`, which I think makes sense.
You are right that the reflink optimization not being applied recursively to files inside subdirectories is not great design. Though in practice the current callers only need to handle flat `solv/` and `repodata/` directories. I can drop the function from the utils header for now, and move it to an ad hoc flat helper in repo.cpp. I'm not sure it's worth complicating it until we have another use-case for it.

m-blaha: @jrohel I've removed the reflink_or_copy_dir from utils and replaced it with `reflink_or_copy_cache` helper directly in repo.cpp. we know that cache dirs are flat, so no recursion included.

libdnf5/repo/repo.cpp
```diff
@@ -544,6 +545,26 @@ void Repo::internalize() {
...
 
+namespace {
+
+// Reflink-or-copy every regular file from a flat cache directory into `dest_dir`.
+void reflink_or_copy_cache(
+    const std::filesystem::path & src_dir, const std::filesystem::path & dest_dir, std::error_code & ec) {
+    for (const auto & entry : std::filesystem::directory_iterator(src_dir, ec)) {
+        if (ec) {
+            return;
+        }
```
ppisar: Is this fist check for _ec_ necessary? std::filesytem::directory_iterator() returns file::system::end() on a failure, skipping the loop body. I recommend removing this branch.
If you want to return from reflink_or_copy_cache() early if non-False _ec_ argument was passed, you should move the branch before the _for_ loop.

m-blaha: Right, sorry.

ppisar (approved):
Good.

PR #2811, opened 2026-07-08
Add --[no-]allow-vendor-change option

Conan-Kudo:
Since allowing vendor changes is now suggested by DNF in certain scenarios where it is disabled by default, this makes it a bit more ergonomic to control.

@jrohel 
Updated/extended CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1917

### DIFF
--- a/dnf5/main.cpp
+++ b/dnf5/main.cpp
@@ -335,6 +335,22 @@ void RootCommand::set_argument_parser() {
 
     no_best->add_conflict_argument(*best);
 
+    auto allow_vendor_change = parser.add_new_named_arg("allow-vendor-change");
+    allow_vendor_change->set_long_name("allow-vendor-change");
+    allow_vendor_change->set_description(_("allow automatic package replacements from different vendors"));
+    allow_vendor_change->set_const_value("true");
+    allow_vendor_change->link_value(&config.get_allow_vendor_change_option());
+    global_options_group->register_argument(allow_vendor_change);
+
+    auto no_allow_vendor_change = parser.add_new_named_arg("no-allow-vendor-change");
+    no_allow_vendor_change->set_long_name("no-allow-vendor-change");
+    no_allow_vendor_change->set_description(_("do not allow automatic package replacements from different vendors"));
+    no_allow_vendor_change->set_const_value("false");
+    no_allow_vendor_change->link_value(&config.get_allow_vendor_change_option());
+    global_options_group->register_argument(no_allow_vendor_change);
+
+    no_allow_vendor_change->add_conflict_argument(*allow_vendor_change);
+
     {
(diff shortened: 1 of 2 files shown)


evan-goode:
Bear with us, we're in the process of swapping out gemini-code-assist...

dnf5/main.cpp
```diff
@@ -335,6 +335,22 @@ void RootCommand::set_argument_parser() {
 
     no_best->add_conflict_argument(*best);
 
+    auto allow_vendor_change = parser.add_new_named_arg("allow-vendor-change");
+    allow_vendor_change->set_long_name("allow-vendor-change");
+    allow_vendor_change->set_description(_("allow automatic package replacements from different vendors"));
```
keszybz, line 340: Maybe "allow automatic installation of upgraded package from a different vendor" ?

keszybz (commented):
Very welcome change in general. Maybe the wording could be adjusted a bit to make it clearer.

dnf5/main.cpp
Conan-Kudo: It affects upgrades and downgrades, hence I was a little less specific on purpose.

keszybz: "allow automatic installation of replacing package from a different vendor" ?

Conan-Kudo: That sounds more awkward to me as an English speaker...

dnf5/main.cpp
```diff
@@ -1160,7 +1176,7 @@ static void print_resolve_hints(dnf5::Context & context) {
         }
 
         if (!conf.get_allow_vendor_change_option().get_value() && vendor_change) {
-            const std::string_view arg{"--setopt=allow_vendor_change=true"};
+            const std::string_view arg{"--allow-vendor-change"};
```
jrohel, line 1179: This line changes the output tested in CI tests. Therefore, the DNF5 CI tests are failing. A PR updating the text in the CI tests needs to be created as well.

jrohel (commented):
The new `--[no-]allow-vendor-change` options are simply shortcuts for `setopt=allow_vendor_change=1/0`. I considered adding a config file with argument aliases rather than bloating the source code, but the PR includes documentation as well. These options seem important and are similar in nature to `--[no-]best`, which are also hardcoded. Weighing all the factors, I agree with this solution.

The only thing left is to update the CI tests to account for the new hint message text.

jrohel:
@Conan-Kudo I am going to update the CI tests.

jrohel (approved):
I've made the necessary adjustments and extended the CI tests, so I can now approve this PR.

dnf5/main.cpp
evan-goode: maybe "allow replacements with packages from different vendors"?

PR #2812, opened 2026-07-08
Distinguish between execute_transaction and execute_trusted_transaction polkit actions

mavit:
Administrators shouldn’t need to go digging through the source code to understand the difference between these actions.

### DIFF
--- a/dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
+++ b/dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
@@ -27,8 +27,8 @@
   </action>
 
   <action id="org.rpm.dnf.v0.rpm.execute_transaction">
-    <description>Execute rpm transaction</description>
-    <message>Authentication is required to execute an rpm transaction.</message>
+    <description>Execute untrusted rpm transaction</description>
+    <message>Authentication is required to execute an rpm transaction with local or signature-unverified packages.</message>
     <defaults>
         <allow_any>auth_admin</allow_any>
         <allow_inactive>auth_admin</allow_inactive>
--- a/dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
+++ b/dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
@@ -18,7 +18,7 @@
 
   <action id="org.rpm.dnf.v0.rpm.execute_trusted_transaction">
     <description>Execute rpm transaction</description>
-    <message>Authentication is required to execute a rpm transaction.</message>
+    <message>Authentication is required to execute an rpm transaction.</message>
     <defaults>
         <allow_any>auth_admin</allow_any>
         <allow_inactive>auth_admin</allow_inactive>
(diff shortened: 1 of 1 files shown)


dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
```diff
@@ -18,7 +18,7 @@
 
   <action id="org.rpm.dnf.v0.rpm.execute_trusted_transaction">
     <description>Execute rpm transaction</description>
-    <message>Authentication is required to execute a rpm transaction.</message>
+    <message>Authentication is required to execute an rpm transaction.</message>
```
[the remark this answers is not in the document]
mavit: I disagree.  Trusted by who, for what?  Safer to simply avoid the word in user-facing messages where possible.

dcantrell (dismissed):

dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
```diff
@@ -27,8 +27,8 @@
   </action>
 
   <action id="org.rpm.dnf.v0.rpm.execute_transaction">
-    <description>Execute rpm transaction</description>
-    <message>Authentication is required to execute a rpm transaction.</message>
+    <description>Execute untrusted rpm transaction</description>
+    <message>Authentication is required to execute an rpm transaction from local or signature-unverified repositories.</message>
```
kontura: Thank you for working on enhancing the docs.

Looking at the code: https://github.com/rpm-software-management/dnf5/blob/main/dnf5daemon-server/services/goal/goal.cpp#L359 I see two things:
1. Authentication is not really required for all local repositories (local repos are treated the same way as remote ones) but rather for packages from a special `COMMANDLINE` repo. I am not sure how much sense the name makes in the context of dnf5daemon but for dnf these are packages passed in direcly: `dnf install ./package.rpm` they are basically packages that are not in any repository. 
2. 'signature-unverified repositories' is also not 100% clear, repository can have signed metadata and/or signed packages, trusted transaction only requires all packages to be from a repo with pkg_gpgcheck option enabled.

Could you amend the text to perhaps something like: `Authentication is required to execute an rpm transaction if it contains any packages from repositories with disabled pkg_gpgcheck option or any packages specified by rpm file path (not present in any repository).

kontura (changes requested):

dnf5daemon-server/polkit/org.rpm.dnf.v0.policy
mavit: > Thank you for working on enhancing the docs.

This isn't documentation, it's user interface.

> Looking at the code: https://github.com/rpm-software-management/dnf5/blob/main/dnf5daemon-server/services/goal/goal.cpp#L359 I see two things:
> 
>     1. Authentication is not really required for all local repositories (local repos are treated the same way as remote ones) but rather for packages from a special `COMMANDLINE` repo. I am not sure how much sense the name makes in the context of dnf5daemon but for dnf these are packages passed in direcly: `dnf install ./package.rpm` they are basically packages that are not in any repository.
> 
>     2. 'signature-unverified repositories' is also not 100% clear, repository can have signed metadata and/or signed packages, trusted transaction only requires all packages to be from a repo with pkg_gpgcheck option enabled.

Fair criticism.

> Could you amend the text to perhaps something like: `Authentication is required to execute an rpm transaction if it contains any packages from repositories with disabled pkg_gpgcheck option or any packages specified by rpm file path (not present in any repository).

I can write something like that, sure, but it seems very verbose and technical for a user-facing dialog box.  Are we sure that's the right thing to do?

kontura: I am not sure but I do also like the text you just updated the PR with so I would go with that.

kontura (approved):

Issue #2797, opened 2026-06-24
rpm-6.0.91: Lua print() output of a scriptlet is not printed


ppisar:
After upgrading rpm-libs from 0:6.0.1-6.fc45.x86_64 to 0:6.0.91-1.fc45.x86_64 dnf/scriptlets.feature fails, e.g.:

~~~~
  Scenario Outline: Install a pkg with a successful scriptlet -- @1.1   # dnf/scriptlets.feature:22
    Given I use repository "scriptlets"                                 # dnf/steps/repo.py:185
    And I set environment variable "FORCE_COLUMNS" to "400"             # dnf/steps/cmd.py:52
    When I execute dnf with args "install Package-pre-ok"               # dnf/steps/cmd.py:109
    Then the exit code is 0                                             # common/output.py:46
    And Transaction is following                                        # dnf/steps/transaction.py:158
      | Action  | Package                       |
      | install | Package-pre-ok-0:1.0-1.x86_64 |
    And stdout contains "pre scriptlet successfully done"               # dnf/steps/cmd.py:217
      ASSERT FAILED: Stdout doesn't contain: pre scriptlet successfully done

----
CAPTURED STDOUT: scenario

Last Command: dnf5 -y --installroot=/tmp/dnf_ci_installroot_e7hltvsb --releasever=29 --setopt=module_platform_id=platform:f29 --disable-plugin='*' --setopt=cachedir=/tmp/dnf_ci_installroot_e7hltvsb/var/cache/dnf install Package-pre-ok

Last Command stdout:
Package         Arch   Version Repository      Size
Installing:
 Package-pre-ok x86_64 0:1.0-1 scriptlets   0.0   B

Transaction Summary:
 Installing:         1 package

Last Command stderr:
Updating and loading repositories:
 scriptlets test repository                                                                                                                                                                                                                                                                                                                                             100% |   4.6 MiB/s |   4.7 KiB |  00m00s
Repositories loaded.
Total size of inbound packages is 6 KiB. Need to download 6 KiB.
After this operation, 0 B extra will be used (install 0 B, remove 0 B).
[1/1] Package-pre-ok-0:1.0-1.x86_64                                                                                                                                                                                                                                                                                                                                     100% |   0.0   B/s |   6.2 KiB |  00m00s
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #2815, opened 2026-07-13
Suppress transient "Signing key not found" error from progress bar

### LINKED ISSUE #2810: Suppress resolvable keyring errors
When running `dnf5` (e.g. for the first time), errors will be shown in red saying that the signing keys are not found when trying to resolve repos with non-local keys.

We think this error message should be suppressed because this causes confusion to users. We constantly receive questions on these error messages.

m-blaha:
When a repository has repo_gpgcheck=1 and the signing key hasn't been imported yet, the progress bar showed a confusing error message `repomd.xml GPG signature verification error: Signing key not found` before the user was prompted to import the key. This error is transient — it is resolved once the key import succeeds.

Suppress the error from the download callbacks only during the first download pass when key import will be attempted (`import_keys` is true). The suppression is controlled by a flag on RepoDownloader that is propagated through CallbackData to the end_cb callbacks. On the second pass (after key import was attempted/refused), the flag is false and
the error is shown normally.

Extract is_signing_key_not_found_error() helper into repo_pgp to share the string matching between the callback suppression and the existing handle_repo_exception() check.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2810


test/data/keys/repo-gpg-key.pub
```diff
@@ -0,0 +1,18 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
ppisar, line 1: This is 2048-bit RSA key. This is not strong enough according to some standards and systems adhering to that standard will reject the key. Please use 4096-bit key.

test/data/repos-repomd/repomd-repo1-gpg/repodata/repomd.xml
```diff
@@ -0,0 +1,35 @@
...
+    <open-size>52054</open-size>
+  </data>
+  <data type="updateinfo">
+    <checksum type="sha256">a1732db95964e07fef567552ca2c895702bdd02aba7110c855d85d2f6378c95c</checksum>
+    <open-checksum type="sha256">a1732db95964e07fef567552ca2c895702bdd02aba7110c855d85d2f6378c95c</open-checksum>
+    <location href="repodata/updateinfo.xml" />
+    <timestamp>1597222003</timestamp>
+    <size>646</size>
+    <open-size>2510</open-size>
+  </data>
```
ppisar: If this repository is only used for the two new tests, only primary.xml is required and the other XML files should be removed to make the test as small as possible.

ppisar (commented):
Besides the in-line comments I left, could you reorder (or squash) the commits, so that bisecting the history which stops between the two commits does not report a test failure?

Otherwise it's good.

libdnf5/repo/repo_downloader.hpp
```diff
@@ -80,6 +80,8 @@ class RepoDownloader {
     // Download the previously added repos.
     std::unordered_map<Repo *, std::vector<std::string>> download();
 
+    void set_suppress_keyring_errors(bool suppress) { suppress_keyring_errors = suppress; }
+
```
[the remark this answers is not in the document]
m-blaha: Yes, but that's how callback data works...

test/data/repos-repomd/repomd-repo1-gpg/repodata/primary.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<metadata xmlns="http://linux.duke.edu/metadata/common" xmlns:rpm="http://linux.duke.edu/metadata/rpm" packages="1">
```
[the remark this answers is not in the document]
m-blaha: will fix that.

test/data/repos-repomd/repomd-repo1-gpg/repodata/repomd.xml
m-blaha: Good point. Will remove them.

m-blaha:
@ppisar squashed into one commit and resolved your comments.

ppisar (approved):
Good. The test failure with RPM 6.1 is unrelated (#2797).

PR #2820, opened 2026-07-17
Remove modularity support from advisories if compiled without modulemd

ppisar:
"dnf5 advisory info --json" used to print "modules" collections in the
JSON output even if DNF5 was built with -DWITH_MODULEMD=OFF.

This patch stops processing and reporting modular data from updateinfo
repository metadata if modularity is disabled at build time. This
involes all "dnf5 advisory" outputs, documentation, D-Bus API, and
libdnf5 API.

To preserve public ABI and API, all modularity classes and class
members were preserved. But AdvisoryCollection::get_modules() now
raises a libdnf5::RuntimeError exception to distinguish from an empty
list of modules.

For the very same reason the modularity code could not been removed
from the D-Bus daemon (it extends libdnf5 API using subclassed templates
and thus needs to implement get_modules() method).

This patch also adapts the tests to expect the exception if modularity
is disabled.

Requires: #2806


ppisar:
Rebased.

kontura (approved):

PR #2821, opened 2026-07-17
Disable modular filtering if modularity support is disabled

ppisar:
This patch disables modular filtering code, "dnf5 repoquery
--disable-modular-filtering" option, and modular filtering from
a documenation if DNF5 is built without modularity support.

A configuration parsers of the actions plugin is kept to ignore
IGNORE_MODULAR_EXCLUDES parameter in order not break old configuration
files.

Requires: #2806


fhbash (approved):
LGTM, thanks!

PR #2823, opened 2026-07-17
Disuse modular configuration options if built without modularity

ppisar:
This patch set hides and stops processing these configuration options:

module_obsoletes
module_platform_id
module_stream_switch

if DNF5 is built without modularity support. Documentation and D-Bus interface is also adapted.

However, accessors to these options are retained in API to preserve compatibility.

Requires: #2806
Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1921


kontura (approved):
Thank you

PR #2829, opened 2026-07-21
zsh-completion: Escape colons in completion candidates

LuanVSO:
prevent full nevra package lines from having a description, which would cause them to rank above package name only lines and have a -EPOCH appended to the package name when selected

### DIFF
--- a/dnf5/zsh-completion/_dnf5
+++ b/dnf5/zsh-completion/_dnf5
@@ -8,6 +8,7 @@ _dnf5() {
         [[ -z "$line" ]] && continue
         word="${line%%[[:space:]]*}"
         [[ -z "$word" ]] && continue
+        word="${word//:/\\:}"
         if [[ "$line" =~ '\((.+)\)$' ]]; then
             entry="${word}:${match[1]}"
         else


m-blaha (commented):
Except for typos in the commit message, it looks good:
characher -> character
comming -> coming

LuanVSO:
> Except for typos in the commit message, it looks good: characher -> character comming -> coming

fixed, sorry 😅

m-blaha (approved):
Thank you!
  
One small ask for future contributions: it helps a lot when the commit message describes what was going wrong and why the fix works. For example here, something like:

```
zsh-completion: Escape colons in completion candidates

zsh's _describe uses : as the delimiter between a completion value
and its description (e.g. candidate:help text). When dnf5 outputs a full
NEVRA like glibc-6:2.41-6.fc43.x86_64, the epoch colon gets
misinterpreted as that delimiter, causing the wrong text to be inserted on
selection.
Escape : as \: so _describe treats it as a literal.
```

That way the commit is self-documenting for anyone reading git log later and doesn't require them to reverse-engineer the reasoning. No big deal for this one, something to keep in mind going forward.

LuanVSO:
> ```
> zsh-completion: Escape colons in completion candidates
> 
> zsh's _describe uses : as the delimiter between a completion value
> and its description (e.g. candidate:help text). When dnf5 outputs a full
> NEVRA like glibc-6:2.41-6.fc43.x86_64, the epoch colon gets
> misinterpreted as that delimiter, causing the wrong text to be inserted on
> selection.
> Escape : as \: so _describe treats it as a literal.
> ```

i agree but some projects don't like long commit messages so i try to limit it a bit (i guess too much)
I’ll push it again with your commit message, since it’s quite thorough.

m-blaha:
> i agree but some projects don't like long commit messages so i try to limit it a bit (i guess too much) I’ll push it again with your commit message, since it’s quite thorough.

No worries and thanks again for contributing!

m-blaha:
The failing test is unrelated to this change.

Issue #2408, opened 2025-08-26
[FEATURE] set systemd inhibitor during transactions


jcpunk:
When DNF is actually installing the packages and running scripts/triggers, it would be helpful if a systemd inhibitor was set.  This would prevent a reboot of the system during sensitive operations that might result in an unusable system.

kontura:
There is an rpm plugin for this: `rpm-plugin-systemd-inhibit`.
Like you mentioned this is especially important during the transaction operations, these are performed by rpm.
I believe the plugin will work also when rpm is used via dnf.

ppisar:
Fedora bug related to the reboot inhibition when rebuilding 3rd-party kernel modules with akmods <https://bugzilla.redhat.com/show_bug.cgi?id=2394774>. I recommended akmods to use post_transaction hook of libdnf5 actions plugin. DNF5 should hold the inhibition not only when saving system state but also when running the plugin hooks.

PR #2832, opened 2026-07-23
Call post_transaction plugin hooks in reverse order

m-blaha:
Plugins are loaded and called in alphabetical order by their config filename. By reversing post_transaction iteration, pre_transaction and post_transaction form symmetric brackets: a plugin that acts first in pre_transaction acts last in post_transaction.

### DIFF
--- a/libdnf5/plugin/plugins.cpp
+++ b/libdnf5/plugin/plugins.cpp
@@ -279,9 +279,9 @@ void Plugins::pre_transaction(const libdnf5::base::Transaction & transaction) {
 }
 
 void Plugins::post_transaction(const libdnf5::base::Transaction & transaction) {
-    for (auto & plugin : plugins) {
-        if (plugin->get_enabled()) {
-            plugin->post_transaction(transaction);
+    for (auto plugin = plugins.rbegin(), stop = plugins.rend(); plugin != stop; ++plugin) {
+        if ((*plugin)->get_enabled()) {
+            (*plugin)->post_transaction(transaction);
         }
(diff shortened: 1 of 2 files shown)


m-blaha:
I'm thinking about how to resolve https://github.com/rpm-software-management/dnf5/issues/2408 and one of options is to create a libdnf5 plugin. For this I need the lock to be acquired early and released as late as possible (ideally first and last of all plugins). For such use-cases, the order of post_transaction execution needs to be reversed.

jrohel:
I understand your specific use case, but I don't like this change in general.

Example: A user has a libdnf5 plugin that only handles the `post_transaction` callback and is run as one of the first. After this change, it will be run as one of the last. An unpleasant change.

Not everyone needs to handle callbacks in pairs (`pre-`, `post-`) and may be surprised that their `post-` handling was done at a different time. Moreover, there are more paired callbacks: `pre_base_setup` - `post_base_setup`, `pre_add_cmdline_packages` - `post_add_cmdline_packages`. Should we also reverse the order in `post_base_setup` and `post_add_cmdline_packages`?

One more universal solution to reverse or even arbitrarily change the order in `post-transaction` or any other callback is to split the plugin into two parts (two plugins). Each can have a different loading order and handle different callbacks. But I understand that it doesn't look very elegant.

m-blaha:
@jrohel that's fair point. I'm probably going to implement the systemd inhibitor using only `pre_pransaction` callback, so the post_transaction ordering is not important.

PR #2834, opened 2026-07-23
Move integration tests from ci-dnf-stack

fhbash:
This commit moves all dnf5 integration tests from ci-dnf-stack repo to dnf5 repo. This will simplify our CI and will keep tests and code in sync.

All integration tests will leave into `integration-tests` folder, so we keep `test` folder as original, with no need to change the unit test was built before.

Created an Makefile to simplify the tests calling, and added `make rpms-mock` as an alternative target. It uses Mock for reproducible builds and supports building for a different Fedora version via MOCK_CONFIG:

```
make rpms-mock                                    # default: fedora-rawhide-x86_64
make rpms-mock MOCK_CONFIG=fedora-43-x86_64       # target a specific version
```

The existing make rpms (rpmbuild) is kept as the simpler/faster option that doesn't require mock to be installed.

Modified packit config to include new tests.


m-blaha:
Only few questions (I haven't checked the diff yet):
- Should we drop test scenarios relevant only for dnf4 and it's plugins?
- What about createrepo_c tests?

fhbash:
> Only few questions (I haven't checked the diff yet):
> 
> * Should we drop test scenarios relevant only for dnf4 and it's plugins?
> * What about createrepo_c tests?

Tests for createrepo_c, librepo, and any other non-dnf5 components should stay in the main branch of ci-dnf-stack. The dnf4 tests still lives int dnf-4-stack branch of ci-dnf-stack and will not be affected.

this change is only for `dnf5` and `dnf5daemon`.

.pre-commit-config.yaml
```diff
@@ -23,6 +23,7 @@ repos:
   hooks:
   - id: autopep8
     args: [-i, --max-line-length=120]
+    exclude: ^integration-tests/
```
evan-goode: I mean, why not run autopep8 on the integration tests too (eventually)?

Makefile
```diff
@@ -0,0 +1,70 @@
...
+	cmake --build $(BUILD_DIR) -j$(NPROC)
+
+test-unit: build
+	cd $(BUILD_DIR) && ctest --output-on-failure $(CTEST_ARGS)
+
+test: test-unit
+
+rpms:
+	rm -f $(RPMS_DIR)/*.rpm
+	git archive --format=tar.gz --prefix=dnf5-$(VERSION)/ HEAD -o /tmp/dnf5-$(VERSION).tar.gz
```
evan-goode: We should build the SRPM in a separate Make job, then `make rpms` will be smart enough to not rerun when the SRPM doesn't change.

evan-goode: The git archive will not include any uncommitted changes, so I'd prefer if `make rpms`/`make srpm` errored when the working directory is not clean.

Makefile
```diff
@@ -0,0 +1,70 @@
...
+	rpmbuild -bb dnf5.spec \
```
evan-goode, line 68: What about using Mock to build these RPMs more reproducibly? That would also allow locally building RPMs for a different Fedora version than the host, and running the integration tests on that Fedora version.

README.md
```diff
@@ -90,6 +90,105 @@ Developing DNF5 by building and testing the code
 
 For details about building and testing DNF5, see the [Development environment setup](CONTRIBUTING.md#setting-up-a-development-environment) section.
 
+
```
evan-goode: Maybe this section should be incorporated into CONTRIBUTING.md instead.

evan-goode (commented):
Thanks for working on this, I think it will be great to be able to work on dnf5 and its tests in the same PR.

I didn't review this in detail either, but to me it's going in the right direction. The top-level "convenience Makefile" is interesting. Admittedly it doesn't look super idiomatic in a CMake project like this, but the repo should include _something_ that lets you build+run integration tests in a single command. Ideally without rebuilding the dnf5 RPMs if dnf5 hasn't changed.

Makefile
fhbash: Done — split into separate `srpm` and `rpms` targets. The tarball is now stored in `integration-tests/.srpm/` so make can skip rebuilding when the source hasn't changed.

fhbash: Done — added a `check-clean-tree` target that `srpm` depends on. It checks `git status --porcelain` and errors with a clear message if there are uncommitted changes.

.pre-commit-config.yaml
fhbash: yeah, I need to remove this, thanks!

I added there just to this initial draft, because for some reason, autopep8 reorder the all imports for all python files.. and some of this reorder fails in python 3.14.. one solution is put `#noqa E402` on the line to not reorder.

Makefile
fhbash: Added `make rpms-mock` as an alternative target. It uses Mock for reproducible builds and supports building for a different Fedora version via `MOCK_CONFIG`:

```
make rpms-mock                                    # default: fedora-rawhide-x86_64
make rpms-mock MOCK_CONFIG=fedora-43-x86_64       # target a specific version
```

The existing `make rpms` (rpmbuild) is kept as the simpler/faster option that doesn't require mock to be installed.

fhbash:
> but the repo should include _something_ that lets you build+run integration tests in a single command. Ideally without rebuilding the dnf5 RPMs if dnf5 hasn't changed.

@evan-goode 
Agreed, my thoughts are: after `make rpms`, save the current commit hash in `integration-tests/.srpm/.built-commit`, after rebuild check if the current HEAD is the same of that saved stamp, if yes skip the full rebuild.

the workflow will be like this:
- `make test-integration`   first time: build RPMs + container + run 
if you edit some `.feature` file:
- `make test-integration`   RPMs build will be skipped, container rebuild + run
if you change some c++ code and commit:
- `make test-integration`   detects new commit, rebuild the rpms + container + run

what you think?

mfocko:
> I didn't review this in detail either, but to me it's going in the right direction. The top-level "convenience Makefile" is interesting. Admittedly it doesn't look super idiomatic in a CMake project like this, but the repo should include something that lets you build+run integration tests in a single command. Ideally without rebuilding the dnf5 RPMs if dnf5 hasn't changed.

It feels like `make` is used mostly like a “runner”. I would advocate for [just](https://just.systems/man/en/introduction.html) here :D it can also be used to auto-generate the help, so it doesn’t need that much of a maintenance

(I’ll try to drop a patch by tomorrow, right now it’s relatively simple, so it shouldn’t be hard to incorporate now rather than later, if it makes sense)

evan-goode:
>  Agreed, my thoughts are: after make rpms, save the current commit hash in integration-tests/.srpm/.built-commit, after rebuild check if the current HEAD is the same of that saved stamp, if yes skip the full rebuild.

I think that would work if the rule building the .built-commit doesn't touch .built-commit unless it changed, since Make works on file timestamps not hashes.

> make test-integration first time: build RPMs + container + run

IMO test-integration should use `rpms-mock`.

> if you edit some .feature file:
> make test-integration RPMs build will be skipped, container rebuild + run

There are also plenty of other things in `integration-tests` that affect test output. The steps, spec files, even container-test itself. Safest would be to always rebuild test container when `integration-tests` has changed. If we require a clean worktree then we could use `git rev-parse HEAD:integration-tests`, I think. 

We could also run `container-test` with `-d` to mount the feature files into the container, then we wouldn't have to rebuild the container when feature files are changed. That would require some fancier change-detection of `integration-tests` than the `git rev-parse`.

> if you change some c++ code and commit:
> make test-integration detects new commit, rebuild the rpms + container + run

> It feels like make is used mostly like a “runner”. I would advocate for [just](https://just.systems/man/en/introduction.html) here :D it can also be used to auto-generate the help, so it doesn’t need that much of a maintenance

Generally my opinion of just is rather negative. Their tagline is "just is not a build system". But just _is_ a build system---a bad one---where if you want any incrementalism you have to hack it together yourself. Which is basically what we are doing here with Makefile too, with the HEAD checking and all. Turns out I'm not really a fan of either Make or just... and I would hesitate to pull in redo or Shake here...

So I actually sort of agree with you :) Just is certainly more ergonomic and would look less out of place. https://github.com/bootc-dev/bootc uses just for a very similar use case. I am curious to hear others' opinions on Make vs just.

Though I would still really appreciate some kind of change detection/smart rebuilding. If the "build RPMs" step is not incremental, then developers will not wait for the RPMs to build every time they run the tests. They will develop the habit of building the RPMs, then running the integration tests without change detection, and occasionally they will change some code and forget to rebuild the RPM, and waste lots of time being confused why the test results don't match what they expect.

Fortunately here we have a very small (high-level) build graph so writing an incremental justfile should be easy enough.

I'm only speaking passionately about this because of my personal experience; after I set up this kind of high-level incremental build system for myself (https://github.com/evan-goode/rsm-pseudomonorepo) it was night and day, I spent way less time managing test runs and silly mistakes.

fhbash:
> > Agreed, my thoughts are: after make rpms, save the current commit hash in integration-tests/.srpm/.built-commit, after rebuild check if the current HEAD is the same of that saved stamp, if yes skip the full rebuild.
> 
> I think that would work if the rule building the .built-commit doesn't touch .built-commit unless it changed, since Make works on file timestamps not hashes.
> 
> > make test-integration first time: build RPMs + container + run
> 
> IMO test-integration should use `rpms-mock`.
> 
> > if you edit some .feature file:
> > make test-integration RPMs build will be skipped, container rebuild + run
> 
> There are also plenty of other things in `integration-tests` that affect test output. The steps, spec files, even container-test itself. Safest would be to always rebuild test container when `integration-tests` has changed. If we require a clean worktree then we could use `git rev-parse HEAD:integration-tests`, I think.
> 
> We could also run `container-test` with `-d` to mount the feature files into the container, then we wouldn't have to rebuild the container when feature files are changed. That would require some fancier change-detection of `integration-tests` than the `git rev-parse`.
> 

It's make sense your suggestions here, I'll implement on next commit so we can test it and see.

Regarding Makefile vs Just, have no strong opinion in this case, I'm good with both approaches.

ppisar:
I would prefer make. Otherwise you might end up maintaining _just_ in EPEL or RHEL in case RHEL QE people will need to run these tests.

fhbash:
Updated with the latest change from ci-stack-dnf

m-blaha (approved):
Thank you!

PR #2836, opened 2026-07-24
Add systemd inhibitor lock plugin

### LINKED ISSUE #2408: [FEATURE] set systemd inhibitor during transactions

m-blaha:
Adds a new plugin API 2.2 (`IPlugin2_2`) with cleanup hooks and a systemd inhibitor lock plugin that uses them.

New plugin interface version with three cleanup hooks guaranteed to run regardless of whether the operation succeeded or failed:

- `post_base_setup_cleanup() noexcept`
- `post_add_cmdline_packages_cleanup() noexcept`
- `post_transaction_cleanup(const Transaction &) noexcept`

Libdnf5 plugin that acquires a systemd logind inhibitor lock (`idle:sleep:shutdown`) during package transactions to prevent system shutdown, sleep, or idle while the system is in a critical state. The lock covers plugin pre/post transaction hooks, RPM transaction execution, system state updates, and history database writes.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2408


libdnf5/plugin/systemd_inhibitor.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+// Keyed by the address of the libdnf5::base::Transaction object passed to a
+// plugin's pre_transaction()/post_transaction() hooks - the same object for
+// both calls during one transaction run.
+std::mutex g_systemd_inhibitor_fds_mutex;
+std::unordered_map<const void *, int> g_systemd_inhibitor_fds;
+
+}  // namespace
+
+void register_systemd_inhibitor_fd(const libdnf5::base::Transaction & transaction, int fd) {
```
jrohel: The method name is `register_systemd_inhibitor`, but in fact, it expects any standard file descriptor. When a new file descriptor is received, the previously registered one is closed. And `close_systemd_inhibitor_fd` calls `close`.

Is that even necessary? Wouldn't it be enough to put `close` into the plugin's `finish` callback method? Or maybe into the plugin destructor - that's a place where the plugin should no longer interact with `libdnf5`, but closing its file descriptor shouldn't be a problem. Just in case you wanted to do the close as late as possible. But even `finish` runs later than the current solution.

jrohel:
Great! You implemented the idea we discussed last time -> adding *_cleanup methods to the libdnf5 plugin interface. The advantage of this solution is that these new methods can be used by other plugins in the future as well. I'll still go through the code.

libdnf5-plugins/systemd-inhibit/systemd-inhibit.cpp
```diff
@@ -0,0 +1,118 @@
...
+        logger.warning("Failed to acquire systemd inhibitor lock: {}", ex.what());
+    }
+}
+
+void SystemdInhibit::post_transaction_cleanup(const libdnf5::base::Transaction &) noexcept {
+    if (inhibit_fd >= 0) {
+        auto & logger = *get_base().get_logger();
+        logger.debug("Releasing systemd inhibitor lock (fd={})", inhibit_fd);
+        close(inhibit_fd);
+        inhibit_fd = -1;
```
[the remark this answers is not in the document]
m-blaha: I don't think close() can realistically fail. The lock fd is just a reference to a kernel object, not an actual file.

libdnf5-plugins/systemd-inhibit/systemd-inhibit.cpp
```diff
@@ -0,0 +1,118 @@
...
+                std::string{"idle:sleep:shutdown"},
+                std::string{"libdnf5"},
+                std::string{"Package transaction is running"},
+                std::string{"block"})
+            .storeResultsTo(fd);
+        inhibit_fd = fd.release();
+        logger.debug("Acquired systemd inhibitor lock (fd={})", inhibit_fd);
+    } catch (const sdbus::Error & ex) {
+        logger.warning("Failed to acquire systemd inhibitor lock: {}", ex.what());
+    }
```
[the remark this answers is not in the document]
m-blaha: Right. I'll broaden the caught exception class.

doc/tutorial/plugins/libdnf5-plugins.rst
jrohel, line 18: I expected a code block to follow this paragraph. However, "Plugin Interface Versions" was inserted before it.

m-blaha: oops, right! i'll fix that.

m-blaha: Actually, after thinking about it once more, I'd rather not complicate the libdnf5 plugin tutorial with these details. I'll only add a link to API reference where the user can read details about interface versioning.

jrohel (approved):
LGTM. Thanks

PR #2837, opened 2026-07-24
Set systemd inhibitor lock during transaction

### LINKED ISSUE #2408: [FEATURE] set systemd inhibitor during transactions

m-blaha:
Acquire a systemd logind inhibitor lock to prevent system shutdown/reboot while a package transaction is in progress. The lock is held from just before pre_transaction plugin hooks run through the end of post_transaction hooks, covering RPM execution, system state update, and history database writes.

libsystemd is loaded on demand via dlopen (through the existing utils::Library helper used for plugin loading) and only the raw sd_bus C API is used, so libdnf5 gains no link-time or runtime dependency on it. If libsystemd is unavailable, or the lock cannot be acquired for any other reason, the failure is logged and the transaction proceeds normally.

Resolves: https://github.com/rpm-software-management/dnf5/issues/2408

### DIFF
--- a/doc/dnf5.conf.5.rst
+++ b/doc/dnf5.conf.5.rst
@@ -300,6 +300,16 @@ repository configuration file should aside from repo ID consists of baseurl, met
 
     Default: ``False``.
 
+.. _inhibit_shutdown_options-label:
+
+``inhibit_shutdown``
+    :ref:`boolean <boolean-label>`
+
+    If enabled, DNF5 will hold a systemd inhibitor lock to prevent the system from
+    shutting down or rebooting while a package transaction is in progress.
+
+    Default: ``True``.
+
 .. _installonlypkgs_options-label:
 
 ``installonlypkgs``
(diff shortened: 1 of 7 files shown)


bluca:
Please consider setting a dlopen elf note to document the optional dependency according to https://uapi-group.org/specifications/specs/elf_dlopen_metadata/
The `systemd/sd-dlopen.h` public header provides a handy macro to do this in one line

m-blaha:
Thanks for the hint! This is PR is more a proof of concept, I'd rather merge the alternative solution using libdnf5 plugin: https://github.com/rpm-software-management/dnf5/pull/2836

m-blaha:
Closed in favor of https://github.com/rpm-software-management/dnf5/pull/2836

PR #2839, opened 2026-07-30
Fix aliging unicode descriptions in a progressbar

### LINKED ISSUE #2838: The progressbar columns are not aligned in locales with non-ASCII characters
中文环境下，不美观，没有对齐

Image

中文：dnf install koji

Image

英文：LANG=C dnf install rpm-build

Image

ppisar:
This patch set fixes #2838.

It adds two new public method:

* `libdnf5::cli::progressbar::ProgressBar::get_description_width()` to obtain a chached column-width of a bar description.
* `libdnf5::cli::progressbar::Widget::to_spanned_string()` to produce properly aligned widgets with respect to multi-byte and wide characters. I added a new method instead of fixing the previously used to_string() function because the old function did not truncated overlong descriptions and changing that would break compability since it is a public API.

I also fixes `libdnf5::cli::progressbar::Widget::to_string()` to correctly count columns.

It also adds a unit test.

### DIFF
--- a/libdnf5-cli/progressbar/progress_bar.cpp
+++ b/libdnf5-cli/progressbar/progress_bar.cpp
@@ -18,10 +18,17 @@
 // along with libdnf.  If not, see <https://www.gnu.org/licenses/>.
 
 
+// For wcwidth()
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE
+#endif
+
 #include "libdnf5-cli/progressbar/progress_bar.hpp"
 
-#include <optional>
+#include <wchar.h>
 
+#include <cwchar>
+#include <optional>
 
 namespace libdnf5::cli::progressbar {
 
@@ -33,8 +40,12 @@ public:
         std::size_t padding;
     };
 
+    void update_description_width() noexcept;
+
     explicit Impl(int64_t total_ticks) : total_ticks{total_ticks} {}
-    Impl(int64_t total_ticks, const std::string & description) : total_ticks{total_ticks}, description{description} {}
+    Impl(int64_t total_ticks, const std::string & description) : total_ticks{total_ticks}, description{description} {
+        update_description_width();
+    }
 
     const MessageMetrics & get_message_metrics(
         std::size_t terminal_width, std::string_view message, std::size_t message_index) noexcept;
@@ -49,6 +60,7 @@ public:
 
     // description
     std::string description;
+    std::size_t description_width = 0;
 
     // messages
     std::vector<ProgressBar::Message> messages;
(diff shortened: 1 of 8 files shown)


libdnf5-cli/progressbar/widgets/widget.cpp
```diff
@@ -20,9 +20,15 @@
...
 
+#include <cwchar>
 #include <iostream>
 #include <string>
 
+// For wcwidth()
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE
+#endif
+#include <wchar.h>
```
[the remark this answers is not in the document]
ppisar: I will move _XOPEN_SOURCE definition at the before including any header files. It's fine to define it without  a value, according to glibc manual.

libdnf5-cli/progressbar/widgets/widget.cpp
```diff
@@ -32,5 +39,44 @@ std::ostream & operator<<(std::ostream & stream, Widget & widget) {
...
+        }
+        start += bytes_consumed;
+        size -= bytes_consumed;
+    }
+    spanned_text = text.substr(0, text.size() - size);
+
+    // Fill the rest of the spanned text with spaces
+    if (text_width < designed_width) {
+        spanned_text.append(designed_width - text_width, ' ');
+    }
```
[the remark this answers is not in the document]
ppisar: Good catch. I will try to catch and handle the integer underflows.

ppisar:
I fixed the two integer underflows and added a test for one of them.

libdnf5-cli/progressbar/widgets/description.cpp
```diff
@@ -34,25 +34,34 @@ std::size_t DescriptionWidget::get_total_width() const noexcept {
...
     }
-    std::ostringstream ss;
-    ss << get_delimiter_before();
-    ss << std::left;
-    ss << std::setw(static_cast<int>(get_total_width() - get_delimiter_before().size()));
-    ss << get_bar()->get_description();
-    return ss.str();
+    std::string output = get_delimiter_before() + get_bar()->get_description();
+    // Pad the text with spaces to fill the designed width.
+    auto text_width = get_delimiter_before().size() + get_bar()->get_description_width();
```
evan-goode, line 59: Theoretically the delimiter could have multibyte chars too but we will probably never do that in practice.

libdnf5-cli/progressbar/progress_bar.cpp
```diff
@@ -119,6 +131,27 @@ const ProgressBar::Impl::MessageMetrics & ProgressBar::Impl::get_message_metrics
     return messages_metrics_cache.at(message_index).value();
 }
 
+void ProgressBar::Impl::update_description_width() noexcept {
+    description_width = 0;
```
evan-goode, line 135: Would be good to have this in `common/utils/string` but doesn't have to be now.

evan-goode (approved):
Thanks!

ペトルさん、あなたの日本語は本当に素晴らしく、尊敬に値するものです。

libdnf5-cli/progressbar/widgets/description.cpp
ppisar: Yes. That's why I did not fixed this. But you are right that set_delimiter_before() is a public method. I will handle it in a different pull request.

libdnf5-cli/progressbar/progress_bar.cpp
ppisar: I also noticed that we implement similar computation at different places. But they are usually for different purposes. I can try later generalize this code, when I have good understanding of requirements on its interface, and move into common/utils/string.

libdnf5-cli/progressbar/widgets/description.cpp
ppisar: Actually that setter is a head-only function. Hooking up a computation of the rendering width would break ABI. So, no I'm not going to do it now.

PR #2841, opened 2026-08-03
conf: Fix OptionStringContainer::add_item and clarify set()/add() semantics for append options, unit tests included

jrohel:
This PR fixes bugs in `OptionStringContainer` that were introduced in commit 26c2b112f ("conf: New classes for append options") and clarifies the semantic difference between `set()` and `add()` methods for append options.

### Problems Fixed

1. **`add_item()` was not updated for new append logic** (commit 26c2b112f)
    - When append option support was added, `add()` method got new priority-based logic
    - `add_item()` was overlooked and continued using old direct insertion
    - This caused `add_item()` to bypass priority handling and not work correctly with append options

2. **Confusing `set()` vs `add()` semantics**
    - Previously, both `set()` and `add()` would clear existing items on empty value/first empty item
    - This made them functionally equivalent for append options, which was semantically incorrect
    - The distinction between "replace" (set) and "append" (add) was lost

### Changes

1. Fix `add_item()` to delegate to `add()`

2. Distinguish set() and add() behavior

   set() - Replace semantics:
   - An empty value or first empty item clears all existing items
   - All empty items are filtered out from the result
  
   add() - Append semantics:
   - Never clears existing items
   - Adds all new items including empty ones

   Examples (OptionStringAppendList option):
   ```
   // set() clears on first empty item and filters all empty items:
   option.set(COMMANDLINE, {""});               // value = [] (cleared!)
   option.set(COMMANDLINE, {"", "pkg1", ""});   // value = ["pkg1"]

   // add() never clears and keeps all items including empty:
   option.add(RUNTIME, {""});                   // value = [""] (empty item appended)
   option.add(RUNTIME, {"", "pkg1", ""});       // value = ["", "pkg1", ""]
   ```

Unit tests included

### DIFF
--- a/libdnf5/conf/option_string_list.cpp
+++ b/libdnf5/conf/option_string_list.cpp
@@ -321,13 +321,8 @@ void OptionStringContainer<T, IsAppend>::add(Priority priority, const std::strin
 
 template <typename T, bool IsAppend>
 void OptionStringContainer<T, IsAppend>::add_item(Priority priority, const std::string & item) {
-    assert_not_locked();
-
-    test_item(item);
-    p_impl->value.insert(p_impl->value.end(), item);
-    if (get_priority() < priority) {
-        set_priority(priority);
-    }
+    T items{item};
+    add(priority, items);
 }
 
 template <typename T, bool IsAppend>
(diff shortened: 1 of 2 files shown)


m-blaha (approved):
Thank you, looks good!

fhbash (approved):
thanks, LGTM

PR #2843, opened 2026-08-03
Fix a crash when a terminal width is forced to negative values

ppisar:
It leads to a crash when the value is used later:

    $ FORCE_COLUMNS=-1 dnf5 -y reinstall dontpanic
    terminate called after throwing an instance of 'std::length_error'
      what():  basic_string::_M_create

    Thread 1 "dnf5" received signal SIGABRT, Aborted.

This patch ignores non-positive values in the FORCE_COLUMNS environment variable and falls back to querying the terminal device.

This patch also guards from a kernel reporting 0 width. Then the code falls back to 80 columns as if the IOCTL failed.

This patch adds a unit test for the crash.

### DIFF
--- a/libdnf5-cli/tty.cpp
+++ b/libdnf5-cli/tty.cpp
@@ -36,7 +36,9 @@ int get_width() {
     char * columns = std::getenv("FORCE_COLUMNS");
     if (columns != nullptr) {
         try {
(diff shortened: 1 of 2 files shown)


kontura (approved):
Looks good to me.

PR #2852, opened 2026-08-07
Add group summary subcommand

### LINKED ISSUE #138: Implement group command

fhbash:
Implement the `group summary` subcommand displaying system-wide statistics: installed/available groups and environments, package counts with user-installed breakdown, total installed size, enabled repository count, and last transaction time.

Also add a count footer to `group list` showing installed/available totals after the table output.

Resolves: https://github.com/rpm-software-management/dnf5/issues/138

### DIFF
--- a/dnf5/commands/group/group.cpp
+++ b/dnf5/commands/group/group.cpp
@@ -23,6 +23,7 @@
 #include "group_install.hpp"
 #include "group_list.hpp"
 #include "group_remove.hpp"
+#include "group_summary.hpp"
 #include "group_upgrade.hpp"
 
 #include <dnf5/shared_options.hpp>
@@ -45,6 +46,7 @@ void GroupCommand::register_subcommands() {
     auto * query_commands_group = get_context().get_argument_parser().add_new_group("group_query_commands");
     query_commands_group->set_header("Query Commands:");
     get_argument_parser_command()->register_group(query_commands_group);
+    register_subcommand(std::make_unique<GroupSummaryCommand>(get_context()), query_commands_group);
     register_subcommand(std::make_unique<GroupListCommand>(get_context()), query_commands_group);
     register_subcommand(std::make_unique<GroupInfoCommand>(get_context()), query_commands_group);
     // software management commands
--- a/dnf5/commands/group/group_list.cpp
+++ b/dnf5/commands/group/group_list.cpp
@@ -25,6 +25,8 @@
 #include <libdnf5/comps/group/query.hpp>
 #include <libdnf5/conf/const.hpp>
 
+#include <iostream>
+
 namespace dnf5 {
 
 using namespace libdnf5::cli;
@@ -100,6 +102,15 @@ void GroupListCommand::print(const libdnf5::comps::GroupQuery & query) {
         items.emplace_back(new libdnf5::cli::output::GroupAdapter(obj));
     }
     libdnf5::cli::output::print_grouplist_table(items);
+
+    std::size_t installed_count = 0;
+    for (const auto & item : items) {
+        if (item->get_installed()) {
+            ++installed_count;
+        }
+    }
+    std::cout << "Groups: " << items.size() << " (Installed " << installed_count << ", Available "
+              << items.size() - installed_count << ")" << std::endl;
 }
 
 }  // namespace dnf5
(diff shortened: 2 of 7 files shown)


dnf5/commands/group/group.cpp
```diff
@@ -57,8 +70,40 @@ void GroupCommand::register_subcommands() {
...
+
+void GroupCommand::configure() {
+    auto & context = get_context();
+    context.set_load_system_repo(true);
+    context.set_load_available_repos(Context::LoadAvailableRepos::ENABLED);
+    context.get_base().get_config().get_optional_metadata_types_option().add_item(
+        libdnf5::Option::Priority::RUNTIME, libdnf5::METADATA_TYPE_COMPS);
+}
+
+void GroupCommand::run() {
```
kontura: I think we are trying to move away from default sub-commands in dnf5, at least no other command behaves this way now.

See also: https://dnf5.readthedocs.io/en/latest/changes_from_dnf4.7.html#optional-subcommands

kontura (commented):
Since we have made it this far without this sub-command I would be tempted to just drop it. To me the functionality doesn't really seem worth it.

If we were to add it would we also add `environment summary`?

Maybe we could discuss this on a devel meeting.

dnf5/commands/group/group.cpp
ppisar: I agree.

ppisar:
I also don't like adding a new subcommand only to print two numbers. The counters could be added to `dnf5 group list {--available|--installed}` output if we think that people cannot count.

If we really want some kind of summary, why not having some comprehensive statistics with number of installed/available packages, groups, explicitly installed packages, occupied space, number of repositories, the time of last dnf system state change, so that people can boast "look at my system, I have so many packages".

dnf5/commands/group/group.cpp
fhbash: Make sense, I'll remove it

fhbash:
> If we really want some kind of summary, why not having some comprehensive statistics with number of installed/available packages, groups, explicitly installed packages, occupied space, number of repositories, the time of last dnf system state change, so that people can boast "look at my system, I have so many packages".

Agree, make sense to me adding more comprehensive statistics, I'll prepare a change and let's start again from this new change..

PR #2854, opened 2026-08-10
doc: Clarify option placement rules for general vs command-specific options

fhbash:
In DNF5, general options (e.g. --disable-repo) can be placed before or after the command name, but command-specific options (e.g. --downloadonly) must come after. This was undocumented and caused user confusion.

So, I updated the synopsis in the main man page to reflect the two option types, added explanatory text to the Options section, and added an "Option placement" subsection to the DNF4-to-DNF5 changes doc with examples.

Closes: SWMBZBUGSM-240

### DIFF
--- a/doc/dnf5.8.rst
+++ b/doc/dnf5.8.rst
@@ -35,7 +35,7 @@
 Synopsis
 ========
 
-``dnf5 <command> [options] [<args>...]``
+``dnf5 [general options] <command> [command options] [<args>...]``
 
 
 Description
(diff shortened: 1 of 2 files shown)


doc/dnf5.8.rst
```diff
@@ -178,7 +178,10 @@ These are available after installing the ``dnf5-plugins`` package.
 Options
```
kontura, line 178: To further clarify the terminology would it make sense to call this heading `General options` and then in each command man page change it to `Command options`? Or even `Command-specific options` to be consistent with the changes doc, if that is not too long?

doc/dnf5.8.rst
```diff
@@ -35,7 +35,7 @@
 Synopsis
 ========
 
-``dnf5 <command> [options] [<args>...]``
+``dnf5 [general options] <command> [command options] [<args>...]``
```
kontura, line 38: If we decide to use `Command-specific options` for the command options heading I think this should be updated to `[command-specific options]`.

PR #2859, opened 2026-08-12
dnf5daemon-server: Validate locale strings

m-blaha:
The session locale is client-supplied and passed to newlocale() without validation. Its safety currently relies on glibc's locale lookup guarding against path traversal, a guarantee this project cannot assume holds for every libc it may run on. Restrict the string to the characters valid in real locale names before it reaches newlocale().

### DIFF
--- a/dnf5daemon-server/threads_manager.cpp
+++ b/dnf5daemon-server/threads_manager.cpp
@@ -88,6 +88,12 @@ void ThreadsManager::finish() {
 
 
 locale_t ThreadsManager::set_thread_locale(const std::string & thread_locale, locale_t & new_locale) {
+    // The locale string comes directly from the D-Bus client and is untrusted.
+    // Reject path traversal attempts: a path separator, or a lone "..".
+    if (thread_locale.find('/') != std::string::npos || thread_locale == "..") {
+        throw libdnf5::Error(M_("Invalid locale \"{}\"."), thread_locale);
+    }
+
     auto no_locale = static_cast<locale_t>(0);
     new_locale = newlocale(LC_ALL_MASK, thread_locale.c_str(), no_locale);
     if (new_locale == no_locale) {


ppisar:
I find this patch not good:

* The locale format is a matter of the standard library. (See setlocale(3) manual, or better POSIX <https://pubs.opengroup.org/onlinepubs/9799919799/functions/setlocale.html>.) You support this patch by a hypothetical flaw in non-glibc libraries, yet you restrict the format to glibc format. That's not portable at all.

* This patch is in a wrong project. If there is a path traversal, it needs to be fixed in the library, not in DNF5.

ppisar:
If you want to sanitize the value against path traversal (if you worry that dnf5daemon-server running under root would open a file which dnf5daemon-client does not have access to and print parts of it if matched a gettext format), then check for `(^|/)../`.

If you really want to sanitize against "valid" locale names supported by the system, `locale -a` command provides a canonical, nonstatic list. But then there are locale aliases, like `czech`, in /usr/share/locale/locale.alias which glibc also supports, but `locale -a` does not report.

m-blaha:
> I find this patch not good:
> 
>     * The locale format is a matter of the standard library. (See setlocale(3) manual, or better POSIX https://pubs.opengroup.org/onlinepubs/9799919799/functions/setlocale.html.) You support this patch by a hypothetical flaw in non-glibc libraries, yet you restrict the format to glibc format. That's not portable at all.
> 
>     * This patch is in a wrong project. If there is a path traversal, it needs to be fixed in the library, not in DNF5.

Yeah, I guess you are right...

m-blaha:
> If you want to sanitize the value against path traversal (if you worry that dnf5daemon-server running under root would open a file which dnf5daemon-client does not have access to and print parts of it if matched a gettext format), then check for `(^|/)../`.
> 
> If you really want to sanitize against "valid" locale names supported by the system, `locale -a` command provides a canonical, nonstatic list. But then there are locale aliases, like `czech`, in /usr/share/locale/locale.alias which glibc also supports, but `locale -a` does not report.

The path traversal is the case. I think just checkinkg the `/` and `../` would do.

ppisar (approved):
Good. The CI failures (resolving domain names, ELN keys) are unrelated.

PR #2863, opened 2026-08-14
Include <cstring> wherever std::strcmp() is used

WOnder93:
There are many places where this function is used that don't include the appropriate header. This actually causes a build failure with fmt 12.2.0, because it removed some includes from its headers.

Fix it by including <cstring> at least where std::strcmp() is used.


m-blaha (approved):
Thank you!

PR #2871, opened 2026-08-19
Unify "no matches"/"no repos enabled" reporting for query commands

### LINKED ISSUE #1075: Unify errors and exit codes of query commands

m-blaha:
Query commands currently print nothing when their output is empty because of no enabled repositories, an unmatched spec, or a mix of both, making it easy to miss a typo or a misconfiguration. This adds shared helpers (`no_repos_enabled()`, `report_no_repos()`, `report_no_matches()`) and wires them into:

- `repo list` / `repo info`
- `group list` / `group info`
- `environment list` / `environment info`
- `module list` / `module info`
- `advisory list` / `advisory info`
- `copr list`

`main()`'s `GoalResolveError` handler and `advisory`'s spec-based  filtering are switched to the same `report_no_repos()` helper instead of their own implementation.

Resolves: https://github.com/rpm-software-management/dnf5/issues/1075

CI tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1939

### DIFF
--- a/dnf5-plugins/copr_plugin/copr_list.cpp
+++ b/dnf5-plugins/copr_plugin/copr_list.cpp
@@ -20,6 +20,8 @@
 #include "copr.hpp"
 #include "copr_repo.hpp"
 
+#include <libdnf5/repo/repo_query.hpp>
+
 #include <iostream>
 
 namespace dnf5 {
@@ -64,7 +66,8 @@ void CoprListCommand::set_argument_parser() {
 
 
 void CoprListCommand::run() {
-    auto & base = get_context().get_base();
+    auto & ctx = get_context();
+    auto & base = ctx.get_base();
     std::unique_ptr<dnf5::CoprConfig> config = std::make_unique<dnf5::CoprConfig>(base);
     // empty string if no --hub is specified
     auto hostname = copr_cmd()->hub();
@@ -73,6 +76,17 @@ void CoprListCommand::run() {
         hostname = config->get_hub_hostname(hostname);
     auto list = RepoListCB(hostname);
     installed_copr_repositories(base, list.list);
+
+    auto & conf = base.get_config();
+    if (conf.get_installroot_option().get_value() != "/" && !conf.get_use_host_config_option().get_value()) {
+        libdnf5::repo::RepoQuery all_repos(base);
+        all_repos.filter_type(libdnf5::repo::Repo::Type::AVAILABLE);
+        if (all_repos.empty()) {
+            ctx.print_info(
+                _("No repositories were loaded from the installroot. To use the configuration and "
+                  "repositories of the host system, pass --use-host-config."));
+        }
+    }
 }
 
 }  // namespace dnf5
--- a/dnf5/commands/advisory/advisory_subcommand.cpp
+++ b/dnf5/commands/advisory/advisory_subcommand.cpp
@@ -19,6 +19,7 @@
 
 #include "advisory_subcommand.hpp"
 
+#include "../no_matches.hpp"
 #include "dnf5/context.hpp"
 
 #include <libdnf5-cli/output/advisorysummary.hpp>
@@ -82,17 +83,28 @@ void AdvisorySubCommand::run() {
     auto & ctx = get_context();
 
     ErrorHandling error_mode = determine_error_mode(ctx, false);
-    auto advisories_opt = advisory_query_from_cli_input(
-        ctx.get_base(),
-        advisory_specs->get_value(),
-        advisory_security->get_value(),
-        advisory_bugfix->get_value(),
-        advisory_enhancement->get_value(),
-        advisory_newpackage->get_value(),
-        advisory_severity->get_value(),
-        advisory_bz->get_value(),
-        advisory_cve->get_value(),
-        error_mode);
+
+    std::optional<libdnf5::advisory::AdvisoryQuery> advisories_opt;
+    if (no_repos_enabled(ctx)) {
+        // Without any enabled repositories no advisory data can be loaded, so
+        // the filtering is guaranteed to return no advisories, only
+        // producing misleading "no advisory found matching ..." messages.
+        if (error_mode != ErrorHandling::SILENT) {
+            report_no_repos(ctx, _("No matches found: no repositories are enabled."));
+        }
+    } else {
+        advisories_opt = advisory_query_from_cli_input(
+            ctx.get_base(),
+            advisory_specs->get_value(),
+            advisory_security->get_value(),
+            advisory_bugfix->get_value(),
+            advisory_enhancement->get_value(),
+            advisory_newpackage->get_value(),
+            advisory_severity->get_value(),
+            advisory_bz->get_value(),
+            advisory_cve->get_value(),
+            error_mode);
+    }
 
     auto advisories = advisories_opt.value_or(libdnf5::advisory::AdvisoryQuery(ctx.get_base()));
 
--- a/dnf5/commands/group/group_list.cpp
+++ b/dnf5/commands/group/group_list.cpp
@@ -19,6 +19,8 @@
 
 #include "group_list.hpp"
 
+#include "../no_matches.hpp"
+
 #include <libdnf5-cli/output/adapters/comps.hpp>
 #include <libdnf5-cli/output/grouplist.hpp>
 #include <libdnf5/comps/group/group.hpp>
@@ -52,24 +54,38 @@ void GroupListCommand::configure() {
 void GroupListCommand::run() {
     auto & ctx = get_context();
 
-    libdnf5::comps::GroupQuery query(ctx.get_base());
+    libdnf5::comps::GroupQuery base_query(ctx.get_base());
     auto group_specs_str = group_specs->get_value();
+    std::set<std::string> unmatched_specs;
+
+    if (installed->get_value()) {
+        base_query.filter_installed(true);
+    } else if (available->get_value()) {
+        base_query.filter_installed(false);
+    }
+
+    libdnf5::comps::GroupQuery query(base_query);
 
     // Filter by patterns if given
     if (group_specs_str.size() > 0) {
-        libdnf5::comps::GroupQuery query_names(query);
-        query_names.filter_name(group_specs_str, libdnf5::sack::QueryCmp::IGLOB);
-        query.filter_groupid(group_specs_str, libdnf5::sack::QueryCmp::IGLOB);
-        query |= query_names;
+        query = libdnf5::comps::GroupQuery(ctx.get_base(), true);
+        for (const auto & spec : group_specs_str) {
+            auto spec_query = base_query;
+            auto spec_query_names = base_query;
+            spec_query.filter_groupid(spec, libdnf5::sack::QueryCmp::IGLOB);
+            spec_query_names.filter_name(spec, libdnf5::sack::QueryCmp::IGLOB);
+            spec_query |= spec_query_names;
+            if (spec_query.empty()) {
+                unmatched_specs.insert(spec);
+            } else {
+                query |= spec_query;
+            }
+        }
     } else if (not hidden->get_value()) {
         // Filter uservisible only if patterns are not given
         query.filter_uservisible(true);
     }
-    if (installed->get_value()) {
-        query.filter_installed(true);
-    } else if (available->get_value()) {
-        query.filter_installed(false);
-    } else {
+    if (!installed->get_value() && !available->get_value()) {
         // to remove duplicities in the output remove from query all available
         // groups with the same groupid as any of the installed groups.
         libdnf5::comps::GroupQuery query_installed(query);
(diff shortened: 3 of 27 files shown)


dnf5/commands/no_matches.cpp
```diff
@@ -0,0 +1,56 @@
...
+    if (!generic_message.empty()) {
+        ctx.print_info(generic_message);
+    }
+    return false;
+}
+
+void report_no_matches(
+    Context & ctx, const std::set<std::string> & unmatched_specs, bool result_is_empty, bool repos_relevant) {
+    if (result_is_empty && repos_relevant && no_repos_enabled(ctx)) {
+        report_no_repos(ctx, _("No matches found: no repositories are enabled."));
```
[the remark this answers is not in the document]
m-blaha: That's intentional. We do not want to add noise to repoquery output.

dnf5/commands/repo/repo_list.cpp
```diff
@@ -47,34 +49,43 @@ void RepoListCommand::set_argument_parser() {
...
     }
 
-    query.filter_type(libdnf5::repo::Repo::Type::AVAILABLE);
-
     // display status because we're printing mix of enabled and disabled repos
     bool with_status = all->get_value();
 
     print(query, with_status);
+
+    report_no_matches(ctx, unmatched_specs, query.empty(), true);
```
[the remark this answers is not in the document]
m-blaha: Fixed.

dnf5/commands/advisory/advisory_subcommand.cpp
```diff
@@ -82,17 +83,28 @@ void AdvisorySubCommand::run() {
...
+        // the filtering is guaranteed to return no advisories, only
+        // producing misleading "no advisory found matching ..." messages.
+        if (error_mode != ErrorHandling::SILENT) {
+            report_no_repos(ctx, _("No advisories found: no repositories are enabled."));
+        }
+    } else {
+        advisories_opt = advisory_query_from_cli_input(
+            ctx.get_base(),
+            advisory_specs->get_value(),
+            advisory_security->get_value(),
```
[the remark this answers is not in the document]
m-blaha: Advisory specs matches reporting is handled by `advisory_query_from_cli_input()`. We did not change that.

dnf5-plugins/copr_plugin/copr_list.cpp
```diff
@@ -73,6 +76,17 @@ void CoprListCommand::run() {
         hostname = config->get_hub_hostname(hostname);
     auto list = RepoListCB(hostname);
     installed_copr_repositories(base, list.list);
+
+    auto & conf = base.get_config();
+    if (conf.get_installroot_option().get_value() != "/" && !conf.get_use_host_config_option().get_value()) {
+        libdnf5::repo::RepoQuery all_repos(base);
+        all_repos.filter_type(libdnf5::repo::Repo::Type::AVAILABLE);
+        if (all_repos.empty()) {
```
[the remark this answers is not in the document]
m-blaha: This is intentional. For copr plugin we only hint in case of empty install root.

dnf5/commands/group/group_list.cpp
```diff
@@ -54,22 +56,36 @@ void GroupListCommand::run() {
...
     libdnf5::comps::GroupQuery query(ctx.get_base());
     auto group_specs_str = group_specs->get_value();
+    std::set<std::string> unmatched_specs;
+
+    if (installed->get_value()) {
+        query.filter_installed(true);
+    } else if (available->get_value()) {
+        query.filter_installed(false);
+    }
+    bool no_candidate_groups = query.empty();
```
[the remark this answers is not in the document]
m-blaha: Fixed.

dnf5/commands/module/module_list.cpp
```diff
@@ -66,25 +75,24 @@ void ModuleListCommand::run() {
...
+    std::string_view no_candidates_message;
+    if (enabled->get_value()) {
+        no_candidates_message = _("No matches found: no module streams are enabled.");
+    } else if (disabled->get_value()) {
+        no_candidates_message = _("No matches found: no module streams are disabled.");
+    } else {
+        no_candidates_message = _("No matches found: no module streams exist.");
     }
+    report_no_matches(
+        ctx, unmatched_module_spec, query.empty(), no_repos_enabled(ctx), no_candidate_modules, no_candidates_message);
```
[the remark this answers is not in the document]
m-blaha: This fix did not touch exit codes. I understand, that "provides" command behaves differently, but it's consistent with what dnf4 does.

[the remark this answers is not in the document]
m-blaha: I believe the current behavior is correct.

dnf5/commands/no_matches.cpp
```diff
@@ -0,0 +1,69 @@
...
+            report_unmatched_specs(ctx, unmatched_specs);
+        }
+        return;
+    }
+    if (no_repos) {
+        report_no_repos(ctx, _("No matches found: no repositories are enabled."));
+    } else if (no_candidates) {
+        ctx.print_info(no_candidates_message);
+    } else {
+        ctx.print_info(_("No matches found."));
```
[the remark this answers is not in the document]
m-blaha: This is intentional. The user would be missing only hints explaining why the output is empty, no actual error is hidden.

dnf5/commands/advisory/advisory_subcommand.cpp
```diff
@@ -82,17 +83,28 @@ void AdvisorySubCommand::run() {
...
-        advisory_cve->get_value(),
-        error_mode);
+
+    std::optional<libdnf5::advisory::AdvisoryQuery> advisories_opt;
+    if (no_repos_enabled(ctx)) {
+        // Without any enabled repositories no advisory data can be loaded, so
+        // the filtering is guaranteed to return no advisories, only
+        // producing misleading "no advisory found matching ..." messages.
+        if (error_mode != ErrorHandling::SILENT) {
+            report_no_repos(ctx, _("No advisories found: no repositories are enabled."));
```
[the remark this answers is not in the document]
m-blaha: OK, unification here means fewer translatable strings and more consistent experience.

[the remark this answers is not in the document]
m-blaha: Fixed.

dnf5/commands/search/search.cpp
```diff
@@ -79,15 +80,17 @@ void SearchCommand::configure() {
...
+        ctx.get_base(),
         patterns->get_value(),
         all->get_value(),
         show_duplicates->get_value(),
         search_name->get_value(),
         search_summary->get_value());
-    libdnf5::cli::output::print_search_results(processor.get_results());
+    auto results = processor.get_results();
+    libdnf5::cli::output::print_search_results(results);
+    report_no_matches(ctx, {}, results.group_results.empty(), no_repos_enabled(ctx), false);
```
kontura, line 93: We could add the `unmatched_specs` here but it would require extending the `SearchProcessor` right?
Perhaps a small follow up issue could be filed?

dnf5/commands/repo/repo_list.cpp
```diff
@@ -47,34 +50,44 @@ void RepoListCommand::set_argument_parser() {
...
-        // filter by repo ID
-        query.filter_id(repo_specs_str, libdnf5::sack::QueryCmp::IGLOB);
-        // union the results
-        query |= query_names;
+        query.clear();
+        for (const auto & spec : repo_specs_str) {
+            auto spec_query = base_query;
+            auto spec_query_names = base_query;
+            spec_query.filter_id(std::vector<std::string>{spec}, libdnf5::sack::QueryCmp::IGLOB);
+            spec_query_names.filter_name(std::vector<std::string>{spec}, libdnf5::sack::QueryCmp::IGLOB);
```
kontura: Is it really required to construct the vectors here? I think there are overloads for just strings.

dnf5/commands/repo/repo_list.cpp
```diff
@@ -47,34 +50,44 @@ void RepoListCommand::set_argument_parser() {
...
     }
 
-    query.filter_type(libdnf5::repo::Repo::Type::AVAILABLE);
-
     // display status because we're printing mix of enabled and disabled repos
     bool with_status = all->get_value();
 
     print(query, with_status);
+
+    bool no_repos = !all->get_value() && !disabled->get_value() && no_repos_enabled(ctx);
```
kontura, line 89: I find it a bit odd that when the user asks for disabled or all repos they no longer get the installroot hint.

Although I see that we want to avoid the message: `No matches found: no repositories are enabled.` when querying disabled repos and there are no matches. 🤔

dnf5/commands/group/group_list.cpp
```diff
@@ -52,24 +54,38 @@ void GroupListCommand::configure() {
...
-        libdnf5::comps::GroupQuery query_names(query);
-        query_names.filter_name(group_specs_str, libdnf5::sack::QueryCmp::IGLOB);
-        query.filter_groupid(group_specs_str, libdnf5::sack::QueryCmp::IGLOB);
-        query |= query_names;
+        query = libdnf5::comps::GroupQuery(ctx.get_base(), true);
+        for (const auto & spec : group_specs_str) {
+            auto spec_query = base_query;
+            auto spec_query_names = base_query;
+            spec_query.filter_groupid(std::vector<std::string>{spec}, libdnf5::sack::QueryCmp::IGLOB);
+            spec_query_names.filter_name(std::vector<std::string>{spec}, libdnf5::sack::QueryCmp::IGLOB);
```
kontura: Unneeded vectors here as well?
For environment info and list as well.

dnf5/commands/provides/provides.cpp
```diff
@@ -126,19 +128,14 @@ void ProvidesCommand::run() {
             }
         }
     }
-    if (!unmatched_specs.empty() && any_match) {
-        for (auto const & spec : unmatched_specs) {
-            std::cerr << "No matches found for " << spec << "." << std::endl;
+    bool no_repos = no_repos_enabled(ctx);
+    report_no_matches(ctx, unmatched_specs, !any_match, no_repos, false);
+    if (!any_match || !unmatched_specs.empty()) {
+        if (any_match || !no_repos) {
```
kontura: I could be wrong but since the `provides` command requires at least one positional argument could the `any_match` be completely dropped from both of these conditions?

kontura (commented):
I was also thinking about exit codes mentioned in https://github.com/rpm-software-management/dnf5/issues/1075 would it make sense if all query commands behaved like `provides` command? So return 1 if at least one arguments doesn't match? 

Though that is definitely out of scope of this PR.

m-blaha:
> I was also thinking about exit codes mentioned in #1075 would it make sense if all query commands behaved like `provides` command? So return 1 if at least one arguments doesn't match?
> 
> Though that is definitely out of scope of this PR.

To be honest, I was scared to even think about changing the exit codes :) But we should definitely discuss this on the team meeting. Returning 1 on "no match" could help some scripts with e.g. detection whether a package is present/installed, on the other hand, it might break current workflows.

dnf5/commands/repo/repo_list.cpp
m-blaha: Yeah, my bad. Will simplify it (here and also on other places).

dnf5/commands/repo/repo_list.cpp
m-blaha: That seems like omission - I juggled quite a lot with repo command to return reasonable results, until I created a set of test scenarios. It seems this one is missing. I'll take a look.

dnf5/commands/search/search.cpp
m-blaha: I'll look into it and either fix it ritght away, or file an issue.

dnf5/commands/provides/provides.cpp
m-blaha: I've completely missed that `ArgumentParser::PositionalArg::AT_LEAST_ONE`.

m-blaha: It's a bit more complicated. In case there are no configured repos, the `filter_spec` returns back `full_package_query` (which is empty because of no repos), causing `unmatched_specs` not to be filled. I'll fix that (and drop the `any_match` conditions as  you pointed out).

dnf5/commands/repo/repo_list.cpp
m-blaha: Should be fixed now (in no_matches.cpp)

dnf5/commands/provides/provides.cpp
kontura: Thank you for updating it, I find it much better and simpler now.

kontura (approved):
Thank you! 
The changes look good. It is not clear to me why some of the F45 tests fail, they do pass locally.
I will investigate that further if it doesn't go away after merging.

dnf5/commands/search/search.cpp
m-blaha: I checked the search command:

- adding unmatched specs cleanly would break API/ABI on `SearchResults` structure
- but most importantly - the semantics of such message would be unclear. By default, the results for individual patterns are combined using intersection (i.e. AND-ed), meaning the first unmatched pattern will clear the output. Only with `--all` they are combined using union (OR-ed) - here the unmatched patterns might make some sense. And on top of that, `--all` also extends a fields set being searched.

That said, I'd rather not use this concept for the search command.

kontura: That makes sense, thank you for the investigation.

PR #2883, opened 2026-08-21
spec: Fix bcond ordering so plugin_systemd_inhibit is enabled

m-blaha:
The %bcond for systemd was declared after plugin_systemd_inhibit
already checked it, so %{with systemd} evaluated to false and
the plugin was silently disabled.

Also reorder all bcond declarations into logical groups (toolchain, core
features, main components, plugins, documentation, language bindings,
testing).

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -3,12 +3,76 @@
 %global project_version_minor 3
 %global project_version_micro 0
 
-%bcond dnf5_obsoletes_dnf %[0%{?fedora} > 40 || 0%{?rhel} > 10]
+# ========== build options ==========
 
+# Toolchain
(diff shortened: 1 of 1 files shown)


kontura (approved):
The changes look good only I think the newly enabled and tested inhibit plugin is tripping up one test.

m-blaha:
> The changes look good only I think the newly enabled and tested inhibit plugin is tripping up one test.

Right, I'll look into that.

m-blaha:
I've created https://github.com/rpm-software-management/ci-dnf-stack/pull/1943 PR to remove the dependency on "action" plugin being loaded first.

PR #2884, opened 2026-08-22
Download order by size with reverse option

rsevero:
**Why this is needed**

`dnf5 download`, `install`, and `upgrade` currently fetch packages in whatever order the resolved transaction happens to produce (essentially NEVRA order), with no way for the user to influence it. That's a real limitation on a couple of common workflows:

- **Slow, unreliable, or metered connections.** Fetching many small packages first gives fast, visible progress and gets the bulk of the package list off the network quickly; if the connection drops partway through, more of the transaction is already downloaded. Fetching largest-first is the opposite trade-off — useful for saturating available bandwidth on a fast link with a few big transfers instead of paying per-request overhead on lots of small ones.
- **Large packages benefit from starting earlier.** Big files tend to take the longest to complete, and if they're also served from a slower mirror or throttled connection, that gap only grows. Kicking them off first — instead of leaving them to be picked up last, after every small package has already been queued — means the download's overall wall-clock time is bounded by the big transfer running in parallel with the rest, rather than being added on top of everything else.
- **Mirroring/offline-prep scripts.** Tools that stage packages ahead of time (e.g. building an offline repo, or `dnf5 download` used as a pre-fetch step before a maintenance window) benefit from predictable, size-ordered output so they can checkpoint or parallelize downstream steps.
- **No workaround exists today.** Users can't get this ordering without post-processing package lists themselves and issuing separate single-package downloads, which defeats dnf5's own parallel downloader and dependency resolution.

**What the change does**

Adds `--sort=size` and `--reverse` to `download`, and — because the sort now lives in `Transaction::download()` itself via two runtime `ConfigMain` options (`download_sort`, `download_sort_reverse`) rather than being local to the download command — the same ordering is available to `install` and `upgrade` for free, with no ABI/config-layout changes.

The new sort helper (`rpm::sort_packages_by_download_size`) is covered by dedicated unit tests, including a repo fixture built specifically to exercise both the forward and `--reverse` orderings.

### DIFF
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -21,11 +21,11 @@
 
 #include "dnf5/shared_options.hpp"
 
-#include <libdnf5-cli/utils/package_download_order.hpp>
 #include <libdnf5/conf/option_string.hpp>
 #include <libdnf5/repo/package_downloader.hpp>
 #include <libdnf5/rpm/arch.hpp>
 #include <libdnf5/rpm/package.hpp>
+#include <libdnf5/rpm/package_download_order.hpp>
 #include <libdnf5/rpm/package_query.hpp>
 #include <libdnf5/rpm/package_set.hpp>
 #include <libdnf5/utils/bgettext/bgettext-mark-domain.h>
@@ -114,31 +114,6 @@ void DownloadCommand::set_argument_parser() {
     debugsource->set_const_value("true");
     debugsource->link_value(debugsource_option);
 
-    sort_valid_options = {"size"};
-    auto sort = parser.add_new_named_arg("sort");
-    sort->set_long_name("sort");
-    sort->set_description("Sort the packages to download by the given key");
-    sort->set_has_value(true);
-    sort->set_arg_value_help("size");
-    sort->set_parse_hook_func(
-        [this](
-            [[maybe_unused]] ArgumentParser::NamedArg * arg, [[maybe_unused]] const char * option, const char * value) {
-            if (sort_valid_options.find(value) == sort_valid_options.end()) {
-                throw libdnf5::cli::ArgumentParserInvalidValueError(M_("Invalid sort option: {}"), std::string(value));
-            }
-            sort_option = value;
-            return true;
-        });
-
-    reverse_option = dynamic_cast<libdnf5::OptionBool *>(
-        parser.add_init_value(std::unique_ptr<libdnf5::OptionBool>(new libdnf5::OptionBool(false))));
-
-    auto reverse = parser.add_new_named_arg("reverse");
-    reverse->set_long_name("reverse");
-    reverse->set_description("To be used together with \"--sort\". Reverses the sort order.");
-    reverse->set_const_value("true");
-    reverse->link_value(reverse_option);
-
     auto url = parser.add_new_named_arg("url");
     url->set_long_name("url");
     url->set_description("Print a URL where the rpms can be downloaded instead of downloading");
@@ -212,8 +187,7 @@ void DownloadCommand::set_argument_parser() {
     cmd.register_named_arg(url);
     cmd.register_named_arg(urlprotocol);
     cmd.register_named_arg(allmirrors);
-    cmd.register_named_arg(sort);
-    cmd.register_named_arg(reverse);
+    create_download_order_options(*this);
     cmd.register_positional_arg(keys);
 }
 
@@ -238,10 +212,7 @@ void DownloadCommand::configure() {
         context.set_load_system_repo(false);
     }
 
-    if (reverse_option->get_value() && sort_option.empty()) {
-        throw libdnf5::cli::ArgumentParserMissingDependentArgumentError(
-            M_("Option \"--reverse\" should be used with \"--sort\""));
-    }
+    check_download_order_options(*this);
 
     if (srpm_option->get_value()) {
         context.get_base().get_repo_sack()->enable_source_repos();
@@ -442,13 +413,15 @@ void DownloadCommand::run() {
     // for download command, we don't want to mark the packages for removal
     downloader.force_keep_packages(true);
 
-    if (sort_option == "size") {
+    auto & config = ctx.get_base().get_config();
+    if (config.get_download_sort_option().get_value() == "size") {
         std::vector<libdnf5::rpm::Package> sorted_packages;
         sorted_packages.reserve(packages_to_download.size());
         for (auto & [nevra, pkg] : packages_to_download) {
             sorted_packages.push_back(pkg);
         }
-        libdnf5::cli::utils::sort_packages_by_download_size(sorted_packages, reverse_option->get_value());
+        libdnf5::rpm::sort_packages_by_download_size(
+            sorted_packages, config.get_download_sort_reverse_option().get_value());
         for (auto & pkg : sorted_packages) {
             downloader.add(pkg);
         }
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -212,8 +212,6 @@ void DownloadCommand::configure() {
         context.set_load_system_repo(false);
     }
 
-    check_download_order_options(*this);
-
     if (srpm_option->get_value()) {
         context.get_base().get_repo_sack()->enable_source_repos();
     }
@@ -424,6 +422,8 @@ void DownloadCommand::run() {
     if (config.get_download_sort_option().get_value() == "size") {
         libdnf5::rpm::sort_packages_by_download_size(
             sorted_packages, config.get_download_sort_reverse_option().get_value());
+    } else if (config.get_download_sort_reverse_option().get_value()) {
+        std::reverse(sorted_packages.begin(), sorted_packages.end());
     }
 
     for (auto & pkg : sorted_packages) {
--- a/dnf5/commands/download/download.cpp
+++ b/dnf5/commands/download/download.cpp
@@ -413,22 +413,21 @@ void DownloadCommand::run() {
     // for download command, we don't want to mark the packages for removal
     downloader.force_keep_packages(true);
 
+    // Packages are collected in NEVRA order (the default); optionally re-sorted below.
+    std::vector<libdnf5::rpm::Package> sorted_packages;
+    sorted_packages.reserve(packages_to_download.size());
+    for (auto & [nevra, pkg] : packages_to_download) {
+        sorted_packages.push_back(pkg);
+    }
+
     auto & config = ctx.get_base().get_config();
     if (config.get_download_sort_option().get_value() == "size") {
-        std::vector<libdnf5::rpm::Package> sorted_packages;
-        sorted_packages.reserve(packages_to_download.size());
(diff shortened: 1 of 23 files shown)


rsevero:
Renamed "--sort" to "--download-sort" and "--reverse" to "download-reverse" to make it clear that these options work on download ordering only.

m-blaha:
Thanks for the PR!
Before diving into code review itself, I'd like to understand the practical benefits better. All the affected commands still need to download **all** the packages, so the total time is determined only by the bandwidth available, submission order does not change that. Scheduling longest processing time items first could theoretically bring some benefits, but I suspect them to be marginal. Did you do some measurements justifying the new switches?

Also note that dnf5 downloads packages via `librepo`, which manages its own parallel scheduling internally. There is no guarantee that it respects the order packages were submitted by dnf5 (although currently it most likely does).

rsevero:
> Thanks for the PR! Before diving into code review itself, I'd like to understand the practical benefits better. All the affected commands still need to download **all** the packages, so the total time is determined only by the bandwidth available, submission order does not change that. Scheduling longest processing time items first could theoretically bring some benefits, but I suspect them to be marginal. Did you do some measurements justifying the new switches?

The main issue that makes this PR worthwhile in my view is that usually, during packages download connections are not fully used. At the end of the download phase of my daily upgrades I usually see two or three big packages that are downloaded way slower than my connection is able.

The total download time is mainly defined not by the usually expected:

1. total_size_of_downloads / connection_download_speed

but by:

2. size_of_big_slow_files / upload_speed_of_big_slow_files_providers

If 1. were true, there would be no reason for this PR but in my experience the download time is mainly defined by 2.

Is this situation, being able to start big files first is a significant advantage.

It is important to mention that:

a) The code changes are really small;
b) If someone is in a situation that 1. above is true, even using this feature creates no problem whatsoever as, in this ideal situation where the connection is really filled, the download order has influence, for good or for bad.

For people where 1. is true, there is no reason to even think about using this feature. For people where 2. is true, this feature can be a very cheap time saver.

kontura:
> a) The code changes are really small;

While the actual functionality is really small it does add two switches and these are not cheap, dnf already has way too many of them and we have to think hard before adding new ones. I would rather just make it the default depending on the downsides of starting with biggest packages.

Some numbers would indeed be nice, at least how much it speeds up the downloads for you. I imagine the effect is more visible for smaller transactions.

This might be a better match for a libdnf5 plugin although that wouldn't work for the download command.

rsevero:
> > a) The code changes are really small;
> 
> While the actual functionality is really small it does add two switches and these are not cheap, dnf already has way too many of them and we have to think hard before adding new ones. I would rather just make it the default depending on the downsides of starting with biggest packages.
> 
> Some numbers would indeed be nice, at least how much it speeds up the downloads for you. I imagine the effect is more visible for smaller transactions.

I had results all over the place as the file providers behaviour changed a lot during the tests but these 4 tests below where produced in rapid succession, one after the other. I believe they represent well the differences this feature can produce on situations where the download link does not get saturated:
`LD_LIBRARY_PATH=libdnf5:libdnf5-cli`
```
# time dnf upgrade --downloadonly -y --destdir=/tmp/dnf_download_dir
Complete!

real    0m51,757s
user    0m2,031s
sys     0m1,132s

# time LD_LIBRARY_PATH=libdnf5:libdnf5-cli ./dnf5/dnf5 upgrade --downloadonly -y --no-plugins --download-sort=size --download-reverse --destdir=/tmp/dnf_download_dir
Complete!

real    0m33,131s
user    0m4,015s
sys     0m1,162s

# time dnf upgrade --downloadonly -y --destdir=/tmp/dnf_download_dir
Complete!

real    0m43,560s
user    0m1,950s
sys     0m1,128s

# time LD_LIBRARY_PATH=libdnf5:libdnf5-cli ./dnf5/dnf5 upgrade --downloadonly -y --no-plugins --download-sort=size --download-reverse --destdir=/tmp/dnf_download_dir
Complete!

real    0m23,026s
user    0m3,923s
sys     0m1,123s
```

The calls prefixed with `LD_LIBRARY_PATH=libdnf5:libdnf5-cli` are my test version. The other is the regular Fedora 44 5.4.43 version.

m-blaha:
To be honest - I'm surprised how large the effect of your change is. Did you cleaned the cache between the runs?
I tend to agree with @kontura and just make this behavior default and do not add switches for it.

rsevero:
> To be honest - I'm surprised how large the effect of your change is. Did you cleaned the cache between the runs?

No. I cleaned the destdir between runs. But I don´t think there were a download cache in effect, as, if there were, I would expect the second, third and last runs being way faster the the first and it didn´t happen.

And I did tons of runs. I don´t expect that there will be that much difference every time, on every file mix. But I believe the effect will be from almost none to very good. I don´t think there will be worst results with this change.

> I tend to agree with @kontura and just make this behavior default and do not add switches for it.

Do you want me to change the PR to make this new code the default behavior?

m-blaha:
I did some measurements (note, that the measured transaction was pretty small - 28 packages, 53 MiB of download, took cca 17s to accomplish), run this command with and without sorting switches:

```
for i in {1..20}; do (dnf clean all >/dev/null 2>&1 && time ./dnf5/dnf5 upgrade --downloadonly --refresh --assumeyes --download-sort=size --download-reverse >/dev/null 2>&1) 2>&1 | grep real; done
```

The result:
```
=== Default ===
N: 20 | Mean: 17.0980s | StdDev: 1.2342s | CV: 7.22%

=== Reversed ===
N: 20 | Mean: 16.9408s | StdDev: 1.1061s | CV: 6.53%

Difference: -0.1572s (-0.92%)
Standard Error: ±0.3706
```

I'll try with a bigger transaction later.

m-blaha:
I've tried on a massive transaction (1776 pkgs, 2GiB of download):

```
=== Default ===
N: 10 | Mean: 120.3690s | StdDev: 9.2328s | CV: 7.67%

=== Ordered ===
N: 10 | Mean: 116.4807s | StdDev: 10.2996s | CV: 8.84%

=== Statistical Significance ===
Difference (Set 2 - Set 1): -3.8883s (-3.23%)
```

The command was (run in a clean F43 container):
```
for i in {1..10}; do (dnf clean packages >/dev/null 2>&1 && time dnf install @workstation-product-environment --downloadonly --assumeyes --download-sort=size --download-reverse >/dev/null 2>&1) 2>&1 | grep real; done
```

rsevero:
> I've tried on a massive transaction (1776 pkgs, 2GiB of download):
> 
> ```
> === Default ===
> N: 10 | Mean: 120.3690s | StdDev: 9.2328s | CV: 7.67%
> 
> === Ordered ===
> N: 10 | Mean: 116.4807s | StdDev: 10.2996s | CV: 8.84%
> 
> === Statistical Significance ===
> Difference (Set 2 - Set 1): -3.8883s (-3.23%)
> ```
> 
> The command was (run in a clean F43 container):
> 
> ```
> for i in {1..10}; do (dnf clean packages >/dev/null 2>&1 && time dnf install @workstation-product-environment --downloadonly --assumeyes --download-sort=size --download-reverse >/dev/null 2>&1) 2>&1 | grep real; done
> ```

Yes, it's no surprise to me that different setups get different results. I bet you have way less situations when a big file does not saturate (or nearly saturate) your connection than me.

This situation is quite common for me.

Even not having a big difference, the difference was always positive to the new method ;)

m-blaha:
> Even not having a big difference, the difference was always positive to the new method ;)

That's right :)

I'll try to build a test scenario closer to your real-world situation. To do that it'd help to know:

- What type of connection do you have (fiber/DSL/mobile/…), and roughly its bandwidth and latency to your mirrors?
- Which Fedora mirror(s) do you hit?

My working theory: if the mirror caps each connection, a big package can't saturate your link on its own, so it makes sense to start it as early as possible and let the small packages download alongside it instead of after it. That also matches what you described - a couple of big files at the tail crawling slower than your connection  can go.

If it's easy, the single most useful number would be: what download speed do those slow big packages get, compared to your overall link speed?

ppisar:
Processing big tasks first and processing small tasks later is a standard optimization technique: The total amount of work to perform is constant regardless of the task scheduling strategy, but if there are more (independent) queues processing the tasks, starting with big ones helps to minimize a time of the longest queue. Technically the strategy makes all the queues similarly long.

I agree that the optimizing behavior should be default and no new options should needed.

m-blaha:
Given that `librepo` determines which download task starts when, should the fix be made there rather than here in dnf5? Although librepo currently starts download jobs in the same order they were created by dnf5, this behavior is not part of the contract.
@ppisar @kontura What do you think?

ppisar:
I'm sorry, I don't know librepo API. If librepo is capable of getting all the download requests from DNF5 ahead*, sort them internally by size (or by estimated time to download), and then schedule the cURL downloads, then it's indeed better to be implemented in librepo.

*) The ahead does not have to be total. DNF5 pushing the request to librepo faster than librepo can complete them could work too. Though the ever changing set of downloads from librepo's point of view would complicate the implementation.

kontura:
I believe librepo basically needs to know all the download requests ahead (be it for metadata or packages).
I would also agree that librepo is a better fit for the functionality.

m-blaha:
@rsevero, do you feel like implementing download targets ordering in https://github.com/rpm-software-management/librepo?

If we take the straightforward approach without adding new options, the implementation should be fairly simple. Basically, targets just need to be sorted before downloads begin in the `lr_download()` function.

Something like:

```
dd.targets = g_slist_sort(dd.targets, cmp_target_size_desc);
```

after `dd.targets` is populated (approximately here: https://github.com/rpm-software-management/librepo/blob/86ed082e25bbd1901d21d75924297c7697b33307/librepo/downloader.c#L2793)

You would also need to implement the comparator, for example:

```
static gint
cmp_target_size_desc(gconstpointer a, gconstpointer b)
{
    gint64 sa = ((const LrTarget *) a)->target->origsize;
    gint64 sb = ((const LrTarget *) b)->target->origsize;
    return (sb > sa) - (sb < sa);
}
```

Of course, we could make this more flexible by adding an `LRO_DOWNLOAD_ORDER` configuration option. And a unit test would also be great.

rsevero:
Yes, I can do it. Can you just clarify the role of LRO_DOWNLOAD_ORDER you mentioned?

m-blaha:
I was thinking that, if we do not want to reorder targets by default, we could add an option for it - `LRO_DOWNLOAD_ORDER` option I mentioned. This would allow librepo API consumers to enable or disable ordering according to their needs. It could accept values like LR_DOWNLOAD_ORDER_NONE (do not reorder, preserve the order in which targets were submitted), LR_DOWNLOAD_ORDER_SIZE_DESC (start with largest targets first).

However, that would require:

1. Adding the new `LRO_DOWNLOAD_ORDER` option (and allowed values enum) to librepo
2. Releasing it
3. Bumping the librepo version requirement in dnf5
4. Using the new option when constructing the librepo handle

I'm not sure it is worth it.

rsevero:
I agree it might not be worthwhile. The new download order would be the new behavior right?

m-blaha:
Yes.

PR #2889, opened 2026-08-24
libdnf5: Fix dangling reference bug in is_vendor_change_allowed

jrohel:
**Bug scenario:**
In `VendorChangeManager::is_vendor_change_allowed()`, two references are obtained sequentially:
1. `const auto & outgoing_vendor_mask = get_vendor_change_masks( outgoing_vendor).outgoing_mask;`
2. `const auto & incoming_vendor_mask = get_vendor_change_masks( incoming_vendor).incoming_mask;`

If `incoming_vendor` is not cached, `get_vendor_change_masks()` calls `vendor_masks.emplace_back()` (`vendor_masks` is `std::vector`) which may reallocate the vector, invalidating the `outgoing_vendor_mask` reference from step 1.

**Fix:**
Replace `std::vector` with `std::map`. `std::map` guarantees that references to existing elements remain valid after inserting new elements (no reallocation like vector).
Secondary benefit: O(log n) lookup instead of O(n) linear search.


m-blaha (approved):
Thank you!

PR #2893, opened 2026-08-26
Fixes and improvements for top-level Makefile

evan-goode:
This is a follow-up to https://github.com/rpm-software-management/dnf5/pull/2834 which added the convenience Makefile for building dnf5 and running integration tests. See commit messages.

My main goal here was to get `make test-integration` to do as little rebuilding as possible while always automatically rebuilding what is necessary to rebuild.

I also swapped the integration tests to use fedora:rawhide by default, mainly to avoid having a hardcoded version number that we have to bump. `registry.fedoraproject.org/fedora:latest` could have also worked, IIUC this always refers to latest stable Fedora. However, I don't believe there is an equivalent Mock config that always means "latest stable".

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+integration-tests export-ignore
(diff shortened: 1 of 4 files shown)


fhbash (approved):
LGTM, thanks!

PR #2759, opened 2026-05-29
filter_latest_unresolved_advisories: fix missing results due to id order

kontura:
The code to get the highest EVR assumed advisory packages vector is sorted by EVRas but it was sorted by libsolv ids of EVRas. The ids depend on repository loading order they don't have to be ascending in string form.

To fix this find the highest EVR explicitly.

Closes: https://github.com/rpm-software-management/dnf5/issues/2754

Issue #2754, opened 2026-05-25
`dnf5 upgrade --security` does not install `kernel` security updates


markusgut:
There are package updates available with correct metadata tagging them as `security` updates:
```
markus@blackbox:~$ dnf5 advisory list --security
Updating and loading repositories:
Repositories loaded.
Name                   Type     Severity                                     Package              Issued
FEDORA-2026-66bba52149 security Important               kernel-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important          kernel-core-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important       kernel-modules-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important  kernel-modules-core-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important kernel-modules-extra-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important         kernel-tools-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important    kernel-tools-libs-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
FEDORA-2026-66bba52149 security Important         python3-perf-7.0.9-205.fc44.x86_64 2026-05-21 23:20:37
```
However, `dnf5 upgrade --security` does not install them:
```
markus@blackbox:~$ sudo dnf upgrade --security
Updating and loading repositories:
Repositories loaded.
No match in selected advisories

Nothing to do.
```
However, not all security updates are missed - sometimes some are installed, while others are not. On systems with `dnf automatic` configured to apply security updates automatically, this may result in unpatched systems. Updates need to be applied manually by applying all updates (`dnf upgrade`) or by specifying a specific advisory (`dnf upgrade --advisory=FEDORA-2026-66bba52149`).

Steps to reproduce - on systems with recent security updates not yet applied:
1. dnf5 advisory list --security
2. dnf5 dnf upgrade --security
3. Compare 

Expected result:
All packages shown for `dnf5 advisory list --security` should be applied when running `dnf upgrade --security`.

Actual result:
Some packages shown for `dnf5 advisory list --security` are not applied when running `dnf upgrade --security`.

Additional info:
```
[...]

PR #2901, opened 2026-08-31
Don't ignore advisory upgrade with lover libsolv id evr

kontura:
For https://github.com/rpm-software-management/dnf5/issues/2754

Requires: https://github.com/rpm-software-management/dnf5/pull/2759


evan-goode (approved):
Thanks! I verified that the new test fails without https://github.com/rpm-software-management/dnf5/pull/2759. Nice catch there.

PR #2902, opened 2026-08-31
spec: Mark all /etc configuration files as noreplacible

ppisar:
Configuration files in /etc can be modified by a local superuser and the modifications should survive updating the dnf5 RPM package.

Therefore this fix adds "noreplace" attribute to the packages configuration files. The attribute causes preserving the old content and adding a new content into *.rpmnew file.


fhbash (approved):
LGTM, thanks!

PR #2905, opened 2026-08-31
manifest plugin: Use new infile format

### LINKED ISSUE #2616: manifest: support new infile format
[libpkgmanifest](https://github.com/rpm-software-management/libpkgmanifest/) supports two infile formats: [a new format](https://github.com/rpm-software-management/libpkgmanifest/#input-files) developed by jan-kolarik based on libpkgmanifest's own conventions, and the [rpm-lockfile-prototype input file format](https://github.com/konflux-ci/rpm-lockfile-prototype?tab=readme-ov-file#whats-the-input_file).

The manifest plugins in DNF4 and DNF5 currently support only the rpm-lockfile-prototype format. The DNF5 plugin (at least?) should be extended to support the libpkgmanifest format too. The rpm-lockfile-prototype format lacks a `version` field and requires a `contentOrigin` field, so a good strategy may be to assume any infile with a `contentOrigin` and no `version` is an rpm-lockfile-prototype infile, and all other files are libpkgmanifest infiles.

Alternatively, we could check for `version` alone, but presumably the libpkgmanifest infile format will supersede the rpm-lockfile-prototype format and it would be nice to error if the `version` field is missing, in case a user accidentally omits it while trying to write a libpkgmanifest infile.

libpkgmanifest infile spec: https://github.com/rpm-software-management/libpkgmanifest/blob/main/schemas/input.json

evan-goode:
Resolves https://github.com/rpm-software-management/dnf5/issues/2616. The new infile schema is still TODO, but we can go ahead and make this change since we'll want to use libpkgmanifest's `parser.parse` (rather than `parser.parse_prototype`) regardless.

The plugin first tries to parse the infile as a libpkgmanifest infile. If that fails, it falls back to parsing as an rpm-lockfile-prototype infile.

The default name of the infile is also changed from `rpms.in.yaml` to `packages.input.yaml`. I'm open to suggestions here. I like `packages.input.yaml` since it matches the lockfile name `packages.manifest.yaml`. Other options include:

- Keep `rpms.in.yaml`. Might be confused more easily with the older rpm-lockfile-prototype infiles.
- `packages.in.yaml`. Slightly shorter.


dnf5-plugins/manifest_plugin/manifest_resolve.cpp
```diff
@@ -75,7 +76,27 @@ void ManifestResolveCommand::set_argument_parser() {
...
+    libpkgmanifest::input::Parser parser;
+    try {
+        // Try parsing as a non-prototype infile first.
+        input = parser.parse(input_path);
+    } catch (const libpkgmanifest::common::ParserError &) {
+        const auto & non_prototype_parse_ep = std::current_exception();
+        try {
+            // If non-prototype infile parsing fails, it may instead be a valid
+            // rpm-lockfile-prototype infile.
+            input = parser.parse_prototype(input_path);
```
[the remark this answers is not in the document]
evan-goode: Yes, if the infile is neither an valid libpkgmanifest infile nor a valid prototype infile, we could try harder to guess which one the user meant to supply. I took a simple approach here since I thought "invalid prototype infile" would be a rare case going forward as hopefully consumers adopt the libpkgmanifest schema. Open to human opinions here.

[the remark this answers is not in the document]
pkratoch: I think this is fine. The version should be parsed in the `parse` or `parse_prototype` functions and the document type is not there for the prototype infile. But I agree that we should check the version at some point, see my comment https://github.com/rpm-software-management/libpkgmanifest/pull/68#pullrequestreview-5208436874 . But it's not necessary in this PR, it would be better as a followup.

Btw, isn't contentOrigin that qodo mentions just for information about repositories in the prototype infile? I don't think this keyword has anything to do with the document type (whether it's prototype or not).

dnf5-plugins/manifest_plugin/manifest.hpp
```diff
@@ -12,7 +12,7 @@
 
 #include <vector>
 
-const std::string DEFAULT_INPUT_FILENAME{"rpms.in.yaml"};
+const std::string DEFAULT_INPUT_FILENAME{"packages.input.yaml"};
```
pkratoch, line 15: I like that it fits the pattern of the other yaml. I don't mind the length, because when used as a default, it picks this file automatically, and when user wants to specify a different file, they can choose a different name anyway.

doc/dnf5_plugins/manifest.8.rst
```diff
@@ -80,7 +80,7 @@ Options
 
 ``--input``
     Specify a custom path for the prototype input file.
-    By default, ``rpms.in.yaml`` is used.
+    By default, ``packages.input.yaml`` is used.
```
pkratoch: Please also remove the "prototype" from the description.

doc/dnf5_plugins/manifest.8.rst
```diff
@@ -112,7 +112,7 @@ Examples
 --------
 
 ``dnf manifest resolve``
-    Resolve the ``rpms.in.yaml`` prototype input file in the current directory
+    Resolve the ``packages.input.yaml`` prototype input file in the current directory
```
pkratoch: Please also remove the "prototype" from the description.

dnf5-plugins/manifest_plugin/manifest_resolve.cpp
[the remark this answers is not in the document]
pkratoch: Ah, the version check would need to happen in libpkgmanifest anyway, didn't realize it in my previous comment.

dnf5-plugins/manifest_plugin/manifest_resolve.cpp
```diff
@@ -75,7 +76,27 @@ void ManifestResolveCommand::set_argument_parser() {
...
+            ctx.print_error(libdnf5::utils::sformat(
+                _("Warning: the rpm-lockfile-prototype input file format is deprecated. "
+                  "Consider migrating to the libpkgmanifest input format."),
+                input_path));
```
pkratoch: The `input_path` is unused, it can be dropped.

dnf5-plugins/manifest_plugin/manifest_resolve.cpp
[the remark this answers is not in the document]
evan-goode: > Btw, isn't contentOrigin that qodo mentions just for information about repositories in the prototype infile? I don't think this keyword has anything to do with the document type (whether it's prototype or not).

I believe it's a required field in the prototype infile schema, so it could be used to identify them.

dnf5-plugins/manifest_plugin/manifest_resolve.cpp
evan-goode: Fixed

doc/dnf5_plugins/manifest.8.rst
evan-goode: Done

doc/dnf5_plugins/manifest.8.rst
evan-goode: Done

PR #2914, opened 2026-09-03
SWIG bindings: Rename (nonmodule) shared files to .swg extension

jrohel:
This PR renames SWIG shared/library files in `bindings/libdnf5/` from `.i` and `.hpp` extensions to `.swg` (SWIG) to better distinguish them from module interface files and standard C++ headers.

SWIG conventions differentiate between:
- **Module interface files (`.i`)** - Generate language bindings for specific modules (e.g., `base.i`, `repo.i`)
- **Shared/library files (`.swg`)** - Contain reusable macros, typemaps, and helpers that are `%include`'d by multiple modules

Note:
I have added an additional commit "doc: Remove shared.i workaround from rtd_build.py". Since files that do not define a module interface were renamed to .swg, the workaround is no longer required.


kontura:
Nice! This works and fixes the `readthedocs` build, it also means we can remove the workaround for `shared.i` in https://github.com/rpm-software-management/dnf5/blob/main/doc/rtd_build.py#L30-L32, can you also do that please?

kontura (approved):

jrohel:
@kontura
> it also means we can remove the workaround for shared.i in https://github.com/rpm-software-management/dnf5/blob/main/doc/rtd_build.py#L30-L32, can you also do that please?

Yes, done. I've added a commit removing the workaround. Thanks for pointing it out - I was just about to look into why `shared.i` worked fine while `my_filesystem.i` caused problems.

kontura:
I am not sure what is going on with packit but the tests pass locally.

PR #2915, opened 2026-09-03
Migrate .gitignore from ci-dnf-stack

kontura:


m-blaha:
There already is the `integration-tests/.gitignore` file. I'd prefer add this content there (or alternatively merge all in one global .gitignore file). WDYT?

kontura:
> There already is the `integration-tests/.gitignore` file. I'd prefer add this content there (or alternatively merge all in one global .gitignore file). WDYT?

Oh I didn't know about the second .gitignore, that is a better fit, thank you!

m-blaha (approved):

PR #2919, opened 2026-09-04
spec: Shorten overlong package summary

ppisar:
rpmlint complained that libdnf5-plugin-local's summary is too long. This patch makes it shorter. It also wraps wide package descriptions.


fhbash (approved):
LGTM, thanks

PR #2920, opened 2026-09-04
spec: Version RPM provides

ppisar:
rpmlint complained that RPM Provides, like "dnf5-command(downgrade)", are not versioned. It's good to version them for case somebody needs to constrain the required RPM symbol, e.g. a new option for "dnf5 downrgade" command available from a specific version.

This patch adds a version-release to all provides.


kontura (approved):
Thank you.

PR #2930, opened 2026-09-09
Makefile: Add set -e to test-integration-build recipe

m-blaha:
Without set -e, a failed "container-test build" was still followed by touch, so the stamp file got updated and "make test-integration" would proceed to run tests against a stale or broken container image.


jrohel (approved):
Thanks

PR #2931, opened 2026-09-09
VendorChangeManager: Allow empty vendor policies in compact format, unit tests included

jrohel:
Users need to mask (disable) distribution vendor policies located in `/usr/share/dnf5/vendors.d/` by creating a file with the same name in `/etc/dnf/vendors.d/`, but without adding any actual vendor change rules. Empty policies make this possible.


kontura (approved):
Thank you.

PR #2933, opened 2026-09-09
spec: Fix shell completions paths

m-blaha:
According to
https://docs.fedoraproject.org/en-US/packaging-guidelines/ShellCompletions/
we must use macros instead of hardcoding paths.

Also, drop ownership of the completion directories, as they are owned by the
"filesystem" package. Claiming ownership of them in dnf5 violates
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_file_and_directory_ownership

### DIFF
--- a/dnf5.spec
+++ b/dnf5.spec
@@ -332,12 +332,8 @@ It supports RPM packages%{?with_modulemd:, modulemd modules,} and comps groups &
 %dir %{_sysconfdir}/dnf/dnf5-plugins
 %doc %{_libdir}/dnf5/plugins/README
 %dir %{_libdir}/libdnf5/plugins
-%dir %{_datadir}/bash-completion/
-%dir %{_datadir}/bash-completion/completions/
-%{_datadir}/bash-completion/completions/dnf*
-%dir %{_datadir}/zsh/
-%dir %{_datadir}/zsh/site-functions/
-%{_datadir}/zsh/site-functions/_dnf5
+%{bash_completions_dir}/dnf*
+%{zsh_completions_dir}/_dnf5
 %license COPYING.md
 %license gpl-2.0.txt
 %doc AUTHORS.md CHANGELOG.md CONTRIBUTING.md README.md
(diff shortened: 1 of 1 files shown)


dnf5.spec
```diff
@@ -332,12 +332,8 @@ It supports RPM packages%{?with_modulemd:, modulemd modules,} and comps groups &
...
 %doc %{_libdir}/dnf5/plugins/README
 %dir %{_libdir}/libdnf5/plugins
-%dir %{_datadir}/bash-completion/
-%dir %{_datadir}/bash-completion/completions/
-%{_datadir}/bash-completion/completions/dnf*
-%dir %{_datadir}/zsh/
-%dir %{_datadir}/zsh/site-functions/
-%{_datadir}/zsh/site-functions/_dnf5
+%{bash_completions_dir}/dnf*
+%{zsh_completions_dir}/_dnf5
```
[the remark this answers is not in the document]
m-blaha: The file defining the macro is provided by redhat-rpm-config package, which should be in the buildroot automatically.

dnf5.spec
```diff
@@ -332,12 +332,8 @@ It supports RPM packages%{?with_modulemd:, modulemd modules,} and comps groups &
 %dir %{_sysconfdir}/dnf/dnf5-plugins
 %doc %{_libdir}/dnf5/plugins/README
 %dir %{_libdir}/libdnf5/plugins
-%dir %{_datadir}/bash-completion/
```
ppisar, line 335: dnf5 package needs to either own all the directories, or require another package which owns them. Since `bash-completion` package is only recommended, dnf5.spec needs to own them.

evan-goode: The completions directories are owned by `filesystem`:

```
dnf5 rq --list filesystem | grep -e completions -e site-functions
/usr/share/bash-completion/completions
/usr/share/fish/vendor_completions.d
/usr/share/zsh/site-functions
```

which is already part of the dnf5 closure, but we could add an explicit dependency on `filesystem`.

evan-goode (changes requested):
~Yeah, let's add `Requires: filesystem`~

dnf5.spec
m-blaha: This is what Fedora packaging guidlines says:
```
Directory ownership is a little more complex than file ownership. Packages must own all directories they put files in, except for:

- any directories owned by the filesystem, man, or other explicitly created -filesystem packages

```

evan-goode: OK, in context I would interpret that as the explicit filesystem dependency is not needed.

evan-goode (approved):

dnf5.spec
m-blaha: On the other hand "explicit is better than implicit". I'm fine with adding `filesystem` requirement.

ppisar: You are right, Evan. The ownership is not needed. I always forget the filesystem package.

ppisar (approved):

ppisar:
> We were not able to find or create Copr project `packit/rpm-software-management-dnf5-2933` specified in the config with the following error:
> 
I thought that the expired Copr key was fixed by Ales yesterday <https://github.com/rpm-software-management/ci-dnf-stack/issues/1945>.

PR #2940, opened 2026-09-14
progressbar: Report progress to terminal taskbar via OSC 9;4

xsuchy:
Emit the ConEmu OSC 9;4 escape sequence so terminals such as Konsole, Windows Terminal, foot, WezTerm and kitty can show overall progress on the tab/card/taskbar entry during downloads and transactions.

A new tty::set_taskbar_progress() helper writes the sequence and is driven from MultiProgressBar's total-bar render.

Key mapping decisions:
- TaskbarProgressState enum values match the protocol numbers directly (CLEAR=0, NORMAL=1, ERROR=2, INDETERMINATE=3, WARNING=4), so emission is a plain static_cast<int> with no lookup table.
- While the total bar runs, emit NORMAL with its get_percent_done(); when the total bar is finished, emit CLEAR so the indicator does not linger after exit. The destructor also emits CLEAR to cover interrupted runs.
- Only emitted when tty::is_interactive(); percent is clamped to 0-100. Emission is tied to total-bar visibility (total_bar_visible_limit).

Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


libdnf5-cli/progressbar/multi_progress_bar.cpp
```diff
@@ -286,6 +288,17 @@ std::ostream & operator<<(std::ostream & stream, MultiProgressBar & mbar) {
...
         }
 
+        // Report overall progress to the terminal taskbar/tab (OSC 9;4).
+        // Remove the indicator once finished so it does not linger after exit.
+        if (is_interactive) {
+            if (mbar_total.is_finished()) {
+                tty::set_taskbar_progress(text_buffer, tty::TaskbarProgressState::CLEAR);
+            } else {
+                tty::set_taskbar_progress(
+                    text_buffer, tty::TaskbarProgressState::NORMAL, mbar_total.get_percent_done());
```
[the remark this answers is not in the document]
xsuchy: I updated the code.

m-blaha:
Nice :)

(The other day, I noticed this in a different app and wondered how it could be achieved.)

Screenshot_20260914_125549

libdnf5-cli/tty.cpp
```diff
@@ -70,6 +70,19 @@ bool is_interactive() {
...
 
+std::ostream & set_taskbar_progress(std::ostream & stream, TaskbarProgressState state, int percent) {
+    if (is_interactive()) {
+        if (percent < 0) {
+            percent = 0;
+        } else if (percent > 100) {
+            percent = 100;
+        }
+        // OSC 9 ; 4 ; <state> ; <percent> BEL
+        stream << "\033]9;4;" << static_cast<int>(state) << ";" << percent << "\007";
```
jrohel: @xsuchy 

Thanks for the UX improvement!

I just have a couple of minor points/suggestions:

1. According to `ECMA-48` and `ISO/IEC 6429` standards, `ST` (String Terminator) is defined as the sequence `ESC \`. I know `BEL` was introduced as a hack (probably in xterm) and is widely supported nowadays, but I'm wondering if it might be better to stick to the actual standard here.

2. We have a bit of a mixed style in this file - we're currently mixing octal and hexadecimal escapes. Personally, I find hexadecimal notation much more readable. What do you think about unifying it like this?

```
// OSC 9 ; 4 ; <state> ; <percent> ST
stream << "\x1b]9;4;" << static_cast<int>(state) << ";" << percent << "\x1b\x5c";
```

xsuchy: np, I will amend it to use hexa notation.

xsuchy: You are right. I checked the flatpak code (that inspired me originally) and indeed they changed a year ago:

> * When using `FLATPAK_TTY_PROGRESS`, terminate OSC escape sequence with
  standard ST sequence instead of xterm-specific BEL (#6092)

xsuchy:
Updated and rebased.

jrohel (approved):
Thank you!

PR #2943, opened 2026-09-14
build: Remove build scripts from literalinclude sources

ppisar:
81f673fca295d1543a40843d6cbcc427ee102c9b ("doc: Copy literalinclude sources into the processed dir") copied files with well-known extensions to "processed" directory with the pretext that all of them are literarly incuded sources for Sphinx. It unintentinally also copied various build scripts written in Python, like doc/atp.py, which were not the literal source files.

This patch excludes those build scripts to prevent from pulluting the "processed" directory.


doc/CMakeLists.txt
```diff
@@ -36,7 +36,9 @@ endforeach()
 file(GLOB_RECURSE LITERALINCLUDE_SOURCES LIST_DIRECTORIES false FOLLOW_SYMLINKS
     RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
     *.cpp *.hpp *.py *.conf CMakeLists.txt)
-list(REMOVE_ITEM LITERALINCLUDE_SOURCES "CMakeLists.txt" "sphinx/CMakeLists.txt")
+list(REMOVE_ITEM LITERALINCLUDE_SOURCES "atp.py" "CMakeLists.txt" "rtd_build.py"
```
kontura: The `rtd_build.py` was just dropped.

ppisar: Yes. I should not have to hurry with this patch.
Now rebased and without rtd_build.py removal.

kontura (approved):

PR #2946, opened 2026-09-15
libdnf5: Move the hardcoded suggest-reboot package list to a configuration file

scaronni:
Replace the hardcoded `CORE_PACKAGE_NAMES` list in `package_query.cpp` with `default.conf`, installed into `/usr/share/dnf5/suggest-reboot.d/`.

Follow up for https://github.com/rpm-software-management/dnf5/pull/2929 as requested by @jrohel.


scaronni:
@jrohel waiting for your input before proceeding, as the request for change from the bot above is not on something introduced by this PR:

 If you're fine I can add the global requested change later, after we sort out about this configuration file.

Thanks.

PR #2947, opened 2026-09-17
Use LR_PACKAGEDOWNLOAD_TRANSIENT for packages that will be deleted

stewartsmith:
When keepcache=0 (and force_keep_packages() hasn't overridden it), downloaded packages are unlinked shortly after the transaction and are never re-verified from disk. Tell librepo not to bother persisting their checksum to an xattr, since nothing will ever read it back - this removes an fsync() per package from the download path with no change in what gets verified.

This results in roughly a 3-43% wall clock time improvement for downloading packages on XFS (tested on the AL2027 Preview release), and 48-68% faster on the Fedora 44 AMI's default btrfs. These tests were for some package sets I guessed at being representative of what one may download to a new machine.
(see librepo commit for full benchmark details)

reposync and `dnf5 download` call force_keep_packages(true) and take the keep_enforced branch, so they never set this flag and their checksum xattrs persist exactly as before (see
reposync-xattrs.feature).

Requires a librepo build carrying LR_PACKAGEDOWNLOAD_TRANSIENT (see the `librepo` pull request: https://github.com/rpm-software-management/librepo/pull/412)


m-blaha:
Let me block it until we have a concrete librepo version number  to require.
But the code itself is correct.

PR #2950, opened 2026-09-17
history: Display times in the local timezone instead of UTC

### LINKED ISSUE #2949: 中文环境 dnf histroy list 显示的时间不对
中文环境 dnf histroy list 显示的时间不对

[环境信息]

Image

Image

[现象]

Image

期望 红框 显示的应该是： 2026-09-17 14:56:24

wangtsingx:
Fixes #2949, addresses the requirements from #2200.

`format_epoch()` formatted a `std::chrono::system_clock::time_point` directly with `fmt::format("{:%F %X}", ...)`. fmt renders `system_clock` time points in **UTC**, so every time shown by `dnf history list` (and `history info`, `dnf status`, advisory/repo info outputs, which share the same helper) was shifted by the local UTC offset — e.g. a transaction that started at 14:56:24 +0800 was displayed as 06:56:24.

The stored timestamps in `transaction_history.sqlite` are correct; only the CLI rendering was wrong.

## Changes

Per the team decision in #2200 (UI timestamps in local timezone honoring `TZ`, offset indicator on all timestamps, logs stay in UTC):

- `common/utils/string.hpp`: `format_epoch()` converts the epoch to a `std::tm` in the local timezone and appends the UTC offset, e.g. `2026-09-17 14:56:24 +0800`. Timestamps that `std::localtime` cannot represent (out-of-range values) keep falling back to the raw `seconds since Unix epoch` output instead of dereferencing a null pointer.
- `libdnf5-cli/session.cpp`: `DateOption` (used by `reposync --min-buildtime` and `changelog --since`) parsed `%F` dates as UTC; it now parses them as local time (`std::get_time` + `std::mktime`), keeping the existing error messages and time_t range checks.
- `libdnf5-cli/output/changelogs.cpp` and `dnf5-plugins/changelog_plugin/changelog.cpp`: changelog dates and the "Listing changelogs since ..." line moved from `gmtime` to local time with an offset.
- Log timestamps already use `%FT%T%z` in UTC and are unchanged.

## Tests

- `test/libdnf5/utils/test_string.{hpp,cpp}`: new `UtilsStringTest::test_format_epoch` asserting one fixed timestamp under `TZ=Asia/Shanghai` (`2026-09-17 14:56:24 +0800`) and `TZ=UTC` (`2026-09-17 06:56:24 +0000`), plus the out-of-range fallback; also fixes the suite registration so `test_rsplit` runs again, and restores the original `TZ` after the test.
- Behave expectations in `updateinfo.feature`, `security-upgrade.feature` and `changelog.feature` updated for the offset suffix (CI runs in UTC).

## Verification

- Full build of `libdnf5`, `libdnf5-cli`, `dnf5` and plugins (except appstream/rhsm/manifest) in a Fedora 46 (rawhide) chroot with fmt 12.
- `test/libdnf5/run_tests`: OK (523), including the restored `test_rsplit` and the new `test_format_epoch` (verified it fails against the old implementation).
- `clang-format --dry-run -Werror` on all changed files.
- `changelog --since=2026-09-17` parses as local midnight under `TZ=UTC` (`+0000`) and `TZ=Asia/Shanghai` (`+0800`); invalid dates keep the original error message.
- On a real Fedora 46 system, `dnf history list` now matches the timestamps stored in `transaction_history.sqlite` (rendered with `datetime(dt,'unixepoch','localtime')`).

Assisted-by: Open My Pi coding agent


ppisar:
This does not address all requirements stated in #2200.

wangtsingx:
Thanks for the pointer to #2200. Updated in f021a468 to address all its requirements:

- **UI timestamps in local timezone honoring TZ**: `format_epoch()` now converts through `std::localtime()`, so all UI timestamps (history, status, advisory, repo info, offline, expired-pgp-keys) follow the process timezone instead of being rendered in UTC.
- **Offset indicator**: per the team decision in #2200, the UTC offset is now appended to UI timestamps (e.g. `2026-09-17 14:56:24 +0800`). Log timestamps were already UTC with an offset (`%FT%T%z`) and are left unchanged.
- **Time parsing honors TZ too**: `DateOption` (used by `reposync --min-buildtime` and `changelog --since`) parsed `%F` as UTC via `std::chrono::parse` into `sys_seconds`; it now parses the date as local time (`std::get_time` + `std::mktime`), keeping the existing error messages and time_t range checks.
- **`dnf changelog`**: the header date and the "Listing changelogs since ..." line also moved from `gmtime` to local time with an offset.
- **Tests**: fixed the CppUnit registration so `test_rsplit` runs again (it was accidentally clobbered by `test_format_epoch`), the new test restores the original `TZ` value, and expectations were extended to cover the offset. Behave test expectations in `updateinfo.feature`, `security-upgrade.feature` and `changelog.feature` updated for the new format.

Verified: full build (`libdnf5`, `libdnf5-cli`, `dnf5` + plugins except appstream/rhsm/manifest), `run_tests` OK (523, including restored `test_rsplit`), `clang-format --dry-run -Werror` on changed files, and manual `changelog --since` parsing under `TZ=UTC` vs `TZ=Asia/Shanghai` (+0000 vs +0800, matching local midnight).

common/utils/string.hpp
```diff
@@ -113,9 +113,14 @@ inline std::string tolower(const std::string & s) {
...
     if (std::in_range<time_t>(epoch_num)) {
-        const auto epoch = static_cast<time_t>(epoch_num);
-        return fmt::format(
-            "{:%F %X}", std::chrono::round<std::chrono::seconds>(std::chrono::system_clock::from_time_t(epoch)));
+        auto epoch = static_cast<time_t>(epoch_num);
+        // Convert to std::tm in the local timezone first. Formatting a
+        // system_clock time point directly would render the time in UTC.
+        // The UTC offset is appended per
+        // https://github.com/rpm-software-management/dnf5/issues/2200.
+        if (const auto * local_time = std::localtime(&epoch)) {
```
ppisar, line 121: One of the reasons why time stamps were not printed in local time zone was that the one-argument std::localtime(epoch) call is not thread-safe. See <https://en.cppreference.com/cpp/chrono/c/localtime> for more details. That might not not a problem for dnf5 program, but it could be a problem for libdnf5-cli. I'd like to hear other's opinion on this problem.

C++26 has a two-argument variant which is thread-safe, but DNF5 right now sticks to C++20.

wangtsingx: Good point. Switched to the POSIX `localtime_r()` in 0e06069b - `format_epoch()` in `common/utils/string.hpp`, the changelog date rendering in `libdnf5-cli/output/changelogs.cpp` and the "Listing changelogs since ..." line in the changelog plugin. On failure it falls back to the raw seconds/epoch value instead of dereferencing null. dnf5 only builds against POSIX targets, so the POSIX wrapper seems fine until the C++26 two-argument `std::localtime` is available; the unit test still covers both timezones.

PR #2957, opened 2026-09-21
fix:  DNF5_FORCE_COLUMNS / non-TTY width in transaction table

fhbash:
Fixes DNF5_FORCE_COLUMNS  environment variable


libdnf5-cli/output/transaction_table.cpp
```diff
@@ -793,6 +798,14 @@ bool print_transaction_table(ITransaction & transaction) {
     print_resolve_logs(transaction);
 
     TransactionTable table(transaction);
+
+    // Only force a specific terminal width when the user explicitly requested one via
+    // DNF5_FORCE_COLUMNS. Otherwise let libsmartcols detect the width itself: it uses
+    // the real terminal width on a TTY and prints at full (untruncated) width when the
+    // output is not a TTY, which keeps long values readable in piped output.
+    if (std::getenv("DNF5_FORCE_COLUMNS") != nullptr) {
```
ppisar, line 806: Checking for the environment variable is unnecessary. The check is performed in libdnf5::cli::tty::get_width(). The call should be unconditional here.

fhbash: I agree, but for some reason if I remove this, we start see several integration tests failing 🤔

ppisar: There are only two integration tests that sets the variable. The other tests were probably written to expect what libsmartcols defaults to.

I think that calling table.set_term_width() forces the table to have that width exactly, while leaving libsmartcols table width unset enables the table to be narrower of wider depending on its content.

Also get_width() defaults to _80_ (if no stdout is not a terminal and that's maybe the case of integration tests) which maybe is different from a libsmartcols' default maximal width.

I consider the unconditional behavior consistent with what we have in _dnf5 automatic_ (see set_term_width() call in AutomaticCommand::run()) and what we have in transaction scriptlet output (see tty::get_width() call in DownloadProgressBar::to_stream()) and therefor better.

But I understand that it's a change in the behavior and therefor we can leave it to independent discussion and pull request.

ppisar (approved):

